PR #1, opened 2019-08-21
Single Hive query nodes

wild-endeavor:
Hive queries were originally bundled into one large dynamic node.  This was a flawed design as it made the execution side of things far more complex.  This now splits out each individual query into different nodes.  See the corresponding change in the IDL https://github.com/lyft/flyteidl/pull/3.

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.1.7'
+__version__ = '0.1.8b0'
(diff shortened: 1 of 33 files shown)


wild-endeavor:
@matthewphsmith - had to roll back the propeller change due to an issue haytham found in the updated plugins but the fix for it should be deployed tomorrow.  After which this change can go out.  Could you take another look please?

Thank you.

wild-endeavor:
ptal @matthewphsmith

wild-endeavor:
wait ptal doesn't do anything anymore.

flytekit/__init__.py
```diff
@@ -1,4 +1,5 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.1.9'
+__version__ = '0.1.10'
```
matthewphsmith: We should make 0.2.0 so we mark this as a feature change.

PR #2, opened 2019-08-21
Implement Hive Unit Test Behavior

matthewphsmith:
Hive unit tests are primarily interesting for validating the queries generated. This change surfaces the queries generated as the output of the `@hive_task` unit test.

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.1.5'
+__version__ = '0.1.6'
--- a/flytekit/engines/unit/engine.py
+++ b/flytekit/engines/unit/engine.py
@@ -7,6 +7,8 @@ import six as _six
 from datetime import datetime as _datetime
 from six import moves as _six_moves
 
+from google.protobuf.json_format import ParseDict as _ParseDict
+from flyteidl.plugins import qubole_pb2 as _qubole_pb2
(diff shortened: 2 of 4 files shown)


matthewphsmith:
PTAL @wild-endeavor

flytekit/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.1.5'
+__version__ = '0.1.6'
```
chanadian: you probably are doing this, but  you can submit this after your other PR so you get both changes the version bump

flytekit/engines/unit/engine.py
```diff
@@ -107,24 +112,32 @@ def _execute_user_code(self, inputs):
...
+        :rtype: T
+        """
+        return outputs
+
     def register(self, identifier, version):
         raise _user_exceptions.FlyteAssertion("You cannot register unit test tasks.")
 
 
 class ReturnOutputsTask(UnitTestEngineTask):
-    def execute(self, inputs, context=None):
```
chanadian: is this no longer needed?

flytekit/__init__.py
matthewphsmith: Yeah in the future I will bump the version via an explicit PR.

flytekit/engines/unit/engine.py
matthewphsmith: It's a refactor so this method is sufficient for transforming the execute function of the super class to behave as we desire.

PR #3, opened 2019-08-22
Update pyflyte.py

honnix:
Is it possible to ignore config file and use environment variables (if exist or fail loudly) instead? The reasoning is to hide config file from workflow developers, so that doesn't need to be in the repository.

### DIFF
--- a/flytekit/clis/sdk_in_container/pyflyte.py
+++ b/flytekit/clis/sdk_in_container/pyflyte.py
@@ -83,8 +83,9 @@ def update_configuration_file(config_file_path):
         set_flyte_config_file(configuration_file.as_posix())
         click.secho('Flyte Admin URL {}'.format(_URL.get()), fg='green')
     else:
-        click.secho("Configuration file '{}' could not be loaded.".format(CONFIGURATION_PATH.get()), color='red')
-        exit(-1)
+        set_flyte_config_file(None)
+        click.secho("Configuration file '{}' could not be loaded. Reading values from environment.".format(CONFIGURATION_PATH.get()),
+                    color='red')
 
 
 main.add_command(register)


matthewphsmith:
Yes indeed! It should be possible not to have a config file as long as the env vars are set correctly. Though I think the change might be a little more involved than this.  Ideally, in the tool that you edited, we only load the config file if specified--either on the command line or via the env var.  Then additionally, we need to make sure the code elsewhere doesn't blow up in the case that a config file isn't found.

Any use of this context manager (https://github.com/lyft/flytekit/search?q=TemporaryConfiguration&unscoped_q=TemporaryConfiguration) could be at risk of blowing up when a file is not used--I'd need to look a little harder.

Another option (if it's more expedient) to hide the config file from users is to store it in a base container and then set the env var for the configuration path to that file path.  Additionally, any individual config can be overridden by setting the appropriate env var.  The resolution order is: env var -> config file -> default (or exception, if required).

matthewphsmith:
Let me know if you'd like additional help on this! Thanks!

honnix:
> Another option (if it's more expedient) to hide the config file from users is to store it in a base container and then set the env var for the configuration path to that file path. Additionally, any individual config can be overridden by setting the appropriate env var. The resolution order is: env var -> config file -> default (or exception, if required).

Yeah this is more or less what we are doing. We put a config file in our deployment container/environment and only use it when registering workflows/tasks. So in user space (git repo, image), there is no config file. If we are not mistaken this file is not used at runtime.

We thought about putting a config file in base image, but we are a bit hesitated to do that, mainly because updating the config file would require users to update their base image.

We will take a look at the PR again to make sure nothing would blow up. Thanks for reviewing and great comments.

honnix:
@matthewphsmith I made a few more changes. PTAL.

PR #6, opened 2019-08-26
Add GCS Support

matthewphsmith:
This adds support for uploading and downloading data to/from GCS.

### DIFF
--- a/flytekit/interfaces/data/gcs/gcs_proxy.py
+++ b/flytekit/interfaces/data/gcs/gcs_proxy.py
@@ -1,0 +1,122 @@
+from __future__ import absolute_import
+
+import os as _os
+import sys as _sys
+import uuid as _uuid
+
+from flytekit.configuration import gcp as _gcp_config
+from flytekit.interfaces import random as _flyte_random
+from flytekit.interfaces.data import common as _common_data
+from flytekit.tools import subprocess as _subprocess
+from flytekit.common.exceptions.user import FlyteUserException as _FlyteUserException
+
+
+if _sys.version_info >= (3,):
+    from shutil import which as _which
+else:
+    from distutils.spawn import find_executable as _which
+
+
+def _update_cmd_config_and_execute(cmd):
+    env = _os.environ.copy()
+    return _subprocess.check_call(cmd, env=env)
+
+
+class GCSProxy(_common_data.DataProxy):
+    _GS_UTIL_CLI = "gsutil"
+
+    @staticmethod
+    def _check_binary():
+        """
+        Make sure that the AWS cli is present
+        """
+        if not _which(GCSProxy._GS_UTIL_CLI):
+            raise _FlyteUserException('gsutil (gcloud cli) not found at Please install.')
+
+    @staticmethod
+    def _split_s3_path_to_bucket_and_key(path):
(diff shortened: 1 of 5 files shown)


honnix:
@matthewphsmith What is the status of this PR? Would you like us to review as well? Thanks.

matthewphsmith:
It's good to go from my perspective, I just need an approval on a review.

wild-endeavor:
+1 bump version plz

tnsetting:
I think the PR works mostly but need to be rebased with master branch.

I verified the PR and the `pyflyte-execute` could download the `input.pb` from gcs and upload the `output.pb` to gcs. But the node still failed because of `OutputsNotFoundError: Outputs not found for task type python-task...`. It looks like the output structure has been changed and the propeller expect the output.pb to be under the data folder.

flytekit/interfaces/data/gcs/gcs_proxy.py
```diff
@@ -0,0 +1,112 @@
...
+    def upload_directory(self, local_path, remote_path):
+        """
+        :param Text local_path:
+        :param Text remote_path:
+        """
+        if not remote_path.startswith("gs://"):
+            raise ValueError("Not an GS Key. Please use FQN (GS ARN) of the format gs://...")
+
+        GCSProxy._check_binary()
+        cmd = [GCSProxy._GS_UTIL_CLI, "cp", "-r", local_path, remote_path]
```
tnsetting: I think we can change `local_path` to `local_path + '\*' to just avoid upload the whole `engine_dir` folder.

honnix: The current behaviour in this PR seems fine to me, because the expectation of this function is to upload the specified local path. Maybe it is the caller who should use `../*` instead?

honnix:
I think there is a caveat using `gsutil`: it doesn't seem to respect `GOOGLE_APPLICATION_CREDENTIALS` environment variable. That means the container should either have a service account activated so it becomes ADC effectively, or carry an identity such as GCE.

flytekit/interfaces/data/gcs/gcs_proxy.py
honnix: We can confirm that `gsutil` and `aws s3` behave differently when copying a dir recursively.

e.g:

```
$ tree test
test
└── test1
    └── test.txt

1 directories, 1 file

$ gsutil cp -r test/* gs://flyte-test
Copying file://test/test1/a/b/test.txt [Content-Type=text/plain]...
/ [1 files][    0.0 B/    0.0 B]
Operation completed over 1 objects.

$ gsutil ls gs://flyte-test
gs://flyte-test//
gs://flyte-test/test1/    <--------
```
vs.
```
$ gsutil cp -r test gs://flyte-test
Copying file://test/test1/a/b/test.txt [Content-Type=text/plain]...
/ [1 files][    0.0 B/    0.0 B]
Operation completed over 1 objects.

$ gsutil ls gs://flyte-test
gs://flyte-test//
gs://flyte-test/test/    <--------
```

honnix:
Rebased and addressed the comments https://github.com/lyft/flytekit/pull/41

PR #11, opened 2019-09-05
Implement client-side behavior for active launch plans

matthewphsmith:
* Implement client endpoints for get and list on active launch plans.
* Implement tool command to activate launch plans regardless of having a schedule or not.
* Implement SDK fetch() method to retrieve the active launch plan when looking for a non-specific version
* Deprecates existing activation command (activate-all-schedules) in favor of the new command.

### DIFF
--- a/flytekit/clients/friendly.py
+++ b/flytekit/clients/friendly.py
@@ -371,6 +371,22 @@ class SynchronousFlyteClient(_RawSynchronousFlyteClient):
             )
         )
 
+    def get_active_launch_plan(self, identifier):
+        """
+        Retrieves the active launch plan entity given a named entity identifier (project, domain, name).  Raises an
+        error if no active launch plan exists.
+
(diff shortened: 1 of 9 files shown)


flytekit/clients/friendly.py
```diff
@@ -469,6 +485,55 @@ def list_launch_plans_paginated(
...
+        lp_list = super(SynchronousFlyteClient, self).list_active_launch_plans_paginated(
+            _launch_plan_pb2.ActiveLaunchPlanListRequest(
+                project=project,
+                domain=domain,
+                limit=limit,
+                token=token,
+                sort_by=None if sort_by is None else sort_by.to_flyte_idl()
+            )
+        )
+        # TODO: tmp workaround
```
chanadian: may want to explain why this is a temp workaround

matthewphsmith: Yeah, that's a copy-pasta (it is used in other parts of the code too). Let me verify if it is needed and remove in one subsequent commit.

PR #15, opened 2019-09-30
Fix Hive Unit Tests

matthewphsmith:
This bug went missing because a test file was inadvertently not prefixed with `test_` and thus was not found by `pytest`.  Once that was corrected, there was a slight bug resulting from https://github.com/lyft/flytekit/pull/1 since queries are now represented as separate nodes.


flytekit/engines/unit/engine.py
```diff
@@ -251,15 +251,18 @@ def _transform_for_user_output(self, outputs):
...
                 for t in futures.tasks
             }
-            return [
-                q.query
-                for q in task_ids_to_defs[futures.nodes[0].task_node.reference_id.name].query_collection.queries
-            ]
+            for node in futures.nodes:
+                queries.extend(
+                    q.query
+                    for q in task_ids_to_defs[node.task_node.reference_id.name].query_collection.queries
```
wild-endeavor: can we just use the `.query.query` field?  it should be there.  We're still creating one element `QueryCollection`'s just for backwards compatibility.

PR #23, opened 2019-10-11
OAuth2 support for flyte-cli and SDK engine

katrogan:
This change adds authentication support for flyte-cli and the pyflyte CLIs.

# New authorization code
Specifically this change introduces an **AuthorizationClient** which implements the [PKCE authorization flow](https://www.oauth.com/oauth2-servers/pkce/authorization-code-exchange/) for untrusted clients. This client handles requesting an initial access token, spinning up a callback server to receive the access token and using that to retrieve an authorization code. The client also handles refreshing expired authorization tokens.

This change also includes a lightweight **DiscoveryClient** for retrieving authorization endpoint metadata defined in the [OAuth 2.0 Authorization Server Metadata](https://tools.ietf.org/id/draft-ietf-oauth-discovery-08.html) draft document.

An authorization client singleton is lazily initialized for use by flyte-cli.

# Pyflyte changes (basic auth)
Requests an authorization token using a username and password.

# Flyte-cli changes (standard auth)
Requests an authorization token using the PKCE flow.

# Raw client changes
Wraps RPC calls to flyteadmin in a retry handler that initiates the appropriate authentication flow defined in the flytekit config in response to `HTTP 401 unauthorized` response codes.


flytekit/clis/auth/auth.py
```diff
@@ -0,0 +1,234 @@
...
+        self._state = state
+        self._credentials = None
+
+        self._params = {
+            # TODO: need an audience param here?
+            "client_id": client_id,  # This must match the Client ID of the OAuth application.
+            "response_type": "code",  # Indicates the authorization code grant
+            "scope": "openid",  # ensures that the /token endpoint returns an ID token
+            # callback location where the user-agent will be directed to.
+            "redirect_uri": urljoin("http://localhost:8088", self._redirect_path),
```
wild-endeavor: can we pick a random port that's unused?  random googling, didn't know about this before and haven't tested. https://unix.stackexchange.com/questions/55913/whats-the-easiest-way-to-find-an-unused-local-port

of course this has to be done first, not sure if it is.  https://support.okta.com/help/s/question/0D50Z00008G7VBgSAN/openid-connect-redirect-uri

flytekit/clients/raw.py
```diff
@@ -35,6 +35,8 @@ def __init__(self, url, insecure=False, credentials=None, options=None):
         :param Text credentials: [Optional] If provided, a secure channel will be opened with the Flyte Admin Service.
         :param dict[Text, Text] options: [Optional] A dict of key-value string pairs for configuring the gRPC core
             runtime.
+        :param [(Text, Text)] metadata: [Optional] metadata pairs to be transmitted to the
```
matthewphsmith: nit: `list[(Text,Text)]`

flytekit/clis/flyte_cli/main.py
```diff
@@ -11,14 +12,15 @@
...
+from flytekit.clis.auth import credentials as _credentials_access
 from flytekit.clis.helpers import construct_literal_map_from_variable_map as _construct_literal_map_from_variable_map, \
     construct_literal_map_from_parameter_map as _construct_literal_map_from_parameter_map, \
     parse_args_into_dict as _parse_args_into_dict, str2bool as _str2bool
 from flytekit.common import utils as _utils, launch_plan as _launch_plan_common
 from flytekit.common.core import identifier as _identifier
 from flytekit.common.types import helpers as _type_helpers
 from flytekit.common.utils import load_proto_from_file as _load_proto_from_file
-from flytekit.configuration import platform as _platform_config
+from flytekit.configuration import creds as _creds_config, platform as _platform_config
```
wild-endeavor: so I think we should keep the underlying config objects regardless (because they'll be useful in pyflyte), but currently none of these files iirc are used in flyte-cli.  That is, the `flytekit.config` style files are never read by flyte-cli, because it's assumed that users of flyte-cli might not have any flyte repos checked out at all - they may only be an administrator.  Auth has a fair amount to config though, it's pretty awkward to pass in everything as command line args.  What does @matthewphsmith think?

flytekit/clis/auth/auth.py
```diff
@@ -0,0 +1,229 @@
...
+from multiprocessing import Process as _Process, Queue as _Queue
+
+try:  # Python 3.5+
+    from http import HTTPStatus as _StatusCodes
+except ImportError:
+    try:  # Python 3
+        from http import client as _StatusCodes
+    except ImportError:  # Python 2
+        import httplib as _StatusCodes
+from BaseHTTPServer import HTTPServer as _HTTPServer, BaseHTTPRequestHandler as _BaseHTTPRequestHandler
```
matthewphsmith: nit: let's import the module. it makes refactoring easier and keeps the namespace cleaner.

```
import BaseHTTPServer as _base_http_server

Later...
_base_http_server.HTTPServer
```

matthewphsmith: same for urlparse below

flytekit/clis/auth/auth.py
```diff
@@ -0,0 +1,229 @@
...
+    code_verifier = _re.sub('[^a-zA-Z0-9]+', '', code_verifier)
+    if len(code_verifier) < 43:
+        raise ValueError("Verifier too short. number of bytes must be > 30.")
+    elif len(code_verifier) > 128:
+        raise ValueError("Verifier too long. number of bytes must be < 97.")
+    return code_verifier
+
+
+def _generate_state_parameter():
+    state = _base64.urlsafe_b64encode(_os.urandom(40)).decode('utf-8')
```
matthewphsmith: nit: constant

flytekit/clis/flyte_cli/main.py
matthewphsmith: I think it should definitely be env var based. You can use the configuration package though to make things easier since the objects are env var aware.

flytekit/configuration/platform.py
```diff
@@ -4,3 +4,7 @@
 
 URL = _config_common.FlyteRequiredStringConfigurationEntry('platform', 'url')
 INSECURE = _config_common.FlyteBoolConfigurationEntry('platform', 'insecure', default=False)
+AUTH = _config_common.FlyteBoolConfigurationEntry('platform', 'auth', default=False)
```
matthewphsmith: Would we actually configure this? or would we get a redirect when trying to connect?

flytekit/configuration/creds.py
```diff
@@ -0,0 +1,30 @@
+from __future__ import absolute_import
+
+from flytekit.configuration import common as _config_common
+
+DISCOVERY_ENDPOINT = _config_common.FlyteStringConfigurationEntry('credentials', 'discovery_endpoint', default=None)
```
matthewphsmith: are all these env vars going to be auto-set via the auth flow? or would a user ever specify any of this? If yes, can you let me know about a few of those cases to help me understand?

flytekit/clis/flyte_cli/main.py
katrogan: I think making it env var based makes it difficult to set-up for an end-user. I'd prefer to have a config that can be easily cloned and written, but I'm happy to move it to a separate directory if it makes sense organizationally.

flytekit/configuration/creds.py
katrogan: These are specific to the hosted Flyte platform - so they can't be set. For example the discovery endpoint is configured in admin. The native Okta app at Lyft we're testing with has a unique client id and a hardcoded, pre-configured redirect URI. The authorization metadata key is a departure we have to make from the standard protocol because Envoy actually behaves funkily if we use the default value.

flytekit/clis/flyte_cli/main.py
wild-endeavor: Haytham's argument was that there is a `.aws/config` and a `.kube/config` and a `.gitconfig`.  Why can't there be a .flyte/config?  I think it's cleaner than having to rely on five or six environment variables.  What's the downside that you're worried about Matt?

flytekit/clis/auth/auth.py
wild-endeavor: but it makes it harder to jump into the definition of the underlying object.

flytekit/configuration/platform.py
katrogan: @wild-endeavor do you know what the plan is? mind weighing in?

flytekit/clis/auth/auth.py
katrogan: hm actually i can still jump around in my ide to the underlying object with refactoring

flytekit/clis/flyte_cli/main.py
kumare3: I think we can definitely have `.flyte/config` The only thing is then we need to provide config modification commands (but with click opening up the editor is simple)
Another option is to have `.flyte/config` as an option otherwise explicitly specify all the args

flytekit/clis/flyte_cli/main.py
```diff
@@ -229,6 +236,23 @@ def _render_schedule_expr(lp):
...
+def _fetch_metadata():
+    """
+    Initializes gRPC metadata according to parameters set in the flyte config.
+    Currently this is used to pass security credentials when authentication is enabled.
+    :return [(Text, Text)]: metadata pairs to be transmitted to the service-side of the RPC.
+    """
+    if not _platform_config.AUTH.get():
+        # nothing to do
+        return None
+    access_token = _keyring.get_password(_metadata, _metadata_access_token)
```
kumare3: That is awesome!

kumare3:
I think before an approval we should either demonstrate the auth flow?

flytekit/configuration/platform.py
wild-endeavor: Sorry what will this be?  Whether or not to use auth?

flytekit/configuration/creds.py
```diff
@@ -15,13 +15,16 @@
 More details here: https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/.
 """
 
-REDIRECT_URI = _config_common.FlyteStringConfigurationEntry('credentials', 'redirect_uri', default=None)
+REDIRECT_URI = _config_common.FlyteStringConfigurationEntry('credentials', 'redirect_uri', default="http://localhost:53593/callback")
 """
-This is the redirect uri registered with the app which handles authorization for a Flyte deployment.
+This is the callback uri registered with the app which handles authorization for a Flyte deployment.
+Please note the hardcoded port number. Ideally we would not do this, but some IDPs do not allow wildcards for
```
katrogan: Haytham suggested that we implement logic to either accept a wildcard port or a list of ports (i.e. if we register a bunch of hard-coded ports with the IDP.) That way we have *some* flexibility with choosing an available system port. I'm happy to save that for the next PR though and leave this comment in for now

flytekit/configuration/platform.py
wild-endeavor: it would be configured.  i think configuring is cleaner than trying without and then doing it automatically.  if admin returns an authorization error, I'd rather print a nice error message asking the user to enable auth.

flytekit/clients/raw.py
```diff
@@ -1,17 +1,46 @@
...
+            except Exception as e:
+                if attempt >= 3:
+                    raise e
+                else:
+                    print('retrying')
+    return handler
 
 
 def _handle_rpc_error(fn):
+    @_try_three_times
```
katrogan: hey @matthewphsmith we introduced this to make sure we don't get stuck in a loop if authentication is persistently failing...but i'm not sure that this is the prettiest code 🙃 let us know if a better way stands out

katrogan: the code's been refactored but i think the original comment still stands

flytekit/clients/raw.py
```diff
@@ -1,15 +1,94 @@
...
+
+
+def _refresh_credentials_standard(flyte_client):
+    """
+    This function is used when the configuration value for AUTH_MODE is set to 'standard'.
+    This either fetches the existing access token or initiates the flow to request a valid access token and store it.
+    :param flyte_client: RawSynchronousFlyteClient
+    :return:
+    """
+    if not _platform_config.AUTH.get():
```
wild-endeavor: but if not set, and admin returns a 401, this function won't do anything.  We still want to attempt to get credentials even if this is not set.  The point of this option has been coopted to mean auth will be tried regardless... that is, even if admin doesn't return a 401, auth will be attempted.

flytekit/clis/auth/auth.py
```diff
@@ -0,0 +1,255 @@
...
+    def access_token(self):
+        return self._access_token
+
+    @property
+    def id_token(self):
+        return self._id_token
+
+
+# TODO:
+#  do we need to support initiate login URI? https://devforum.okta.com/t/initiate-login-uri-for-all-subdomain-urls/3766
```
wild-endeavor: i think we can remove this.

wild-endeavor:
+1 from me aside from comments.  let's wait til matt gets back and make him look at this as his first priority.

wild-endeavor:
and conflict resolution ofc

flytekit/clients/raw.py
katrogan: ah sorry i didn't follow the original logic - i can bump this block back to where it was

flytekit/clis/sdk_in_container/basic_auth.py
```diff
@@ -0,0 +1,83 @@
...
+from flytekit.common.exceptions.base import FlyteException as _FlyteException
+from flytekit.configuration.creds import (
+    CLIENT_CREDENTIALS_SECRET_LOCATION as _CREDENTIALS_SECRET_FILE,
+    CLIENT_CREDENTIALS_SECRET as _CREDENTIALS_SECRET,
+)
+
+_utf_8 = 'utf-8'
+
+
+class FlyteAuthenticationException(_FlyteException):
```
matthewphsmith: NIt: should we define this in flyte.exceptions.user?

flytekit/clis/flyte_cli/main.py
```diff
@@ -30,11 +32,30 @@
 
 _tt = _six.text_type
 
+# Similar to how kubectl has a config file in the users home directory, this Flyte CLI will also look for one.
+# The format of this config file is the same as a workflow's config file, except that the relevant fields are different.
+# Please see the example.config file
+_default_config_file_path = ".flyte/config"
```
matthewphsmith: I think this is already set as ‘~/.Flyte’ in the configuration.internal. Perhaps use that, but I’d also support changing if to ‘~.flyte/config’

requirements.txt
```diff
@@ -1,3 +1,4 @@
 pytest==4.6.6
 mock==3.0.5
 six==1.12.0
+responses==0.10.7
```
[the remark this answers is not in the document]
wild-endeavor: btw did we add keyring to setup.py?  @matthewphsmith how should we handle that?  @katrogan can fill you in in person.

flytekit/clis/flyte_cli/example.config
```diff
@@ -0,0 +1,11 @@
+# This is an example of what a config file may look like.
+# Place this in ~/.flyte/config and flyte-cli will pick it up automatically
+
+[platform]
+url=flyte.company.com
+auth=true
+
+[credentials]
+discovery_endpoint=http://corp.idp.com/.well-known/oauth-authorization-server
```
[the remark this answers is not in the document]
wild-endeavor: Just to clarify, this discovery endpoint should be pointed at `flyte.corp.net` or something... and that should redirect to whatever IDP corp.net is using.  Or it can return the values itself without redirecting, but the point is, you only hard-code the main flyte server.  Thinking about it, maybe we should nix this @katrogan and just assume the extension off of the platform URL.  That's probably the cleanest thing to do.

The redirect endpoint is configured in the IDP.  It's a hardcoded fully-specified URL.  This means that the callback url (including the port) that the local server that's stood up at has to match exactly what the IDP has.  This information is not returned by the metadata/discovery endpoint.  Why i'm not sure... it may be a security concern.

Ideally, the metadata endpoint would be less restrictive and allow us to return more things.  Because of the limitations of the RFC, we also need to hardcode this in for example: https://github.com/lyft/flyteexamples/pull/97/files#diff-2e41be19539d0ba6fda0e5ce814a9123R22.  This is the scope that will be requested by pyflyte when it makes the token request.  Again, like the callback url, it has to match a hardcoded string set within the IDP, otherwise it'll fail.  It doesn't really offer anything, but we have to have it.

requirements.txt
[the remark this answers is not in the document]
katrogan: I'll move both and add to setup.py

flytekit/clis/flyte_cli/example.config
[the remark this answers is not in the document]
katrogan: Re: assuming the extension - I'm happy to do this if we can safely assume this is the major convention (which I guess sounds like it is?)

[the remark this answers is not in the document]
wild-endeavor: yup https://tools.ietf.org/html/rfc8414#section-3

flytekit/clis/flyte_cli/main.py
katrogan: Looks like it's just flytekit.config https://github.com/lyft/flytekit/blob/d4cfedc4c580f08bf904e6e474a0b948a4608737/flytekit/configuration/internal.py#L13 - do we want to change the default as part of this change? i'm worried about unrolling out too much one in go (and breaking stuff in the process)

flytekit/clis/sdk_in_container/basic_auth.py
katrogan: @wild-endeavor can we use https://github.com/lyft/flytekit/pull/23/files#diff-7fbb98dd27586954c518f84994ed987cR77?

flytekit/clis/sdk_in_container/basic_auth.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    if _CREDENTIALS_SECRET_FILE.get():
+        return get_file_contents(_CREDENTIALS_SECRET_FILE.get())
+    elif _CREDENTIALS_SECRET.get():
+        return _CREDENTIALS_SECRET.get()
+    raise FlyteAuthenticationException('No secret could be found in either {} or the {} env variable'.format(
+        _CREDENTIALS_SECRET_FILE.get(), _CREDENTIALS_SECRET.env_var))
+
+
+def get_basic_authorization_header(client_id, client_secret):
```
katrogan: @wild-endeavor is this used anywhere outside fo tests?

flytekit/clients/raw.py
```diff
@@ -1,15 +1,99 @@
...
 from grpc import insecure_channel as _insecure_channel, secure_channel as _secure_channel, RpcError as _RpcError, \
     StatusCode as _GrpcStatusCode, ssl_channel_credentials as _ssl_channel_credentials
 from flyteidl.service import admin_pb2_grpc as _admin_service
 from flytekit.common.exceptions import user as _user_exceptions
+from flytekit.configuration.creds import (
+    CLIENT_ID as _CLIENT_ID,
+    CLIENT_CREDENTIALS_SCOPE as _SCOPE,
+)
+from flytekit.clis.sdk_in_container import basic_auth
+import logging
```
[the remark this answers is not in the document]
katrogan: done

flytekit/clis/flyte_cli/example.config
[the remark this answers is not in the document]
matthewphsmith: Hmm very interesting...well seems like you have a handle on it, so I defer to whatever you think is best though I am biased to less config wherever possible.

flytekit/clis/flyte_cli/main.py
matthewphsmith: Yeah, I think that's probably best. Looking at the way the code is written below, I think it's good actually. It will fall back to `flytekit.config` when `~/.flyte/config` doesn't exist so there isn't any breaking change (and if that isn't found, which i think 99.99% likely it won't be found, it will go to env vars).

flytekit/clis/sdk_in_container/basic_auth.py
```diff
@@ -0,0 +1,64 @@
...
+        'Authorization': authorization_header,
+        'Cache-Control': 'no-cache',
+        'Accept': 'application/json',
+        'Content-Type': 'application/x-www-form-urlencoded'
+    }
+    body = {
+        'grant_type': 'client_credentials',
+        'scope': scope,
+    }
+    response = _requests.post(token_endpoint, data=body, headers=headers)
```
jonathanburns: I think you want to use "json" instead of "data" here, or else the None scope will get translated to literally "None"

flytekit/clients/raw.py
```diff
@@ -1,15 +1,90 @@
...
+    This function is used by the _handle_rpc_error decorator, depending on the AUTH_MODE config object. This handler
+    is meant for SDK use-cases of auth (like pyflyte, or when users call SDK functions that require access to Admin,
+    like when waiting for another workflow to complete from within a task). This function uses basic auth, which means
+    the credentials for basic auth must be present from wherever this code is running.
+
+    :param flyte_client: RawSynchronousFlyteClient
+    :return:
+    """
+    auth_endpoints = _credentials_access.get_authorization_endpoints()
+    token_endpoint = auth_endpoints.token_endpoint
```
jonathanburns: Guessing you'll have to do something here to support relative URLs.

wild-endeavor: Not sure what you mean?

jonathanburns: if the token endpoint is "/token", the code doesn't work, I think

jonathanburns:
Any chance you can ask for password over stdin if password is required but not password is passed? Something like this:

https://github.com/lyft/flytekit/pull/56/files#diff-39ff0a124da92894da1f07bd5ce13c67R46-R47

PR #25, opened 2019-10-23
Serialization and registration for workflow closures

kumare3:
PROPOSAL: Split serialization and registration into 2 steps.

Serialization of a workflow results in a closure file. This can be registered independently without access to the code. This makes the workflow definitions more portable and splits the pyflyte monolith registration into two steps.

Eventually we feel that the registration step is always part of flyte code agnostic cli but serialization is implemented per language (python, go etc) or flavors (jupyter notebook, flytekit, other python sdks) of SDK.


flytekit/clis/flyte_cli/main.py
```diff
@@ -1442,5 +1443,43 @@ def register_project(identifier, name, host, insecure):
...
+    Register a new workflow from a serialized workflow-closure protobuf specified in
+    \b
+      https://github.com/lyft/flyteidl/blob/master/protos/flyteidl/core/workflow_closure.proto
+    \b
+      Argument: `closure` Path to the Workflow closure protobuf file that should be registered with flyteadmin
+
+    """
+    _welcome_message()
+
+    c = _workflow_closure_pb2.WorkflowClosure()
```
kumare3: i think a problem here is the order of registration

matthewphsmith: Like what order?

It looks like this script can only register tasks and workflows.  All tasks are registered first and tasks can't depend on one another so that's fine.  Then the workflow is registered, so the order should be fine.  If you start adding in launch plans and stuff, that is where it will get more confusing in terms of ordering.

flytekit/clis/flyte_cli/main.py
```diff
@@ -1442,5 +1443,43 @@ def register_project(identifier, name, host, insecure):
...
+    with _platform_config.URL.get_patcher(host), _platform_config.INSECURE.get_patcher(_tt(insecure)):
+        for t in wc.tasks:
+            _click.echo(t.short_string())
+            tk = _task.SdkTask.promote_from_model(t)
+            _click.echo("Registering task... {}:{}:{}:{}".format(t.id.project, t.id.domain, t.id.name, t.id.version))
+            tk.register(t.id.project, t.id.domain, t.id.name, t.id.version)
+        _click.echo("All tasks registered, building workflow...")
+
+        w = wc.workflow
+        _click.echo(w)
```
matthewphsmith: nit: too much print

flytekit/common/workflow.py
```diff
@@ -182,7 +187,8 @@ def promote_from_model(cls, base_model):
                 [],
                 n.inputs,
                 n.metadata,
-                sdk_task=None,
+                sdk_task=_SdkTask.fetch(n.task_node.reference_id.project, n.task_node.reference_id.domain,
```
matthewphsmith: Can you put a TODO that we should cache the fetches? That way we don't overload the service on workflows that have a lot of replicated tasks.

flytekit/clis/flyte_cli/main.py
kumare3: ya, debugging, will remove

flytekit/common/workflow.py
kumare3: ahh good point, we can create a loadingcache / readhtrough cache

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -33,12 +34,24 @@ def serialize_workflows(pkgs):
         tmap[t.id] = t
 
     for m, k, w in iterate_registerable_entities_in_order(pkgs, include_entities={_workflow.SdkWorkflow}):
-        click.echo('Serializing {}'.format(_utils.fqdn(m.__name__, k, entity_type=w.resource_type)))
+        name = _utils.fqdn(m.__name__, k, entity_type=w.resource_type)
+        click.echo('Serializing {}'.format(name))
         task_templates = []
         for n in w.nodes:
             if n.task_node is not None:
                 task_templates.append(tmap[n.task_node.reference_id])
```
wild-endeavor: not this pr, but this isn't correct right?  @matthewphsmith?  there can be duplicates

flytekit/common/workflow.py
```diff
@@ -176,13 +181,16 @@ def promote_from_model(cls, base_model):
...
+        # TODO: To save on task fetches (for workflows with a lot of repeated tasks), we should cache the results of
+        #  the fetch, just in-process
         node_map = {
             n.id: _nodes.SdkNode(
                 n.id,
                 [],
                 n.inputs,
                 n.metadata,
-                sdk_task=None,
+                sdk_task=_task.SdkTask.fetch(n.task_node.reference_id.project, n.task_node.reference_id.domain,
```
matthewphsmith: should only be called if task_node is set.

flytekit/common/workflow.py
```diff
@@ -192,12 +200,16 @@ def promote_from_model(cls, base_model):
         for v in _six.itervalues(node_map):
             v.upstream_nodes[:] = [node_map[k] for k in v.upstream_node_ids]
 
+        iface = _interface.TypedInterface.promote_from_model(base_model.interface)
+        inputs = [_Input(name=k, sdk_type=_type_helpers.get_sdk_type_from_literal_type(v.type), required=True) for k, v
+                  in _six.iteritems(iface.inputs)]
+        outputs = [Output(name=o.var, value=None, binding_data=o.binding,
```
matthewphsmith: perhaps we can use the above created node map to simply create these as normal `Output(node_map[o.binding.node_id].outputs[o.binding.var], sdk_type=...)` and not alter the constructor (see above comment).

flytekit/common/workflow.py
```diff
@@ -8,19 +8,21 @@
...
-from flytekit.models import interface as _interface_models, literals as _literal_models,\
+from flytekit.models import interface as _interface_models, literals as _literal_models, \
     launch_plan as _launch_plan_models
 from flytekit.models.core import workflow as _workflow_models, identifier as _identifier_model
 
 
 class Output(object):
 
-    def __init__(self, name, value, sdk_type=None, help=None):
+    def __init__(self, name, value, sdk_type=None, binding_data=None, help=None):
```
matthewphsmith: I think we can avoid changing the definition of this constructor to have divergent. See comment below. If not, we should place`_binding_data`  after `help` and precede with an underscore since it's a private constructor behavior.

PR #26, opened 2019-10-23
Support for Notebook Tasks

akhurana001:
Adds support for basic Python and Spark Notebooks to be consumed as Flyte tasks.


flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+    bool: _p.Boolean,
+    float: _p.Float,
+    str: _p.String,
+    _datetime.datetime: _p.Datetime,
+    _datetime.timedelta: _p.Timedelta,
+}
+
+OUTPUT_NOTEBOOK = 'output_notebook'
+
+def convert_outputs(outputs=None):
```
kumare3: _

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+            args=[],
+            resources=_task_models.Resources(limits=limits, requests=requests),
+            env=environment,
+            config={}
+        )
+
+
+
+
+def spark_notebook(
```
kumare3: add an optional conf

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+from flytekit.models import literals as _literal_models, task as _task_models
+from google.protobuf.json_format import MessageToDict as _MessageToDict
+from flytekit.sdk.types import Types
+
+from flytekit.common.types.helpers import pack_python_std_map_to_literal_map as _packer
+from flytekit.common.types import primitives as _p
+from google.protobuf import text_format
+from flytekit.common import sdk_bases
+from flytekit.common.tasks import task as base_tasks
+from flytekit.engines import loader as _engine_loader
```
matthewphsmith: nit: for all imports, let's import the module itself and not the leaf attribute. it makes it easier to refactor things.  

2. If possible, use commas if they have the same package.  example:

```
from flytekit.common.tasks import sdk_runnable as _sdk_runnable, spark_task as _spark_task
```

3.  Always prefix anything meant to be private to this package with underscore to prevent it leaking into other namespaces.

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+        tm[k] = type_map[t]
+    return _packer(outputs, tm)
+
+
+def record_outputs(outputs=None):
+    p = convert_outputs(outputs)
+    return p.to_flyte_idl()
+
+# TODO: Support Client Mode
+def get_spark_context(spark_conf):
```
kumare3: comments

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+    for k, v in _six.iteritems(outputs):
+        t = type(v)
+        if t not in type_map:
+            raise ValueError(
+                "Currently only primitive types {} are supported for recording from notebook".format(type_map))
+        tm[k] = type_map[t]
+    return _packer(outputs, tm)
+
+
+def record_outputs(outputs=None):
```
kumare3: comments

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+    """
+    outputs: dict
+    Records the output
+    """
+    if outputs is None:
+        return _packer({}, {})
+    tm = {}
+    for k, v in _six.iteritems(outputs):
+        t = type(v)
+        if t not in type_map:
```
matthewphsmith: I think this is unnecessarily restrictive--the SDK's type engine can convert a rich set of object types into literals. And we can rely on that logic to raise an exception if the cast is not possible.

flytekit/common/tasks/notebook.py
matthewphsmith: Should this be upstreamed to spark task? Wouldn't that suffer from the same issue.

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+            gpu_limit,
+            memory_limit,
+            discoverable,
+            timeout,
+            environment,
+            custom
+    ):
+
+        # Add output_notebook as an implicit output to the task.
+        outputs[OUTPUT_NOTEBOOK] = Types.Blob
```
kumare3: check None

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+
+        self._notebook_path = notebook_path
+        super(SdkNotebookTask, self).__init__(
+            task_type,
+            _task_models.TaskMetadata(
+                discoverable,
+                _task_models.RuntimeMetadata(
+                    _task_models.RuntimeMetadata.RuntimeType.FLYTE_SDK,
+                    __version__,
+                    'python'
```
kumare3: python-notebook

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+            gpu_limit=gpu_limit,
+            memory_limit=memory_limit,
+            discoverable=cache,
+            timeout=timeout or _datetime.timedelta(seconds=0),
+            environment=environment,
+            custom={})
+
+
+class SdkNotebookTask(
+        _six.with_metaclass(sdk_bases.ExtendedSdkType, base_tasks.SdkTask)):
```
matthewphsmith: Can we make this derive from SdkRunnableTask?  If not, we should massage SdkRunnableTask to allow this use-case to work.

flytekit/common/tasks/notebook.py
matthewphsmith: We should verify there is no name collision.

Also, lets make a TODO to introduce a Notebook type that derives from Blob so we can eventually know to render things nicely in the UI/etc.

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+            environment=None,
+    ):
+
+        spark_exec_path = _os.path.abspath(spark_executor.__file__)
+        if spark_exec_path.endswith('.pyc'):
+            spark_exec_path = spark_exec_path[:-1]
+        self._notebook_path = notebook_path
+
+        # Parse Spark_conf from notebook
+        with open(notebook_path) as json_file:
```
kumare3: we should try and patch the notebook path by first checking if it is relative or absolute
https://stackoverflow.com/questions/3320406/how-to-check-if-a-path-is-absolute-path-or-relative-path-in-cross-platform-way-w

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+            working directory (with the names provided), which will in turn allow Flyte Propeller to push along the
+            workflow.  Where as local engine will merely feed the outputs directly into the next node.
+        """
+        inputs_dict = _type_helpers.unpack_literal_map_to_sdk_python_std(inputs, {
+            k: _type_helpers.get_sdk_type_from_literal_type(v.type) for k, v in _six.iteritems(self.interface.inputs)
+        })
+
+        # Execute Notebook via Papermill.
+        input_notebook_path = self._notebook_path
+        output_notebook_path = input_notebook_path[:len(input_notebook_path) - 6] + '-out' + input_notebook_path[len(
```
matthewphsmith: can we guarantee this `- 6`? what does it represent?

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+
+            spark_job = _task_models.SparkJob(
+                spark_conf=spark_conf,
+                hadoop_conf={},
+                application_file="local://" + spark_exec_path,
+                executor_path=_sys.executable,
+            ).to_flyte_idl()
+
+        # Add output_notebook as an implicit output to the task.
+        outputs[OUTPUT_NOTEBOOK] = Types.Blob
```
kumare3: remove it

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+
+        input_variables = {k: Variable(v.to_flyte_literal_type(), k) for k, v in _six.iteritems(inputs)}
```
kumare3: remove

tests/flytekit/common/workflows/notebook.py
```diff
@@ -0,0 +1,26 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from flytekit.sdk.types import Types
```
matthewphsmith: Unit tests too plz

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+        retries=0,
+        deprecated='',
+        cache=False,
+        timeout=None,
+        notebook_path=None,
+        inputs=None,
+        outputs=None,
+        environment=None,
+):
+    """
```
kumare3: lets add what is to be required and how to use it

For example, mark cell as conf, have a variable called spark_conf

kumare3:
Linter fix and couple comments. Please add some docs. This looks awesome 👍

flytekit/common/tasks/notebook.py
matthewphsmith: i presume `.ipynb`? will ipython run a notebook i give named `mynotbook.notebook`? if yes, we should probably not make this assumption about path and instead just append `.flyteout`.  alternatively, we could explicitly enforce that it must end in `.ipynb`--in which . case `.ipynb` should be a constant.

matthewphsmith:
I think a meta-comment I have is that we shouldn't need to duplicate so much code for this plugin. However, with the current state of SDK, it might be necessary and then it is an AI on me to refactor things so you can trivially implement notebook behavior (perhaps as a mixin).

So let's chat and get this into the best state we can to reduce duplicate code based on the current state of things.  And then make a TODO for me to simplify--including broader changes needed to support that.

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
+from __future__ import absolute_import
```
matthewphsmith: can we put this in `flytekit.contrib.notebooks` or similar as an example of a 'contrib'?  Also, that can buy us some flexibility in terms of the simplification work I want to take on to make this easier to accomplish since it won't have the same guarantees of our 1st class tasks. Once stabilized, we can upstream into the 1st class package. @kumare3 what do you think?

flytekit/common/tasks/notebook.py
```diff
@@ -0,0 +1,542 @@
...
+
+    @property
+    def container(self):
+        """
+        If not None, the target of execution should be a container.
+        :rtype: Container
+        """
+
+        # Find task_name
+        task_module = _importlib.import_module(self.instantiated_in)
```
matthewphsmith: Can we use this to resolve relative paths too?

`os.path.join(os.path.dirname(task_module.__file__), user_specified_path)`

flytekit/common/tasks/notebook.py
matthewphsmith: Also, I think we want to use the SDK type engine logic because it is designed to be extensible and allow people to introduce their own type behavior. So doing it this way would block that for notebook tasks.

flytekit/common/tasks/notebook.py
akhurana001: removed

flytekit/common/tasks/notebook.py
akhurana001: done

flytekit/common/tasks/notebook.py
akhurana001: done, I might have missed some but hopefully got them all.

flytekit/common/tasks/notebook.py
akhurana001: done

flytekit/common/tasks/notebook.py
akhurana001: done

flytekit/common/tasks/notebook.py
akhurana001: no, this is only for notebook_spark_tasks

flytekit/common/tasks/notebook.py
akhurana001: defaulted to {}

flytekit/common/tasks/notebook.py
akhurana001: done

flytekit/common/tasks/notebook.py
akhurana001: It is quite a bit of work and would involve most of SdkRunnableTask methods to be overridden. leaving as-is based on our discussion

flytekit/common/tasks/notebook.py
akhurana001: done

flytekit/common/tasks/notebook.py
akhurana001: done

flytekit/common/tasks/notebook.py
akhurana001: replaced with just appending an `.out` . works fine

flytekit/common/tasks/notebook.py
akhurana001: removed

flytekit/common/tasks/notebook.py
akhurana001: removed

flytekit/common/tasks/notebook.py
akhurana001: moved

flytekit/common/tasks/notebook.py
akhurana001: done

notebooks/spark-notebook-pi.ipynb
```diff
@@ -0,0 +1,110 @@
...
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": [
+     "outputs"
+    ]
+   },
+   "outputs": [],
+   "source": [
+    "from flytekit.common.tasks.notebook  import record_outputs\n",
```
bnsblue: flytekit.contrib.notebook

PR #30, opened 2019-10-25
initial commit for SensorTask

datability-io:


datability-io:
🔨

flytekit/contrib/task.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self, **kwargs):
+        super(SensorTask, self).__init__(**kwargs)
+
+    def _execute_user_code(self, context, inputs):
+        sensor = super()._execute_user_code(context=context, inputs=inputs)
+        if sensor is not None:
+            if not isinstance(sensor, _Sensor):
+                raise _user_exceptions.FlyteTypeException(
+                    received_type=type(sensor),
+                    expected_type=type(_Sensor),
```
matthewphsmith: I think this should be `expected_type=_Sensor`

flytekit/contrib/task.py
```diff
@@ -0,0 +1,60 @@
...
+                )
+            succeeded = sensor.sense()
+            if not succeeded:
+                raise _user_exceptions.FlyteRecoverableException()
+
+
+def sensor_task(
+    _task_function=None,
```
matthewphsmith: nit: For consistency, let's not set non-empty/non-zero defaults in code.  And since this really is just a special case of a python task, we should support all the same kwargs (as shown here: https://github.com/lyft/flytekit/blob/master/flytekit/sdk/tasks.py#L114-L129)

Also, let's give a descriptive docstring describing how it works.

flytekit/contrib/task.py
```diff
@@ -0,0 +1,60 @@
...
+from flytekit.contrib.base_sensor import Sensor as _Sensor
+
+
+class SensorTask(_sdk_runnable.SdkRunnableTask):
+
+    def __init__(self, **kwargs):
+        super(SensorTask, self).__init__(**kwargs)
+
+    def _execute_user_code(self, context, inputs):
+        sensor = super()._execute_user_code(context=context, inputs=inputs)
```
matthewphsmith: nit: `super(SensorTask,  self)`

PR #35, opened 2019-10-29
Fix tests and instructions in README

slai:
I just went through the README to get myself setup for this repo and found a few issues. This PR fixes -

* ~requirements.txt so it contains all the packages needed to run the tests (presumably this file is only used for dev as it previously had pytest in it)~ looks like this was fixed recently
* tests so they assert exception messages properly. _Looks like this was changed between pytest 4 and pytest 5._ The repo now pins to pytest 4, so these tests currently work, although this change is backwards compatible
* README to reference `flytekit` instead of `flytekit2`

---

I should've gotten the latest changes before creating this PR :) The remaining changes are pretty minor, so don't mind if we close this instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,14 +69,14 @@ Flytekit is Python 2.7+ compatible, so when feasible, it is recommended to test 
 
 #### Setup (Do Once)
 ```bash
-virtualenv ~/.virtualenvs/flytekit2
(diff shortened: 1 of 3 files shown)


slai:
@matthewphsmith TIL about GitHub's stale review dismissal feature while rebasing. Can I get another green tick?

PR #37, opened 2019-10-29
Implement the register --test arg

slai:
This implements the `--test` functionality for the register subcommand. As it is now, the `--test` flag does nothing and workflows and tasks will be registered regardless of whether it is set.

I know there's some discussion about the usefulness of `--test` at all as full validation only happens server-side. I think there's still value to catch simple errors like missing imports, and also shows what Flyte has detected (to catch bad module paths, sys.path or missing `__init__.py` issues). Might need a better name though for clarity.

I also fixed a bug with a test mock - it mocks out flytekit.tools.module_loader.iterate_modules, which takes a sequence of names, but the mock only took in a single module name. It worked because there was also a mistake when specifying the value in the context.


slai:
@matthewphsmith sorry, had to rebase this one because it was stacked.

PR #40, opened 2019-11-04
Fix upload content instead of directory

tnsetting:


flytekit/interfaces/data/gcs/gcs_proxy.py
```diff
@@ -95,7 +95,7 @@ def upload_directory(self, local_path, remote_path):
             raise ValueError("Not an GS Key. Please use FQN (GS ARN) of the format gs://...")
 
         GCSProxy._check_binary()
-        cmd = [GCSProxy._GS_UTIL_CLI, "cp", "-r", local_path, remote_path]
+        cmd = [GCSProxy._GS_UTIL_CLI, "cp", "-r", local_path + "/*", remote_path]
```
matthewphsmith: Perhaps we should use `_os.path.join(local_path, '*')` just so we don't have to worry about weird edgecases around path format?

honnix:
Sorry I ended up working on https://github.com/lyft/flytekit/pull/41 fixing the same issue. :(

PR #44, opened 2019-11-07
Ensure Spark Entrypoint Exits Cleanly

matthewphsmith:


scripts/flytekit_spark_entrypoint.sh
```diff
@@ -42,10 +42,12 @@ case "$SPARK_K8S_CMD" in
       shift 1
       ;;
     "")
+      exit
```
kumare3: We should probably echo that received empty command?

PR #46, opened 2019-11-08
Fix Tool to Allow Partial Activation of Launch Plans

matthewphsmith:


flytekit/tools/module_loader.py
```diff
@@ -90,7 +95,8 @@ def iterate_registerable_entities_in_order(pkgs, ignore_entities=None, include_e
         entities will be taken.  Only one of ignore_entities or include_entities can be set.
     :param set[type] include_entities: If specified, include these entities while doing a topological sort.  All
         other entities will be ignored.  Only one of ignore_entities or include_entities can be set.
-    :param bool only_include_defs: If false, we will possibly return duplicate references to an entity.
```
chanadian: why is this removed too?

matthewphsmith: I changed the name and doc string

PR #49, opened 2019-11-12
Remove hidden arg from register --pkgs option

slai:
The `hidden=True` arg on `@click.option` was only introduced in Click 7.0. Bumping flytekit up to this version led to an incompatibility with lyftlearnclient, a package commonly installed with flytekit, which requires v6.x.

Ideally lyftlearnclient would be updated to use Click 7.x, but I don't have the time to learn another package, so replacing the `hidden=True` arg with an angry `DEPRECATED` message is the next best option. It was only there for cosmetic reasons, so this doesn't affect the reason for the original change.

Slack thread - https://lyft.slack.com/archives/C3BLDHN92/p1573498574188800


matthewphsmith:
Thanks for circling back on this.

PR #50, opened 2019-11-13
Add missing format call on exception message

slai:
Missing the output name in the message, making the message pretty useless :)

```
/usr/local/lib/python3.6/dist-packages/flytekit/common/tasks/sdk_runnable.py:495: in _validate_outputs
    ???
E   flytekit.common.exceptions.user.FlyteValidationException: The output named '{}' was not specified in the task function.  Therefore, this output cannot be provided to the task.
```


EngHabu:
Thank you!!

PR #54, opened 2019-11-15
Propagating error code instead of unknown

tnsetting:
When we tried to raise our exception inherited from `FlyteUserException`, the error code is not propagated to the task execution as we always get `USER:unknown`. The PR is to address the issue.


wild-endeavor:
@tnsetting does the squash and merge button not show up for you?

tnsetting:
> e button not show up for you?

Since it was merged, I didn't get the chance to see the `squash and merge button`. I can check if it is there next time.

PR #55, opened 2019-11-23
Ability to customize image name

wild-endeavor:
This allows an override so that when we try to build images for git repos with multiple container images, we can customize the name.

### DIFF
--- a/scripts/flytekit_build_image.sh
+++ b/scripts/flytekit_build_image.sh
@@ -19,9 +19,11 @@ echo ""
 # Go into the directory representing the user's repo
 pushd $1
 
-# Grab the repo name from the argument
+# Grab the repo name from the argument if not already defined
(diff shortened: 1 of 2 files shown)


kumare3:
cool

scripts/flytekit_build_image.sh
```diff
@@ -19,9 +19,11 @@ echo ""
 # Go into the directory representing the user's repo
 pushd $1
 
-# Grab the repo name from the argument
+# Grab the repo name from the argument if not already defined
 # Note that this repo name will be the name of the Docker image.
-IMAGE_NAME=${PWD##*/}
+if [ -z "IMAGE_NAME" ]; then
```
honnix: Is this supposed to be `-z "${IMAGE_NAME}"`?

wild-endeavor: oops

PR #58, opened 2019-12-03
fetch latest task

honnix:
The use case is alway fetching and running the latest task.

@kumare3 wdty?

### DIFF
--- a/flytekit/engines/flyte/engine.py
+++ b/flytekit/engines/flyte/engine.py
@@ -108,16 +108,16 @@ class FlyteEngineFactory(_common_engine.BaseExecutionEngineFactory):
             insecure=_platform_config.INSECURE.get()
         ).client.get_task(task_id)
 
-    def fetch_latest_task(self, task):
+    def fetch_latest_task(self, named_task):
         """
-        Fetches latest tasks
-        :param flytekit.models.common.NamedEntityIdentifier task: NamedEntityIdentifier to fetch
+        Fetches the latest task
+        :param flytekit.models.common.NamedEntityIdentifier named_task: NamedEntityIdentifier to fetch
         :rtype: flytekit.models.task.Task
         """
         task_list, _ = _FlyteClientManager(
             _platform_config.URL.get(), insecure=_platform_config.INSECURE.get()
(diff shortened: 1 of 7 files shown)


flytekit/engines/flyte/engine.py
```diff
@@ -107,6 +108,21 @@ def fetch_task(self, task_id):
             insecure=_platform_config.INSECURE.get()
         ).client.get_task(task_id)
 
+    def fetch_latest_task(self, named_task):
```
honnix: Would it be better we put this logic into `fetch_task`? E.g. if there is no `version` specified, we fetch the latest one.

honnix:
@kumare3 PTAL again. Thanks.

flytekit/engines/flyte/engine.py
honnix: Or should we at least stick to `Identifier` instead of `NamedEntityIdentifier` at this layer? Sort of like what `fetch_launch_plan` does.

matthewphsmith: I think your way is correct just to prevent accidental problematic patterns from emerging. Launch plan has the special case because there is a concept of 'active' which is basically a way of indicating what is the currently 'deployed entrypoint' version to a workflow. This makes it easy to indicate which schedule should be active, and write command line tools to kick off workflows.

Tasks don't currently have the active concept and fetch is instead used to create workflows. Creating a workflow with a loose reference _can_ be problematic if you're not prepared for what that entails. Because then registering a workflow isn't always idempotent--if the task version received on different attempts at registration of the workflow, the registration will fail.

So this is pretty explicit and I think it is good for now. If we want to elevate tasks to have a concept like 'active', this change doesn't preclude that and we can do it later--but that is a larger change.

honnix: Thanks for the detailed explanation. I totally agree.

NatashaL:
👍

honnix:
@matthewphsmith Could you please help us merge it? Thanks.

honnix:
@matthewphsmith ping

PR #60, opened 2019-12-05
Remove `setup_requires` of pip

asottile:
This slows down installation and is too late to have an effect on the actual package build


asottile:
> Should this be enforced using docs only?

as far as I can tell that range isn't even necessary -- it works fine with newer and older pip (I only tried pip 9.x)

PR #62, opened 2019-12-05
Env var (from pkce-auth pr) option 1

wild-endeavor:


test_config.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python
```
matthewphsmith: Can we put this test in the `tests.flytekit.unit.configuration` package and also make it assert something. Just for good measure.

PR #64, opened 2019-12-05
Env var (from pkce-auth pr) option 2

wild-endeavor:


test_config.py
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env python
+
+from flytekit.configuration import creds
+
+secret = creds.CLIENT_CREDENTIALS_SECRET.get()
+print("Secret is {}".format(secret))
```
matthewphsmith: can we make this a real test? And shove it in the `tests.unit.flytekit.configuration`

PR #67, opened 2019-12-09
try asking for password over stdin if not provided

jonathanburns:


matthewphsmith:
Since this CLI is called from inside a container, presumably by calling `docker run` on it. Will this work in that case?

The original plan was to have yet another CLI which sits outside the container which either calls `docker run` and could pass through credentials _somehow_ or just causes the container to dump it's Flyte spec to a file or stdout, which is then forwarded to the Admin service from outside the container. Just to give some context on how this internal CLI was intended for use.

PR #68, opened 2019-12-10
Changes to Auth configurations

wild-endeavor:
I realized while testing things a bit more locally that just having one endpoint was insufficient - flyte-cli and pyflyte currently leverage the gRPC Python Flyte Admin client.  However, discovery needs the HTTP endpoint.  Even though in production both are served on the same endpoint and traffic is handled through ingress, this is not always the case, and we want to support the use-case where gRPC Admin and HTTP Admin are hosted separately.

We were debating just adding a separate config for discovery alone, but ultimately it makes more sense to split out the existing URL config into two objects, especially if the above pattern is something we want to continue to support.  However, in the interest of not making people migrate, and not writing migration code in this PR, we're just introducing the HTTP one for now.

Also, we should lowercase the metadata header before sending it to Admin in the gRPC calls.

### DIFF
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -129,7 +129,8 @@ class RawSynchronousFlyteClient(object):
             self.force_auth_flow()
 
     def set_access_token(self, access_token):
-        # Always set the header to lower-case regardless of what the config is.
+        # Always set the header to lower-case regardless of what the config is. The grpc libraries that Admin uses
+        # to parse the metadata don't change the metadata, but they do automatically lower the key you're looking for.
         self._metadata = [(_creds_config.AUTHORIZATION_METADATA_KEY.get().lower(), "Bearer {}".format(access_token))]
 
     def force_auth_flow(self):
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -129,7 +129,8 @@ class RawSynchronousFlyteClient(object):
             self.force_auth_flow()
 
     def set_access_token(self, access_token):
-        self._metadata = [(_creds_config.AUTHORIZATION_METADATA_KEY.get(), "Bearer {}".format(access_token))]
+        # Always set the header to lower-case regardless of what the config is.
+        self._metadata = [(_creds_config.AUTHORIZATION_METADATA_KEY.get().lower(), "Bearer {}".format(access_token))]
 
     def force_auth_flow(self):
         refresh_handler_fn = _get_refresh_handler(_creds_config.AUTH_MODE.get())
--- a/flytekit/clis/auth/credentials.py
+++ b/flytekit/clis/auth/credentials.py
@@ -1,12 +1,15 @@
 from __future__ import absolute_import
+
+import logging
+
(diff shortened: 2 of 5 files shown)


wild-endeavor:
ptal @matthewphsmith @EngHabu @katrogan

flytekit/clis/auth/credentials.py
```diff
@@ -1,12 +1,15 @@
 from __future__ import absolute_import
+
+import logging
```
katrogan: i'm guessing matt wants `as _logging`?

wild-endeavor: @matthewphsmith katrina is calling you out on your pattern.

matthewphsmith: indeed he does ;)

flytekit/clients/raw.py
```diff
@@ -129,7 +129,8 @@ def __init__(self, url, insecure=False, credentials=None, options=None):
             self.force_auth_flow()
 
     def set_access_token(self, access_token):
-        self._metadata = [(_creds_config.AUTHORIZATION_METADATA_KEY.get(), "Bearer {}".format(access_token))]
+        # Always set the header to lower-case regardless of what the config is.
```
katrogan: can you add why this must be lower-case please

flytekit/clis/auth/credentials.py
```diff
@@ -17,10 +20,24 @@
...
-    return _urlparse.urljoin('https://{}/'.format(_URL.get()), discovery_endpoint_path)
+def _get_discovery_endpoint(http_config_val, platform_url_val, insecure_val):
+
+    if http_config_val:
+        scheme, netloc, path, _, _, _ = _urlparse.urlparse(http_config_val)
+        if not scheme:
+            scheme = 'http' if insecure_val else 'https'
+
+        if path:
+            path = '/'.join([path, discovery_endpoint_path])
```
katrogan: @EngHabu is this okay by you?

EngHabu: Can we use this? 
https://www.programcreek.com/python/example/83428/six.moves.urllib.parse.urljoin

EngHabu: I guess you already know about this given the deleted code above... Must have a good reason

flytekit/clis/auth/credentials.py
```diff
@@ -26,18 +26,17 @@ def _get_discovery_endpoint(http_config_val, platform_url_val, insecure_val):
...
-        return _urlparse.urljoin('http://{}/'.format(platform_url_val), discovery_endpoint_path)
-    return _urlparse.urljoin('https://{}/'.format(platform_url_val), discovery_endpoint_path)
+    else:  # Use the main _URL config object effectively
+        scheme = 'http' if insecure_val else 'https'
+        netloc = platform_url_val
+        path = ''
+
+    computed_endpoint = _urlparse.urlunparse((scheme, netloc, path, None, None, None))
+    # The urljoin function needs a trailing slash in order to append things correctly. Also, having an extra slash
+    # at the end is okay, it just gets stripped out.
```
katrogan: sigh

PR #70, opened 2019-12-11
fix IMAGE_NAME and enforce shellcheck

honnix:
As commented https://github.com/lyft/flytekit/pull/55/files#r356599004

### DIFF
--- a/scripts/flytekit_build_image.sh
+++ b/scripts/flytekit_build_image.sh
@@ -17,11 +17,11 @@ echo "------------------------------------"
 echo ""
 
 # Go into the directory representing the user's repo
-pushd $1
+pushd "$1"
 
 # Grab the repo name from the argument if not already defined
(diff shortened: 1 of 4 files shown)


scripts/flytekit_build_image.sh
```diff
@@ -17,11 +17,11 @@ echo "------------------------------------"
 echo ""
 
 # Go into the directory representing the user's repo
-pushd $1
+pushd "$1"
 
 # Grab the repo name from the argument if not already defined
 # Note that this repo name will be the name of the Docker image.
-if [ -z "IMAGE_NAME" ]; then
+if [ -z "${IMAGE_NAME}" ]; then
```
honnix: This is the fix.

wild-endeavor:
this worked before right?  just not for all shells?

honnix:
I doubt. Without $ it is just a string.

PR #72, opened 2019-12-19
Add option to write default config file

katrogan:
This change introduces a lookup against the admin `config/v1/flyte_client` endpoint to discover values necessary to write a config when it doesn't exist.

### DIFF
--- a/flytekit/clis/flyte_cli/main.py
+++ b/flytekit/clis/flyte_cli/main.py
@@ -1531,6 +1531,11 @@ def register_project(identifier, name, description, host, insecure):
 @_host_option
 @_insecure_option
 def setup_config(client_id, redirect_uri, authorization_metadata_key, host, insecure):
+    """
+    Set-up a default config file.
+
+    """
+    _welcome_message()
     config_file = _get_config_file_path()
     if _get_user_filepath_home() and _os.path.exists(config_file):
         _click.secho("Config file already exists at {}".format(_tt(config_file)), fg='blue')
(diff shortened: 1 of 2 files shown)


flytekit/clis/flyte_cli/main.py
```diff
@@ -511,12 +518,50 @@ def _flyte_cli(ctx, project, domain, name, host, insecure):
...
+            configuration option, otherwise, we'd just hardcode the callback path as a constant.
+            FYI, to see if a given port is already in use, run `sudo lsof -i :<port>` if on a Linux system.
+            More details here: https://www.oauth.com/oauth2-servers/redirect-uris/."""
+)
+
+_authorization_metadata_key_option = _click.option(
+    "--authorization-metadata-key",
+    required=False,
+    type=str,
+    default="authorization",
```
wild-endeavor: can we make this flyte-authorization?

flytekit/clis/flyte_cli/main.py
```diff
@@ -511,12 +518,50 @@ def _flyte_cli(ctx, project, domain, name, host, insecure):
...
+    default=None,
+    help="""This is the public identifier for the app which handles authorization for a Flyte deployment.
+            More details here: https://www.oauth.com/oauth2-servers/client-registration/client-id-secret/."""
+)
+
+_redirect_uri_option = _click.option(
+    "--redirect-uri",
+    required=False,
+    type=str,
+    default=None,
```
wild-endeavor: This needs to be set.  we should pick something.  I like `http://localhost:53593/callback`

flytekit/clis/flyte_cli/main.py
katrogan: comes from admin now

PR #73, opened 2019-12-30
Fixing host and insecure flag handling

wild-endeavor:
Basically, this makes it so that,

* If you don't have a `~/.flyte/config` file...
  * `flyte-cli -h host.com` will use `host.com`
  * Not having a `-h` switch will fail
  * `flyte-cli --insecure` will use insecure connections
  * `flyte-cli` without `--insecure` will use secure.

* If you do have a `~/.flyte/config` file...
  * `flyte-cli -h host.com` will use `host.com`
  * Not having a `-h` switch will default to the value set in the config file.  If not set in the config file, the command will fail.
  * `flyte-cli --insecure` will use insecure connections
  * `flyte-cli` without `--insecure` will use the setting in the config file.  If not set in the config file, it will pick up the default, which is `False` and only use secure connections.

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.4.2'
+__version__ = '0.4.3'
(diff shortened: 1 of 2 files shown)


flytekit/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.4.2'
```
katrogan: rebase? 0.4.3 already exists https://pypi.org/project/flytekit/0.4.3/

PR #75, opened 2020-01-13
Workflow fetch and sub-workflows

wild-endeavor:
Please see https://github.com/lyft/flyte/issues/139 for additional context.

* Completes the SdkWorkflow.promote_from_model functionality.  Promoting from model now optionally takes subworkflows and tasks.  If specified, the ones provided will be used instead of fetching from Admin.
* Enables the sub-workflow behavior
* Updates flyteidl to a version that has sub-workflows in the workflow creation request to Admin.
* Calling an SdkWorkflow now produces a node.
* `SdkTaskNode` and `SdkWorkflowNode` have been moved into a separate file from nodes.py.

Related things but don't have time to put into this PR:
* Proper serialization of components (we want to use the same construct as the current registration call path)
* How do you not register subworkflows, but do register them if they're also standalone workflows, and always use the correct name, and also always have the correct dependency structure when doing the topologic sort.
* Discovered that we have duplicate `CompiledTask` models while writing this PR but will fix later.  Left a todo.


flytekit/common/nodes.py
```diff
@@ -219,8 +272,49 @@ def promote_from_model(cls, model):
         :param flytekit.models.core.workflow.Node model:
         :rtype: SdkNode
         """
-        raise _user_exceptions.FlyteAssertion("An SDK node cannot be instantiated merely from a data model object "
-                                              "because it must be contextualized within a workflow.")
+        id = model.id
+        # This should never be called
+        if id == "start-node" or id == "end-node":
```
matthewphsmith: I think we should embed this check in the workflow code since it's really a workflow concept.

flytekit/common/nodes.py
```diff
@@ -219,8 +272,49 @@ def promote_from_model(cls, model):
...
+            _logging.warning("Should not call promote from model on a start node or end node {}".format(model))
+            return None
+
+        sdk_task_node, sdk_workflow_node = None, None
+        if model.task_node is not None:
+            sdk_task_node = SdkTaskNode.promote_from_model(model.task_node)
+        elif model.workflow_node is not None:
+            sdk_workflow_node = SdkWorkflowNode.promote_from_model(model.workflow_node)
+        else:
+            raise Exception("bad Node model")
```
matthewphsmith: let's use the system exceptions `flytekit.common.exceptions.system` just so we can build better visibility in as we go through a common set of exceptions.

flytekit/common/nodes.py
```diff
@@ -173,6 +225,7 @@ def __init__(
         :param flytekit.common.tasks.task.SdkTask sdk_task: The task to execute in this
             node.
         :param flytekit.common.workflow.SdkWorkflow sdk_workflow: The workflow to execute in this node.
+        Question: does this really need to be the sdkrunnable launch plan?
```
matthewphsmith: No I don't think it does...SdkLaunchPlan should suffice.

flytekit/common/nodes.py
wild-endeavor: it's there as well.

flytekit/common/nodes.py
wild-endeavor: will do, this was just a placeholder.

flytekit/clis/sdk_in_container/register.py
```diff
@@ -18,6 +17,9 @@ def register_all(project, domain, pkgs, test, version):
     click.echo('Running task, workflow, and launch plan registration for {}, {}, {} with version {}'.format(
         project, domain, pkgs, version))
 
+    # m = module (i.e. python file)
+    # k = value of dir(m), type str
+    # o = object (e.g. SdkWorkflow)
```
matthewphsmith: yOoUu dDonTt't uNdErrRstAnDd mY sHorTthaNndD?? 😛

flytekit/common/nodes.py
```diff
@@ -216,13 +166,65 @@ def executable_sdk_object(self):
...
+        elif model.workflow_node is not None:
+            sdk_workflow_node = _component_nodes.SdkWorkflowNode.promote_from_model(
+                model.workflow_node,sub_workflows, tasks)
+        else:
+            raise _system_exceptions.FlyteSystemException("Bad Node model, neither task nor workflow detected")
+
+        # When WorkflowTemplate models (containing node models) are returned by Admin, they've been compiled with a
+        # start node.  In order to make the promoted SdkWorkflow look the same, we strip the start-node text back out.
+        for i in model.inputs:
+            if i.binding.promise is not None and i.binding.promise.node_id == _constants.START_NODE_ID:
```
matthewphsmith: we should really tag nodes as system nodes or something when they are inserted by the compiler so we can avoid these types of checks. but for another day, i suppose

tests/flytekit/unit/common_tests/test_workflow.py
```diff
@@ -252,13 +248,13 @@ def my_list_task(wf_params, a, b):
...
 
-    # Test that launch plan ID ref is flexible
+    # Test that workflow is saved in the node
     w._id = 'fake'
     assert n.workflow_node.sub_workflow_ref == 'fake'
     w._id = None
 
     # Test that outputs are promised
-    n.assign_id_and_return('node-id')
+    n.assign_id_and_return('node-id*')  # dns'ified
```
matthewphsmith: Whats this testing?

tests/flytekit/unit/common_tests/test_helpers.py
```diff
@@ -0,0 +1,8 @@
+from __future__ import absolute_import
+
+from os import path as _path
+
+from flyteidl.core import workflow_pb2 as _workflow_pb2
+
+from flytekit.models.core import workflow as _workflow_model
+
```
matthewphsmith: any tests?

also, I can't comment on the bin files, but maybe move to `resources/protos/`? Just in case other things come in

flytekit/common/tasks/task.py
```diff
@@ -91,6 +91,10 @@ def promote_from_model(cls, base_model):
             custom=base_model.custom,
             container=base_model.container
         )
+        # Override the newly generated name if one exists in the base model
+        if base_model.id:
```
matthewphsmith: does this check work? I imagine it will always be true.

flytekit/common/workflow.py
```diff
@@ -171,33 +213,39 @@ def fetch(cls, project, domain, name, version=None):
...
+                continue
+            else:
+                node_map[n.id] = _nodes.SdkNode.promote_from_model(n, sub_workflows, tasks)
+
+        # Set upstream nodes for each node
+        for n in base_model.nodes:
+            if n.id == _constants.START_NODE_ID or n.id == _constants.END_NODE_ID:
+                continue
+            child_node = node_map[n.id]
+            child_node.upstream_nodes[:] = [node_map[upstream_id] for upstream_id in n.upstream_node_ids]
```
matthewphsmith: Recommend using the `>>` operators to note the edges just in case we add more logic there in the future.

flytekit/common/workflow.py
```diff
@@ -171,33 +213,39 @@ def fetch(cls, project, domain, name, version=None):
...
+            if n.id == _constants.START_NODE_ID or n.id == _constants.END_NODE_ID:
+                # The workflow compilation process done by Admin/Propeller will add a fake start-node to the graph
+                # so we need to strip them back out here.
+                continue
+            else:
+                node_map[n.id] = _nodes.SdkNode.promote_from_model(n, sub_workflows, tasks)
+
+        # Set upstream nodes for each node
+        for n in base_model.nodes:
+            if n.id == _constants.START_NODE_ID or n.id == _constants.END_NODE_ID:
```
matthewphsmith: nit: put everything in a conditional like `if n.id not in {START_N)DE_ID, END_NODE_ID}`...and maybe make that a function just in case we add more special case nodes in the future.

matthewphsmith: or pre-prune out the system nodes.

flytekit/common/component_nodes.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        Takes the idl wrapper for a TaskNode and returns the hydrated Flytekit object for it by fetching it from the
+        engine.
+
+        :param flytekit.models.core.workflow.TaskNode base_model:
+        :param list[flytekit.models.task.TaskTemplate] tasks:
+        :rtype: SdkTaskNode
+        """
+        from flytekit.common.tasks import task as _task
+        tasks = tasks or []
```
matthewphsmith: nit: if we're creating this before passing into the function, maybe we make it a dictionary of id -> task in the caller and use it for all tasks.

flytekit/common/component_nodes.py
```diff
@@ -0,0 +1,140 @@
...
+        domain = base_model.reference.domain
+        name = base_model.reference.name
+        version = base_model.reference.version
+        if base_model.launchplan_ref is not None:
+            sdk_launch_plan = _launch_plan.SdkLaunchPlan.fetch(project, domain, name, version)
+            return cls(sdk_launch_plan=sdk_launch_plan)
+        elif base_model.sub_workflow_ref is not None:
+            sub_workflows = sub_workflows or []
+            # The workflow templates for sub-workflows should have been included in the original response
+            for sw in sub_workflows:
```
matthewphsmith: nit: same comment as above about caller making this a dict.

tests/flytekit/unit/common_tests/test_helpers.py
wild-endeavor: forgot to delete this file.  sure, will move.

tests/flytekit/unit/common_tests/test_workflow.py
wild-endeavor: that the dns filter strips it out i guess.

flytekit/common/workflow.py
wild-endeavor: @matthewphsmith you mean this right? https://github.com/lyft/flytekit/pull/75/commits/ebb89222559f11609803593ad6b2792a985e6809

flytekit/common/workflow.py
wild-endeavor: https://github.com/lyft/flytekit/pull/75/commits/f499dd1c5c6dfaefd539d964ed86b47a7ea7794f please check @matthewphsmith

flytekit/common/component_nodes.py
wild-endeavor: @matthewphsmith look through this commit? https://github.com/lyft/flytekit/pull/75/commits/cb8b07790397713b31e7aa6ef11992da1d8d5c6f

flytekit/common/tasks/task.py
wild-endeavor: feel more comfortable with this, leaving for now unless you really want me to remove it.

matthewphsmith: The check is fine, i just don't think it will work even if the ID is "empty"

flytekit/common/workflow.py
matthewphsmith: yup!

flytekit/models/core/identifier.py
```diff
@@ -62,6 +62,13 @@ def version(self):
         """
         return self._version
 
+    @property
+    def is_empty(self):
```
matthewphsmith: oh i meant you could put this in FlyteIdlEntity and then do something like:

```
@property 
def is_empty(self):
    return len(self.to_flyte_idl().SerializeToString()) == 0
```

wild-endeavor: gotcha. thanks.

PR #79, opened 2020-02-05
Relax constraints on sortedcontainers for python3.8

th0114nd:
As in version 2.0 the method was changed from `iteritems` to `items`, it seemed easiest to connect the two in `__getattr__` as `_six.items` doesn't exist.

The reason to upgrade sortedcontainers is to support python3.8: in that version, the path that it imports ABCs from will no longer exist: 
```
/srv/venvs/service/trusty/service_venv_python3.7/lib/python3.7/site-packages/sortedcontainers/sortedlist.py:9
  /srv/venvs/service/trusty/service_venv_python3.7/lib/python3.7/site-packages/sortedcontainers/sortedlist.py:9: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
    from collections import Sequence, MutableSequence
```

In addition, `super(ParameterMapper, self).__getattr_` doesn't exist.

### DIFF
--- a/flytekit/common/nodes.py
+++ b/flytekit/common/nodes.py
@@ -58,7 +58,7 @@ class ParameterMapper(_six.with_metaclass(_common_models.FlyteABCMeta, _SortedDi
         self._initialized = True
 
     def __getattr__(self, key):
-	if key == 'iteritems' and hasattr(super(ParameterMapper, self), 'items'):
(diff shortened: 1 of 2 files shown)


th0114nd:
No problem

PR #85, opened 2020-02-27
Remove the travis version of pypi push now that we have github workflows

wild-endeavor:
We use this now: https://github.com/lyft/flytekit/blob/master/.github/workflows/pythonpublish.yml


wild-endeavor:
ptal @matthewphsmith

kumare3:
LGTM

PR #86, opened 2020-03-02
Mtoledo/add interruptible parameter

migueltol22:


matthewphsmith:
Are we going to expose this configuration to the user?

migueltol22:
> Are we going to expose this configuration to the user?

Yes, the idea is that this can be set at the workflow or task level.

matthewphsmith:
This part looks good, do you want to do that plumbing in the same PR? Or make those changes separately?

migueltol22:
> This part looks good, do you want to do that plumbing in the same PR? Or make those changes separately?

I can do those changes here as well. Was going to sync with Yee to test end to end today and then will iterate off of that.

flytekit/__init__.py
```diff
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.5.1b3'
+__version__ = '0.5.1b4'
```
wild-endeavor: bump again, already past this on master.

wild-endeavor: one more time.

PR #88, opened 2020-03-04
Don't modify resource names when generating sidecar pod spec

katrogan:
Fixes https://github.com/lyft/flyte/issues/180 so that the vanilla flytekit `gpu_request` & `gpu_limit` attributes get respected in the primary sidecar container.

For background, we replace [GPU](https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L22) requests/limits with the[ nvidia/gpu resource name](https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L24).

Capitalization doesn't affect pod spec unmarshaling (tested in flyteplugins)

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.5.1'
+__version__ = '0.5.2'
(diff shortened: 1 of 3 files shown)


kumare3:
Can we add a unit test?

katrogan:
@kumare3 unit tests are already updated

giordyb:
Hi, I tested this and it's still not working, see my last comment on https://github.com/lyft/flyte/issues/180. It looks like this https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L22 never actually happens with the sidecar_task

PR #89, opened 2020-03-05
Add support for the new Presto task

varshaparthay:
* JIRA - https://jira.lyft.net/browse/AVSW-41872 

This PR addresses the following - 

* Presto task SDK 
* This is dependent on https://github.com/lyft/flyteidl/pull/45/files 
* Once the above PR is merged, we can finish the integration and do unit test


tests/flytekit/unit/common_tests/tasks/test_presto_task.py
```diff
@@ -0,0 +1,16 @@
+from datetime import timedelta as _timedelta
+
+from flytekit.common.tasks.presto_task import SdkPrestoTask
+
+
+def test_abc():
+    from flytekit.models.qubole import QuboleHiveJob, HiveQuery
+    # This should change ofc to be the plugin idl object for Presto
+    q = QuboleHiveJob(HiveQuery(query="select * from test", timeout_sec=10, retry_count=1), cluster_label="default",
```
lu4nm3: We might not need to include this if  the tests are for Presto

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,60 @@
...
+    literals as _literals, types as _types, task as _task_model
+
+class SdkPrestoTask(_base_task.SdkTask):
+    """
+    This class includes the logic for building a task that executes as a presto task.
+    """
+
+    def __init__(
+            self,
+            hive_job,
```
matthewphsmith: presto?

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,60 @@
...
+            _task_model.RuntimeMetadata(_task_model.RuntimeMetadata.RuntimeType.FLYTE_SDK, "1.0.0", "python"),
+            timeout,
+            _literals.RetryStrategy(retries),
+            discovery_version,
+            "This is deprecated!"
+        )
+
+        super(SdkPrestoTask, self).__init__(
+            _constants.SdkTaskType.PRESTO_TASK,
+            metadata,
```
matthewphsmith: How are we generating the query to be run?

varshaparthay:
@matthewphsmith - The PR is just to understand the task. this is not a final PR, needs cleanup and unit test.

lu4nm3:
@wild-endeavor @EngHabu can you guys take a look? It's pretty much what we pair-programmed together but I cleaned it up. It's ready for review

flytekit/common/tasks/presto_task.py
wild-endeavor: that's done in the plugin, in the sdk, the query is just provided statically by the user.  However it supports templating as per haytham's suggestion, so you can do things like `select * from blah where ds = {{ds}}`

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,141 @@
...
+            i,
+            _MessageToDict(presto_query.to_flyte_idl()),
+        )
+
+        # Set user provided inputs
+        task_inputs(self)
+
+    # Override method in order to set the implicit inputs
+    def __call__(self, *args, **kwargs):
+        kwargs["implicit_routing_group"] = self.routing_group
```
wild-endeavor: why are we calling it `implicit_`?  Can we remove the implicit_s?

lu4nm3: I did this in order to support interpolation in the `routing_group`, `catalog`, and `schema`. Because of this, i use the `implicit` prefix so that I can filter these out in the  backend and avoid a circular dependency problem during interpolation

lu4nm3: @wild-endeavor I  added an example of this in this PR:
```
presto_task = SdkPrestoTask(
    query="SELECT * FROM hive.city.fact_airport_sessions WHERE ds = '{{ds}}' LIMIT 10",
    output_schema=schema.schema_type,
    routing_group="{{ routing_group }}",
    catalog="hive",
    schema="city",
    task_inputs=inputs(ds=Types.String, routing_group=Types.String),
)
```
Notice how the `routing_group` parameter is given a mustache argument. 

I  figure this is a good way to give users more flexibility

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,141 @@
...
+
+from flytekit.common import constants as _constants
+from flytekit.common.tasks import task as _base_task
+from flytekit.models import (
+    interface as _interface_model
+)
+from flytekit.models import literals as _literals, types as _types, \
+    task as _task_model
+
+from flytekit.common import interface
```
matthewphsmith: nit: interface as _interface

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,141 @@
...
+import datetime as _datetime
+from flytekit.models.presto import PrestoQuery
```
matthewphsmith: nit: `from flytekit.models import presto as _presto_models`

flytekit/sdk/tasks.py
```diff
@@ -9,6 +9,7 @@
     spark_task as _sdk_spark_tasks, hive_task as _sdk_hive_tasks, sidecar_task as _sdk_sidecar_tasks
 from flytekit.common.types import helpers as _type_helpers
 from flytekit.models import interface as _interface_model
+from flytekit.common.tasks import presto_task
```
wild-endeavor: let's do the _ dance in all the things

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,141 @@
...
+        presto_query = PrestoQuery(
+            routing_group=routing_group or "",
+            catalog=catalog or "",
+            schema=schema or "",
+            statement=query
+        )
+
+        # Here we set the routing_group, catalog, and schema as implicit
+        # parameters for caching purposes
+        i = interface.TypedInterface(
```
matthewphsmith: curious, why use `flytekit.models.interface.TypedInterface` instead of the richer object defined in `flytekit.common.interface`?

flytekit/sdk/tasks.py
```diff
@@ -35,7 +36,7 @@ def my_task(wf_params, in1, in2, out1, out2):
     """
 
     def apply_inputs_wrapper(task):
-        if not isinstance(task, _sdk_runnable_tasks.SdkRunnableTask):
+        if not isinstance(task, _sdk_runnable_tasks.SdkRunnableTask) and not isinstance(task, presto_task.SdkPrestoTask):
```
wild-endeavor: add a comment here?  Presto tasks are not sdkrunnable, they're just tasks, but we still want to be able to apply inputs.  If there's ever more, this checking logic should probably be moved to a separate abstract base class or something.

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,141 @@
...
+            metadata,
+            i,
+            _MessageToDict(presto_query.to_flyte_idl()),
+        )
+
+        # Set user provided inputs
+        task_inputs(self)
+
+    # Override method in order to set the implicit inputs
+    def __call__(self, *args, **kwargs):
```
matthewphsmith: if this task can't accept additional inputs when creating a node, we should just make the method `def __call__(self)`

flytekit/sdk/tasks.py
matthewphsmith: this logic to prevent additional application of inputs to a presto task should be contained within the presto_task object. Consider overriding the `add_inputs` method in presto task

tests/flytekit/common/workflows/presto.py
```diff
@@ -0,0 +1,26 @@
...
+
+schema = Types.Schema([("a", Types.String), ("b", Types.Integer)])
+
+presto_task = SdkPrestoTask(
+    query="SELECT * FROM hive.city.fact_airport_sessions WHERE ds = '{{ds}}' LIMIT 10",
+    output_schema=schema.schema_type,
+    routing_group="{{ routing_group }}",
+    catalog="hive",
+    schema="city",
+    task_inputs=inputs(ds=Types.String, routing_group=Types.String),
```
matthewphsmith: Using a decorator like this seems a bit confusing to me. Perhaps it would be better just to pass a dict like `{'ds': Types.String, 'routing_group': Types.String}`

flytekit/common/tasks/presto_task.py
lu4nm3: `TypedInterface` is coming from this import:
```
from flytekit.common import interface as _interface
```
So I think it is the one defined in `flytekit.common.interface`

tests/flytekit/common/workflows/presto.py
matthewphsmith: though i do see how it is nice to use it in the constructor above. so i'll leave the ultimate determination up to you.

tests/flytekit/common/workflows/presto.py
```diff
@@ -0,0 +1,26 @@
...
+)
+
+@workflow_class()
+class PrestoWorkflow(object):
+    ds = Input(Types.String, required=True, help="Test string with no default")
+    routing_group = Input(Types.String, required=True, help="Test string with no default")
+
+    p_task = presto_task(ds=ds, routing_group=routing_group)
+
+    output_a = Output(p_task.outputs.results, sdk_type=Types.Schema())
```
wild-endeavor: yeah i think we can just do `schema` here too.

flytekit/common/tasks/presto_task.py
lu4nm3: *When I say it comes from `from flytekit.common import interface as _interface`, that's what it will look like once I make your suggested change above.

flytekit/common/tasks/presto_task.py
matthewphsmith: Just as a suggestion for convention, perhaps we should give the implicit parameters names like `__implicit` and start documenting that as a standard for naming in such situations.

flytekit/common/tasks/presto_task.py
matthewphsmith: cool, you should do this for hive tasks too (one for read, one for write, and one for arbitrary queries)

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,141 @@
...
+        """
+        return self._catalog
+
+    @property
+    def schema(self):
+        """
+        The schema to set for the given Presto query
+        :rtype: Text
+        """
+        return self._schema
```
matthewphsmith: We'll need to implement the `unit_test` behavior for this task.

i.e.

```
presto_task = SdkPrestoTask(
    query="SELECT * FROM hive.city.fact_airport_sessions WHERE ds = '{{ds}}' LIMIT 10",
    output_schema=schema.schema_type,
    routing_group="{{ routing_group }}",
    catalog="hive",
    schema="city",
    task_inputs=inputs(ds=Types.String, routing_group=Types.String))

output_query = presto_task.unit_test(ds='2020-02-02', routing_group='route_me')
assert output_query == "SELECT * FROM hive.city.fact_airport_sessions WHERE ds = '2020-02-02' LIMIT 10"
```

flytekit/common/tasks/presto_task.py
matthewphsmith: Or maybe it'd be more clean if we had a way to alter the Propeller-side caching logic per plugin. Then shove these parameters into the `custom` struct and pop them out and include them in the cache descriptor.

tests/flytekit/common/workflows/presto.py
lu4nm3: Sorry, what do you mean by this comment?

flytekit/sdk/tasks.py
matthewphsmith: OK I see a little bit more of what you are trying to do here now. I think the best way forward depends on whether you stick with passing `inputs(...)` or go to `{'in1': Types.String}` style when calling the constructor. But we really shouldn't stack isinstance checks where we can avoid it (they really should just exist for robust error messaging--not specifically for uncontained logic)

flytekit/common/tasks/presto_task.py
matthewphsmith: Oh you are totally correct, my bad!

flytekit/common/tasks/presto_task.py
matthewphsmith: (maybe follow up in a separate PR though--it might be a little involved)

tests/flytekit/common/workflows/presto.py
wild-endeavor: as in the flytetools repo PR - we can do just `schema` I think, since it's defined above.

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,144 @@
...
+            _literals.RetryStrategy(retries),
+            interruptible,
+            discovery_version,
+            "This is deprecated!"
+        )
+
+        presto_query = _presto_models.PrestoQuery(
+            routing_group=routing_group or "",
+            catalog=catalog or "",
+            schema=schema or "",
```
matthewphsmith: i don't think schema can ever be an empty string. we should raise an exception if not provided

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,144 @@
...
+        :param Text discovery_version: String describing the version for task discovery purposes
+        :param int retries: Number of retries to attempt
+        :param datetime.timedelta timeout:
+        """
+
+        # Set as class fields which are used down below to configure implicit
+        # parameters
+        self._routing_group = routing_group or ""
+        self._catalog = catalog or ""
+        self._schema = schema or ""
```
matthewphsmith: schema should always be non-null and can't default to a string.

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,144 @@
...
+        self._inputs = task_inputs
```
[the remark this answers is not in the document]
lu4nm3: It's  used down below

flytekit/common/tasks/presto_task.py
lu4nm3: Makes sense. I'll add  a check earlier in the code

lu4nm3:
@wild-endeavor @matthewphsmith @EngHabu Thank you all for the great feedback! I went ahead and made changes from your comments. Please take another look at your earliest convenience :)

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,144 @@
...
+                    description="The routing group set as an implicit input"
+                ),
+                "__implicit_catalog": _interface_model.Variable(
+                    type=_types.LiteralType(simple=_types.SimpleType.STRING),
+                    description="The catalog set as an implicit input"
+                ),
+                "__implicit_schema": _interface_model.Variable(
+                    type=_types.LiteralType(simple=_types.SimpleType.STRING),
+                    description="The schema set as an implicit input"
+                )
```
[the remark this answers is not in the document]
wild-endeavor: I thought we did want them... cuz if they change, we don't want a cache hit.

flytekit/common/tasks/presto_task.py
wild-endeavor: No no, this is a poor name that we unfortunately didn't realize earlier.  This is not the flytekit schema, this is a Presto schema.  @lu4nm3 if these three inputs are `None` will the backend plugin work correctly?

flytekit/common/tasks/presto_task.py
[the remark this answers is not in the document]
wild-endeavor: I think what haytham is saying is that `task_inputs` is used, but there's no point in saving it to `_inputs`.  Can we try removing this and testing?  I feel like it should work.  I don't see it being used anywhere including the base class.

wild-endeavor:
Two more comments in response to Haytham's and Matt's comments above - tiny things, but would feel more comfortable testing it one more time.

flytekit/common/tasks/presto_task.py
lu4nm3: Oops my bad.  @wild-endeavor is correct. This is a Presto "schema" which is equivalent to a  "db" in other database  systems

lu4nm3:
Alright, I went ahead and applied the last of the suggestions and tested locally and everything appears to work well. 

When `routing_group` is not specified, flyteplugins provides the default value of `adhoc` which will target the various adhoc clusters that the routing group points to. In the case of `catalog` and `schema`, these values  are optional because a  user can also  specify them directly in their query such as:
```
SELECT * FROM {catalog}.{schema}.{table_name}
```
e.g.
```
SELECT * FROM hive.city.foo_bar
```
In this case, user's wouldn't need to set the `catalog` nor the `schema` and so we shouldn't set any default for them. If their query  looked something like:
```
SELECT * FROM fact_airport_sessions
```
and neither `catalog` nor `schema` is specified, then the query will result in an error and users will have to set these values directly  in the query or as parameters.

This is what the Presto UI shows when `catalog` and  `schema` are specified directly in the query:

Screen Shot 2020-03-20 at 9 37 39 PM

In Flyte, this is what the inputs will look like (notice that none of the explicit `routing_groupg`, `catalog`, and `schema` inputs are there and that the implicit inputs are all empty):

Screen Shot 2020-03-20 at 9 47 12 PM

And the output will show the external location along with the schema that the user specified:

Screen Shot 2020-03-20 at 9 49 44 PM

I await a final look-over  :)

flytekit/common/tasks/presto_task.py
EngHabu: Is it too late to rename?

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,157 @@
...
+        # parameters
+        self._routing_group = routing_group or ""
+        self._catalog = catalog or ""
+        self._schema = schema or ""
+
+        metadata = _task_model.TaskMetadata(
+            discoverable,
+            # This needs to have the proper version reflected in it
+            _task_model.RuntimeMetadata(
+                _task_model.RuntimeMetadata.RuntimeType.FLYTE_SDK, "1.0.0",
```
EngHabu: We should use [this](https://github.com/lyft/flytekit/blob/9f3b48f2577f727f3cf1eb1292c5267d1749a3a2/flytekit/common/tasks/sdk_runnable.py#L210) for a version.

flytekit/common/tasks/presto_task.py
```diff
@@ -0,0 +1,157 @@
...
+            _literals.RetryStrategy(retries),
+            interruptible,
+            discovery_version,
+            "This is deprecated!"
+        )
+
+        presto_query = _presto_models.PrestoQuery(
+            routing_group=routing_group or "",
+            catalog=catalog or "",
+            schema=schema,
```
EngHabu: should we have a `or ""` here?

flytekit/common/tasks/task.py
```diff
@@ -189,14 +190,7 @@ def validate(self):
...
     def add_inputs(self, inputs):
-        """
-        Adds the inputs to this task.  This can be called multiple times, but it will fail if an input with a given
-        name is added more than once, a name collides with an output, or if the name doesn't exist as an arg name in
-        the wrapped function.
-        :param dict[Text, flytekit.models.interface.Variable] inputs: names and variables
-        """
-        self._validate_inputs(inputs)
-        self.interface.inputs.update(inputs)
+        raise _user_exceptions.FlyteUserException("You can not add inputs to this task")
```
EngHabu: Why is this an SDKRunnableTask-only behavior?

flytekit/models/presto.py
```diff
@@ -0,0 +1,77 @@
...
+
+    @property
+    def schema(self):
+        """
+        :rtype: int
+        """
+        return self._schema
+
+    @property
+    def statement(self):
```
EngHabu: Why is it called Statement here and query up there?

flytekit/common/tasks/presto_task.py
lu4nm3: Not sure. Even if it wasn't, I would rather do the renaming as a separate PR if possible  as I feel this one has been dragged out for a while

flytekit/common/tasks/task.py
lu4nm3: I'll let @wild-endeavor chime in on this one

flytekit/models/presto.py
lu4nm3: I'll rename it  to `statement`. In Presto, a query is made  up  of not  just the main query "statement"  but also the  routing group, catalog,  etc

flytekit/common/tasks/task.py
wild-endeavor: @matthewphsmith ?

lu4nm3:
@EngHabu could you take one final look when you get a chance?

PR #90, opened 2020-03-07
Revert "Don't modify resource names when generating sidecar pod spec"

katrogan:
Reverts lyft/flytekit#88

Originally this was a misguided fix for https://github.com/lyft/flyte/issues/180, because flyteplugins updates all "[GPU](https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L22)" resources with the appropriate nvdia full resource name.

However, https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/api/core/v1/types.go#L4623 uses the lower case name for referencing cpu.

Furthermore, when we build Flyte container tasks we never even use "GPU", but rather the full nvidia name anyways (https://github.com/lyft/flytepropeller/blob/f487cfe6e4d35b52f6f9cb399cceb5abf4a0595a/pkg/utils/k8s.go#L58)

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.5.2'
+__version__ = '0.5.1'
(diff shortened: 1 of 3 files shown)


kumare3:
Lets add some description please. I am working on creating PR templates.

katrogan:
oops, yes - updated the PR description

PR #92, opened 2020-03-22
Finish dynamic yielding of launch plans

wild-endeavor:
# TL;DR
Finish the dynamic behavior of launch plans so that users can yield a launch plan node from within a dynamic task.  See the example [here](https://github.com/lyft/flyteexamples/pull/101/files#diff-53225d5fa74dbe2a2ec4eaa259b06775R41) for more information.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Adding to some of the `repr` functions to make them more complete
* Adding a temporary workaround to bypass the unit test engine when a workflow node is detected.
* Dynamic task logic can now produce nodes in the `DynamicJobSpec` that are not just task nodes.
* Split out the iteration of modules from the rest of the registration step. This pattern makes splitting out registration and serialization easier in the future as well.
* Adding a function to discover the variable name an object was assigned to after the fact.  Without this, when a launchplan is `__call__`ed to produce a node in a dynamic task at runtime, it will only have a random name, not anything recognizable by Admin.
* Move the register function from `SdkLaunchPlan` to `SdkRunnableLaunchPlan`.
* Move the `__call__` function from `SdkRunnableLaunchPlan` to `SdkLaunchPlan`.
* Added an early return to the unit test engine so that dynamic tasks that have launchplan/subwf nodes just return early.  This just makes testing easier for now - should fill out in the future.

Using this as an example dynamic task
```
@inputs(num=Types.Integer)
@outputs(out=Types.Integer, imported_output=Types.Integer, fetched_output=Types.Integer)
@dynamic_task
def lp_yield_task(wf_params, num, out, imported_output, fetched_output):
    wf_params.logging.info("Running inner task... yielding a launchplan")
    wf_params.logging.info("{} {}".format(id_lp._id, id_lp.workflow_id))

    # Test one launch plan defined in this file, but from a workflow imported from another file
    identity_lp_execution = id_lp(a=num)
    yield identity_lp_execution
    out.set(identity_lp_execution.outputs.task_output)

    # Test another defined in another file and just imported.
    imported_lp_execution = secondary_sibling_identity_lp()
    yield imported_lp_execution
    imported_output.set(imported_lp_execution.outputs.task_output)

    # Test a launch plan that's fetched
    fetched_lp_execution = fetched_identity_lp(a=15)
    yield fetched_lp_execution
    fetched_output.set(fetched_lp_execution.outputs.task_output)
```
the resulting dynamic job spec has been attached.
[dynamic_job_spec.txt](https://github.com/lyft/flytekit/files/4422886/dynamic_job_spec.txt)

## Tracking Issue
https://github.com/lyft/flyte/issues/139

## Follow-up issue
Still need to revamp the serialization behavior - https://github.com/lyft/flyte/issues/225
This PR has made the dynamic job spec code messier - will need to do a broader cleanup - https://github.com/lyft/flyte/issues/226

### DIFF
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -1,12 +1,9 @@
 from __future__ import absolute_import
 
-import logging as _logging
-
 import click
 
 from flytekit.clis.sdk_in_container.constants import CTX_PROJECT, CTX_DOMAIN, CTX_TEST, CTX_PACKAGES, CTX_VERSION
 from flytekit.common import utils as _utils
-from flytekit.common.core import identifier as _identifier
 from flytekit.common.tasks import task as _task
 from flytekit.configuration.internal import look_up_version_from_image_tag as _look_up_version_from_image_tag, \
     IMAGE as _IMAGE
@@ -16,31 +13,21 @@ from flytekit.tools.module_loader import iterate_registerable_entities_in_order
 def register_all(project, domain, pkgs, test, version):
     if test:
         click.echo('Test switch enabled, not doing anything...')
+
     click.echo('Running task, workflow, and launch plan registration for {}, {}, {} with version {}'.format(
         project, domain, pkgs, version))
 
     # m = module (i.e. python file)
     # k = value of dir(m), type str
     # o = object (e.g. SdkWorkflow)
-    loaded_entities = []
     for m, k, o in iterate_registerable_entities_in_order(pkgs):
         name = _utils.fqdn(m.__name__, k, entity_type=o.resource_type)
-        _logging.debug("Found module {}\n   K: {} Instantiated in {}".format(m, k, o._instantiated_in))
-        o._id = _identifier.Identifier(
-            o.resource_type,
-            project,
-            domain,
-            name,
-            version
-        )
-        loaded_entities.append(o)
 
-    for o in loaded_entities:
         if test:
-            click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
+            click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), name))
         else:
-            click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
-            o.register(project, domain, o.id.name, version)
+            click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), name))
+            o.register(project, domain, name, version)
 
 
 def register_tasks_only(project, domain, pkgs, test, version):
@@ -59,7 +46,6 @@ def register_tasks_only(project, domain, pkgs, test, version):
         else:
             click.echo("Registering task {:20} {}".format("{}:".format(t.entity_type_text), name))
             t.register(project, domain, name, version)
-
 
 @click.group('register')
 # --pkgs on the register group is DEPRECATED, use same arg on pyflyte.main instead
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -46,13 +46,8 @@ def assign_registerable_identifiers(project, domain, pkgs, test, version):
             o._index = idx
         idx += 1
         name = _utils.fqdn(m.__name__, k, entity_type=o.resource_type)
-        # import ipdb; ipdb.set_trace()
         # click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), name))
         print("\nIdx: {} Module {}\n   K: {} Instantiated in {}".format(o._index, m, k, o._instantiated_in))
-        if o._id is not None:
-            print("   Current id: {}\n   New id: {}".format(o._id, name))
-        else:
-            print("   Current id: None New id: {}".format(name))
         o._id = _identifier.Identifier(
             o.resource_type,
             project,
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -37,9 +37,9 @@ def register_all(project, domain, pkgs, test, version):
 
     for o in loaded_entities:
         if test:
-            click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), name))
+            click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
         else:
-            click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), name))
+            click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
             o.register(project, domain, o._id.name, version)
 
 
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -3,12 +3,12 @@ from __future__ import absolute_import
 import click
 
 from flytekit.clis.sdk_in_container.constants import CTX_PROJECT, CTX_DOMAIN, CTX_TEST, CTX_PACKAGES, CTX_VERSION
-from flytekit.common import utils as _utils
-from flytekit.common.tasks import task as _task
+from flytekit.common import utils as _utils, workflow as _workflow, launch_plan as _launch_plan
 from flytekit.common.core import identifier as _identifier
+from flytekit.common.tasks import task as _task
 from flytekit.configuration.internal import look_up_version_from_image_tag as _look_up_version_from_image_tag, \
     IMAGE as _IMAGE
-from flytekit.models.core import workflow as _workflow_models, identifier as _identifier_model
+from flytekit.models.core import identifier as _identifier_model
 from flytekit.tools.module_loader import iterate_registerable_entities_in_order
 
 
@@ -40,6 +40,7 @@ def assign_registerable_identifiers(project, domain, pkgs, test, version):
     # k = value of dir(m), type str
     # o = object (e.g. SdkWorkflow)
     idx = 0
+    loaded_entities = []
     for m, k, o in iterate_registerable_entities_in_order(pkgs):
         if not hasattr(o, "_index"):
             o._index = idx
@@ -59,6 +60,14 @@ def assign_registerable_identifiers(project, domain, pkgs, test, version):
             name,
             version
         )
+        loaded_entities.append(o)
+
+    print("==================== Second Pass ====================")
+    for o in loaded_entities:
+        o.register(project, domain, o._id.name, version)
+        print("\no.ID {}\n   Instantiated in {}".format(o._id, o._instantiated_in))
+        if isinstance(o, _launch_plan.SdkLaunchPlan):
+            print("LP: {} workflow_id {}".format(o.id, o.workflow_id))
 
 
 
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -40,7 +40,7 @@ def register_all(project, domain, pkgs, test, version):
             click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
         else:
             click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
-            o.register(project, domain, o._id.name, version)
+            o.register(project, domain, o.id.name, version)
 
 
 def register_tasks_only(project, domain, pkgs, test, version):
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -54,7 +54,7 @@ def assign_registerable_identifiers(project, domain, pkgs, test, version):
         else:
             print("   Current id: None New id: {}".format(name))
         o._id = _identifier.Identifier(
-            _identifier_model.ResourceType.WORKFLOW,
+            o.resource_type,
             project,
             domain,
             name,
@@ -68,7 +68,6 @@ def assign_registerable_identifiers(project, domain, pkgs, test, version):
         print("\no.ID {}\n   Instantiated in {}".format(o._id, o._instantiated_in))
         if isinstance(o, _launch_plan.SdkLaunchPlan):
             print("LP: {} workflow_id {}".format(o.id, o.workflow_id))
-
 
 
 def register_tasks_only(project, domain, pkgs, test, version):
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -1,15 +1,15 @@
 from __future__ import absolute_import
 
-import click
 import logging as _logging
 
+import click
+
 from flytekit.clis.sdk_in_container.constants import CTX_PROJECT, CTX_DOMAIN, CTX_TEST, CTX_PACKAGES, CTX_VERSION
-from flytekit.common import utils as _utils, workflow as _workflow, launch_plan as _launch_plan
+from flytekit.common import utils as _utils
 from flytekit.common.core import identifier as _identifier
 from flytekit.common.tasks import task as _task
 from flytekit.configuration.internal import look_up_version_from_image_tag as _look_up_version_from_image_tag, \
     IMAGE as _IMAGE
-from flytekit.models.core import identifier as _identifier_model
 from flytekit.tools.module_loader import iterate_registerable_entities_in_order
 
 
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -5,8 +5,10 @@ import click
 from flytekit.clis.sdk_in_container.constants import CTX_PROJECT, CTX_DOMAIN, CTX_TEST, CTX_PACKAGES, CTX_VERSION
 from flytekit.common import utils as _utils
 from flytekit.common.tasks import task as _task
+from flytekit.common.core import identifier as _identifier
 from flytekit.configuration.internal import look_up_version_from_image_tag as _look_up_version_from_image_tag, \
     IMAGE as _IMAGE
+from flytekit.models.core import workflow as _workflow_models, identifier as _identifier_model
 from flytekit.tools.module_loader import iterate_registerable_entities_in_order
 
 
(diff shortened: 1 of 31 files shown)


flytekit/common/launch_plan.py
```diff
@@ -225,12 +206,44 @@ def execute_with_literals(self, project, domain, literal_inputs, name=None, noti
...
+        # Take the default values from the launch plan
+        default_inputs = {
+            k: v.sdk_default
+            for k, v in _six.iteritems(self.default_inputs.parameters) if not v.required
+        }
+        default_inputs.update(input_map)
+
+        bindings, upstream_nodes = self.interface.create_bindings_for_inputs(default_inputs)
+
+        # TODO: Remove DEADBEEF
```
EngHabu: Yes plz. I would take a look at the `rename_and_return` method on the SdkNode I believe, this is where we rewrite the id, it should also set the name in metadata...

flytekit/common/tasks/sdk_dynamic.py
```diff
@@ -150,10 +157,28 @@ def _produce_dynamic_job_spec(self, context, inputs):
...
             visited_nodes.add(sub_task_node)
+            executable = sub_task_node.executable_sdk_object
+
+            # If the executable object that we're dealing with is registerable (ie, either an SdkRunnableLaunchPlan or
+            # an SdkWorkflow), then it should have the ability to give itself a name.  After assigning itself the name,
+            # also make sure the id is properly set according to current config values.
+            if isinstance(executable, _registerable.RegisterableEntity):
+                executable.auto_assign_name()
+                executable._id = _identifier.Identifier(
+                    _identifier_model.ResourceType.LAUNCH_PLAN,
```
EngHabu: Why is it always a launch plan? is this correct?

flytekit/common/tasks/sdk_dynamic.py
```diff
@@ -165,44 +190,62 @@ def _produce_dynamic_job_spec(self, context, inputs):
...
-                    literals={binding.var: binding.binding.to_literal_model() for binding in
-                              sub_task_node.inputs})
+                    # Upload inputs to working directory under /array_job.input_ref/inputs.pb
+                    input_path = _os.path.join(node.id, _constants.INPUT_FILE_NAME)
+                    generated_files[input_path] = _literal_models.LiteralMap(
+                        literals={binding.var: binding.binding.to_literal_model() for binding in
+                                  sub_task_node.inputs})
+
+            # Handling tasks
             else:
```
EngHabu: elif ?

flytekit/engines/unit/engine.py
```diff
@@ -148,6 +148,16 @@ def _transform_for_user_output(self, outputs):
 
 class DynamicTask(ReturnOutputsTask):
 
+    def __init__(self, *args, **kwargs):
+        self._has_workflow_node = False
+        super(DynamicTask, self).__init__(*args, **kwargs)
+
+    def _transform_for_user_output(self, outputs):
+        if self.has_workflow_node:
+            # If a workflow node has been detected, then we skip any transformation
```
EngHabu: unclear to me why...

flytekit/tools/module_loader.py
```diff
@@ -2,10 +2,16 @@
 
 import importlib
 import pkgutil
+import logging as _logging
```
EngHabu: I don't see code change in this file... any of these additions needed?

tests/flytekit/unit/models/test_dynamic_wfs.py
```diff
@@ -0,0 +1,54 @@
...
+    launch_plan_node = dj_spec.nodes[0]
+    node_id = launch_plan_node.id
+    assert "models-test-dynamic-wfs-id-lp" in node_id
+    assert node_id.endswith("-0")
+
+    # Assert that the output of the dynamic job spec is bound to the single node in the spec, the workflow node
+    # containing the launch plan
+    assert dj_spec.outputs[0].var == "out"
+    assert dj_spec.outputs[0].binding.promise.node_id == node_id
+    assert dj_spec.outputs[0].binding.promise.var == "task_output"
```
EngHabu: Do you mind also writing a dynamic WF in test Workflows? I understand we can't run and register it.. but at least we will test with it when Propeller breaks because we forgot to implement something :D

flytekit/clis/sdk_in_container/register.py
```diff
@@ -1,33 +1,46 @@
 from __future__ import absolute_import
 
 import click
+import logging as _logging
```
matthewphsmith, line 4: maybe revert the changes in this file with the exception of the debug logs as I don't think they are necessary (and somewhat divergent from keeping logic associated with the objects themselves).

flytekit/common/component_nodes.py
```diff
@@ -67,7 +67,14 @@ def __init__(self, sdk_workflow=None, sdk_launch_plan=None):
         """
         self._sdk_workflow = sdk_workflow
         self._sdk_launch_plan = sdk_launch_plan
-        super(SdkWorkflowNode, self).__init__()
```
matthewphsmith, line 70: You shouldn't remove this in case we ever add `__init__` logic in the super

flytekit/common/mixins/registerable.py
```diff
@@ -73,3 +80,55 @@ def instantiated_in(self):
...
+        It then scans all objects in the module, and when an object match is found, it knows it's found the right
+        variable name.
+        """
+        _logging.debug("Running name auto assign")
+        m = _importlib.import_module(self.instantiated_in)
+
+        for k in dir(m):
+            if getattr(m, k) == self:
+                _logging.debug("Auto-assigning name to {}".format(k))
+                self._platform_valid_name = "{}.{}".format(self.instantiated_in, k)
```
matthewphsmith, line 130: there is a util function somewhere that does this (and it is important for our legacy systems at Lyft because they use a different naming convention). Search for `fqdn`, I think.

flytekit/common/mixins/registerable.py
```diff
@@ -73,3 +80,55 @@ def instantiated_in(self):
...
+    def platform_valid_name(self):
+        """
+        :rtype: Text
+        """
+        return self._platform_valid_name
+
+    def auto_assign_name(self):
+        """
+        This function is a bit of trickster Python code that goes hand in hand with the _InstanceTracker metaclass
+        defined above. Thanks @matthewphsmith for this bit of ingenuity.
```
matthewphsmith, line 101: lol don't blame me for this. git blame will tell the truth :p

flytekit/common/tasks/sdk_dynamic.py
```diff
@@ -150,10 +157,28 @@ def _produce_dynamic_job_spec(self, context, inputs):
...
         for sub_task_node in _itertools.chain(yielded_sub_tasks, upstream_nodes):
             if sub_task_node in visited_nodes:
                 continue
             visited_nodes.add(sub_task_node)
+            executable = sub_task_node.executable_sdk_object
+
+            # If the executable object that we're dealing with is registerable (ie, either an SdkRunnableLaunchPlan or
+            # an SdkWorkflow), then it should have the ability to give itself a name.  After assigning itself the name,
+            # also make sure the id is properly set according to current config values.
+            if isinstance(executable, _registerable.RegisterableEntity):
```
matthewphsmith, line 173: this should hit for task, launch plan, and workflow since they're all registerable

flytekit/common/tasks/sdk_dynamic.py
```diff
@@ -165,44 +190,62 @@ def _produce_dynamic_job_spec(self, context, inputs):
...
-                if sub_task_node.executable_sdk_object in array_job_index:
-                    array_job, node = array_job_index[sub_task_node.executable_sdk_object]
-                    array_job.size += 1
-                    array_job.min_successes = int(math.ceil((1 - effective_failure_ratio) * array_job.size))
-                else:
-                    array_job = self._create_array_job(inputs_prefix=unique_node_id)
-                    node = sub_task_node.assign_id_and_return(unique_node_id)
-                    array_job_index[sub_task_node.executable_sdk_object] = (array_job, node)
+            # Handling cases where the yielded nodes are launch plan or subworkflow nodes
+            if isinstance(sub_task_node.executable_sdk_object, _launch_plan.SdkLaunchPlan) \
```
matthewphsmith, line 194: nit: `isinstance(sub_task_node.executable_sdk_object, (_launch_plan.SdkLaunchPlan, _workflow.SdkWorkflow))`

flytekit/engines/unit/engine.py
```diff
@@ -188,6 +203,13 @@ def _execute_user_code(self, inputs):
             )
         return results
 
+    @property
+    def has_workflow_node(self):
```
matthewphsmith, line 207: don't think we should expose this as a property since it isn't relevant outside the class. also not sure if the indication of this is best represented as an instance variable.

flytekit/engines/unit/engine.py
matthewphsmith: theoretically, we shouldn't need a check like this. Can you explain the reasoning a bit?

tests/flytekit/unit/models/test_dynamic_wfs.py
```diff
@@ -0,0 +1,54 @@
...
+@_tasks.inputs(num=_Types.Integer)
+@_tasks.outputs(out=_Types.Integer)
+@_tasks.python_task
+def inner_task(wf_params, num, out):
+    wf_params.logging.info("Running inner task... setting output to input")
+    out.set(num)
+
+
+@_workflow.workflow_class()
+class IdentityWorkflow(object):
```
matthewphsmith, line 18: can you add a test in `user_scenarios/unit_testing` where a dynamic task blows up with a useful exception if you try to execute a unit_test on a dynamic task with a launch plan or sub-workflow

flytekit/common/launch_plan.py
wild-endeavor: Yup, i'll just make it empty to start.

flytekit/common/launch_plan.py
```diff
@@ -27,6 +27,7 @@ class SdkLaunchPlan(
 ):
     def __init__(self, *args, **kwargs):
         super(SdkLaunchPlan, self).__init__(*args, **kwargs)
+        # Why do we set this to None?
```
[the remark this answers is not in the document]
wild-endeavor: Right, got it, thanks.

flytekit/tools/module_loader.py
wild-endeavor: reverting

flytekit/common/mixins/registerable.py
wild-endeavor: 😈

flytekit/common/mixins/registerable.py
wild-endeavor: I was following this: https://github.com/lyft/flytekit/blob/master/flytekit/common/tasks/sdk_runnable.py#L233

flytekit/engines/unit/engine.py
wild-endeavor: removing the property.  Not quite sure what you mean by the second comment.

flytekit/common/tasks/sdk_dynamic.py
wild-endeavor: right, updated the comment.

flytekit/clis/sdk_in_container/register.py
wild-endeavor: but i like this...

flytekit/common/tasks/sdk_dynamic.py
wild-endeavor: nope, not correct.

flytekit/common/component_nodes.py
wild-endeavor: https://github.com/lyft/flytekit/pull/92/commits/a3a389627a6cda42fc9a0b86b33c0dec7f7319c7#diff-11888e82c8c56a7d42b7bc092340d1baR70?

flytekit/common/tasks/sdk_dynamic.py
wild-endeavor: do i have to?

flytekit/common/tasks/sdk_dynamic.py
wild-endeavor: done

tests/flytekit/unit/models/test_dynamic_wfs.py
wild-endeavor: I have a test, and it doesn't blow up, because I added that bit that was obviously very confusing about how it only runs half the steps... if a launch plan or workflow node is detected, the unit test will end early.  I like being able to test the dynamic job spec that's generated, but I also don't want to/don't know how to finish the entire dynamic task unit test engine right now that can follow through on lps/subwfs.

flytekit/common/mixins/registerable.py
wild-endeavor: Nm, I think I got it.  You mean that when a launch plan is registered right now, the name that is assigned to it comes from register.py, which uses the fqdn function.
https://github.com/lyft/flytekit/pull/92/commits/76a812cd3ba43353f2ed040d4b5b4ae912554f4b

flytekit/engines/unit/engine.py
wild-endeavor: I'm going to leave this here for now, if that's okay and add a lot of warning logs and comments.  I'm basically short-circuiting the unit test engine.  I want to be able to at least test the dynamic job spec that's produced, but don't have time to finish the rest of the unit test engine right now.  Is this okay?

flytekit/clis/sdk_in_container/register.py
```diff
@@ -13,21 +16,31 @@
...
         project, domain, pkgs, version))
 
     # m = module (i.e. python file)
     # k = value of dir(m), type str
     # o = object (e.g. SdkWorkflow)
+    loaded_entities = []
     for m, k, o in iterate_registerable_entities_in_order(pkgs):
         name = _utils.fqdn(m.__name__, k, entity_type=o.resource_type)
-
+        _logging.debug("Found module {}\n   K: {} Instantiated in {}".format(m, k, o._instantiated_in))
```
kumare3, line 28: so the default logging is to the console, would this be weird to look at?
How about you enable click.echo when debug is on?

kumare3: what I mean is click.debug_echo?

flytekit/common/component_nodes.py
```diff
@@ -67,7 +67,18 @@ def __init__(self, sdk_workflow=None, sdk_launch_plan=None):
...
+        if sdk_workflow:
+            super(SdkWorkflowNode, self).__init__(sub_workflow_ref=sdk_workflow.id)
```
kumare3, line 71: Why does it matter if we set both the values and just default them to None? easier to read isn't it?

kumare3: Again to explain
wf_id = sdk_workflow.id if sdk_workflow else None
lp = sdk_launch_plan.id if sdk_launch_plan else None
and then init

wild-endeavor: can't dereference `None`.  The underlying constructor already has default None values, but we need to pass in the `.id` of the subworkflow/launch plan object, not the object itself.

flytekit/clis/sdk_in_container/register.py
wild-endeavor: Normally this wouldn't print at all actually, it'll only show up if you set your logging level elsewhere to debug, which I do when I'm working on this.  But I think too many people have complained about this file, I think I'll just revert everything here.

kumare3:
Awesome, Thank you for implementing this.

flytekit/common/component_nodes.py
wild-endeavor: updated

flytekit/clis/sdk_in_container/register.py
wild-endeavor: reverted, saving the diff for the next pr.

flytekit/common/component_nodes.py
```diff
@@ -67,7 +67,17 @@ def __init__(self, sdk_workflow=None, sdk_launch_plan=None):
         """
         self._sdk_workflow = sdk_workflow
         self._sdk_launch_plan = sdk_launch_plan
-        super(SdkWorkflowNode, self).__init__()
+        sdk_wf_id = sdk_workflow.id if sdk_workflow else None
+        sdk_lp_id = sdk_launch_plan.id if sdk_launch_plan else None
+        super(SdkWorkflowNode, self).__init__(launchplan_ref=sdk_lp_id, sub_workflow_ref=sdk_wf_id)
```
kumare3, line 72: what happens if both are set?

kumare3: Is that not a violation?

tests/flytekit/unit/models/test_dynamic_wfs.py
kumare3: what does that mean? We dont need to register the dynamic workflow but I agree we should run that as a test? Where is the example

flytekit/common/component_nodes.py
wild-endeavor: I can add an exception

flytekit/common/tasks/sdk_dynamic.py
```diff
@@ -150,10 +157,28 @@ def _produce_dynamic_job_spec(self, context, inputs):
...
+            # If the executable object that we're dealing with is registerable (ie, SdkRunnableLaunchPlan, SdkWorkflow
+            # SdkTask, or SdkRunnableTask), then it should have the ability to give itself a name. After assigning
+            # itself the name, also make sure the id is properly set according to current config values.
+            if isinstance(executable, _registerable.RegisterableEntity):
+                executable.auto_assign_name()
+                executable._id = _identifier.Identifier(
+                    executable.resource_type,
+                    _internal_config.TASK_PROJECT.get() or _internal_config.PROJECT.get(),
+                    _internal_config.TASK_DOMAIN.get() or _internal_config.DOMAIN.get(),
+                    executable._platform_valid_name,
```
matthewphsmith, line 179: we should make this a property on the RegisterableEntity object so that we don't read access a 'private' attribute. Someday I'll figure out the right pattern so we don't have to do this `executable._id` stuff too :(

flytekit/engines/unit/engine.py
matthewphsmith: 👍

matthewphsmith: just add a `TODO:` so it searchable please

matthewphsmith: oo you did below. nevermind.

flytekit/engines/unit/engine.py
matthewphsmith: it's kinda weird to set the value on the instance itself since theoretically you might call it multiple times and it might yield different nodes on each run (like if you were unit testing someday). I think you can just put a `TODO: remove when unit testing is implemented for workflow/launch plan nodes`

tests/flytekit/unit/models/test_dynamic_wfs.py
matthewphsmith: ah ok got it.

flytekit/common/tasks/sdk_dynamic.py
EngHabu: no

PR #101, opened 2020-04-10
 Generic Spark Integration

akhurana001:
Makes it easier to integrate a non-Python Spark Job as a task in Flyte. Issue : https://github.com/lyft/flyte/issues/266 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/common/tasks/generic_spark_task.py
+++ b/flytekit/common/tasks/generic_spark_task.py
@@ -141,7 +141,7 @@ class SdkGenericSparkTask( _base_tasks.SdkTask):
             args.append("{{{{.Inputs.{}}}}}".format(k))
 
         return _task_models.Container(
-            image= _internal_config.IMAGE.get(),
+            image=_internal_config.IMAGE.get(),
             command=[],
             args=args,
             resources=_task_models.Resources([], []),
--- a/flytekit/common/tasks/generic_spark_task.py
+++ b/flytekit/common/tasks/generic_spark_task.py
@@ -9,11 +9,24 @@ from flytekit import __version__
 import sys as _sys
 import six as _six
 from flytekit.common.tasks import task as _base_tasks
+from flytekit.common.types import helpers as _helpers, primitives as _primitives
+
 from flytekit.models import literals as _literal_models, task as _task_models
 from google.protobuf.json_format import MessageToDict as _MessageToDict
 from flytekit.common import interface as _interface
-from flytekit.models import interface as _interface_model
+from flytekit.common.exceptions import user as _user_exceptions
+from flytekit.common.exceptions import scopes as _exception_scopes
+
 from flytekit.configuration import internal as _internal_config
+
+input_types_supported = { _primitives.Integer,
+                          _primitives.Boolean,
+                          _primitives.Float,
+                          _primitives.String,
+                          _primitives.Datetime,
+                          _primitives.Timedelta,
+                        }
+
 
 class SdkGenericSparkTask( _base_tasks.SdkTask):
     """
@@ -56,14 +69,11 @@ class SdkGenericSparkTask( _base_tasks.SdkTask):
         spark_job = _task_models.SparkJob(
             spark_conf=spark_conf,
             hadoop_conf=hadoop_conf,
-            type = spark_type,
+            spark_type = spark_type,
             application_file=main_application_file,
             main_class=main_class,
             executor_path=_sys.executable,
         ).to_flyte_idl()
-
-        # No output support
-        input_variables = {k: _interface_model.Variable(v.to_flyte_literal_type(), k) for k, v in _six.iteritems(task_inputs)}
 
         super(SdkGenericSparkTask, self).__init__(
             task_type,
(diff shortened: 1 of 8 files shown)


wild-endeavor:
issue please.

tests/flytekit/common/workflows/scala_spark.py
```diff
@@ -0,0 +1,40 @@
+from __future__ import absolute_import
+from __future__ import division
+from __future__ import print_function
+
+from flytekit.sdk.tasks import spark_task, inputs, python_task
+from flytekit.sdk.types import Types
+from flytekit.sdk.workflow import workflow_class, Input
+
+
+scala_spark = spark_task(spark_type="SCALA",
```
kumare3, line 10: I think we should do one of 2 things,
Make Spark_type an enum, which it seems it is already in the proto def
OR
we should make special task wrappers like scala_spark etc which fixes the type

flytekit/sdk/tasks.py
```diff
@@ -406,12 +406,16 @@ def spark_task(
         cache_version='',
         retries=0,
         interruptible=None,
+        inputs=None,
```
kumare3, line 409: Is this handleded differently as compared to @inputs decorator? If so we should document that correctly, as this is very confusing.

ALSO at the moment I think only a few input types are supported like primitives and blobs right?

flytekit/common/tasks/generic_spark_task.py
```diff
@@ -0,0 +1,112 @@
...
+            ),
+            _interface.TypedInterface(input_variables, {}),
+            _MessageToDict(spark_job),
+            container=self._get_container_definition(
+                task_inputs= task_inputs,
+                environment=environment
+            )
+        )
+
+    def _get_container_definition(
```
kumare3, line 91: I think this is where you will have to check the types and raise an alert. You can make this as a common method for now, when we add support for all types, we can remove this check

akhurana001: done, added it in the validate_inputs below

flytekit/sdk/tasks.py
akhurana001: Moved to use the inputs decorator. Added the check for primitives for now

tests/flytekit/common/workflows/scala_spark.py
akhurana001: made an enum .

flytekit/sdk/tasks.py
matthewphsmith: I wouldn't mix the decorator with the generic usage. The decorator patterns are part of the SDK's basic python programming model so the assumption is that they will always be python. It's easy enough to expose a helper function or class for Scala spark jobs elsewhere.

flytekit/common/tasks/spark_task.py
```diff
@@ -88,6 +89,8 @@ def __init__(
             hadoop_conf=hadoop_conf,
             application_file="local://" + spark_exec_path,
             executor_path=_sys.executable,
+            main_class="",
```
matthewphsmith, line 92: perhaps None to make more clear that it is unset?

flytekit/sdk/tasks.py
matthewphsmith: (if you want to make available through this file, that's fine too. I just think mixing a decorator and common object definitions will eventually cause a problem)

matthewphsmith: and continuing Ketan's thought, I would also just avoid adding the input arg that way all together. Especially since as your test below shows, you use the input decorator as a helper function--which is a bit confusing. Better to use a new interface and just a normal `{'a': Types.Integer}` style annotation

flytekit/sdk/types.py
```diff
@@ -1,8 +1,15 @@
 from __future__ import absolute_import
 from flytekit.common.types import primitives as _primitives, blobs as _blobs, schema as _schema, helpers as _helpers, \
     proto as _proto, containers as _containers
+import enum
 
 
+class SparkType(enum.Enum):
```
matthewphsmith, line 7: Can this be defined in flyteidl?

matthewphsmith: Looks like it is and you map it above. This file is really for I/O typing so I wouldn't introduce them here.

flytekit/common/tasks/spark_task.py
matthewphsmith: Nevermind

flytekit/sdk/tasks.py
akhurana001: I originally wasn't using the decorator but changed after Ketan's comment to be in-sync with how we do this for Presto as well : https://github.com/lyft/flytekit/blob/master/tests/flytekit/common/workflows/presto.py#L11

flytekit/sdk/types.py
akhurana001: will move to a different file

flytekit/sdk/tasks.py
akhurana001: Updated to stop using the existing spark_task decorator. I do think we should make the generic_spark_task available from this file as well. Added it separately as a helper function.

kumare3:
Looks good

flytekit/sdk/tasks.py
```diff
@@ -488,6 +490,45 @@ def wrapper(fn):
         return wrapper
 
 
+def generic_spark_task(
```
kumare3, line 493: I dont think we need this method right?
We can directly use - SdkGenericSparkTask?

Unless you are defaulting somethings. (i see timeout)

rmcarthur:
I'm a new Flyte (and FlyteKit) user. Is this PR going to be merged and is it going to be the canonical way to use scala spark with Flyte?

flytekit/sdk/tasks.py
akhurana001: Yeah, we are defaulting. In-addition, I also want users to be able to look-up all task_types supported from a single place.

tests/flytekit/common/workflows/scala_spark.py
```diff
@@ -0,0 +1,38 @@
...
+@python_task(cache_version='1')
+def print_every_time(workflow_parameters, date_triggered):
+    print("My input : {}".format(date_triggered))
+
+
+@workflow_class
+class SparkTasksWorkflow(object):
+    triggered_date = Input(Types.Datetime)
+    partitions = Input(Types.Integer)
+    sparkTask = scala_spark(partitions=partitions)
```
matthewphsmith, line 36: nit: no camel case

matthewphsmith:
You're gonna need to update `flytekit.__init__.py` to have a beta version to releas

matthewphsmith:
Looks good, but I will have to approve after you update the version or else my review will get cleared.

tests/flytekit/common/workflows/scala_spark.py
akhurana001: done

akhurana001:
> You're gonna need to update `flytekit.__init__.py` to have a beta version to releas

Bumped version

PR #104, opened 2020-04-30
Serialization to produce registerable entities

wild-endeavor:
# TL;DR
This is step 1 of the issue linked below.  This corrects the serialization component of pyflyte.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Minor update to the registration code to just aggregate everything into a list first, and then run register.
* This PR basically copies the registration process, but instead of registering against admin, will create the objects necessary for registration at a later time.
* The same topo sort is used to correctly ascertain the order in which things should be registered later, enforced by adding an index to the start of the filename.
* This code path does not rely on the engine.

## Tracking Issue
https://github.com/lyft/flyte/issues/225

## Follow-up issue
NA

### DIFF
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -1,9 +1,12 @@
 from __future__ import absolute_import
 
+import logging as _logging
+
 import click
 
 from flytekit.clis.sdk_in_container.constants import CTX_PROJECT, CTX_DOMAIN, CTX_TEST, CTX_PACKAGES, CTX_VERSION
 from flytekit.common import utils as _utils
+from flytekit.common.core import identifier as _identifier
 from flytekit.common.tasks import task as _task
 from flytekit.configuration.internal import look_up_version_from_image_tag as _look_up_version_from_image_tag, \
     IMAGE as _IMAGE
@@ -13,21 +16,31 @@ from flytekit.tools.module_loader import iterate_registerable_entities_in_order
 def register_all(project, domain, pkgs, test, version):
     if test:
         click.echo('Test switch enabled, not doing anything...')
-
     click.echo('Running task, workflow, and launch plan registration for {}, {}, {} with version {}'.format(
         project, domain, pkgs, version))
 
     # m = module (i.e. python file)
     # k = value of dir(m), type str
     # o = object (e.g. SdkWorkflow)
+    loaded_entities = []
     for m, k, o in iterate_registerable_entities_in_order(pkgs):
         name = _utils.fqdn(m.__name__, k, entity_type=o.resource_type)
+        _logging.debug("Found module {}\n   K: {} Instantiated in {}".format(m, k, o._instantiated_in))
+        o._id = _identifier.Identifier(
+            o.resource_type,
+            project,
+            domain,
+            name,
+            version
+        )
+        loaded_entities.append(o)
 
+    for o in loaded_entities:
         if test:
-            click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), name))
+            click.echo("Would register {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
         else:
-            click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), name))
-            o.register(project, domain, name, version)
+            click.echo("Registering {:20} {}".format("{}:".format(o.entity_type_text), o.id.name))
+            o.register(project, domain, o.id.name, version)
 
 
 def register_tasks_only(project, domain, pkgs, test, version):
@@ -46,6 +59,7 @@ def register_tasks_only(project, domain, pkgs, test, version):
         else:
             click.echo("Registering task {:20} {}".format("{}:".format(t.entity_type_text), name))
             t.register(project, domain, name, version)
+
 
 @click.group('register')
 # --pkgs on the register group is DEPRECATED, use same arg on pyflyte.main instead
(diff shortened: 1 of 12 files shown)


flytekit/common/workflow.py
```diff
@@ -286,6 +287,20 @@ def register(self, project, domain, name, version):
             self._id = old_id
             raise
 
+    @_exception_scopes.system_entry_point
+    def serialize(self):
```
matthewphsmith, line 291: if this is going to be a required method on each registerable object, let's add it as an abstract method to the RegisterableEntity mixin.

flytekit/common/workflow.py
```diff
@@ -286,6 +287,20 @@ def register(self, project, domain, name, version):
...
+        """
+        Serializing a workflow should produce an object similar to what the registration step produces, in preparation
+        for actual registration to Admin.
+
+        :rtype: flyteidl.admin.workflow_pb2.WorkflowSpec
```
matthewphsmith, line 296: Also, if we're going to serialize, maybe go straight to bytes and call `SerializeToString()`?

matthewphsmith: And if we want to keep it as the IDL object, let's refactor the engine to call this method so we don't have duplicate code.

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -2,49 +2,120 @@
...
+            project,
+            domain,
+            name,
+            version
+        )
+        loaded_entities.append(o)
+
+    i = 0
+    zero_padded_length = _determine_text_chars(len(loaded_entities))
+    for entity in loaded_entities:
```
matthewphsmith, line 92: nit: `for i, entity in enumerate(..):`

flytekit/common/workflow.py
wild-endeavor: I'm gonna go with option 2 here.

flytekit/common/workflow.py
wild-endeavor: good point

flytekit/common/workflow.py
wild-endeavor: Actually, can I just leave it as is?  It's not quite the same (serialize calls to_flyteidl already).

So my plan is to remove the register call completely in a future PR (since registration happens from a container, we can control this, and only will do this after all refactorator PRs are done, or possibly rewrite how pyflyte register itself works).

But I also don't want bytes quite yet.  I haven't quite worked out how it'll work but I want the option to serialize these all to a pipe.  To do this, the serialize command will also need to emit the size of each object (since protobuf doesn't have a native delimiter).  I might try to do that then.  If i change it now then i can't use the current `write_proto_to_file` function too.

matthewphsmith: you do you. i'm not sure if fully deprecating the register call is worth the pain though. Adding serialize is a nice option, but I don't think I see the value in removing the existing registration guts and then dealing with the migration.

flytekit/clis/sdk_in_container/register.py
```diff
@@ -13,21 +16,31 @@
...
         project, domain, pkgs, version))
 
     # m = module (i.e. python file)
     # k = value of dir(m), type str
     # o = object (e.g. SdkWorkflow)
+    loaded_entities = []
     for m, k, o in iterate_registerable_entities_in_order(pkgs):
         name = _utils.fqdn(m.__name__, k, entity_type=o.resource_type)
-
+        _logging.debug("Found module {}\n   K: {} Instantiated in {}".format(m, k, o._instantiated_in))
```
katrogan, line 28: Did you mean to leave this debug and click echo calls below?

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -1,50 +1,132 @@
...
+    """
+    :param Text project:
+    :param Text domain:
+    :param list[Text] pkgs:
+    :param Text version:
+    :param Text folder:
+
+    :return:
+    """
+    # m = module (i.e. python file)
```
katrogan, line 32: could just use descriptive variable names in the first place to avoid these comments :P

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -1,50 +1,132 @@
...
+        serialized = entity.serialize()
+        fname_index = str(i).zfill(zero_padded_length)
+        fname = '{}_{}.pb'.format(fname_index, entity._id.name)
+        click.echo('  Writing {} to\n    {}'.format(entity._id, fname))
+        _write_proto_to_file(serialized, fname)
+
+        # Not everything serialized will necessarily have an identifier field in it, even though some do (like the
+        # TaskTemplate). To be more rigorous, we write an explicit identifier file that reflects the choices (like
+        # project/domain, etc.) made for this serialize call. We should not allow users to specify a different project
+        # for instance come registration time, to avoid mismatches between potential internal ids like the TaskTemplate
```
katrogan, line 113: i think admin populates the id in the task template anyways

flytekit/clis/sdk_in_container/register.py
wild-endeavor: sorry didn't see this, yeah i did, it's helpful maybe one day.

flytekit/clis/sdk_in_container/serialize.py
wild-endeavor: true.

PR #108, opened 2020-05-05
File-based registration

wild-endeavor:
# TL;DR
As detailed in the linked issue, this will take the output of the [previous pr](https://github.com/lyft/flytekit/pull/104) and register them with `flyte-cli`.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This uses the same core logic as in the old registration path.  The only difference is that the data is read from files - files which follow a specific set of rules (detailed in help) - and the `to_flyte_idl()` call has already been made.

## Tracking Issue
https://github.com/lyft/flyte/issues/225

## Follow-up issue
NA

### DIFF
--- a/flytekit/clients/friendly.py
+++ b/flytekit/clients/friendly.py
@@ -5,15 +5,40 @@ import six as _six
 from flyteidl.admin import task_pb2 as _task_pb2, common_pb2 as _common_pb2, workflow_pb2 as _workflow_pb2, \
     launch_plan_pb2 as _launch_plan_pb2, execution_pb2 as _execution_pb2, node_execution_pb2 as _node_execution_pb2, \
     task_execution_pb2 as _task_execution_pb2, project_pb2 as _project_pb2
+from flyteidl.core import identifier_pb2 as _identifier_pb2
 
 from flytekit.clients.raw import RawSynchronousFlyteClient as _RawSynchronousFlyteClient
 from flytekit.models import filters as _filters, common as _common, launch_plan as _launch_plan, task as _task, \
     execution as _execution, node_execution as _node_execution
 from flytekit.models.core import identifier as _identifier
 from flytekit.models.admin import workflow as _workflow, task_execution as _task_execution
+from flytekit.common.exceptions.user import FlyteAssertion as _FlyteAssertion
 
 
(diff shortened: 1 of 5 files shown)


flytekit/clients/friendly.py
```diff
@@ -5,16 +5,41 @@
...
 from flytekit.models import filters as _filters, common as _common, launch_plan as _launch_plan, task as _task, \
     execution as _execution, node_execution as _node_execution
 from flytekit.models.core import identifier as _identifier
 from flytekit.models.admin import workflow as _workflow, task_execution as _task_execution
+from flytekit.common.exceptions.user import FlyteAssertion as _FlyteAssertion
 
 
 class SynchronousFlyteClient(_RawSynchronousFlyteClient):
 
+    def register_entities(self, id_entity_list):
```
matthewphsmith, line 20: I think this is a little weird to have in the client because it is a non-transactional method and doesn't map directly to a single API call.

wild-endeavor: i can definitely move this back to the core cli code.

scripts/flytekit_build_image.sh
```diff
@@ -61,8 +61,11 @@ if [ -n "$REGISTRY" ]; then
   if [[ "${REGISTRY}" == "docker.io"*  && -z "${NOPUSH}" ]]; then
     docker login --username="${DOCKERHUB_USERNAME}" --password="${DOCKERHUB_PASSWORD}"
   fi
-  docker push "${FLYTE_INTERNAL_IMAGE}"
-  echo "${FLYTE_INTERNAL_IMAGE} pushed to remote"
+
+  if [[ -z "${NOPUSH}" ]]; then
```
wild-endeavor, line 65: Clobbing this along this PR - just a minor bug fix that makes the building script respect the `NOPUSH` flag.

flytekit/clients/friendly.py
```diff
@@ -51,6 +53,35 @@ def create_task(
             )
         )
 
+    def create_task_raw(
```
matthewphsmith, line 56: Why do we need to duplicate this? if you want to use this, why not use the raw client directly?

PR #109, opened 2020-05-07
Add condition to generation of dynamic job spec

wild-endeavor:
# TL;DR
Please see the issue for a proper description.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
If a launch plan/subworkflow node does not have outputs, currently there is no dynamic job spec created.  This was the main issue this PR is meant to address.

Also discovered that if the workflow/launchplan yielded in a dynamic node does not have outputs, but does have inputs, those inputs would not be generated.  This was from improper indentation from the first time this code was created.

This code needs to be cleaned up... a lot.  See the follow-up issue.

## Tracking Issue
https://github.com/lyft/flyte/issues/282

## Follow up issue
https://github.com/lyft/flyte/issues/226

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -1,4 +1,4 @@
 from __future__ import absolute_import
 import flytekit.plugins
 
-__version__ = '0.7.0'
+__version__ = '0.7.1'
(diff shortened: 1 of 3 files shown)


feiyu-meng:
Thanks!

PR #112, opened 2020-05-08
pytorch plugin implementation

igorvalko:
# TL;DR
Pytorch plugin for Flyte that leverages https://github.com/kubeflow/pytorch-operator to run distributed pytorch jobs on k8s.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [X] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description
Pytorch plugin for Flyte that leverages https://github.com/kubeflow/pytorch-operator to run distributed pytorch jobs on k8s.
Related PRs:
https://github.com/lyft/flyteidl/pull/61
https://github.com/lyft/flyteplugins/pull/85

## Tracking Issue
_NA_

## Follow-up issue
_NA_


flytekit/plugins/__init__.py
```diff
@@ -46,3 +48,9 @@
     ],
     [hmsclient]
 )
+
+_lazy_loader.LazyLoadPlugin(
+    "pytorch",
+    ["torch>=1.0.0,<1.6.0"],
```
EngHabu, line 54: Is there a breaking change after 1.6? generally speaking, if they are following semantic versioning, up to 2.0.0 should be fine, right?

flytekit/sdk/tasks.py
```diff
@@ -1000,3 +1001,156 @@ def wrapper(fn):
...
+def pytorch_task(
+        _task_function=None,
+        cache_version='',
+        retries=0,
+        interruptible=True,
+        deprecated='',
+        cache=False,
+        timeout=None,
+        workers_count=1,
+        instance_storage_request="",
```
EngHabu, line 1015: Any reason to call it "instance_" ? is it to signify this request is per worker instead of for the entire task?

flytekit/sdk/tasks.py
```diff
@@ -1000,3 +1001,156 @@ def wrapper(fn):
         return wrapper(_task_function)
     else:
         return wrapper
+
+
+def pytorch_task(
+        _task_function=None,
+        cache_version='',
+        retries=0,
+        interruptible=True,
```
EngHabu, line 1010: Is it so? does it have the concept of checkpointing?

flytekit/common/tasks/pytorch_task.py
```diff
@@ -0,0 +1,121 @@
...
+import six as _six
+from flytekit.common import constants as _constants
+from flytekit.common.exceptions import scopes as _exception_scopes
+from flytekit.common.tasks import output as _task_output, sdk_runnable as _sdk_runnable
+from flytekit.common.types import helpers as _type_helpers
+from flytekit.models import literals as _literal_models, task as _task_models
+from google.protobuf.json_format import MessageToDict as _MessageToDict
+
+
+class SdkRunnablePytorchkContainer(_sdk_runnable.SdkRunnableContainer):
```
wild-endeavor, line 17: what does the k stand for?

flytekit/common/tasks/pytorch_task.py
```diff
@@ -0,0 +1,121 @@
...
+    def _get_container_definition(
+            self,
+            **kwargs
+    ):
+        """
+        :rtype: SdkRunnableSparkContainer
+        """
+        return super(SdkPyTorchTask, self)._get_container_definition(cls=SdkRunnablePytorchkContainer, **kwargs)
+
+    def _get_kwarg_inputs(self):
```
wild-endeavor, line 119: Is this different than in the base SdkRunnableTask class?  Is it supposed to be?

flytekit/common/tasks/pytorch_task.py
```diff
@@ -0,0 +1,121 @@
...
+            gpu_limit=instance_gpu_limit,
+            memory_limit=instance_memory_limit,
+            discoverable=discoverable,
+            timeout=timeout,
+            environment=environment,
+            custom=_MessageToDict(pytorch_job)
+        )
+
+    @_exception_scopes.system_entry_point
+    def execute(self, context, inputs):
```
wild-endeavor, line 74: sorry same thing here, is this function different than what happens in the base class?

flytekit/common/tasks/pytorch_task.py
```diff
@@ -0,0 +1,121 @@
...
+                literals={k: v.sdk_value for k, v in _six.iteritems(outputs_dict)}
+            )
+        }
+
+    def _get_container_definition(
+            self,
+            **kwargs
+    ):
+        """
+        :rtype: SdkRunnableSparkContainer
```
matthewphsmith, line 115: SdkRunnablePytorchkContainer

flytekit/common/tasks/pytorch_task.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        return super(SdkPyTorchTask, self)._get_container_definition(cls=SdkRunnablePytorchkContainer, **kwargs)
+
+    def _get_kwarg_inputs(self):
+        # Trim off first parameter as it is reserved for workflow_parameters
+        return set(_getargspec(self.task_function).args[1:])
```
matthewphsmith, line 121: nitpick: end with a newline. don't know if we have linters set up yet for these things.

flytekit/plugins/__init__.py
```diff
@@ -17,6 +17,8 @@
 hmsclient = _lazy_loader.lazy_load_module("hmsclient")  # type: types.ModuleType
 type(hmsclient).add_sub_module("genthrift.hive_metastore.ttypes")
 
+torch = _lazy_loader.lazy_load_module("torch")  # type: types.ModuleType
```
matthewphsmith, line 20: could you mention this in the README.md so people know it is available!

flytekit/sdk/tasks.py
igorvalko: Yes, you've got it right: it's just to stress an accent that it's for requesting resources per instance (both for worker(s) and master)

igorvalko:
Thanks everyone for the review. Latest changes should resolve all the requests.

flytekit/sdk/tasks.py
kumare3: maybe for the user we could provide an interface that says "master_cpu" "worker_cpu". But for now internally map it. Eventually I think we want to use - https://github.com/pytorch/elastic/tree/master/kubernetes, which will not require the master/worker model

kumare3: actually thinking more about it, lets replace "instance" -> "node"? and that should be it. WE could move to elastic later on?

igorvalko: Huh, I see that in related docs (at least [here](https://pytorch.org/tutorials/intermediate/dist_tuto.html) and [here](https://www.kubeflow.org/docs/components/training/pytorch/)) processes in distributed job are referred as `replicas`. Hence, may be let's go with `per_replica_*`? Any objections?

EngHabu: Do we create master/workers? I don't see that, I only see "workers" created by the operator.
Either way, I think the naming between `X_count` and `per_X_cpu_request` should be consistent... I'm fine with worker, so `workers_count` and `per_worker_cpu_request`... etc. The operator docs are all over the place and inconsistent. Look at this line `You can scale the number of nodes by adjusting .spec.replicaSpecs[Worker].replicas and applying the change.` they have 3 names for pretty much the same thing in one sentence... 
I'm fine with replicas too... as long as we change the count to be `replicas_count` and `per_replica_cpu_request` ... etc.

igorvalko: > I only see "workers" created by the operator
Are you referring to extra argument `workers_count`? There is no respective argument for master since there could be the only one. But eventually 1 master and N workers gonna be spawned.

wild-endeavor: i think Haytham means that we should change all the `instance_blah_limit` to `per_worker_blah_limit`

wild-endeavor: since these are per worker right?

kumare3: @EngHabu I think what @igorvalko is trying to do is use same cpu/mem for both master and worker. Initially he had all resources as part of the cRD and i said you could re-purpose our "container" resources for one of them.
So Igor made them the same. 
Now maybe the way to think about this this, the container in "flyte" for a distributed job refers to the master or worker?

EngHabu: I would like to avoid the Spark issue we had where people would assign huge resources to the "driver" pod the same way they assign them to the executor pods but they end up barely using any on the driver pod... by essentially forcing users to use the same for both, are we getting into the same situation?

igorvalko: I have no hands-on experience with pytorch, but it looks like master in pytorch does the same kind of job as workers with some extra communication arrangement work. In spark different approach is used: driver in most cases is a lightweight coordinator and all the heavy stuff happens on executors.
Hence I assume that's uncommon thing to request uneven resources for master and workers. Input from anyone having experience with distributed pytorch would be very welcome.

bnsblue: From what I read, I think there are two possible paradigms: (1) a master node does what a worker does + the coordination and the communication, (2) a master node does only the reduction. But I think either of these requires the master node to have the same/similar resource requirement as workers.

bnsblue: (I am assuming we are using `torch.nn.parallel.DistributedDataParallel()` which builds on top of `torch.distributed` package, which should supposedly use the NCCL backend according to https://pytorch.org/docs/stable/distributed.html#which-backend-to-use)

PR #114, opened 2020-05-18
Add _sync_closure call to reduce load from wait_for_completion calls.

katrogan:
# TL;DR
The `wait_for_completion` method calls a full sync every poll period even though a full sync is really only needed after the underlying closure indicates the artifact has transitioned to complete.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/312

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/common/nodes.py
+++ b/flytekit/common/nodes.py
@@ -414,3 +414,11 @@ class SdkNodeExecution(
                 _task_executions.SdkTaskExecution.promote_from_model(te) for te in ne.get_task_executions()
             ]
             # TODO: Sub-workflows too once implemented
+
+    def _sync_closure(self):
+        """
+        Syncs the closure of the underlying execution artifact with the state observed by the platform.
+        :rtype: None
+        """
+        ne = _engine_loader.get_engine().get_node_execution(self)
+        ne.sync()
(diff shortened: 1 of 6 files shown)


flytekit/plugins/__init__.py
```diff
@@ -25,7 +25,7 @@
 
 _lazy_loader.LazyLoadPlugin(
     "sidecar",
-    ["k8s-proto>=0.0.2,<1.0.0"],
+    ["k8s-proto>=0.0.3,<1.0.0"],
```
katrogan, line 28: This is unrelated but matches what we have internally

flytekit/common/nodes.py
```diff
@@ -414,3 +414,11 @@ def sync(self):
...
             ]
             # TODO: Sub-workflows too once implemented
+
+    def _sync_closure(self):
+        """
+        Syncs the closure of the underlying execution artifact with the state observed by the platform.
+        :rtype: None
+        """
+        ne = _engine_loader.get_engine().get_node_execution(self)
+        ne.sync()
```
wild-endeavor, line 424: so node execution sync() just syncs the closure?  and there's no need to return anything?

wild-endeavor: indeed it does, cool
```
self.sdk_node_execution._closure = client.get_node_execution(self.sdk_node_execution.id).closure
```

katrogan:
> Can you test locally to make sure it works?

yep, done. i can share the script & execution if that's useful.

PR #121, opened 2020-06-04
Improve logging in friendly client

wild-endeavor:
# TL;DR
In cases where we have a registration conflict, we should print out what it is that the user was trying to register so that the user can compare against what's stored in Admin.

In the future, we should have Admin return a proper diff.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Adding logging.
Auto-formatted

## Tracking Issue
https://github.com/lyft/flyte/issues/340

## Follow-up issue
NA

### DIFF
--- a/flytekit/clients/friendly.py
+++ b/flytekit/clients/friendly.py
@@ -1,6 +1,9 @@
 from __future__ import absolute_import
 
 import six as _six
+import logging as _logging
+import grpc as _grpc
+from google.protobuf.json_format import MessageToJson as _MessageToJson
 
 from flyteidl.admin import task_pb2 as _task_pb2, common_pb2 as _common_pb2, workflow_pb2 as _workflow_pb2, \
     launch_plan_pb2 as _launch_plan_pb2, execution_pb2 as _execution_pb2, node_execution_pb2 as _node_execution_pb2, \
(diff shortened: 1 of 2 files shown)


kumare3:
👍 LGTM

flytekit/clients/friendly.py
```diff
@@ -353,12 +370,19 @@ def create_launch_plan(
...
+        create_request = _launch_plan_pb2.LaunchPlanCreateRequest(
+            id=launch_plan_identifer.to_flyte_idl(),
+            spec=launch_plan_spec.to_flyte_idl()
         )
+        try:
+            super(SynchronousFlyteClient, self).create_launch_plan(create_request)
+        except _grpc.RpcError as e:
+            # When Admin detects a conflicting entry or otherwise invalid request during registration, it returns an
+            # invalid argument code. We should print out the thing we were trying to create in these cases.
+            if e.code() == _grpc.StatusCode.INVALID_ARGUMENT:
```
kumare3, line 382: why not dump all other error codes here? This way we could probably make the error message better?

wild-endeavor:
Did a bit of a compromise.  @matthewphsmith ptal - I moved it to the raw client, and i'm using a decorator, but not the same decorator.  I think it's better to have a separate decorator than pick through *args/**kwargs.

PR #123, opened 2020-06-09
Adding SageMaker samples as tests

bnsblue:
# PLEASE HELP COMMENTING ON the sample usages in `tests/flytekit/common/workflows/sagemaker.py`

# TL;DR
This PR adds some sample usages for SageMaker tasks as tests. This PR serves as a standalone feedback collection PR, which might not be merged. The possible outcome is that the code in this PR will be copied to https://github.com/lyft/flytekit/pull/120 and merged together with https://github.com/lyft/flytekit/pull/120

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,192 @@
...
+@inputs(
+    extra_input1=Types.Integer,
+    extra_input2=Types.Integer,
+)
+@outputs(
+    extra_output1=Types.Blob,
+)
+@trainingjob_task(
+    trainingjob_conf={
+        "InstanceType": "ml.m4.xlarge",
```
chengxuanw, line 96: Customer maybe more inclined to specify cores and memory instead of `InstanceType`, since they may don't know how much resource the instance have.

bnsblue: Thanks for the comment! 

I had briefly discussed this with Ketan and Haythem. The problem is that SageMaker currently doesn't support fractional resources (that is, you have to be billed for the entire instance even though your program uses only a part of it). Basically it kind of conflicts with the current semantic in flytekit around resource. So If we allow users to specify  `cpu_limit` and `memory_limit` for SageMaker tasks, users might be misled and found that they are "overcharged" for the resource they didn't claim.

Exposing `InstanceType` and `InstanceCount` to the users is currently the most viable workaround I can think of to address the above issue. 

Any suggestion? :D

chengxuanw: got it. I don't know the solution either. :D
I was thinking calculate similarity based on user's `cpu_limit` and `memory_limit` between each instance type's. and choose the smallest based on some distance ( maybe x*y/||x||*||y||. (x= (core, memory)) ). but thought it is too complicated and error prone. 

so really i don't know. :(

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+        "InstanceCount": 1,
+        "VolumeSizeInGB": 25,
+    },
+    stopping_condition={
+        "MaxRuntimeInSeconds": 43200,
+        "MaxWaitTimeInSeconds": 43200
+    },
+    algorithm_specification={
+        "TrainingInputMode": "File",
```
kumare3, line 67: what are the options here? Should this be an enumeration?

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+        "AlgorithmName": "xgboost",
```
kumare3, line 68: samething, an enumeration and for algorithm besides the one supported directly (byo) just have "OTHERS"/"BYO" as an option

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+        "Version": "0.72",
```
kumare3, line 69: what is this version for?

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+    #
+    # This aforementioned billing policy of SageMaker conflicts with the current semantic
+    # in flytekit around resources. So If we allow users to specify cpu_limit and memory_
+    # limit for SageMaker tasks, users might be misled and found that they are "overcharged"
+    # for the resource they didn't claim.
+    #
+    # Exposing InstanceType and InstanceCount to the users seems to be the most viable workaround
+    # I can think of at the moment to address the above issue.
+    trainingjob_conf={
+        "InstanceType": "ml.m4.xlarge",
```
kumare3, line 58: I agree, sadly this is Sagemaker and this is what we should expose

tests/flytekit/common/workflows/sagemaker.py
bnsblue: Algorithm/framework Version. For example PISCO requires XGBoost to be pinned at 0.72 for downstream compatibility (e.g. modelexec)

tests/flytekit/common/workflows/sagemaker.py
bnsblue: Yes this should be a enum

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+    b = Input(Types.Integer, required=False, default="1", help="Test integer with default")
+
+    my_simple_xgboost_trainingjob_task = simple_xgboost_trainingjob_task(
+        train='s3://my-bucket/training.csv',
+        validation='s3://my-bucket/validation.csv',
+        static_hyperparameters=example_hyperparams,
+    )
+
+    my_simple_hpojob_task = simple_hpojob_task(
+        trainingjob_task=my_simple_xgboost_trainingjob_task,  # TODO: this doesn't seem totally reasonable
```
bnsblue, line 169: @kumare3 I don't think this works actually... but I haven't come up with a intuitive way to do this

tests/flytekit/common/workflows/sagemaker.py
kumare3: my preference is to keep the interface like Sagemaker has with the instance ID, we can document that (just point to aws documentation) and if they eventually support cpu_limit then we can support it

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+        extra_input2=b,
+        train='s3://my-bucket/training.csv',
+        validation='s3://my-bucket/validation.csv',
+        static_hyperparameters=example_hyperparams,
+    )
+
+    my_custom_task = simple_hpojob_task(
+        trainingjob_task=my_custom_trainingjob_task,  # TODO: this doesn't seem totally reasonable
+        hyperparameter_ranges={  # hyperparameters included here will automatically override the static counterpart
+            # TODO: Need to figure out how to do decide precedence
```
kumare3, line 193: what does the TODO mean?

bnsblue: This is like a reminder for myself that the implementation of this might be tricky for BYOC use case. 

For example, if in `custom_trainingjob_task` user has some code to calculate the value of a hyperparameter, then what is specified might be overriden if not implemented carefully

tests/flytekit/common/workflows/sagemaker.py
bnsblue: I think one example here is that if a user specifies `cpu_request=5`, but there are only a 4-core option and a 8-core option available among all the instance types. 

In this case, we cannot choose the 4-core option because users specifies a minimum requirement of 5 cores; therefore, we have to choose the 8-core option because that's the smallest instance that satisfies user's requirement. That choice, unfortunately, causes overcharging; the bigger problem is that this overcharging is not easily visible to the users and may causes "surprises" :(

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+    algorithm_specification={
+        "TrainingInputMode": "File",
+        "AlgorithmName": "xgboost",
+        "Version": "0.72",
+    },
+    cache_version='1',
+    cachable=True,
+)
+
+simple_hpojob_task = SdkSimpleHPOJobTask(
```
bnsblue, line 75: Wrap trainingjob in this

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+        "InstanceCount": 1,
+        "VolumeSizeInGB": 25,
+    },
+    stopping_condition={
+        "MaxRuntimeInSeconds": 43200,
+        "MaxWaitTimeInSeconds": 43200
+    },
+    algorithm_specification={
+        "TrainingInputMode": "File",
+        "AlgorithmName": "xgboost",
```
bnsblue, line 117: don't need this

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+    with train as reader:
+        train_df = reader.read(concat=True)
+        dtrain_x = xgb.DMatrix(train_df[:-1])
+        dtrain_y = xgb.DMatrix(train_df[-1])
+    with validation as reader:
+        validation_df = reader.read(concat=True)
+        dvalidation_x = xgb.DMatrix(validation_df[:-1])
+        dvalidation_y = xgb.DMatrix(validation_df[-1])
+
+    static_hyperparameters["num_round"] = extra_input1 + extra_input2
```
bnsblue, line 144: This is an anti-pattern

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+    for i in enumerate(y_predicted):
+        if y_predicted[i] != y_true[i]:
+            err_count += 1
+
+    return "err rate", err_count/len(y_predicted)
+
+
+@inputs(
+    extra_input1=Types.Integer,
+    extra_input2=Types.Integer,
```
bnsblue, line 100: Look into how to support this

bnsblue: rename to `custom_input`

tests/flytekit/common/workflows/sagemaker.py
```diff
@@ -0,0 +1,203 @@
...
+
+    my_simple_xgboost_trainingjob_task = simple_xgboost_trainingjob_task(
+        train='s3://my-bucket/training.csv',
+        validation='s3://my-bucket/validation.csv',
+        static_hyperparameters=example_hyperparams,
+    )
+
+    my_simple_hpojob_task = simple_hpojob_task(
+        trainingjob_task=my_simple_xgboost_trainingjob_task,  # TODO: this doesn't seem totally reasonable
+        hyperparameter_ranges={   # hyperparameters included here will automatically override the static counterpart
```
bnsblue, line 170: Allow this to be an input

bnsblue:
We do not aim to merge this. This PR serves as a place to host discussions around the design of the SageMaker-on-Flyte user interface.

PR #126, opened 2020-06-16
Register and launch single task execution

katrogan:
# TL;DR
This change implements register_and_launch for single task executions featuring containerless tasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The use case for this feature is in-notebook task execution. Because only containerless tasks do not need to update an image during task iteration, only those are supported for register_and_launch.

## Tracking Issue
https://github.com/lyft/flyte/issues/256

## Follow-up issue
https://github.com/lyft/flyte/issues/221

### DIFF
--- a/flytekit/common/mixins/registerable.py
+++ b/flytekit/common/mixins/registerable.py
@@ -158,12 +158,13 @@ class RegisterableEntity(_six.with_metaclass(_InstanceTracker, object)):
                     self._platform_valid_name = _utils.fqdn(m.__name__, k, entity_type=self.resource_type)
                     _logging.debug("Auto-assigning name to {}".format(self._platform_valid_name))
                     return
-            except ValueError:
+            except ValueError as err:
                 # Empty pandas dataframes behave weirdly here such that calling `m.df` raises:
                 # ValueError: The truth value of a {type(self).__name__} is ambiguous. Use a.empty, a.bool(), a.item(),
                 #   a.any() or a.all()
                 # Since dataframes aren't registrable entities to begin with we swallow any errors they raise and
                 # continue looping through m.
+                _logging.warning("Caught ValueError {} while attempting to auto-assign name".format(err))
                 pass
 
         _logging.error("Could not auto-assign name")
--- a/flytekit/common/mixins/registerable.py
+++ b/flytekit/common/mixins/registerable.py
@@ -153,6 +153,7 @@ class RegisterableEntity(_six.with_metaclass(_InstanceTracker, object)):
         m = _importlib.import_module(self.instantiated_in)
 
         for k in dir(m):
+            print("calling getattr({}, {})".format(m, k)
             if getattr(m, k) == self:
                 self._platform_valid_name = _utils.fqdn(m.__name__, k, entity_type=self.resource_type)
                 _logging.debug("Auto-assigning name to {}".format(self._platform_valid_name))
--- a/flytekit/common/mixins/registerable.py
+++ b/flytekit/common/mixins/registerable.py
@@ -153,11 +153,18 @@ class RegisterableEntity(_six.with_metaclass(_InstanceTracker, object)):
         m = _importlib.import_module(self.instantiated_in)
 
         for k in dir(m):
-            print("calling getattr({}, {})".format(m, k)
-            if getattr(m, k) == self:
-                self._platform_valid_name = _utils.fqdn(m.__name__, k, entity_type=self.resource_type)
-                _logging.debug("Auto-assigning name to {}".format(self._platform_valid_name))
-                return
+            try:
+                if getattr(m, k) == self:
+                    self._platform_valid_name = _utils.fqdn(m.__name__, k, entity_type=self.resource_type)
+                    _logging.debug("Auto-assigning name to {}".format(self._platform_valid_name))
+                    return
+            except ValueError:
+                # Empty pandas dataframes behave weirdly here such that calling `m.df` raises:
+                # ValueError: The truth value of a {type(self).__name__} is ambiguous. Use a.empty, a.bool(), a.item(),
+                #   a.any() or a.all()
+                # Since dataframes aren't registrable entities to begin with we swallow any errors they raise and
+                # continue looping through m.
+                pass
 
         _logging.error("Could not auto-assign name")
         raise _system_exceptions.FlyteSystemException("Error looking for object while auto-assigning name.")
--- a/flytekit/common/tasks/presto_task.py
+++ b/flytekit/common/tasks/presto_task.py
@@ -117,11 +117,20 @@ class SdkPrestoTask(_base_task.SdkTask):
         # Set user provided inputs
         task_inputs(self)
 
+    def _add_implicit_inputs(self, inputs):
+        """
+        :param dict[Text,Any] inputs:
+        :param inputs:
+        :return:
+        """
+        inputs["__implicit_routing_group"] = self.routing_group
+        inputs["__implicit_catalog"] = self.catalog
+        inputs["__implicit_schema"] = self.schema
+        return inputs
+
     # Override method in order to set the implicit inputs
     def __call__(self, *args, **kwargs):
-        kwargs["__implicit_routing_group"] = self.routing_group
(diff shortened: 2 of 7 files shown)


flytekit/common/tasks/presto_task.py
```diff
@@ -122,6 +127,21 @@ def __call__(self, *args, **kwargs):
             *args, **kwargs
         )
 
+    # Override method in order to set the implicit inputs
+    def _python_std_input_map_to_literal_map(self, inputs):
+        """
```
wild-endeavor, line 132: Can you explain what the stuff in the __call__ function does and why it's not enough please, for me and for posterity to understand?

katrogan: ordinarily we use `__call__` to create an SDK node, which is formerly the only the way a task was transformed for execution.

but when we call `launch` on a task to trigger a single task execution [here](https://github.com/lyft/flytekit/blob/master/flytekit/common/mixins/launchable.py#L32) we don't use this code path and the implicit inputs are never added

flytekit/models/common.py
```diff
@@ -56,7 +56,7 @@ def __str__(self):
         return self.verbose_string()
 
     def __hash__(self):
-        return hash(self.to_flyte_idl().SerializeToString())
+        return hash(self.to_flyte_idl().SerializeToString(deterministic=True))
```
wild-endeavor, line 59: @matthewphsmith is there a reason this wasn't here initially?

flytekit/common/tasks/task.py
```diff
@@ -268,6 +273,58 @@ def _python_std_input_map_to_literal_map(self, inputs):
             for k, v in _six.iteritems(self.interface.inputs)
         })
 
+    def _produce_deterministic_version(self, version=None):
+        """
+        :param Text version:
+        :return Text:
+        """
+
+        if self.container is not None:
```
wild-endeavor, line 282: This won't work for Ketan's [raw container tasks](https://github.com/lyft/flytekit/pull/107/files#diff-0b4aabc3c6488fa90c0e7f3afca65ae7R214) btw.  How about a whitelist of allowed task types (strings) instead?

flytekit/common/tasks/task.py
```diff
@@ -268,6 +273,58 @@ def _python_std_input_map_to_literal_map(self, inputs):
...
+    def register_and_launch(self, project, domain, name=None, version=None, inputs=None):
+        """
+        :param Text project: The project in which to register and launch this task.
+        :param Text domain: The domain in which to register and launch this task.
+        :param Text name: The name to give this task.
+        :param Text version: The version in which to register this task
+
+        :param dict[Text, Any] inputs: A dictionary of Python standard inputs that will be type-checked, then compiled
+            to a LiteralMap.
+        """
```
wild-endeavor, line 306: Add the return type?

flytekit/common/tasks/task.py
```diff
@@ -268,6 +273,58 @@ def _python_std_input_map_to_literal_map(self, inputs):
...
+        self.validate()
+        version = self._produce_deterministic_version(version)
+
+        if name is None:
+            try:
+                self.auto_assign_name()
+                generated_name = self._platform_valid_name
```
wild-endeavor, line 313: can remove the _ if you want

flytekit/common/tasks/task.py
kumare3: @wild-endeavor raw container tasks are not task type. You just have to check a different, thing, as this proposal
```suggestion
        if self.container is not None and self.container. data_config is None:
```

flytekit/common/tasks/task.py
kumare3: I want to start using python3 typing.

flytekit/models/common.py
kumare3: ohh wow this looks like a bug? But, we never really use this right, we always use the git-sha as the version

kumare3:
Is there an example execution of single task from command line / jupyter?

flytekit/common/tasks/presto_task.py
EngHabu: Should we "unify" the setting of those implicit inputs between those two call paths? maybe make __call__ call this function or something?

flytekit/common/tasks/task.py
EngHabu: Is it true that client side task versions do not work with container tasks? why not? I thought that's exactly what we do today... (we grab the Image version and make it the task version)...

katrogan:
> Is there an example execution of single task from command line / jupyter?

@kumare3 adding docs separately :) https://github.com/lyft/flyte/pull/354

flytekit/common/tasks/presto_task.py
katrogan: `__call__` only creates bindings for inputs, rather than converting them to sdk types but good point i can refactor the behavior to a shared function

flytekit/common/tasks/task.py
katrogan: done

flytekit/common/tasks/task.py
katrogan: thanks for the catch/suggestion! done

kumare3:
@katrogan do we want to add the example notebook that you had to this? Or maybe I will just update my raw container task example to use your single task execution :)

katrogan:
@kumare3 do we have a designated place in the docs or flytekit for adding example notebooks?

kumare3:
> @kumare3 do we have a designated place in the docs or flytekit for adding example notebooks?

I am actually creating one in my Raw Container PR. For now I think you can continue merge it, I will use it in my PR

flytekit/models/common.py
wild-endeavor: yeah git sha is used for task version, this is just used for equality.

flytekit/common/mixins/registerable.py
```diff
@@ -153,10 +153,18 @@ def some_task()
...
+                    self._platform_valid_name = _utils.fqdn(m.__name__, k, entity_type=self.resource_type)
+                    _logging.debug("Auto-assigning name to {}".format(self._platform_valid_name))
+                    return
+            except ValueError:
+                # Empty pandas dataframes behave weirdly here such that calling `m.df` raises:
+                # ValueError: The truth value of a {type(self).__name__} is ambiguous. Use a.empty, a.bool(), a.item(),
+                #   a.any() or a.all()
+                # Since dataframes aren't registrable entities to begin with we swallow any errors they raise and
+                # continue looping through m.
+                pass
```
wild-endeavor, line 167: Can you add a debugging log here just in case.

flytekit/common/tasks/task.py
wild-endeavor: I think revamping everything to use Python 3 typing will be a separate effort - one that will probably take a while to do.  We can incrementally add along the way sure, either way, no preference.

flytekit/common/tasks/task.py
```diff
@@ -268,6 +273,59 @@ def _python_std_input_map_to_literal_map(self, inputs):
             for k, v in _six.iteritems(self.interface.inputs)
         })
 
+    def _produce_deterministic_version(self, version=None):
+        """
+        :param Text version:
+        :return Text:
+        """
+
+        if self.container is not None and self.container.data_config is None:
```
wild-endeavor, line 282: so no white list?  can you add a comment explaining why data config can't be none?

flytekit/common/mixins/registerable.py
katrogan: done

flytekit/common/tasks/task.py
katrogan: done. i'd rather not add a whitelist because barring the raw container task anything that's containerless (e.g. Sagemaker) in the future should just seamlessly be able to call this function without a code change here

PR #139, opened 2020-07-16
Add a --watch argument to execute-launch-plan command and add a watch-execution command

bnsblue:
# TL;DR
Adding a command `execute-launch-plan-and-wait`, by using which users kick off an execution and enforce a wait for the completion of the execution. This can be useful for kicking off workflows as tests in CI.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/clis/flyte_cli/main.py
+++ b/flytekit/clis/flyte_cli/main.py
@@ -1087,9 +1087,8 @@ def update_launch_plan(state, host, insecure, urn=None):
 @_principal_option
 @_verbose_option
 @_watch_option
-@_optional_urns_only_option
 @_click.argument('lp_args', nargs=-1, type=_click.UNPROCESSED)
-def execute_launch_plan(project, domain, name, host, insecure, urn, principal, verbose, watch, urns_only, lp_args):
+def execute_launch_plan(project, domain, name, host, insecure, urn, principal, verbose, watch, lp_args):
     """
     Kick off a launch plan. Note that the {project, domain, name} specified in the command line
     will be for the execution.  The project/domain for the launch plan are specified in the urn.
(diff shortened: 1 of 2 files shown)


kumare3:
I think we should not add a combined method. We should have an execute and a watch/wait as separate methods

wild-endeavor:
yeah i second ketan's thought.  if we want to do this, we should just add a watch command that can watch any execution that's given (probably in urn form).

This means that there will need to be logic that handles the handoff between the launching and the watching.  I'm okay either letting them handle that for now, or adding a 'quiet' flag to execute that makes it so that the only output is the urn, and then you can | it to watch.

And we should call it watch instead of wait.

Alternatively we can add a flag to the get-execution command, -f for follow.

@kumare3

flytekit/clis/flyte_cli/main.py
```diff
@@ -1098,17 +1110,42 @@ def execute_launch_plan(project, domain, name, host, insecure, urn, principal, v
...
 
-        inputs = _construct_literal_map_from_parameter_map(lp.default_inputs, _parse_args_into_dict(lp_args))
-        # TODO: Implement notification overrides
-        # TODO: Implement label overrides
-        # TODO: Implement annotation overrides
-        execution = lp.launch_with_literals(project, domain, inputs, name=name)
-        _click.secho("Launched execution: {}".format(_tt(execution.id)), fg='blue')
-        _click.echo("")
+
+@_flyte_cli.command('execute-launch-plan-and-wait', cls=_FlyteSubCommand)
```
EngHabu, line 1118: I think this should just be a flag on the existing execute-lp command (`--wait`)

PR #144, opened 2020-07-24
Remove test breaking numpy 1.19

wild-endeavor:
# TL;DR
Basically if you `pip install numpy==1.18` this works.

```
python -m pytest -svv tests/flytekit/unit/common_tests/types/impl/test_schema.py::test_type_promoted_schema_read_with_fastparquet                                                                                  
```

If you `pip install numpy==1.19`

it fails.  Basically in 1.18 and earlier, if you declared a bool column and had Nones in the data, it would automatically promote the column to an object.  Starting with 1.19, they're no longer do this.  The 1.19 behavior makes more sense IMO, since this is how most typed languages work.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Removed the offending test.

## Tracking Issue
https://github.com/lyft/flyte/issues/430

## Follow-up issue
In working on this, also discovered that the fastparquet test wasn't actually setting the fastparquet engine.  Doing so ended up failing the test though it might be a simple case of missing dependencies.  Created this issue to follow up later.
https://github.com/lyft/flyte/issues/433

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -2,4 +2,4 @@ from __future__ import absolute_import
 
 import flytekit.plugins
 
-__version__ = "0.10.8"
+__version__ = "0.10.9"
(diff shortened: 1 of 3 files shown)


akhurana001:
Can we merge this please ?

PR #149, opened 2020-08-01
[Bug fix] remove erroneous decorator code

bnsblue:
# TL;DR
Remove an erroneously checked-in decorator.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
In the previous PR I erroneously checked in a decorate that is supposed to be pushed into another PR, which is in sdk/tasks so it causes failure when using other decorators such as python_task

## Follow-up issue
_NA_


bnsblue:
@wild-endeavor I just ran optimize import on sagemaker files. Could you please take another look?

PR #153, opened 2020-08-11
Refactor interfaces and engine code

wild-endeavor:
# TL;DR
Changes are:

* Formalizing the `SdkTask`, `SdkWorkflow`, and `SdkLaunchPlan` classes as control plane smart objects.
* Removal of the engine layer as it pertains to accessing the control plane from those smart objects.  For instance, `SdkTask` would directly access the Flyte Admin client to `fetch` a task instead of going through the `FlyteTask` wrapper layer.
* Introduction of a new `SdkRunnableWorkflow` class mean to encapsulate the local workflow object (i.e. you only have this object when you have the code locally available, and imported into the Python process).  Despite the name, it has no bearing on whether or not a workflow is runnable, either locally or otherwise.
    * This is in a separate file and user-space level objects like `Output` was also moved, but we don't have to if it breaks people's imports.
    * The original constructor to `SdkWorkflow` has changed.  Hopefully not too many people are using it.
* Cleaning up the abstract base classes by creating a `Registerable` ABC meant to encapsulate control plane capability and a `LocalEntity` ABC meant to encapsulate local-only functionality.
* Removing all the `future` stuff.  We only used stuff that is [mandatory in 3.0](https://docs.python.org/2/library/__future__.html) and we've decided we're no longer supporting Python 2.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Please see the above.

## Tracking Issue
https://github.com/lyft/flyte/issues/226

## Follow-up issue
https://github.com/lyft/flyte/issues/226


EngHabu (approved):

flytekit/__init__.py
```diff
@@ -1,5 +1,20 @@
...
 import flytekit.plugins  # noqa: F401
 
-__version__ = "0.12.6"
+__version__ = "0.13.0b2"
+
+logger = _logging.getLogger("flytekit")
+
+# create console handler and set level to debug
+ch = _logging.StreamHandler()
+ch.setLevel(_logging.DEBUG)
```
kumare3, line 11: do you want the level to be DEBUG?

flytekit/common/launch_plan.py
```diff
@@ -66,6 +72,29 @@ def promote_from_model(cls, model):
             raw_output_data_config=model.raw_output_data_config,
         )
 
+    @_exception_scopes.system_entry_point
+    def register(self, project, domain, name, version):
```
kumare3, line 76: can we change register to take in a context variable, maybe a separate PR. I want to use this context to specify endpoints, image etc as inputs instead of automatically figuring it out

flytekit/common/local_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+    def var(self):
+        """
+        :rtype: flytekit.models.interface.Variable
+        """
+        return self._var
+
+
+class PythonWorkflow(SdkWorkflow):
+    """
+    Wrapper class for locally defined Python workflows
```
kumare3, line 82: what does locally defined mean? Isn't everything locally defined?

flytekit/common/local_workflow.py
```diff
@@ -0,0 +1,371 @@
...
+        """
+        :rtype: list[flytekit.common.promise.Input]
+        """
+        return self._user_inputs
+
+    def create_launch_plan(
+        self,
+        default_inputs: Dict[str, _promise.Input] = None,
+        fixed_inputs: Dict[str, Any] = None,
+        schedule=None,
```
kumare3, line 233: can we introduce type hints here?

flytekit/common/mixins/registerable.py
```diff
@@ -171,3 +141,52 @@ def some_task()
...
+        workflow, it will still hit Admin).
+        """
+        pass
+
+    @property
+    def has_registered(self) -> bool:
+        return self._has_registered
+
+
+class HasDependencies(object):
```
kumare3, line 175: we are just modeling a graph - albeit poorly

flytekit/common/tasks/sdk_runnable.py
```diff
@@ -103,7 +97,11 @@ def execution_id(self):
         return self._execution_id
 
 
-class SdkRunnableContainer(_six.with_metaclass(_sdk_bases.ExtendedSdkType, _task_models.Container)):
+class SdkRunnableContainer(_task_models.Container, metaclass=_sdk_bases.ExtendedSdkType):
```
kumare3, line 100: ❤️ getting rid of metaclasses :)

flytekit/common/tasks/task.py
```diff
@@ -199,7 +210,14 @@ def fetch_latest(cls, project, domain, name):
         :rtype: SdkTask
         """
         named_task = _common_model.NamedEntityIdentifier(project, domain, name)
-        admin_task = _engine_loader.get_engine().fetch_latest_task(named_task)
+        client = _flyte_engine._FlyteClientManager(
```
kumare3, line 213: I think this is increasing the blast radius of creating the client manager. I see this code fragment repeated again again and again.

I think 
`fetch` `register` etc should take in a context that has the client in it. But I understand that is a large change. In the interim have a helper method that creates the client, instead of passing the url and insecure flag repeatedly?

flytekit/common/workflow_execution.py
```diff
@@ -36,9 +42,26 @@ def inputs(self):
...
+                _platform_config.URL.get(), insecure=_platform_config.INSECURE.get()
+            ).client
+            execution_data = client.get_execution_data(self.id)
+
+            # Inputs are returned inline unless they are too big, in which case a url blob pointing to them is returned.
+            if bool(execution_data.full_inputs.literals):
+                return execution_data.full_inputs
+
+            if execution_data.inputs.bytes > 0:
+                with _common_utils.AutoDeletingTempDir() as t:
```
kumare3, line 55: I actually want to get rid of the concept of AutoDeletingTempDir in the framework code. I think this should be a user decision

kumare3:
This is a huge change and I love it! A few comments, mostly to push for next steps. I dont think any of the feedback needs action right away.

flytekit/__init__.py
wild-endeavor: Yup.  At least for the beta releases right?  I'd like to add a lot more debug statements and some info statements too.  I'll make it info for the non-beta release.

flytekit/common/local_workflow.py
wild-endeavor: let me know what name to use?  happy to rename.  SdkRunnableWorkflow?

flytekit/common/launch_plan.py
wild-endeavor: yup, next PR.

flytekit/common/mixins/registerable.py
wild-endeavor: yeah.  dependency tracking in flytekit is confusing.  We have nodes, which have upstream nodes, but we also have tasks/workflows/launchplans, which also keep track of upstream entities.

I'd like to get to the point where we only use nodes to keep track of registration order and such.  This should make the topological sort easier to reason about as well.

It's not quite just the workflow graph though... if you have a workflow that statically references another workflow, somehow you'll need to know to process the second workflow first.  same for default launch plans.  we'll probably have to build up a registration graph which contains these inter-entity dependencies.  This is effectively what the entity level upstream entity array keeps track of now, but i feel like this is cleaner.  anyways, not this pr.

flytekit/common/tasks/task.py
wild-endeavor: https://github.com/lyft/flytekit/pull/153/commits/4a3604fb04acded509c2e157e634f85eabeeb5b4

flytekit/common/tasks/sdk_runnable.py
wild-endeavor: i wish... one day

flytekit/common/local_workflow.py
wild-endeavor: added it here, but it's missing from so many places... let's address in future PRs.

EngHabu (dismissed):

EngHabu (dismissed):

flytekit/tools/module_loader.py
```diff
@@ -1,13 +1,11 @@
-from __future__ import absolute_import
-
 import importlib
 import pkgutil
 
 import six
 
 from flytekit.common.exceptions import user as _user_exceptions
+from flytekit.common.local_workflow import SdkRunnableWorkflow as _PythonWorkflow
```
kumare3, line 7: the `_` name is different? Intentional?

flytekit/sdk/workflow.py
```diff
@@ -112,11 +114,13 @@ def workflow(nodes, inputs=None, outputs=None, cls=None, on_failure=None):
     :param dict[Text,Output] outputs: [Optional] A dictionary of output descriptors for a workflow.
     :param T cls: This is the class that will be instantiated from the inputs, outputs, and nodes. This will be used
         by users extending the base Flyte programming model. If set, it must be a subclass of
-        :py:class:`flytekit.common.workflow.SdkWorkflow`.
+        :py:class:`flytekit.common.local_workflow.PythonWorkflow`.
```
kumare3, line 117: wrong type?

kumare3 (approved):

PR #156, opened 2020-08-13
Add Sagemaker Runner script

EngHabu:
# TL;DR
This PR aims to adapt Flyte containers to the SageMaker execution environment.

### For details about how SageMaker runs a custom container, refer to https://github.com/lyft/flyte/issues/454.

In short, the contract that SageMaker follows is to this command to run the container:
```shell
docker run <image> train
```
where `train` is a python executable that will be put into `/usr/local/bin` inside the container  when you install the `sagemaker-training` library, which is just a very simple python file that calls into the `sagemaker_training` library, the content of which is shown as follows:
```python
#!/usr/bin/python3

# -*- coding: utf-8 -*-
import re
import sys

from sagemaker_training.cli.train import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```
and this script will eventually call into the script you specified with `SAGEMAKER_PROGRAM` in the Dockerfile.

### Why do we need the runner script
Two reasons:
1. This contract command cannot be changed. If we want to use our virtual environment `service_venv`, we need to have an intermediate layer to kick that off
2. The only way to pass non-blob-type inputs to our container is to use the hyperparameter field of the CRD. With that, the inputs will be passed in as command-line arguments. That means we have to pass in the container's `args` and `env_vars` via command line. 

To pass in 
```
`env1=val1 env2=val2 service_venv pyflyte-execute --task-module blah --task-name bloh --output-prefix s3://fake-bucket --inputs s3://fake-bucket`
```

Our flyteplugin need to rewrite it into and our runner script needs to parse it from the following format:
```
 --__FLYTE_ENV_VAR_env1__ val1 --__FLYTE_ENV_VAR_env2__ val2
 --__FLYTE_CMD_0_service_venv__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_1_pyflyte-execute__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_2_--task-module__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_3_blah__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_4_--task-name__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_5_bloh__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_6_--output-prefix__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_7_s3://fake-bucket__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_8_--inputs__ __FLYTE_CMD_DUMMY_VALUE__
 --__FLYTE_CMD_9_s3://fake-bucket__ __FLYTE_CMD_DUMMY_VALUE__
```
We added the prefix and suffix in order to prepare for the future hyperparameter jobs.

### Notable observations and changes:
1. SageMaker will parse everything the user specifies in the hyperParameters field of a TrainingJob CRD or the staticHyperparameters field of a HyperparameterTuningJob CRD into a JSON object. So we can create hyperparameter names and values freely as long as they comply with JSON format.
2. If a hyperparameter has an empty value `""`, it will be ignored by SageMaker.
3. SageMaker currently doesn't support running side-car containers nor custom AMIs, so statsD is not an option. We add a flag the plugin will set (not flytekit) so it can control what to set it to (if we deploy custom AMIs that has statsD relay on localhost, the plugin can override the config)

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/453


scripts/flytekit_sagemaker_runner.py
```diff
@@ -0,0 +1,16 @@
+from __future__ import absolute_import
+import argparse
+import logging
+import subprocess
+
+parser = argparse.ArgumentParser(description="Running sagemaker task")
+parser.add_argument('--__FLYTE_SAGEMAKER_CMD__', dest='flyte_sagmaker_cmd',
```
kumare3, line 7: Are you sure this will work. Are there limits on the size of the argument?

bnsblue: I don't see their CRD posing any restrictions on that front: https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/release/rolebased/installer.yaml#L1658

The operating system could pose a limit on the length `ARG_MAX` on a command line when the command is evaluated by the `exec` function, but that's usually in tens to hundreds of thousands of byte/characters 
https://stackoverflow.com/a/19355351
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html

but I don't know if there's any implicit/intrinsic limitation posed by SageMaker backend

bnsblue: @kumare3 
I don't see Sagemaker's CRD posing any restrictions on that front: https://github.com/aws/amazon-sagemaker-operator-for-k8s/blob/master/release/rolebased/installer.yaml#L1658

The operating system could pose a limit on the length `ARG_MAX` on a command line when the command is evaluated by the `exec` function, but that's usually in tens to hundreds of thousands of byte/characters 
https://stackoverflow.com/a/19355351
https://pubs.opengroup.org/onlinepubs/009695399/basedefs/limits.h.html

but I don't know if there's any implicit/intrinsic limitation posed by SageMaker backend

scripts/flytekit_sagemaker_runner.py
```diff
@@ -0,0 +1,26 @@
...
+import subprocess
+from os import environ
+
+parser = argparse.ArgumentParser(description="Running sagemaker task")
+parser.add_argument('--__FLYTE_SAGEMAKER_CMD__', dest='flyte_sagmaker_cmd',
+                    help='The entrypoint selector argument')
+args, unknowns = parser.parse_known_args()
+
+# Extending the command with the rest of the command-line arguments
+subprocess_cmd = args.flyte_sagmaker_cmd.split('+')
```
bnsblue, line 13: Why can't we use blank space?

scripts/flytekit_sagemaker_runner.py
bnsblue: Unfortunately, there is 
```yaml
status:
additional: "Unable to create training job: Unable to create Training Job: ValidationException:
1 validation error detected: Value '{__FLYTE_SAGEMAKER_CMD__=service_venv+pyflyte-execute+--task-module+workflows.custom_sagemaker_training+--task-name+custom_training_task+--inputs+s3://<prefix>/custom/data/inputs.pb+--output-prefix+s3://<prefix>+--my_input+hello
world}' at 'hyperParameters' failed to satisfy constraint: Map value must satisfy
constraint: [Member must have length less than or equal to 256, Member must have
length greater than or equal to 0, Member must satisfy regular expression pattern:
.*]\n\tstatus code: 400, request id: ff068295-a7ce-4fcf-8453-053a3fa9bc31"
```

flytekit/common/tasks/sagemaker/hpo_job_task.py
```diff
@@ -21,13 +23,13 @@
...
-        max_number_of_training_jobs: int,
-        max_parallel_training_jobs: int,
-        training_job: SdkBuiltinAlgorithmTrainingJobTask,
-        retries: int = 0,
-        cacheable: bool = False,
-        cache_version: str = "",
+            self,
+            max_number_of_training_jobs: int,
+            max_parallel_training_jobs: int,
+            training_job: typing.Union[SdkBuiltinAlgorithmTrainingJobTask, CustomTrainingJobTask],
```
bnsblue, line 29: This is really smart! Could you please add a small unit test for that if you haven’t done that already?

scripts/flytekit_sagemaker_runner.py
```diff
@@ -0,0 +1,43 @@
...
+import logging
+import subprocess
+from os import environ
+
+FLYTE_ARG_PREFIX = "--__FLYTE"
+FLYTE_ENV_VAR_PREFIX = f"{FLYTE_ARG_PREFIX}_ENV_VAR_"
+FLYTE_CMD_PREFIX = f"{FLYTE_ARG_PREFIX}_CMD_"
+FLYTE_ARG_SUFFIX = "__"
+
+# An example for a valid command:
```
EngHabu, line 13: Please take a look at this commit @bnsblue

bnsblue: @EngHabu This is not entirely correct. We currently support the following format
```
--__FLYTE_SAGEMAKER_CMD__ service_venv+pyflyte-execute
--__FLYTE_CMD_ARG_arg1__ value1 --__FLYTE_CMD_ARG_arg2__ value2 --__FLYTE_CMD_ARG_arg3__ value3 --__FLYTE_CMD_ARG_flag__ --__FLYTE_ENV_VAR_env_var1__ envvalue1 --__FLYTE_ENV_VAR_env_var2__ envvalue2 
```

bnsblue: The reason for doing this is that in a end2end2 flow, the variable name and values will be passed to the runner script via the Hyperparameter field of the SM CRD. However, we found that there's a limit on the size of hyperparameter values (and I believe the's also a limit on the size of the hyperparameter name). Using the format that we currently support, we are trying to minimize both the size of the variable name and the size of the value.

scripts/flytekit_sagemaker_runner.py
```diff
@@ -6,53 +6,32 @@
...
 
 FLYTE_ARG_PREFIX = "--__FLYTE"
-FLYTE_ENV_VAR_PREFIX = "{}_ENV_VAR_".format(FLYTE_ARG_PREFIX)
-FLYTE_CMD_ARG_PREFIX = "{}_CMD_ARG_".format(FLYTE_ARG_PREFIX)
+FLYTE_ENV_VAR_PREFIX = f"{FLYTE_ARG_PREFIX}_ENV_VAR_"
+FLYTE_CMD_PREFIX = f"{FLYTE_ARG_PREFIX}_CMD_"
 FLYTE_ARG_SUFFIX = "__"
 
 parser = argparse.ArgumentParser(description="Running sagemaker task")
-parser.add_argument("--__FLYTE_SAGEMAKER_CMD__", dest="flyte_sagemaker_cmd", help="The entrypoint selector argument")
```
bnsblue: This shouldn't be removed. See my previous comment

scripts/flytekit_sagemaker_runner.py
```diff
@@ -6,53 +6,32 @@
...
 args, unknowns = parser.parse_known_args()
 
-# Extending the command with the rest of the command-line arguments
-subprocess_cmd = []
-if args.flyte_sagemaker_cmd is not None:
-    subprocess_cmd = args.flyte_sagemaker_cmd.split("+")
```
bnsblue: These should not be removed either.

scripts/flytekit_sagemaker_runner.py
bnsblue: We don't want to assume that the hyperparameters are passed in the order we injected them. What we are doing here is we make sure we know what the command is so that we can reconstruct them in a reasonable way.

scripts/flytekit_sagemaker_runner.py
```diff
@@ -6,53 +6,32 @@
 from os import environ
 
 FLYTE_ARG_PREFIX = "--__FLYTE"
-FLYTE_ENV_VAR_PREFIX = "{}_ENV_VAR_".format(FLYTE_ARG_PREFIX)
-FLYTE_CMD_ARG_PREFIX = "{}_CMD_ARG_".format(FLYTE_ARG_PREFIX)
+FLYTE_ENV_VAR_PREFIX = f"{FLYTE_ARG_PREFIX}_ENV_VAR_"
+FLYTE_CMD_PREFIX = f"{FLYTE_ARG_PREFIX}_CMD_"
```
bnsblue, line 10: why is it necessary to change it from CMD_ARG to CMD?

scripts/flytekit_sagemaker_runner.py
bnsblue: Admittedly, if we do this, instead of making assumption of SM's behavior, we are making assumption about the format of the command carried by task template's container (http://github.com/lyft/flyteplugins/blob/ad68fea4fc2f19f447567d315b700ea308c12d27/go/tasks/plugins/k8s/sagemaker/utils.go#L287-L300). But I think it's either one way or the other. Making assumption about our own code base seems to be safer to deal with. WDYT @EngHabu ?

bnsblue: Let me show you an example of what I mean: 
In our SM flyteplugin, I injected the task template's container `args` first, in the same order as they appear in the task template `args` list. And then I injected env variables.

But looking at the following actual command that SM used to invoke the runner script (though I removed the lyft-internal SHAs and s3 paths), you can probably notice that SM sorts all the arguments based on the alphabetical order of the arguments' names.

```
/usr/bin/python3 /srv/service/venv/bin/flytekit_sagemaker_runner.py --__FLYTE_CMD_ARG_inputs__ s3://<path>/inputs.pb --__FLYTE_CMD_ARG_output_prefix__ s3://<path> --__FLYTE_CMD_ARG_task_module__ workflows.custom_sagemaker_training --__FLYTE_CMD_ARG_task_name__ custom_training_task --__FLYTE_ENV_VAR_FLYTE_INTERNAL_CONFIGURATION_PATH__ /code/flyteplayground/flyte_configs/staging.config --__FLYTE_ENV_VAR_FLYTE_INTERNAL_DOMAIN__ development --__FLYTE_ENV_VAR_FLYTE_INTERNAL_IMAGE__ <image> --__FLYTE_ENV_VAR_FLYTE_INTERNAL_PROJECT__ <project> --__FLYTE_ENV_VAR_FLYTE_INTERNAL_VERSION__ <version> --__FLYTE_SAGEMAKER_CMD__ service_venv+pyflyte-execute
```

scripts/flytekit_sagemaker_runner.py
bnsblue: see this as well: https://github.com/lyft/flytekit/pull/156#discussion_r479418016

EngHabu: It's not necessary... just borrowing from k8s & docker concepts... there are commands and there are args...
I'm now assuming you will pass `service_venv` as well as `pyflyte-execute` as just regular __FLYTE_CMD_* params... nothing is special about these two..

and the reason for that is, I don't like [this](https://github.com/lyft/flyteplugins/pull/113/files#diff-d8c750da896887beb6e43f80268f2351R286). I think the plugin should not make any assumptions about the command and how it looks like... so I would rather keep the "protocol" simple between the two to say:
- The SDK will write a container proto with cmd & args
- The plugin will pass all the CMDs and Args as `__FLYTE_CMD_*`
- The SDK must then reconstruct the original command from those...

Remember that FlyteKit might not even be in the container the plugin will launch... so keeping this protocol clear and easy to implement (in a bash script) is just as important...

Let me make the env var format work... that's the only difference, right?

bnsblue: see this: https://github.com/lyft/flytekit/pull/156#discussion_r479428335

scripts/flytekit_sagemaker_runner.py
```diff
@@ -0,0 +1,68 @@
...
+# --__FLYTE_CMD_0_service_venv__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_1_pyflyte-execute__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_2_--task-module__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_3_blah__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_4_--task-name__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_5_bloh__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_6_--output-prefix__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_7_s3://fake-bucket__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_8_--inputs__ __FLYTE_CMD_DUMMY_VALUE__
+# --__FLYTE_CMD_9_s3://fake-bucket__ __FLYTE_CMD_DUMMY_VALUE__
```
bnsblue, line 25: This example concludes the format @EngHabu and I agreed on

flytekit/common/tasks/sagemaker/custom_training_job_task.py
```diff
@@ -0,0 +1,84 @@
+from __future__ import absolute_import
```
wild-endeavor, line 1: 🗡️

flytekit/common/tasks/sagemaker/hpo_job_task.py
```diff
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
```
wild-endeavor, line 1: 🗡️

scripts/flytekit_sagemaker_runner.py
```diff
@@ -0,0 +1,87 @@
+from __future__ import absolute_import
```
wild-endeavor, line 1: 🗡️ 
Also, can you add a comment to this script explaining when/where it's run?  Is this something users run from their laptops?

flytekit/configuration/statsd.py
```diff
@@ -4,3 +4,4 @@
 
 HOST = _common_config.FlyteStringConfigurationEntry("statsd", "host", default="localhost")
 PORT = _common_config.FlyteIntegerConfigurationEntry("statsd", "port", default=8125)
+DISABLED = _common_config.FlyteBoolConfigurationEntry("statsd", "disabled", default=False)
```
wild-endeavor, line 7: Can you explain why this is necessary?  Were you seeing an error in sagemaker?  If so, what sets it?  Should we set it by default for certain types of jobs?

flytekit/common/tasks/sagemaker/custom_training_job_task.py
kumare3, line 1:
```suggestion
```

flytekit/configuration/statsd.py
EngHabu: SageMaker currently doesn't support running side-car containers nor custom AMIs.. so statsD is not an option. It's a flag the plugin will set (not flytekit) so it can control what to set it to (if we deploy custom AMIs that has statsD relay on localhost, the plugin can override the config) I don't think flytekit should make assumptions about the execution environment... generally speaking

wild-endeavor (approved):

wild-endeavor (approved):

bnsblue (approved):

PR #160, opened 2020-08-17
Add proposed output def change

EngHabu:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/annotated/sample.py
```diff
@@ -1,13 +1,14 @@
 from flytekit.annotated.stuff import task, workflow, WorkflowOutputs
+from flytekit.annotated.type_engine import outputs
 
 
-@task(outputs=['s_out'])
-def x(s: int) -> int:
+@task
+def x(s: int) -> outputs(s_out=int):
```
kumare3, line 6: tbh I do not like this, because this needs the user to change the interface of their tasks :(

EngHabu: Can you elaborate? Change the interface of current flytekit style tasks? Or of py3 style functions?
This is, from python perspective, equivalent to `(int, str)`

EngHabu: And if you return a single output, you don't need to do this and we will automatically name it just "output"

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -0,0 +1,95 @@
...
+    
+#     assert my_task.unit_test() == {'output': 'Hello world'}
+
+
+@flyte_test
+def test_read_file_known_location():
+    # Option 1
+    # users receive a file handle as the parameter to their task
+    @python_task
+    def my_task(ctx, fh: typing.BinaryIO):
```
kumare3, line 53: Love this!

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -0,0 +1,95 @@
...
+        with open(file_path.local_path, mode='rb') as fh:
+            lines = fh.readlines()
+            # assert 
+
+    assert my_task.unit_test(file_path=CustomPathLike('s3://bucket-my/mytest')) == {}
+
+    # Option 2.2
+    # Provide a custom open() that encapsulates download and open actions
+    @python_task
+    def my_task(ctx, file_path: CustomPathLike):
```
kumare3, line 83: this is standard, so love this

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -0,0 +1,95 @@
...
+        with file_path.open() as fh:
+            lines = fh.readlines()
+            # assert 
+
+
+    # Option 2.3
+    # Flyte always downloads the file and users can use the regular open() function to read it.
+    @python_task
+    def my_task(ctx, file_path: CustomPathLike):
```
kumare3, line 92: i actually dont want to assume this, as this has perf implications

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -0,0 +1,95 @@
...
+    # Option 1.2
+    # Users pass a Path-Like object that flyte knows how to interpret and open or youo
+    assert my_task.unit_test(fh=CustomPathLike('/mytest', format="csv")) == {}
+    
+    # Option 2
+    # Users receive a Path-Like type as a parameter to their function (much like how Blobs work today)
+    # Option 2.1
+    # Users will have to call downlooad then open
+    @python_task
+    def my_task(ctx, file_path: CustomPathLike):
```
kumare3, line 72: i think this is an overhead

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -0,0 +1,95 @@
...
+from io import BufferedWriter
+
+from flytekit.sdk.test_utils import flyte_test
+from flytekit.sdk.tasks import python_task
+from flytekit.annotated.type_engine import outputs
+
+@flyte_test
+def test_simple_input_output():
+    @python_task
+    def my_task(ctx, a: int) -> outputs(b=int, c=str):
```
kumare3, line 12: do we need to wrap with `outputs()`
I think this is more intrusive?

flytekit/annotated/sample.py
kumare3: What I mean, is that they have to have an explicit dependency on 
```python
from flytekit. ... import outputs
#then use outputs
```

What I want is ideally just annotate with one thing and you are done, if you remove the annotation, your code works as is?
Maybe unreasonable

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -0,0 +1,110 @@
...
+#     assert my_task.unit_test(fh=CustomPathLike('/mytest', format="csv")) == {}
+    
+#     # Option 2
+#     # Users receive a Path-Like type as a parameter to their function (much like how Blobs work today)
+#     # Option 2.1
+#     # Users will have to call downlooad then open
+#     @python_task
+#     def my_task(ctx, file_path: CustomPathLike):
+#         file_path.download()
+#         with open(file_path.local_path, mode='rb') as fh:
```
wild-endeavor, line 89: if this is how you open the locally downloaded file, what would `print(file_path)` show?  I feel like it should be the remote path.  That is, either there's a `remote_path` or there's a `local_path` but not both, and the naked object should show the other one.

flytekit/sdk/tasks.py
```diff
@@ -16,6 +16,16 @@
 from flytekit.sdk.spark_types import SparkType as _spark_type
 
 
+def current_context() -> _sdk_runnable_tasks.ExecutionParameters:
```
wild-endeavor, line 19: I want to move this into the engine.

flytekit/common/utils.py
```diff
@@ -205,6 +207,30 @@ def __exit__(self, exc_type, exc_val, exc_tb):
         return False
 
 
+T = typing.TypeVar('T')
+
+
+class StackableContext(typing.Generic[T]):
```
wild-endeavor, line 213: why is this a generic?

PR #164, opened 2020-08-24
[pr-into-pr] please ignore

wild-endeavor:
wip


flytekit/annotated/type_engine.py
```diff
@@ -1,39 +1,62 @@
...
-def type_to_literal_type(t: type) -> _type_models.LiteralType:
-    if t in SIMPLE_TYPE_LOOKUP_TABLE:
-        return SIMPLE_TYPE_LOOKUP_TABLE[t]
-    return _type_models.LiteralType(
-        simple=_type_models.SimpleType.NONE
-    )
+# These are not supported in the context of Flyte's type management, because there is no way to peek into what the
+# inner type is. Also unsupported are typing's Tuples. Even though you can look inside them, Flyte's type system
+# doesn't support these currently.
+UNSUPPORTED_CONTAINERS = [tuple, list, dict, typing.Tuple, typing.NamedTuple]
```
EngHabu, line 59: We can serialize NamedTuples as Proto Structs, right? perhaps in a future PR :)

flytekit/bin/entrypoint.py
```diff
@@ -86,6 +86,17 @@ def _execute_task(task_module, task_name, inputs, output_prefix, test):
                     _literal_models.LiteralMap.from_flyte_idl(input_proto),
                     context={'output_prefix': output_prefix}
                 )
+            elif not test and hasattr(task_def, "_task_style"):
+                local_inputs_file = input_dir.get_named_tempfile('inputs.pb')
+
+                _data_proxy.Data.get_data(inputs, local_inputs_file)
+                raw_inputs = _utils.load_proto_from_file(_literals_pb2.LiteralMap, local_inputs_file)
+                from flytekit.engine import ExecutionContextProvider
```
EngHabu, line 94: nit: move up in the file?

flytekit/common/tasks/sdk_runnable.py
```diff
@@ -400,19 +400,30 @@ def execute(self, context, inputs):
                     literals={k: v.sdk_value for k, v in _six.iteritems(outputs_dict)}
                 )
             }
+        # New style - V1: inputs are just inputs
         else:
-            outputs = self._execute_user_code(context, inputs_dict)
-            sdk_types = {k: _type_helpers.get_sdk_type_from_literal_type(v.type) for k, v in self.interface.outputs.items()}
+            new_style_inputs = flyte_engine.idl_literal_map_to_python_value(context, inputs)
```
EngHabu, line 405: nit: drop the "new_style" name

flytekit/common/tasks/sdk_runnable.py
```diff
@@ -400,19 +400,30 @@ def execute(self, context, inputs):
...
+            outputs = self._execute_user_code(context, new_style_inputs)
             expected_output_names = list(self.interface.outputs.keys())
             if len(expected_output_names) == 1:
                 literals = {expected_output_names[0]: outputs}
             else:
-                literals = {expected_output_names[i]: outputs[i] for i, _ in enumerate(outputs)} 
+                # Question: How do you know you're going to enumerate them in the correct order? Even if autonamed, will
```
EngHabu, line 411: ah because you are saying there is no expectation from proto representation of the interface about ordering to begin with?
{"a": int, "b": str} is not the same as {"b": str, "a": int} from proto perspective I agree... however, in go, for example, we don't get to choose in which ordering does it get loaded into a go map... so if we ever build a golang SDK, we won't be able to apply that same logic...

Maybe the option to reinspect the annotations to match up the names is not a bad one... I was hoping we don't need to do that during execution though... I was hoping we could, one day, do something like:
```
run(my_mock_python_func, idl_task_template)
```
but maybe that's a bit of wishful thinking... we can still achieve the same by requiring the mock func to adhere to the same interface with the same names... that's also ok....

Alright, sorry for writing down my thoughts as they flow :) I think I like re-inspecting the func python interface... just not right here... probably somewhere a bit higher when we load the python module & func and match it up to the idl interface before proceeding...

flytekit/engine.py
```diff
@@ -0,0 +1,317 @@
...
+    # TODO: Move this into the execution context
+    local_path = _generate_local_path()
+
+    def _downloader():
+        _data_proxy.Data.get_data(blob.uri, local_path, is_multipart=False)
+
+    if blob.metadata.type.format == flyte_typing.FlyteFileFormats.CSV.value:
+        return flyte_typing.FlyteCSVFilePath(local_path, _downloader, blob.uri)
+
+    if blob.metadata.type.format == flyte_typing.FlyteFileFormats.TEXT_IO:
```
EngHabu, line 161: Add binary :D ?

flytekit/typing.py
```diff
@@ -81,13 +84,22 @@ def format(cls) -> str:
...
+        :param remote_path: Just in case the user wants it for some reason.
+        """
         self._abspath = os.path.abspath(path)
+        self._downloader = downloader
         self._remote_path = remote_path
         logger.debug(f"Path is: {self._abspath}")
 
     def __fspath__(self):
         # This is where a delayed downloading of the file will happen
+        self._downloader()
```
EngHabu, line 102: nit: check if we ```downloaded``` already

PR #168, opened 2020-08-26
Use lazy_loader for pyspark

akhurana001:
# TL;DR
Use lazy_loaded for pyspark so that non-spark notebooks can run without pyspark installed.

Also added a dependency on `ipykernel` . Executions in my local venv were failing with a `No module named ipykernel_launcher` . 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Use lazy_loader for pyspark

## Tracking Issue
https://github.com/lyft/flyte/issues/488

## Follow-up issue
_NA_
OR


kumare3:
Please update the description completely

setup.py
```diff
@@ -50,6 +50,7 @@
         "urllib3>=1.22,<2.0.0",
         "wrapt>=1.0.0,<2.0.0",
         "papermill>=1.2.0",
+        "ipykernel>=5.0.0",
```
kumare3, line 53: ?

flytekit/contrib/notebook/helper.py
```diff
@@ -1,10 +1,10 @@
 import os as _os
 
 import six as _six
-from pyspark import SparkConf, SparkContext
 
 from flytekit.common.types.helpers import pack_python_std_map_to_literal_map as _packer
 from flytekit.contrib.notebook.supported_types import notebook_types_map as _notebook_types_map
+from flytekit.plugins import pyspark as _pyspark
```
kumare3, line 7: this will work because only when we access pyspark.`x` will it load the real module?

setup.py
akhurana001: See the description.

flytekit/contrib/notebook/helper.py
akhurana001: Yeah, that's my understanding of lazy_loading here: https://github.com/lyft/flytekit/blob/master/flytekit/plugins/__init__.py#L5

PR #169, opened 2020-08-27
feat; debug enabled optionally for aws s3 operations

kumare3:
# TL;DR
If an S3 op fails, a request ID is required. Currently there is no way to get requestID without running aws cli in debug mode

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


wild-endeavor (dismissed):

apda:
👍

wild-endeavor:
@jeevb Mind taking a quick look at this if you get a chance?  if not it's okay.  I ran into an issue when trying to use flytekit internally with a private repo.  When I ran the equivalent of the `make requirements` step here, I got errors like

```
Could not find a version that satisfies the requirement pyarrow==0.15.0 (from flytekit[all]->-r requirements.in ...
```

The issue is that it was reading flytekit's `requirements.txt` and taking those versions as hard and exact requirements, rather than using the version ranges found in `setup.py`.  We're temporarily removing the requirements if that's okay - I'll try to dig deeper on our end as well to ascertain if it's just something we're doing on our side.  Hope this doesn't cause any consternation on your end.

jeevb:
We previously had `pyarrow==0.17.1` installed: https://github.com/lyft/flytekit/blob/9dc9911d14a8610d998be8a322e285677fd8153f/requirements.txt#L46

What is constraining it to `0.15.0`? Would help to see the full error message.

wild-endeavor:
Okay, reverted the other changes.

Still pinning black to 0.19.0b0 in both the dev-requirements, and in setup.  If I didn't pin in setup, pip-compile was failing because papermill itself requires black and it was finding v20.

PR #170, opened 2020-08-28
Add a flag to disable default launch plan generation

EngHabu:
# TL;DR
Adds a flag to control whether a default launchplan should be generated for the workflow at registration time. The default behavior is to do so.

E.g.
```
@workflow_class(disable_default_launch_plan=True)
def MyWorkflow():
   ...

my_lp = MyWorkflow.create_launch_plan(...)

#my_lp will be the only launchplan created for this workflow
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/462

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -2,4 +2,4 @@ from __future__ import absolute_import
 
 import flytekit.plugins  # noqa: F401
 
-__version__ = "0.12.4"
+__version__ = "0.12.5"
--- a/flytekit/common/workflow.py
+++ b/flytekit/common/workflow.py
@@ -93,15 +93,16 @@ class SdkWorkflow(
     )
 ):
     def __init__(
-        self,
-        inputs,
-        outputs,
-        nodes,
-        id=None,
-        metadata=None,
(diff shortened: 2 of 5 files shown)


EngHabu:
PTAL @wild-endeavor

flytekit/__init__.py
```diff
@@ -2,4 +2,4 @@
 
 import flytekit.plugins  # noqa: F401
 
-__version__ = "0.12.4"
+__version__ = "0.12.5"
```
RubenBarragan, line 5: nit question, do we have to change the version every change?

RubenBarragan (approved):

flytekit/__init__.py
EngHabu: yes

honnix:
Looks like this feature is gone. Is there a replacement? Thanks.

PR #173, opened 2020-09-15
Enabling bare-bones distributed training for SageMaker and prepare for enabling distributed training for PyTorch / Tensorflow plugins

bnsblue:
# TL;DR
Adding components to properly enable bare-bones distributed training for SageMaker tasks, and laying out the foundation for distributed training with PyTorch and Tensorflow plugins.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
1. Adding a `distributed_training_context` to the engine context
2. Constructing the `distributed_training_context` by either getting the values from the environment (https://github.com/aws/sagemaker-training-toolkit/blob/master/ENVIRONMENT_VARIABLES.md) or directly parsing the `resourceconfig.json` file (https://docs.aws.amazon.com/sagemaker/latest/dg/your-algorithms-training-algo-running-container.html#your-algorithms-training-algo-running-container-dist-training)
3. Wrapping a retry logic around the `distributed_training_context`-getting function
4. Adding a `output-persist-predicate` field to the `custom_training_job_task` decorator and the `CustomTrainingJobTask` class. This field should receive a Callable, which would decide whether the output of the current instance should be persisted to S3. The Callable takes a `distributed_training_context` dictionary and should return a boolean. Class `CustomTrainingJobTask`'s `execute` function will invoke this Callable after the body of the user task function finishes its execution. If the Callabe returns `True`, it means the output of this instance should be persisted; otherwise the output would be dropped.
5. Adding a default output-persist predicate (https://github.com/lyft/flytekit/pull/173/commits/af709ecded33a6f6424748d23fe5864507a3ccd7)

* I (@bnsblue) decided not to add an extra `distributed_custom_training_job_task` type because when users specify an `instance_count` > 1 it should already indicates that the users would want to run the training in a distributed manner. if we define a new class, this `instance_count` would then become redundant.

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/bin/entrypoint.py
```diff
@@ -17,6 +19,16 @@
 from flytekit.models import literals as _literal_models
 
 
+def _run_with_retries(func, expected_exception, max_retries=10):
```
EngHabu, line 22: Do we not have something that does this already :-D ? I think there is a whole python library to do retries.. not sure if we use it...

flytekit/common/tasks/sagemaker/custom_training_job_task.py
```diff
@@ -77,6 +88,69 @@ def __init__(
...
+
+        if (
+            self._is_distributed()
+            and self._output_persist_predicate
+            and context.distributed_training_context
+            and self.output_persist_predicate(context.distributed_training_context) is True
+        ):
+            return ret
+        else:
+            return {}
```
EngHabu, line 127: Maybe log here that we are skipping persisting the output...

flytekit/common/tasks/sagemaker/custom_training_job_task.py
```diff
@@ -77,6 +88,69 @@ def __init__(
...
+
+        return _exception_scopes.user_entry_point(self.task_function)(
+            _ExecutionParameters(
+                execution_date=context.execution_date,
+                # TODO: it might be better to consider passing the full struct
+                execution_id=_six.text_type(WorkflowExecutionIdentifier.promote_from_model(context.execution_id)),
+                stats=context.stats,
+                logging=context.logging,
+                tmp_dir=context.working_directory,
+                distributed_training_context=context.distributed_training_context,
```
EngHabu, line 153: I think this is where you should actually attempt to build the context instead?

EngHabu (commented):
I think the biggest question is if you tested not returning outputs from a task that supposedly returns outputs... does that cause it to write an empty outputs file, or fail upstream in flytekit somehow?

flytekit/common/constants.py
```diff
@@ -36,3 +36,9 @@ class SdkTaskType(object):
 class CloudProvider(object):
     AWS = "aws"
     GCP = "gcp"
+
+
+class DistributedTrainingContextKey(object):
+    CURRENT_HOST = "current_host"
+    HOSTS = "hosts"
+    NETWORK_INTERFACE_NAME = "network_interface_name"
```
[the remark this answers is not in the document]
bnsblue: My thought was that this is for all the plugins that would use distributed training so that's why I didn't put in under `sagemaker` package. Does that not make sense?

flytekit/common/distribution.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+# The default output-persisting predicate.
+# With this predicate, only the copy running on the first host in the list of hosts would persist its output
+class DefaultOutputPersistPredicate(object):
+    def __call__(self, distributed_training_context):
+        return (
+            distributed_training_context[_common_constants.DistributedTrainingContextKey.CURRENT_HOST]
+            == distributed_training_context[_common_constants.DistributedTrainingContextKey.HOSTS][0]
+        )
```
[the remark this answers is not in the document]
bnsblue: Ditto. This can be used by all the tasks/plugins that would have a use for distributed training, right?

flytekit/common/tasks/sagemaker/custom_training_job_task.py
bnsblue: Unfortunately, the distributed_training_context is needed not only by user task_function, but also the execute() function of the CustomTrainingJobTask. If we create `distributed_training_context` here, the execute() function would not be able to access it. We can, of course, create it from the env var again, but I am not sure if that is preferred.

flytekit/bin/entrypoint.py
bnsblue: Changing to using the retry library

flytekit/bin/entrypoint.py
```diff
@@ -80,9 +92,16 @@ def _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data
 
                 _data_proxy.Data.get_data(inputs, local_inputs_file)
                 input_proto = _utils.load_proto_from_file(_literals_pb2.LiteralMap, local_inputs_file)
+                _distributed_training_context = _run_with_retries(
```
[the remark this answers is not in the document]
bnsblue: It will only be passed into the ExecutionParameters of CustomTrainingJobTasks. The reason why we need it in the EngineContext object is that we need to feed this information to `execute()` function so that the `output_persist_predicate` callable can take this information and decide whether or not it wants to persist the output of the task

flytekit/common/tasks/sagemaker/custom_training_job_task.py
bnsblue: What I can do is to fill `distributed_training_context` into the context object only when CustomTrainingJobTasks' execute() is called. But again I am not sure if that is preferred

bnsblue:
> I think the biggest question is if you tested not returning outputs from a task that supposedly returns outputs... does that cause it to write an empty outputs file, or fail upstream in flytekit somehow?

We do not know the answer yet. We need to have the example running to know this.

flytekit/common/constants.py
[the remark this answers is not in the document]
bnsblue: Moved

flytekit/bin/entrypoint.py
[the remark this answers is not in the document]
bnsblue: This is fixed

flytekit/common/distribution.py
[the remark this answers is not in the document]
bnsblue: Moved. Resolving

wild-endeavor:
What is `flytekit/common/distribution.py`?

bnsblue:
Removed. Missed that file when doing the house keeping

tests/flytekit/unit/sdk/tasks/test_sagemaker_tasks.py
```diff
@@ -211,3 +222,149 @@ def my_task(wf_params, input_1, model):
...
+        with mock.patch.dict(
+            os.environ,
+            {
+                _sm_distribution.SM_ENV_VAR_CURRENT_HOST: "algo-1",
+                _sm_distribution.SM_ENV_VAR_HOSTS: '["algo-0", "algo-1", "algo-2"]',
+            },
+            clear=True,
+        ):
+
+            print(os.environ)
```
wild-endeavor, line 311: nix

flytekit/common/tasks/sagemaker/custom_training_job_task.py
bnsblue: Update: @EngHabu and I discussed and agreed to build a training-job-specific context here: https://github.com/lyft/flytekit/pull/173/files#diff-3d88a183363264f577728ca8af28fd16R121

flytekit/common/tasks/sdk_runnable.py
```diff
@@ -26,12 +26,13 @@ class ExecutionParameters(object):
     decorated function.
     """
 
-    def __init__(self, execution_date, tmp_dir, stats, execution_id, logging):
+    def __init__(self, execution_date, tmp_dir, stats, execution_id, logging, distributed_training_context=None):
```
wild-endeavor, line 29: Not a huge fan of this but I'm okay with it if it's required.  Possible alternative... 
1. we can subclass EngineContext, and then we can just add any field we want
2. change the name to `task_specific_context` or something, but i’m struggling with how other tasks might use it and where/how tasks would set it.

bnsblue: I like the subclass idea. I've made the change in the following commits

flytekit/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 import flytekit.plugins  # noqa: F401
 
-__version__ = "0.13.0"
+__version__ = "0.13.1"
```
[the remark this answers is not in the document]
bnsblue: fixed

wild-endeavor (approved):
+1 from me, not sure if @kumare3 wants to take a look as well?

setup.py
```diff
@@ -44,6 +44,7 @@
         "papermill>=1.2.0",
         "ipykernel>=5.0.0",
         "black==19.10b0",
+        "retry==0.9.2",
```
kumare3, line 47: ?

bnsblue: For this https://github.com/lyft/flytekit/pull/173/files#diff-fcefe0b613289d8720e4090b8f4caa85R15-R16

bnsblue: @kumare3 ^

flytekit/common/tasks/sagemaker/distribution.py
```diff
@@ -0,0 +1,99 @@
...
+    distributed_training_context[DistributedTrainingContextKey.CURRENT_HOST] = _os.environ.get(SM_ENV_VAR_CURRENT_HOST)
+    distributed_training_context[DistributedTrainingContextKey.HOSTS] = _json.loads(_os.environ.get(SM_ENV_VAR_HOSTS))
+    distributed_training_context[DistributedTrainingContextKey.NETWORK_INTERFACE_NAME] = _os.environ.get(
+        SM_ENV_VAR_NETWORK_INTERFACE_NAME
+    )
+
+    return distributed_training_context
+
+
+@_retry.retry(exceptions=FileNotFoundError, delay=1, tries=10, backoff=1)
```
kumare3, line 34: why the retries?

flytekit/common/tasks/sagemaker/distribution.py
```diff
@@ -0,0 +1,99 @@
+import json as _json
```
kumare3, line 1: the name of the file is weird?
distributed_training.py?

flytekit/common/tasks/sagemaker/distribution.py
bnsblue: The files we try to read here is mounted by SageMaker so it is not guaranteed to be immediately available

kumare3:
Looks good to me barring couple nits

flytekit/common/tasks/sagemaker/distribution.py
bnsblue: fixed

setup.py
bnsblue: SageMaker suggests that hostname information might not be immediately available to the processing container. They recommend adding a retry policy on hostname resolution operations as nodes become available in the cluster.
See https://docs.aws.amazon.com/sagemaker/latest/dg/build-your-own-processing-container.html#byoc-config

I added a comment to document this

wild-endeavor (dismissed):

katrogan (approved):

PR #175, opened 2020-09-16
wip-ignore

wild-endeavor:
None of this code is tested, I'm sure it doesn't work yet.  Still going through things.

* Added a high level FlyteContext class - this is all in the context_manager.py file, take a look through this first.
* Added executors.py - this is pretty small, basically a combination of logic from the current engine code and sdkrunnabletask.
* Then take a look at entrypoint - instead of fetching an engine and then calling FlyteTask.execute which calls SdkRunnableTask.execute, it creates an executor and calls the execute method.
* In stuff.py:
  * Get rid of the WorkflowOutputs class - this was the class that workflow functions formerly returned that allowed us to inspect the stack but the addition of the context means we don't have to do all this crap.  Also got rid of the stack crawling stuff.
  * In some places where we were relying on heavier weight objects, I've moved over to the model objects directly.
* The NodeOutput object is going to be the thing that holds the wrapper for actual values when running a workflow locally.  We can no-op out all the with_cpu-like calls when it just wraps a value.
* Removing stuff from SdkRunnableTask if we're going to rely on the executor to do the setup work.


flytekit/annotated/context_manager.py
```diff
@@ -0,0 +1,101 @@
...
+            parent = FlyteContext.OBJS[-1]
+
+        self._parent: FlyteContext = parent
+        self._local_file_access = local_file_access
+        self._remote_data_proxy = remote_data_proxy
+        self._compilation_state = compilation_state
+        self._flyte_client = flyte_client
+
+    def __enter__(self):
+        FlyteContext.OBJS.append(self)
```
EngHabu, line 42: Semantically won't matter but should this __enter__ forcefully set `self._parent = FlyteContext.OBJS[-1]`

tests/flytekit/unit/annotated/test_context_manager.py
```diff
@@ -0,0 +1,22 @@
...
+
+def test_levels():
+    s = SampleTestClass(value=1)
+    with FlyteContext(flyte_client=s) as ctx:
+        assert ctx.flyte_client.value == 1
+        with FlyteContext(flyte_client=SampleTestClass(value=2)) as ctx:
+            assert ctx.flyte_client.value == 2
+
+        with FlyteContext(compilation_state=CompilationState()) as ctx:
+            assert ctx.flyte_client.value == 1
```
EngHabu, line 17: Do you want to retrieve current_context() here and make sure it matches?

flytekit/annotated/context_manager.py
wild-endeavor: you decide.

wild-endeavor: the whole array keeping track of everything is annoying actually, i'd rather not have it.  (but I think it's preferable to using thread locals).  the only place it's really used (read as, "the only place where context isn't explicitly passed as as an arg"), is when a task is called, since we don't inject a context arg anymore (which is a good thing imo btw).

flytekit/annotated/context_manager.py
```diff
@@ -0,0 +1,259 @@
...
+from flytekit.engines.unit import mock_stats as _mock_stats
+
+
+class CompilationState(object):
+    def __init__(self):
+        self.nodes: List[SdkNode] = []
+        self.mode = 1  # TODO: Turn into enum in the future, or remove if only one mode.
+
+
+class ExecutionState(object):
```
kumare3, line 31: i am really liking where we are heading

kumare3: i just think the context stack should be separate from the actual context - composition this way will make unit test easier.

So create ExecutionContext
ExecutionContextStack?

kumare3: actually i mean this comment for FlyteContext

flytekit/annotated/executors.py
```diff
@@ -0,0 +1,54 @@
...
+from flytekit import engine as _flyte_engine
+from flytekit.annotated import context_manager, stuff as _annotated
+from flytekit.common import utils as _utils
+from flytekit.interfaces.data import data_proxy as _data_proxy
+from flytekit.models import literals as _literal_models
+
+
+class BaseExecutor(object):
+    # Should these be class methods?
+    def execute(self, ctx: context_manager.FlyteContext, task: _annotated.PythonTask, inputs_path: str):
```
kumare3, line 14: if there is inheritance then why class methods? let me be regular methods so that overriding, is easy and you can have class state

flytekit/annotated/context_manager.py
wild-endeavor: I agree, it's awkward - can you think of a way around it?

PR #188, opened 2020-09-30
Cron schedule with offset

honnix:
# TL;DR
Support validate and propagate CronScheduleWithOffset.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
**Motivation: Why do you think this is important?**
When integrating with our scheduler Styx, we will need to extend the semantics of schedule to support Styx cron syntax, cron aliases, as well as offset.

**Goal: What should the final outcome look like, ideally?**
Scheduleprotobuf definition is extended to support Styx style schedule.

**Describe alternatives you've considered**
We tried reusing existing cron but the syntax is a bit different and flytekit doesn't support Styx style cron aliases; and more importantly, we need to support offset which is one the most important concepts in Styx.

More details can be found in [this issue](https://github.com/lyft/flyte/issues/529).

Related and depended change in flyteidl is https://github.com/lyft/flyteidl/pull/83.

## Tracking Issue
https://github.com/lyft/flyte/issues/529


kumare3:
cc @wild-endeavor

dev-requirements.txt
```diff
@@ -12,7 +12,6 @@ coverage==5.3             # via -r dev-requirements.in
 flake8-black==0.2.1       # via -r dev-requirements.in
 flake8-isort==4.0.0       # via -r dev-requirements.in
 flake8==3.8.3             # via -r dev-requirements.in, flake8-black, flake8-isort
-importlib-metadata==1.7.0  # via -c requirements.txt, flake8, pluggy, pytest
```
honnix, line 15: I ran `pip-compile dev-requirements.in`.

requirements.txt
```diff
@@ -7,30 +7,31 @@
 -e file:.#egg=flytekit    # via -r requirements.in
 ansiwrap==0.8.4           # via papermill
 appdirs==1.4.4            # via black
-appnope==0.1.0            # via ipykernel, ipython
```
honnix, line 10: I ran `pip-compile requirements.in`.

honnix:
@wild-endeavor I think this is ready for review. Please take a look and check whether this change would break backward compatibility in some way. Thanks a lot.

dev-requirements.txt
honnix: Maybe because I ran it under Python 3.8? I can revert the change if seems risky.

requirements.txt
wild-endeavor: same as if you run `make requirements` right?

tests/flytekit/unit/common_tests/test_launch_plan.py
```diff
@@ -127,19 +128,31 @@ def test_no_additional_inputs():
     assert lp.default_inputs.parameters["required_input"].required is True
 
 
-def test_schedule():
+@_pytest.mark.parametrize(
+    "schedule,cron_expression,cron_schedule",
+    [
+        (_schedules.CronSchedule("* * ? * * *"), "* * ? * * *", None),
```
wild-endeavor, line 134: Can we add a test here for `'0/15 * * * ? *'`

flytekit/__init__.py
```diff
@@ -2,7 +2,7 @@
 
 import flytekit.plugins  # noqa: F401
 
-__version__ = "0.13.0b10"
+__version__ = "0.13.0b11"
```
wild-endeavor, line 5: and will need to bump here now that we bumped the other pr

requirements.txt
honnix: Interesting. Actually not. First of all, I had to fix `Makefile` because of `_install-piptools` that should be `install-piptools`; then `requirements.txt` file header became:
```
#
# This file is autogenerated by pip-compile
# To update, run:
#
#    make requirements.txt
#
```

I will make a commit to reflect the difference.

honnix: I see only a few things been upgraded because of `define PIP_COMPILE`.

tests/flytekit/unit/common_tests/test_launch_plan.py
honnix: Added in 25d4ef8e62c44b1ad5b1c049e2773fb809ac6041. Sorry I messed up commits and had to force push.

wild-endeavor (approved):
Let's see... it's a beta anyways - if anything weird happens we can take a look at the dependency changes.

honnix:
Thanks a lot for approving this!

PR #200, opened 2020-10-07
Replace with promises throughout and allow tuples to be returned

kumare3:
Annotations promise implementation

### DIFF
--- a/flytekit/engine.py
+++ b/flytekit/engine.py
@@ -2,6 +2,7 @@ import typing
 from datetime import datetime, timedelta
 
 from flytekit import typing as flyte_typing
+from flytekit.annotated.promise import Promise
 from flytekit.common import promise as _promise
 from flytekit.common.exceptions import user as _user_exceptions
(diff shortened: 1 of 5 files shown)


flytekit/engine.py
```diff
@@ -192,16 +193,9 @@ def binding_from_python_std(ctx: _flyte_context.FlyteContext, var_name: str, exp
...
-            _user_exceptions.FlyteTypeException(
-                incoming_value_type,
-                expected_literal_type,
-                additional_msg="When binding node output: {}".format(t_value)
-            )
-        promise = t_value
-        binding_data = _literals_models.BindingData(promise=promise)
+    elif isinstance(t_value, Promise):
+        if not t_value.is_ready:
+            binding_data = _literals_models.BindingData(promise=t_value.ref)
```
wild-endeavor, line 198: add a note to add type checking in the future?  or do we not need to?  not sure if that adds any value actually.

wild-endeavor (approved):

PR #203, opened 2020-10-08
Nested Subworkflow

kumare3:


flytekit/annotated/task.py
```diff
@@ -82,12 +85,41 @@ def _create_and_link_node(self, ctx: FlyteContext, *args, **kwargs):
...
+        for k, v in kwargs.items():
+            if isinstance(v, Promise):
+                if not v.is_ready:
+                    raise BrokenPipeError(
+                        f"Expected an actual value from the previous step, but received an incomplete promise {v}")
+                kwargs[k] = v.val
+
+        input_literal_map = _literal_models.LiteralMap(literals=kwargs)
+
+        outputs_literal_map = self.dispatch_execute(ctx, input_literal_map)
```
wild-endeavor, line 103: So i know this was done in another PR, when you removed the stupid BaseExecutor thing, but it harkens back to this discussion here: https://lyft.slack.com/archives/GAG9AAS9K/p1600964346001300 - where we were discussing whether a) tasks knew how to execute themselves b) they relied on an executor to execute them.

I feel like we all seemed inclined at the time to go with the latter option - you kinda wrote something like it in the google doc https://docs.google.com/document/d/1nJprPSZtxbXQLoafD2xeN0JIueHnEb6TNIsJiXHF6rY/edit

We're sure we're okay with this change?  This is the right thing to do?

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -170,6 +170,36 @@ def my_wf(a: int, b: str) -> (int, str):
...
+    @workflow
+    def my_subwf(a: int) -> (str, str):
+        x, y = t1(a=a)
+        u, v = t1(a=x)
+        return y, v
+
+    @workflow
+    def my_wf(a: int, b: str) -> (int, str, str):
+        x, y = t1(a=a)
+        u, v = my_subwf(a=x)
```
wild-endeavor, line 192: Do you end up with two `node-0`'s in the current scheme?

PR #207, opened 2020-10-14
[ignore] Annotations changes

wild-endeavor:
Ignore unless you're ketan/haytham.

* Delete all the freeform param stuff
* Remove the fancier (original, on master) Interface object in favor of the straight model
* Factor out the loop that makes things into Promise objects
* Remove nested subwf execution arg
* Move transformation of wf outputs in local exec to the call function (search for "unpack result")
* Add SdkNode creation on subworkflow call
* Add list binding creation
* Change workflow binding to use the binding creator in the flytekit_engine file
* Allow workflow compile to be given native inputs

### DIFF
--- a/flytekit/annotated/context_manager.py
+++ b/flytekit/annotated/context_manager.py
@@ -93,7 +93,6 @@ class FlyteContext(object):
         self._execution_state = execution_state
         self._flyte_client = flyte_client
         self._user_space_params = user_space_params
-        self._freeform_params = []
 
     def __enter__(self):
         # Should we auto-assign the parent here?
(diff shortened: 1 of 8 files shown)


wild-endeavor:
After more consideration I still want to make the local dynamic execution two steps - I think we should compile and then execute.  The reason is that this is what will happen in production, and we should mimic to the extent possible - I think it eliminates a whole class of problems that can come up.

wild-endeavor:
After writing the above comment, I realized that we can't... or rather, we can compile, but the thing produced is a workflow, as in a workflow template.  We can't execute it without writing a full local execution engine for workflows... it's only possible now because we're leveraging the Python interpreter as the workflow engine.

PR #213, opened 2020-10-15
[ignore] annotations serialization

wild-endeavor:
ignore unless ketan


flytekit/annotated/interface.py
```diff
@@ -1,7 +1,7 @@
 import copy
 import inspect
 from collections import OrderedDict
-from typing import Dict, Generator, Union, Type, Tuple, List, Any, _GenericAlias, TypeVar
```
wild-endeavor, line 4: _GenericAlias gives problems in cookbook, I think cuz the cookbook image uses 3.6.  I don't think we should support 3.6, but I also don't like the idea of importing _GenericAlias... Take a look at the change below to see if it's acceptable.

flytekit/annotated/interface.py
```diff
@@ -260,3 +260,14 @@ def t(a: int, b: str) -> None: ...
 
     return_map = OrderedDict(zip(list(output_name_generator(len(return_types))), return_types))
     return return_map
+
+
+# Doesn't really belong here, not sure where else to put it
+class ControlPlaneSettings(object):
```
kumare3, line 266: should this not be in the context? as one more layer?

flytekit/annotated/task.py
```diff
@@ -208,6 +221,31 @@ def execute(self, **kwargs) -> Any:
...
+                   self._task_function.__name__,
+                   "--inputs",
+                   "{{.input}}",
+                   "--output-prefix",
+                   "{{.outputPrefix}}",
+                   "--raw-output-data-prefix",
+                   "{{.rawOutputDataPrefix}}",
+               ]
+        env =            {
+                _internal_config.CONFIGURATION_PATH.env_var: _internal_config.CONFIGURATION_PATH.get(),
```
kumare3, line 243: can we take this also as an input?

kumare3: and in the higher layer if you want translate it from the config?

tests/flytekit/unit/use_scenarios/unit_testing/test_type_hints.py
```diff
@@ -172,6 +172,28 @@ def my_wf(a: int, b: str) -> (int, str):
     }
 
 
+def test_wf1_with_list_of_inputs():
```
kumare3, line 169: we should also probably add a test that says 
"workflow.serialize()"
"task.serialize()"

flytekit/annotated/interface.py
```diff
@@ -244,9 +242,11 @@ def t(a: int, b: str) -> None: ...
         logger.debug(f'Task returns a single output of type {return_annotation}')
         return {default_output_name(): return_annotation}
 
-    if isinstance(return_annotation, _GenericAlias) and return_annotation.__origin__ in [list, dict]:
+    if hasattr(return_annotation, '_name') and return_annotation._name == 'List' and return_annotation.__origin__ == list:
```
kumare3, line 245: haha not import _GenericAlias, but look for a private property :)

flytekit/annotated/promise.py
```diff
@@ -1,36 +1,65 @@
...
+            b = task_2(in1=5, in2=[a, in2])
+            return b
+
+    Here, in task_2, during execution we'd have a list of Promises. We have to make sure to give task2 a Flyte
+    LiteralCollection (Flyte's name for list), not a Python list of Flyte literals.
     """
-    # TODO: Are there any pass-by-reference considerations we need to think about?  I don't think so. Write unit
-    #  tests to be sure.
+
+    def extract_value(ctx: FlyteContext, input_val: Any, flyte_literal_type: _type_models.LiteralType) -> Any:
```
kumare3, line 34: why is the return type Any? It seems that the return is always a Literal?

flytekit/annotated/interface.py
wild-endeavor: one more layer?  you mean like an object inside the CompilationState object?  I think that would make the most sense, a registration settings object or something...

flytekit/annotated/promise.py
wild-endeavor: cuz i was wrong.

flytekit/annotated/interface.py
kumare3: yup, like in the compilation state. I thought we had discussed this. This would also make it possible to serialize with different compilation states?

flytekit/annotated/task.py
```diff
@@ -208,6 +221,36 @@ def execute(self, **kwargs) -> Any:
     def native_interface(self) -> Interface:
         return self._native_interface
 
+    def get_registerable_entity(self, settings: ControlPlaneSettings) -> _registerable.RegisterableEntity:
+        if self._registerable_entity is not None:
+            return self._registerable_entity
+
+        from flytekit.configuration import internal as _internal_config
+        args = [
```
kumare3, line 229: I think we should make the get_command an overridable function so that spark can easily override it?

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -79,18 +83,48 @@ def serialize_all(project, domain, pkgs, version, folder=None):
     # m = module (i.e. python file)
     # k = value of dir(m), type str
     # o = object (e.g. SdkWorkflow)
+
+    # Clear annotated Flyte entities
+    # TODO: Decide if we should reload, and if so, move this clear after old style entities have been dealt with.
+    flyte_context.FlyteEntities.entities.clear()
```
kumare3, line 89: i want to understand what this is for?

kumare3:
I really like the idea of getting a registrable entity. 
So now this is an overridable property and we should just split things in there so that those can be overriden easily

flytekit/clis/sdk_in_container/serialize.py
wild-endeavor: this should probably be deleted.

flytekit/annotated/task.py
wild-endeavor: Sure sounds good.  Currently the entire get_registerable_entity function is overridable.  We can change this to any combination of get container, get command, get entity, etc.

flytekit/annotated/interface.py
wild-endeavor: doing this half-way for now.  moved it into FlyteContext but leaving it out of compilation for now.  Couldn't work out how to mesh compilation and registration.  Can revisit after tomorrow.

PR #215, opened 2020-10-20
[ignore] Annotations - change file access pattern

wild-endeavor:
* Remove the data proxy object in the context manager and replace it with a heavier weight file access manager (`FileAccessProvider`) class that encapsulates all the various data proxies.
* Update engine code to remove the last traces of the original `DataProxy` opting instead for always retrieving data proxies from the context.
* Add a temporary `node_name` override hack to get demo to work.  We can clean this up when the functionality is expanded more generally.
* Removing some tests that have been commented for a long time.

Hopefully this makes developing flytekit a bit easier.  I'm now able to run stuff like this.
```
docker run -e PROJECT=flytesnacks -v `pwd`:/root -v /Users/ytong/go/src/github.com/lyft/flytekit/flytekit:/opt/venv/lib/python3.8/site-packages/flytekit -v /tmp/output/flytekit:/tmp/output -it flytecookbook:latest bash
```
and then within the container
```
FLYTE_INTERNAL_CONFIGURATION_PATH=/root/sandbox.config FLYTE_PLATFORM_CLOUD_PROVIDER=local pyflyte-execute --task-module recipes.aaa.simple --task-name t1 --inputs file:///tmp/output/t1/inputs.pb --output-prefix file:///tmp/output/t1
```


kumare3:
@wild-endeavor why do we have to run the 2nd command inside the container? if we have a virtualenv is that not enough?

flytekit/bin/entrypoint.py
```diff
@@ -140,17 +143,34 @@ def _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data
                     tmp_dir=user_workspace_dir,
                 )
 
-                with ctx.new_data_proxy_by_cloud_provider(
-                    cloud_provider=cloud_provider, raw_output_data_prefix=raw_output_data_prefix
-                ) as ctx:
+                if cloud_provider == _constants.CloudProvider.AWS:
```
kumare3, line 146: how do we get the cloudprovider? Should this not be based on the prefix in the file?

kumare3:
Thank you for doing this. So I eventually want it to be just local file based

kumare3 (approved):

flytekit/bin/entrypoint.py
wild-endeavor: It's retrieved here https://github.com/lyft/flytekit/pull/215/files#diff-932d345851356237c198d139fa9a11d53d68257dc05387e933680ed2e764a50fR106

flytekit/interfaces/data/data_proxy.py
```diff
@@ -167,3 +172,179 @@ def get_remote_directory(cls):
...
+            # Note that we default to the local one here, not the remote one.
+            return self.local_access
+        elif path.startswith("/"):
+            # Note that we default to the local one here, not the remote one.
+            return self.local_access
+        raise Exception(f"Unknown file access {path}")
+
+    @property
+    def aws(self) -> _s3proxy.AwsS3Proxy:
+        if self._aws is None:
```
honnix, line 227: Should this rather be `hasattr(self, "_aws")` ?

PR #218, opened 2020-10-23
Lazy load Sagemaker-training

bnsblue:


flytekit/plugins/__init__.py
```diff
@@ -15,6 +15,8 @@
 hmsclient = _lazy_loader.lazy_load_module("hmsclient")  # type: _lazy_loader._LazyLoadModule
 type(hmsclient).add_sub_module("genthrift.hive_metastore.ttypes")
 
+sagemaker_training = _lazy_loader.lazy_load_module("sagemaker-training")  # type: _lazy_loader._LazyLoadModule
+
```
matthewphsmith, line 19: Is the hyphen right? I think you need an underwcore

bnsblue: Good catch. thanks!

flytekit/plugins/__init__.py
```diff
@@ -26,3 +28,5 @@
 )
 
 _lazy_loader.LazyLoadPlugin("hive_sensor", ["hmsclient>=0.0.1,<1.0.0"], [hmsclient])
+
+_lazy_loader.LazyLoadPlugin("sagemaker-training", ["sagemaker-training>=3.6.2,<4.0.0"], [sagemaker_training])
```
EngHabu, line 32: `_` or `-` ?
```suggestion
_lazy_loader.LazyLoadPlugin("sagemaker_training", ["sagemaker-training>=3.6.2,<4.0.0"], [sagemaker_training])
```

EngHabu (approved):

flytekit/plugins/__init__.py
```diff
@@ -15,6 +15,8 @@
 hmsclient = _lazy_loader.lazy_load_module("hmsclient")  # type: _lazy_loader._LazyLoadModule
 type(hmsclient).add_sub_module("genthrift.hive_metastore.ttypes")
 
+sagemaker_training = _lazy_loader.lazy_load_module("sagemaker-training")  # type: _lazy_loader._LazyLoadModule
```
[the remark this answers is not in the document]
bnsblue: @kumare3 Maybe I misunderstood what Lazy Load is used for? `sagemaker_training` needs to be installed in the container if the users wants to do custom training on SM because it will put a `train` script in PATH. It's not really used in user code but it needs to be installed.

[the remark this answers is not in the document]
bnsblue: cc @EngHabu ^

EngHabu (approved):

PR #219, opened 2020-10-23
[ignore] Annotations dynamic

wild-endeavor:
unless ketan/haytham


flytekit/annotated/context_manager.py
```diff
@@ -48,10 +48,15 @@ def env(self) -> Dict[str, str]:
...
 
 class CompilationState(object):
-    def __init__(self):
+    def __init__(self, prefix: str):
         self.nodes: List[Node] = []
+        self._prefix = prefix
         self.mode = 1  # TODO: Turn into enum in the future, or remove if only one mode.
 
+    @property
+    def prefix(self) -> str:
```
kumare3, line 57: what does prefix in compilation context mean? Can we add a small doc?

flytekit/annotated/node.py
```diff
@@ -36,7 +38,8 @@ def get_registerable_entity(self) -> SdkNode:
             raise Exception("Node flyte entity none")
 
         if self._flyte_entity._registerable_entity is None:
-            raise Exception("Node registerable entity has not been set")
+            print("WARNING!!!!!!!!!!!!!!!!!!!  not really")
```
kumare3, line 41: ?

flytekit/annotated/task.py
```diff
@@ -86,7 +88,7 @@ def _compile(self, ctx: FlyteContext, *args, **kwargs):
         # TODO: Make the metadata name the full name of the (function)?
         sdk_node = Node(
             # TODO
-            id=f"node-{len(ctx.compilation_state.nodes)}",
+            id=f"{ctx.compilation_state.prefix}node-{len(ctx.compilation_state.nodes)}",
```
kumare3, line 91: should we just add a method to compilation context that generates the node id? and internally uses the prefix? This seems error prone

flytekit/annotated/task.py
```diff
@@ -433,6 +469,55 @@ def execute(self, **kwargs) -> Any:
...
+        try:
+            native_outputs = self.execute(**native_inputs)
+        except Exception as e:
+            logger.exception("Exception when executing", e)
+            raise e
+        logger.info(f"Task executed successfully in user level, outputs: {native_outputs}")
+
+        # Short circuit the translation to literal map cuz what's returned is a dj spec (or an already-constructed
+        # LiteralMap if the dynamic task was a no-op), not python native values
+        if ctx.execution_state and ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION:
```
kumare3, line 498: so basically you want to modify the dispatch execute to return dynamicjobspec if that exists or return literalMap. Why cant we check for the djspec type?

flytekit/annotated/node.py
wild-endeavor: i'm removing this if block entirely so this becomes fully recursive.  I was kinda worried about an infinite loop situation for some reason but I don't think that can happen... not unless new entities get created somehow, but at least then we'll run out of memory too.

flytekit/annotated/task.py
wild-endeavor: i'm trying to resolve the case where you do

```
@workflow
def wf():
  @workflow
  def wf2():
    ...
```

flytekit/annotated/task.py
wild-endeavor: done.

PR #221, opened 2020-10-24
Simple Schema type support -> pandas dataframe

kumare3:
Adds a new pandas dataframe transformer. This results in a generic schema

### DIFF
--- a/flytekit/annotated/workflow.py
+++ b/flytekit/annotated/workflow.py
@@ -1,7 +1,7 @@
 import datetime
 import inspect
 import typing
-from typing import Callable, Dict, List, Optional, Tuple, Union, Any
+from typing import Any, Callable, Dict, List, Optional, Tuple, Union
 
 import flytekit.annotated.promise
 import flytekit.annotated.type_engine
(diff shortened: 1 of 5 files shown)


flytekit/annotated/workflow.py
```diff
@@ -152,9 +152,12 @@ def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], P
...
 
-        if len(output_names) != len(function_outputs):
-            # Length check, clean up exception
-            raise Exception(f"Length difference {len(output_names)} {len(function_outputs)}")
+        if len(output_names) > 1 and isinstance(function_outputs, tuple):
+            if len(output_names) != len(function_outputs):
+                # Length check, clean up exception
+                raise Exception(f"Length difference {len(output_names)} {len(function_outputs)}")
+        elif len(output_names) == 1 and isinstance(function_outputs, Promise):
+            return function_outputs
```
wild-endeavor, line 160: is the stuff below not needed?

kumare3: what do you mean? its like elif right? the indentation here looks weird, but seems fine in the code

wild-endeavor: But it returns here right, so the stuff below won't run.  Should it not run?  I added some Promise renaming stuff but maybe it's not needed?

kumare3: done

PR #223, opened 2020-10-27
Raw container support

kumare3:
ContainerTask implemented and works!


flytekit/annotated/task.py
```diff
@@ -472,7 +593,7 @@ def execute(self, **kwargs) -> Any:
 
 class DynamicWorkflowTask(PythonFunctionTask):
     def __init__(self, dynamic_workflow_function: Callable, metadata: _task_model.TaskMetadata, *args, **kwargs):
-        super().__init__(dynamic_workflow_function, metadata, *args, **kwargs)
+        super().__init__(dynamic_workflow_function, metadata, task_type="dynamic_task", *args, **kwargs)
```
wild-endeavor, line 596: Can we copy in the current strings in `SdkTaskType` so that the rest of the system respects it as well?

wild-endeavor (commented):
did you want to do a mocking example in this PR?

flytekit/annotated/task.py
kumare3: sure will do

kumare3:
I will do the mocking as a separate PR.

PR #226, opened 2020-10-28
[ignore] Annotations launch plans

wild-endeavor:
ignore unless ketan/haytham.

* Adds a new Launch Plan class in the same pattern as the new Task and Workflow classes
* Remove the identifier_model.Identifier instantiations in Task and Workflow now that we handle it at serialization time.
* Call the type engine to create literals when making the ParameterMap in interface.
* Stupid oneof in Parameter tripped me up for hours. Thanks Haytham.

### DIFF
--- a/flytekit/annotated/context_manager.py
+++ b/flytekit/annotated/context_manager.py
@@ -19,7 +19,16 @@ from flytekit.models.core import identifier as _identifier
 
 
 class RegistrationSettings(object):
-    def __init__(self, project: str, domain: str, version: str, image: str, env: Optional[Dict[str, str]], iam_role: Optional[str] = None, service_account: Optional[str] = None):
+    def __init__(
+        self,
+        project: str,
+        domain: str,
+        version: str,
+        image: str,
+        env: Optional[Dict[str, str]],
+        iam_role: Optional[str] = None,
+        service_account: Optional[str] = None,
+    ):
         self._project = project
         self._domain = domain
         self._version = version
@@ -49,11 +58,11 @@ class RegistrationSettings(object):
         return self._env
 
     @property
-    def iam_role(self)-> Optional[str]:
+    def iam_role(self) -> Optional[str]:
         return self._iam_role
 
     @property
-    def service_account(self)-> Optional[str]:
+    def service_account(self) -> Optional[str]:
         return self._service_account
 
 
--- a/flytekit/annotated/context_manager.py
+++ b/flytekit/annotated/context_manager.py
@@ -16,6 +16,7 @@ from flytekit.configuration import sdk as _sdk_config
 from flytekit.engines.unit import mock_stats as _mock_stats
 from flytekit.interfaces.data import data_proxy as _data_proxy
 from flytekit.models.core import identifier as _identifier
+from flytekit.models.common import RawOutputDataConfig
 
 
 class RegistrationSettings(object):
@@ -28,6 +29,7 @@ class RegistrationSettings(object):
         env: Optional[Dict[str, str]],
         iam_role: Optional[str] = None,
         service_account: Optional[str] = None,
+        raw_output_data_config: Optional[str] = None,
     ):
         self._project = project
         self._domain = domain
@@ -36,6 +38,7 @@ class RegistrationSettings(object):
         self._env = env or {}
         self._iam_role = iam_role
         self._service_account = service_account
+        self._raw_output_data_config = raw_output_data_config
 
     @property
     def project(self) -> str:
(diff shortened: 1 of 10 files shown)


flytekit/annotated/launch_plan.py
```diff
@@ -0,0 +1,168 @@
+from __future__ import annotations
```
kumare3, line 1: why __future__

flytekit/annotated/launch_plan.py
```diff
@@ -0,0 +1,168 @@
...
+from flytekit.models import common as _common_models
+from flytekit.models import interface as _interface_models
+from flytekit.models import launch_plan as _launch_plan_models
+from flytekit.models import literals as _literal_models
+from flytekit.models import schedule as _schedule_model
+from flytekit.models.core import identifier as _identifier_model
+
+
+class LaunchPlan(object):
+    # The reason we cache is simply because users may get the default launch plan twice for a single Workflow. We
```
kumare3, line 23: how is that possible, Flyteadmin will reject it right?

flytekit/annotated/launch_plan.py
wild-endeavor: this is what you do if you want delayed type checking in 3.7, so that you can do `-> ClassName` instead of `-> "ClassName"`

flytekit/annotated/launch_plan.py
wild-endeavor: Admin will be fine yeah, i just mean locally.  I think it makes more sense to return the same object if for instance the same launch plan is used in two different parent workflows.

flytekit/annotated/context_manager.py
```diff
@@ -16,15 +16,29 @@
...
 class RegistrationSettings(object):
-    def __init__(self, project: str, domain: str, version: str, image: str, env: Optional[Dict[str, str]]):
+    def __init__(
+        self,
+        project: str,
+        domain: str,
+        version: str,
+        image: str,
+        env: Optional[Dict[str, str]],
+        iam_role: Optional[str] = None,
```
kumare3, line 30: can we make this a union of [ServiceAccount | IAMRole]

kumare3: for string we cannot i guess. So ignore

flytekit/annotated/launch_plan.py
kumare3: aah TIL

flytekit/annotated/launch_plan.py
```diff
@@ -0,0 +1,233 @@
...
+    # don't want to create two defaults, could be confusing.
+    CACHE = {}
+
+    @staticmethod
+    def native_kwargs_to_literal_map(
```
kumare3, line 25: can we add this method to TypeEngine ?

flytekit/annotated/launch_plan.py
```diff
@@ -0,0 +1,233 @@
...
+            workflow=workflow,
+            parameters=parameter_map,
+            fixed_inputs=_literal_models.LiteralMap(literals={}),
+        )
+
+        LaunchPlan.CACHE[workflow.name] = lp
+        return lp
+
+    @staticmethod
+    def create(
```
kumare3, line 55: typically a class method :)

flytekit/annotated/launch_plan.py
```diff
@@ -0,0 +1,233 @@
...
+        lp = LaunchPlan(name=name, workflow=workflow, parameters=wf_signature_parameters, fixed_inputs=fixed_lm)
+
+        # This is just a convenience - we'll need the fixed inputs LiteralMap for when serializing the Launch Plan out
+        # to protobuf, but for local execution and such, why not save the original Python native values as well so
+        # we don't have to reverse it back every time.
+        default_inputs.update(fixed_inputs)
+        lp._saved_inputs = default_inputs
+
+        if name in LaunchPlan.CACHE:
+            logger.warning(f"Launch plan named {name} was already created! Make sure your names are unique.")
```
kumare3, line 92: so we are no overriding?

kumare3: should we raise an AssertionError?

flytekit/annotated/launch_plan.py
wild-endeavor: done

flytekit/annotated/launch_plan.py
wild-endeavor: sure

flytekit/annotated/launch_plan.py
wild-endeavor: Interface already depends on TypeEngine.  Don't want a circular dependency.  How about we add to Interface?

kumare3: in that case maybe just a helper method?

PR #238, opened 2020-11-06
[ignore] Annotations filepath handling cleanup

wild-endeavor:


flytekit/annotated/type_engine.py
```diff
@@ -391,31 +404,57 @@ def to_literal(
...
+            if remote_path is not None:
+                ctx.file_access.put_data(source_path, remote_path, is_multipart=False)
+            meta = BlobMetadata(type=self._blob_type(format=FlyteFilePathTransformer.get_format(python_type)))
+            return Literal(scalar=Scalar(blob=Blob(metadata=meta, uri=remote_path)))
 
     def to_python_value(
         self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[flyte_typing.FlyteFilePath]
     ) -> flyte_typing.FlyteFilePath:
+
+        uri = lv.scalar.blob.uri
```
kumare3, line 441: this should never happen right?

kumare3: looks weird to me

wild-endeavor: hmm?  is this comment on the right line?  I'm just making a local variable cuz there's too many dots.

PR #239, opened 2020-11-06
Annotations file arrangement

wild-endeavor:
* Added a new `create_and_link_node` function in node.py to replace the compile method in task, launch plan, and workflow.
* Fix the file handling behavior
* Removed the `native_kwargs_to_literal_map` in launch plan cuz I realized we can use the `translate_inputs_to_literals` function instead.
* Added a ReferenceTask, no workflows or launch plans yet.
* Started moving things around.  Removed all the flytekit imports from setup.py and replaced with literals - we can work out a better solution in the future, maybe ask around for suggestions.
* Moved logger out of flytekit init into a separate file, but imported it back into init, along with most things users will need.
* Made typing a folder and moved flytefile into a file.  Renamed `FlyteFilePath` to `FlyteFile`


Makefile
```diff
@@ -32,7 +32,9 @@ lint: ## Run linters
 
 .PHONY: test
 test: lint ## Run tests
-	pytest tests/flytekit/unit
+	pytest tests/flytekit/unit -k "not test_task"
```
wild-endeavor, line 35: will remove before merging.

tests/flytekit/unit/models/test_dynamic_job.py
```diff
@@ -29,7 +29,7 @@
 ]
 
 
-@pytest.mark.parametrize("task", LIST_OF_DYNAMIC_TASKS)
+@pytest.mark.parametrize("task", LIST_OF_DYNAMIC_TASKS[0:100])
```
wild-endeavor, line 32: will remove before merging

flytekit/annotated/type_engine.py
```diff
@@ -368,64 +368,103 @@ def to_literal(
...
+        if not ctx.file_access.is_remote(uri):
+            return uri
 
+        # Since no delayed downloading is possible with strings, always download immediately.
+        ctx.file_access.get_data(lv.scalar.blob.uri, local_destination_path, is_multipart=False)
+        return local_destination_path
 
-class FlyteFilePathTransformer(TypeTransformer[flyte_typing.FlyteFilePath]):
+
+class FlyteFilePathTransformer(TypeTransformer[flytekit.typing.flyte_file.FlyteFile]):
```
kumare3, line 396: should this be kept next to the `typing` code?

kumare3:
lgtm

PR #240, opened 2020-11-06
Fix fetched workflows

wild-endeavor:
# TL;DR
Fetched workflows come with their underlying tasks.  When a fetched workflow is then used in a parent workflow, those tasks become "upstream entities" of the parent workflow.  At registration time, flytekit should know to skip those.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
See https://flyte-org.slack.com/archives/CNMKCU6FR/p1604681225118200 for more information.

## Tracking Issue
NA

## Follow-up issue
NA

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -2,7 +2,7 @@ import logging as _logging
 
 import flytekit.plugins  # noqa: F401
 
-__version__ = "0.14.3"
+__version__ = "0.14.4"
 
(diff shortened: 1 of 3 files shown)


wild-endeavor:
was accidentally merged directly into master.

PR #242, opened 2020-11-09
Add retries to Get/Idempotent Calls

wild-endeavor:
# TL;DR
Add retries to get/idempotent calls to safely handle any transient network/admin issues.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/664

## Follow-up issue
_NA_
OR


flytekit/clients/raw.py
```diff
@@ -86,7 +87,11 @@ def handler(*args, **kwargs):
                         refresh_handler_fn = _get_refresh_handler(_creds_config.AUTH_MODE.get())
                         refresh_handler_fn(args[0])
                     else:
-                        raise
+                        if i == (retries - 1):
+                            raise
+                        else:
+                            _logging.error(f"Non-auth RPC error {e}, sleeping 5 seconds and retrying")
+                            time.sleep(5)
```
anandswaminathan, line 94: Wait, we are sleeping for 5 seconds? 

That's extremely high for all calls. Can we do it only on waitforcompletion. ? This will kill all synchronous calls to Flyteadmin. Imagine a service using flytekit to call flyte. 5 seconds will kill the callers of the service

akhurana001: So even a get status request can be invoked behind a service ? That way we can't do retries at all or maybe just 1-2

anandswaminathan: Also do it only for Get calls

wild-endeavor: if that's the case, maybe just scrap this PR entirely.  Maybe we can just do retries within the wait_for_completion function?  Where is that function defined?

akhurana001: I am planning to create another handler here with retries to use for all get/idempotent calls only.. 

Also would retry with lower wait times (a few 100ms with exponential back off)

akhurana001: done, PTAL

flytekit/clients/raw.py
```diff
@@ -96,6 +97,46 @@ def handler(*args, **kwargs):
...
+                    else:
+                        if i == (retries - 1):
+                            raise
+                        else:
+                            # Start with 200ms wait-time and exponentially back-off upto 1 second.
+                            wait_time = min(200 * (2 ** retries), max_wait_time)
+                            _logging.error(f"Non-auth RPC error {e}, sleeping {wait_time}ms and retrying")
+                            time.sleep(wait_time / 1000)
+        except _RpcError as e:
+            if e.code() == _GrpcStatusCode.ALREADY_EXISTS:
```
anandswaminathan, line 132: This should not happen for GEt right?

flytekit/clients/raw.py
```diff
@@ -96,6 +97,46 @@ def handler(*args, **kwargs):
...
+        """
+        retries = 3
+        max_wait_time = 1000
+
+        try:
+            for i in range(retries):
+                try:
+                    return fn(*args, **kwargs)
+                except _RpcError as e:
+                    if e.code() == _GrpcStatusCode.UNAUTHENTICATED:
```
anandswaminathan, line 117: what's this special case for UNAUTHENTICATED?

anandswaminathan (dismissed):

flytekit/clients/raw.py
akhurana001: Not just for get, hope that this to be used for any mutating calls as well as long as they are idempotent

flytekit/clients/raw.py
akhurana001: This already exited in the legacy handler above, keeping it as is. I believe the diff is no backoff here, not sure why. Would let @wild-endeavor  comment on this

wild-endeavor: this is what triggers the code to attempt to authenticate, either by popping up a browser in the flyte-cli case, or by using a username password in the pyflyte case.

wild-endeavor:
+1.  want to bump the version too? (version is currently in two places, the version file and setup.py).

Also since I opened this PR i can't approve.  You can approve on your end or open another PR?

wild-endeavor:
@akhurana001 how'd testing go?  Feel free to approve and merge when you feel comfortable.

akhurana001 (approved):

PR #252, opened 2020-11-13
Annotations preserve extensions

wild-endeavor:
* Keep file extension on download
* Moving the FlyteFile related things out to a separate file.


flytekit/types/flyte_file.py
```diff
@@ -237,13 +238,22 @@ def to_python_value(
...
             return expected_python_type(uri)
 
         # For the remote case, return an FlyteFile object that can download
-        local_path = ctx.file_access.get_random_local_path()
+        source_file_name = ctx.file_access.get_filename_only(uri)
+        if source_file_name:
+            # Let's check to make sure we're not overwriting anything
+            local_path = os.path.join(ctx.file_access.local_access.sandbox, source_file_name)
+            if os.path.exists(local_path):
+                flytekit.logger.info(f"Intended path already exists {local_path}, using random")
```
kumare3, line 253: this might lead to circular dependencies. I think the logger should not be at the flytekit level, but underneath

flytekit/types/flyte_file.py
```diff
@@ -237,13 +238,22 @@ def to_python_value(
...
         uri = lv.scalar.blob.uri
+
         # This is a local file path, like /usr/local/my_file, don't mess with it. Certainly, downloading it doesn't
         # make any sense.
         if not ctx.file_access.is_remote(uri):
             return expected_python_type(uri)
 
         # For the remote case, return an FlyteFile object that can download
-        local_path = ctx.file_access.get_random_local_path()
+        source_file_name = ctx.file_access.get_filename_only(uri)
```
kumare3, line 248: this is only for download, how about upload?

flytekit/interfaces/data/local/local_file_proxy.py
```diff
@@ -29,6 +29,10 @@ def __init__(self, sandbox):
         """
         self._sandbox = sandbox
 
+    @property
+    def sandbox(self) -> str:
```
kumare3, line 33: i thought you dont like sandbox ?

kumare3:
lgtm 👍

flytekit/interfaces/data/local/local_file_proxy.py
wild-endeavor: i don't... but it's already there!

PR #253, opened 2020-11-13
Multi image support

wild-endeavor:
Example of Multi-Image support
https://github.com/lyft/flytesnacks/pull/44


flytekit/annotated/context_manager.py
```diff
@@ -52,8 +113,8 @@ def version(self) -> str:
         return self._version
 
     @property
-    def image(self) -> str:
-        return self._image
+    def image(self) -> ImageConfig:
```
wild-endeavor, line 116: can we rename this to image_config?

flytekit/annotated/task.py
```diff
@@ -764,6 +826,7 @@ def task(
     interruptible: bool = False,
     deprecated: str = "",
     timeout: Union[_datetime.timedelta, int] = 0,
+    container_image: str = None,
```
wild-endeavor, line 829:
```suggestion
    container_image: Optional[str] = None,
```

wild-endeavor: and in the @property too

tests/flytekit/unit/annotated/test_serialization.py
```diff
@@ -102,11 +118,47 @@ def my_wf(a: int) -> str:
...
     )
     with ctx.current_context().new_registration_settings(registration_settings=registration_settings):
         wf = my_wf.get_registerable_entity()
         assert wf is not None
         assert len(wf.nodes) == 2
         assert wf.nodes[1].branch_node is not None
+
+
+def test_serialization_images():
+    @task(container_image="{{.image.xyz.fqn}}:{{.image.default.version}}")
```
wild-endeavor, line 137: can we add a couple tests with hardcoded strings as well?  like without the {{}}

kumare3: done!

flytekit/annotated/task.py
kumare3: Done

kumare3:
the check was successful, but it does not update here

PR #260, opened 2020-11-19
Annotations reference workflow

wild-endeavor:


flytekit/annotated/workflow.py
```diff
@@ -289,11 +291,57 @@ def get_registerable_entity(self) -> _SdkWorkflow:
         return self._registerable_entity
 
 
-def workflow(_workflow_function=None):
+class ReferenceWorkflow(Workflow):
```
kumare3, line 294: should we make it such that a ReferenceWorkflow can be created inline, without needing a function. Something like this
```python
w = Workflow(reference=WorkflowReference(...), inputs=kwtypes(), outputs=kwtypes())
```

kumare3: I know you made the task also, to use an empty function, but if we factor it so, then we can implement both methods?

tests/flytekit/unit/annotated/test_type_hints.py
```diff
@@ -1057,3 +1083,33 @@ def my_wf(a: int) -> str:
 
     with pytest.raises(TypeError):
         my_wf(a=5)
+
+
+def test_fff():
```
kumare3, line 1088: can we name it better?

flytekit/annotated/testing.py
```diff
@@ -35,3 +37,28 @@ def _log(*args, **kwargs):
...
+    if not isinstance(target, PythonTask) and not isinstance(target, Workflow):
+        raise Exception("Can only use mocks on tasks/workflows declared in Python.")
+
+    def wrapper(test_fn):
+        def new_test(*args, **kwargs):
+            logger.warning(f"Invoking mock method for task: '{target.name}'")
+            m = MagicMock()
+            if isinstance(target, PythonTask):
+                saved = target.execute
+                target.execute = m
```
kumare3, line 52: I made a dummy method that logs that the mock has been invoked,  do you want to add that here to? Helps in debugging IMO

flytekit/annotated/testing.py
```diff
@@ -35,3 +37,28 @@ def _log(*args, **kwargs):
     t.execute = _log
     yield m
     t.execute = _captured_fn
+
+
+def patch(target: Union[PythonTask, Workflow]):
+    if not isinstance(target, PythonTask) and not isinstance(target, Workflow):
```
kumare3, line 43: like this type-check. Can you add it to `task_mock` as well?

kumare3:
couple comments, then lgtm

flytekit/annotated/testing.py
wild-endeavor: copied the warning above

flytekit/annotated/workflow.py
wild-endeavor: How about just a `ReferenceEntity` but keep the other ones as is?

flytekit/annotated/testing.py
wild-endeavor: sure.

flytekit/annotated/testing.py
kumare3: where is the copy?

wild-endeavor: line 52

flytekit/annotated/testing.py
```diff
@@ -35,3 +41,29 @@ def _log(*args, **kwargs):
...
+        raise Exception("Can only use mocks on tasks/workflows declared in Python.")
+
+    def wrapper(test_fn):
+        def new_test(*args, **kwargs):
+            logger.warning(f"Invoking mock method for task: '{target.name}'")
```
wild-endeavor, line 52:
```suggestion
            logger.warning(f"Invoking mock method for target: '{target.name}'")
```

PR #262, opened 2020-11-20
Annotations - Hive

wild-endeavor:
https://github.com/lyft/flyte/issues/631

### DIFF
--- a/flytekit/taskplugins/hive/task.py
+++ b/flytekit/taskplugins/hive/task.py
@@ -4,7 +4,7 @@ from google.protobuf.json_format import MessageToDict
 
 from flytekit.annotated.task import SQLTask, TaskPlugins, PythonTask
 from flytekit.models import task as _task_model
-from flytekit.models.qubole import QuboleHiveJob
+from flytekit.models.qubole import QuboleHiveJob, HiveQuery
 from flytekit.types.schema import FlyteSchema
 
 
@@ -19,6 +19,7 @@ class HiveTask(SQLTask):
         cluster_label: str,
         inputs: Dict[str, Type],
         output_schema_type: Type[FlyteSchema],
+        stage_query_template: str,
         query_template: str,
         *args,
         **kwargs,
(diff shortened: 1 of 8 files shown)


tests/flytekit/unit/annotated/test_type_hints.py
```diff
@@ -273,13 +273,13 @@ def t1() -> datetime.datetime:
...
-    def my_wf() -> str:
+    def my_wf() -> FlyteSchema:
         dt = t1()
         return sql(ds=dt)
 
     with task_mock(sql) as mock:
-        mock.return_value = "Hello"
-        assert my_wf() == "Hello"
+        mock.return_value = pandas.DataFrame(data={"x": [1, 2], "y": ["3", "4"]})
+        assert (my_wf().open().all() == pandas.DataFrame(data={"x": [1, 2], "y": ["3", "4"]})).all().all()
```
wild-endeavor, line 282: is this the right way to use the FlyteSchema?

flytekit/taskplugins/hive/task.py
```diff
@@ -0,0 +1,106 @@
...
+        DROP TABLE {{{{ .PerRetryUniqueKey }}}};
+        """
+
+    def __init__(
+        self,
+        name: str,
+        cluster_label: str,
+        inputs: Dict[str, Type],
+        select_query: str,
+        output_schema_type: Type[FlyteSchema],
```
kumare3, line 88: What if we dont want an output? I guess that will be a separate type of hiveTask?

tests/flytekit/unit/annotated/test_type_hints.py
kumare3: yes, but you cannot just compare 2 pandas dataframes like this
you have to compare, `==` which returns a new dataframe, that you assert has all `True`

Look at some examples of how I am doing it

kumare3:
cool 🆒

flytekit/taskplugins/hive/task.py
wild-endeavor: This is the select task though... if they don't want an output, they can always use the barebones hive task.

PR #266, opened 2020-11-30
Improved spark sessions handling & user context params

kumare3:


EngHabu (approved):
I wish we add more doc strings in all of the new types....

PR #267, opened 2020-12-02
Spark DataFrames handled as a type if using spark

kumare3:
- This is also an overhaul of the schema system to support remote IO
   based dataframes and regular pandas style dataframes. All of them can
be added at the user layer by implementing 2 classes SchemaReader
/ SchemaWriter.
 - If the new dataframe is to be added as a supported type then, it
   contributor needs to implement the TypeTransformer interface as well

### DIFF
--- a/flytekit/types/schema.py
+++ b/flytekit/types/schema.py
@@ -4,6 +4,7 @@ import datetime as _datetime
 import os
 import typing
 from abc import abstractmethod
+from dataclasses import dataclass
 from enum import Enum
 from typing import Type
 
@@ -11,7 +12,6 @@ import numpy as _np
 
 from flytekit import FlyteContext
 from flytekit.annotated.type_engine import T, TypeEngine, TypeTransformer
-from flytekit.configuration import sdk
 from flytekit.models.literals import Literal, Scalar, Schema
 from flytekit.models.types import LiteralType, SchemaType
 from flytekit.plugins import pandas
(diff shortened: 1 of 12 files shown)


flytekit/types/schema.py
```diff
@@ -132,17 +259,31 @@ def open(
...
-            self._downloader(self._remote_path, self._local_path)
-            self._downloaded = True
-        return SchemaEngine.open(t=dataframe_fmt, s=self, mode=override_mode if override_mode else self._supported_mode)
+        mode = override_mode if override_mode else self._supported_mode
+        h = SchemaEngine.get_handler(dataframe_fmt)
+        if not h.handles_remote_io:
+            # The Schema Handler does not manage its own IO, and this it will expect the files are on local file-system
+            if self._supported_mode == SchemaOpenMode.READ and not self._downloaded:
+                # Only for readable objects if they are not downloaded already, we should download them
+                # WRite objects should already have everything written to
```
wild-endeavor, line 276:
```suggestion
                # Write objects should already have everything written to
```

tests/flytekit/unit/annotated/test_type_hints.py
```diff
@@ -1104,3 +1105,53 @@ def wf(x: int, y: int) -> Datum:
...
+        new_df = session.createDataFrame([("Bob", 10)], my_schema.column_names())
+        return df.union(new_df)
+
+    @workflow
+    def my_wf() -> my_schema:
+        df = my_dataset()
+        return my_spark(df=df)
+
+    x = my_wf()
+    reader = x.open(pandas.DataFrame)
```
wild-endeavor, line 1152: what happens if `pandas.DataFrame` is not specified?

wild-endeavor (approved):
some minor comments

tests/flytekit/unit/annotated/test_type_hints.py
kumare3: default is pandas.DataFrame actually, I can drop it

PR #273, opened 2020-12-04
Silence output when optimistically checking for fast-register archive

katrogan:
# TL;DR
As an optimization, fast-register checks to see if a remote distribution exists before uploading it. Confusingly this prints a 404 in the reasonable case where the distribution does not exist. Annoyingly, if aws retries in a flytekit config are set to > 1 this 404 prints multiple times and is pretty alarming. This change silences the output of that command.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
https://docs.google.com/document/d/1n7DmKXaOv8dAb9eHD4TKqka6FsdJNs3Tugpf1I6nvMg/edit?usp=sharing

## Tracking Issue
https://github.com/lyft/flyte/issues/554

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/tools/fast_registration.py
+++ b/flytekit/tools/fast_registration.py
@@ -1,5 +1,7 @@
+import contextlib as _contextlib
 import os as _os
 import subprocess as _subprocess
+import sys as _sys
 import tarfile as _tarfile
 import tempfile as _tempfile
 from pathlib import Path as _Path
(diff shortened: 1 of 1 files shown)


wild-endeavor (commented):
is this the right approach?  redirecting to dev/null seems kinda heavy handed.  Also, is there a way to not retry four times?  are you referring to the awscli setting?

PR #274, opened 2020-12-09
Support map as input/output type

honnix:
# TL;DR
Add native map support as input/output type.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Largely copied implementation of list, and fixed where `NotImplementedError` was raised.

## Tracking Issue
https://github.com/lyft/flyte/issues/637

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/common/types/containers.py
+++ b/flytekit/common/types/containers.py
@@ -149,3 +149,162 @@ class TypedListImpl(ListImpl, metaclass=TypedCollectionType):
             len(self.collection.literals),
             ",\n\t\t".join("\n\t\t".join(v.verbose_string().splitlines()) for v in self.collection.literals),
         )
+
+
+class MapType(_base_sdk_types.FlyteSdkType):
+    pass
+
+
+class TypedMapType(MapType):
+    @property
+    def sub_type(cls):
+        """
+        :rtype: flytekit.common.types.base_sdk_types.FlyteSdkType
+        """
+        return cls._sub_type
+
+    def __eq__(cls, other):
+        return hasattr(other, "sub_type") and cls.sub_type == other.sub_type
+
+    def __hash__(cls):
+        # Python 3 checks complain if hash isn't implemented at the same time as equals
+        return super(TypedMapType, cls).__hash__()
+
+
+def Map(sdk_type):
+    """
+    :param flytekit.common.types.base_sdk_types.FlyteSdkType sdk_type:
+    :rtype: flytekit.common.types.base_sdk_types.FlyteSdkType
+    """
+
+    class TMap(TypedMapImpl):
+        _sub_type = sdk_type
+
+    # TODO: Figure out generics and type-hinting
+    return TMap
+
+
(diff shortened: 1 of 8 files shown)


flytekit/common/types/containers.py
```diff
@@ -149,3 +149,162 @@ def verbose_string(self):
...
+        # loading naturally interprets all layers, not just the outer layer.
+        return cls({k: cls.sub_type.from_python_std(v) for k, v in dictionary.items()})
+
+    @classmethod
+    def is_castable_from(cls, other):
+        """
+        :param flytekit.common.types.base_literal_types.FlyteSdkType other:
+        :rtype: bool
+        """
+        return type(cls) == type(other) and cls.sub_type.is_castable_from(other.sub_type)
```
honnix, line 233: I'm not entirely sure this is the right way to implement. I checked the code of `list` but it didn't pass any unit test I tried.

honnix:
cc @kumare3

kumare3:
@honnix do you mind waiting on the new flytekit api for adding map type support. It is already supported in the new api

kumare3:
Here is the code for it
https://github.com/lyft/flytekit/blob/annotations/flytekit/annotated/type_engine.py#L303

And the usage
https://github.com/lyft/flytekit/blob/annotations/tests/flytekit/unit/annotated/test_type_hints.py#L948-L1024

honnix:
@kumare3 Thanks for taking a look. Yes I took a look at the new type engine before and it looks a lot simpler and cleaner!

This is not that urgent because we have a workaround to skip returning a map as output. We would love to use the new API/SDK. What is the expected time to get out of alpha?

kumare3:
@honnix alpha0 is out https://pypi.org/project/flytekit/0.16.0a0/. @wild-endeavor and I are working on releasing alpha1 which makes it almost feature complete (except for array/map tasks - different from dynamic tasks). We are hoping to have alpha1 early next week?

honnix:
@kumare3 That sounds great. I guess it won't hurt that we start using (at least trying) alpha version.

BTW, is there any example showing how to use `FlyteContext` to set `auth_role`? Also how do we set `schedule` now when creating a launch plan (the factory method doesn't have such argument as I can see). I guess we can take this on Slack. Sorry for derailing. :)

honnix:
There is a problem with the new API, as I see it requires Python 3.7+ due to `from __future__ import annotations`. Is that absolutely needed?

wild-endeavor:
@honnix I thought I asked everyone that, sorry!  Can you not do 3.7?  We changed our minds a while back and decided to do 3.7 instead of just 3.6 and there's better typing support (even if hidden behind a future import).

honnix:
@wild-endeavor Yeah I remember that conversion. It's not a big deal and we can prepare to get move to 3.7+. I will close this for now, unless we get an urgent need for it (things like we cannot wait for readiness of the new API, but unlikely to happen).

PR #275, opened 2020-12-11
Proper setting of fetched entities

wild-endeavor:
# TL;DR
Entities fetched from the control plane don't behave the same as entities declared locally.  This change circumvents some of the behavior.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Please see the [Slack](https://flyte-org.slack.com/archives/CNMKCU6FR/p1607624069238500) conversation for background.  This was discovered when a user yielded a launch plan that was fetched from within a dynamic task.  The code in sdk_dynamic was attempting to assign a name to the fetched launch plan (which didn't work because the fetching was done inside a function). If it had been done outside the function, the name would've been set to the variable that the fetched launch plan was assigned to, which is even worse.

## Tracking Issue
NA

## Follow-up issue
NA

### DIFF
--- a/flytekit/common/launch_plan.py
+++ b/flytekit/common/launch_plan.py
@@ -51,7 +51,7 @@ class SdkLaunchPlan(
         :param flytekit.models.launch_plan.LaunchPlanSpec model:
         :rtype: SdkLaunchPlan
         """
-        return cls(
+        launch_plan = cls(
             workflow_id=_identifier.Identifier.promote_from_model(model.workflow_id),
             default_inputs=_interface_models.ParameterMap(
                 {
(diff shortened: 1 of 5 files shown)


flytekit/common/launch_plan.py
```diff
@@ -116,12 +116,13 @@ def fetch(cls, project, domain, name, version=None):
 
         sdk_lp = cls.promote_from_model(lp.spec)
         sdk_lp._id = lp.id
+        sdk_lp.assign_name(name)
```
matthewphsmith, line 119: is name not merely a shortcut to the name field in the ID? Or is there another action taken as part of assigning a name?

wild-endeavor: another field is set, something that is read in sdk_dynamic.

PR #280, opened 2020-12-16
Add notifications, schedules to new flytekit

katrogan:


wild-endeavor:
Can we broaden this PR a bit?  Instead of just adding notifications, can we also add schedules, auth (probably as two fields - iam role or service account), and rawoutputdataconfig?

Also why do we need to create a new Notification class?  Can we use `flytekit.models.common.Notification`, `flytekit.models.schedule.Schedule`, `flytekit.models.common.AuthRole`, and `flytekit.models.common.RawOutputDataConfig`

Also, where did we land on stuff having both iam role and service account?   @anandswaminathan is that a thing we decided to do?

EngHabu:
@anandswaminathan has a doc on deprecating KIAM... I'm not sure if it's public (I think it's ok to make public... if so, please do @anandswaminathan)... I do NOT recall that we actually decided this is the way to go... here is my recollection:
- First iteration: Let user repos provision these service accounts into flyte clusters.. CON: we keep fiddling with these clusters all the time, what happens if we move them to different clusters, or we add new clusters... etc. etc.
- Second iteration: Admin will provision these service accounts before it launches a launchplan.. CON: additional steps, more prune to errors

Honorable mentioned:
- Admin provisions these service accounts along with cluster resource manager templates. CON: Launchplans can define their own IAM roles that where never previously provisioned with admin and we like to keep this flexibility.

Orthogonal to all of this: We continue to need IAM roles for AWS Batch plugin.

@anandswaminathan ?

katrogan:
@wild-endeavor schedules, roles, etc is all in the pipeline :) i'm still not very familiar with new flytekit so i prefer a piecemeal approach for PRs so i can get feedback before doing a lot more work

> Also why do we need to create a new Notification class? 

For validation that notification phases are only terminal.

katrogan:
@wild-endeavor mind taking another look? the lp auth role is already implemented https://github.com/lyft/flytekit/blob/annotations/flytekit/annotated/launch_plan.py#L189,L194

flytekit/annotated/notification.py
```diff
@@ -0,0 +1,68 @@
+from typing import List
+
+from flytekit.models import common as _common_model
+from flytekit.models.core import execution as _execution_model
+
+
+class Notification(_common_model.Notification):
```
wild-endeavor, line 7: Add a comment that this is duplicate because we don't want to deal with metaclass?

flytekit/annotated/schedule.py
```diff
@@ -0,0 +1,150 @@
+import datetime
+import re as _re
+
+import croniter as _croniter
+
+from flytekit.models import schedule as _schedule_models
+
+
+class CronSchedule(_schedule_models.Schedule):
```
wild-endeavor, line 9: Add a comment that this is duplicate because we don't want to deal with metaclass?

flytekit/models/schedule.py
```diff
@@ -158,7 +158,7 @@ def from_flyte_idl(cls, pb2_object):
         :param flyteidl.admin.schedule_pb2.Schedule pb2_object:
         :rtype: Schedule
         """
-        return cls(
+        return Schedule(
```
wild-endeavor, line 161: This is cuz you explicitly don't want the inheritance?  Add a comment if that's the case?

flytekit/annotated/notification.py
katrogan: done

katrogan:
@honnix

honnix:
Thanks for implementing this! Regarding auth role, how would a user pass in when creating a launch plan?

PR #281, opened 2020-12-17
Annotations ref entity

wild-endeavor:
This reworks how reference entities are done.

* I realized that for reference entities, it doesn't matter if they're a workflow, a launch plan, or a task, the behavior we want is roughly the same.  The simplest thing to do is to mimic what a task does.  As such, I've basically copied the task's logic into a new class called `ReferenceEntity`, the only difference is that I've renamed `dispatch_execute` to `unwrap_literal_map_and_execute`.  There's definitely some refactoring to do in the future, perhaps even between tasks and workflows.  We know that the `create_and_link_node` behavior is already common across all three.
* Added an `execute` function to the `Workflow` - this allows us to simplify the mocking/patch functions.
  * Note that the mock functions don't handle LaunchPlans yet.  This is because LaunchPlans don't have an execute function, the `__call__` is delegated to the workflow.
* All three reference objects now derive from the same base. Note that the constructor for `ReferenceEntity` does not call `super()` and that the `Workflow/LaunchPlan/Task` is always second in the inheritance order. This means that the constructor for the `Workflow/LaunchPlan/Task` is not called for ReferenceEntities.
* Because of circular dependencies, I couldn't put `ReferenceTask` alongside `PythonFunctionTask` so I had to make a separate file for it.
* Bumping version - but no we won't release a new alpha just yet.
* Reorder the mypy check - it prints out too much stuff right now, putting it at the top so we don't have to scroll to see linting errors.
* Ran `make requirements`


flytekit/annotated/launch_plan.py
```diff
@@ -228,3 +229,36 @@ def get_registerable_entity(self) -> SdkLaunchPlan:
             version=settings.version,
         )
         return self._registerable_entity
+
+
+class ReferenceLaunchPlan(ReferenceEntity, LaunchPlan):
```
kumare3, line 234: can we get some docs on this? Especially as this is a user facing entity

flytekit/annotated/launch_plan.py
```diff
@@ -228,3 +229,36 @@ def get_registerable_entity(self) -> SdkLaunchPlan:
...
+    def __init__(
```
kumare3, line 235: how is `latest` version handled?

flytekit/annotated/launch_plan.py
```diff
@@ -228,3 +229,36 @@ def get_registerable_entity(self) -> SdkLaunchPlan:
...
+        self, project: str, domain: str, name: str, version: str, inputs: Dict[str, Type], outputs: Dict[str, Type]
+    ):
+        super().__init__(_identifier_model.ResourceType.LAUNCH_PLAN, project, domain, name, version, inputs, outputs)
+
+    def get_registerable_entity(self) -> SdkLaunchPlan:
```
kumare3, line 240: another instance where I would love to move the Code from here into a separate package

flytekit/annotated/launch_plan.py
```diff
@@ -228,3 +229,36 @@ def get_registerable_entity(self) -> SdkLaunchPlan:
             version=settings.version,
         )
         return self._registerable_entity
+
+
+class ReferenceLaunchPlan(ReferenceEntity, LaunchPlan):
+    def __init__(
+        self, project: str, domain: str, name: str, version: str, inputs: Dict[str, Type], outputs: Dict[str, Type]
```
kumare3, line 236: should we make the inputs / outputs as `interface`? and eventially `interface=auto` could be a supported thing?

flytekit/annotated/reference_entity.py
```diff
@@ -0,0 +1,200 @@
...
+from flytekit.annotated.type_engine import TypeEngine
+from flytekit.common.exceptions import user as _user_exceptions
+from flytekit.loggers import logger
+from flytekit.models import dynamic_job as _dynamic_job
+from flytekit.models import interface as _interface_models
+from flytekit.models import literals as _literal_models
+from flytekit.models.core import identifier as _identifier_model
+
+
+class Reference(object):
```
kumare3, line 16: how about making this a dataclass? I think that is what it is right?

flytekit/annotated/reference_entity.py
```diff
@@ -0,0 +1,200 @@
...
+        self, resource_type: int, project: str, domain: str, name: str, version: str, *args, **kwargs,
+    ):
+        self._id = _identifier_model.Identifier(resource_type, project, domain, name, version)
+
+    @property
+    def id(self) -> _identifier_model.Identifier:
+        return self._id
+
+
+class TaskReference(Reference):
```
kumare3, line 27: these could all be simple derived dataclasses

flytekit/annotated/reference_entity.py
```diff
@@ -0,0 +1,200 @@
...
+    def __init__(
+        self, project: str, domain: str, name: str, version: str, *args, **kwargs,
+    ):
+        super().__init__(_identifier_model.ResourceType.WORKFLOW, project, domain, name, version, *args, **kwargs)
+
+
+class ReferenceEntity(object):
+    def __init__(
+        self,
+        resource_type,
```
kumare3, line 51: instead of Resource_type, project, domain etc, can we not just accept `Reference` type?

flytekit/annotated/reference_task.py
```diff
@@ -1,43 +1,21 @@
...
-
-class WorkflowReference(Reference):
-    def __init__(
-        self, project: str, domain: str, name: str, version: str, *args, **kwargs,
-    ):
-        super().__init__(_identifier_model.ResourceType.WORKFLOW, project, domain, name, version, *args, **kwargs)
-
-
-class ReferenceTask(PythonTask):
+class ReferenceTask(ReferenceEntity, PythonFunctionTask):
```
kumare3, line 14: can we call this TypedReferenceTask? as after this, a ReferenceTask is not required to have a interface declaration?

flytekit/annotated/reference_task.py
```diff
@@ -46,49 +24,49 @@ def __init__(
...
 
-    def execute(self, **kwargs) -> Any:
-        raise Exception("Remote reference tasks cannot be run locally. You must mock this out.")
+        self._registerable_entity: Optional[SdkTask] = None
 
-    @property
-    def reference(self) -> TaskReference:
-        return self._reference
+    @classmethod
+    def create_from_get_entity(
```
kumare3, line 47: thats a weird method name - with is `*_get_entity` Also dont understand the usecase for this method

kumare3:
A few comments

flytekit/annotated/launch_plan.py
wild-endeavor: Sure.

flytekit/annotated/launch_plan.py
wild-endeavor: It's not... can we do it in another PR?

flytekit/annotated/launch_plan.py
wild-endeavor: yes, agreed.

flytekit/annotated/launch_plan.py
wild-endeavor: We can but then it's one more thing for the user to have to import in addition to `kwtypes` and the `Interface` object is kinda an internal detail right?  I think we can support automatic in the future by just not specifying inputs/outputs.  Either way.

flytekit/annotated/reference_entity.py
wild-endeavor: sure

flytekit/annotated/reference_task.py
wild-endeavor: The ReferenceTask class is hooked up to the task decorator (the bit that picks the constructor to use based on the config class type).  That function calls the task in a specific way

```
task_instance = TaskPlugins.find_pythontask_plugin(type(task_config))(
       task_config, fn, _metadata, container_image=container_image, *args, **kwargs
)
```

This means that the constructor for ReferenceTask is different from launch plans and workflows, so I wanted another object creation function that has the same signature.

flytekit/annotated/reference_task.py
wild-endeavor: sure

flytekit/annotated/reference_entity.py
wild-endeavor: I think so, let me do this.

flytekit/annotated/launch_plan.py
kumare3: sure

flytekit/annotated/launch_plan.py
kumare3: ok, what have we done for hive etc, lets keep it the same

flytekit/annotated/reference_task.py
kumare3: and we have to use `@task` annotation for this? Do you think a different one makes sense, especially because `container_image` is really weird?

wild-endeavor: Nope, don't have to use @task for it at all.  I can rip that out completely so that users have to use the `get_reference_entity` function for tasks.  Just thought the empty function body style looked kinda nice but if it's just getting in the way, let's get rid of it.

flytekit/annotated/reference_entity.py
honnix: This design is a bit weird and not so consistent. I guess it is due to we use both class hierarchy and  property to differentiate types, so there are cases `Reference(_identifier_model.ResourceType.TASK, "foo", "bar", "foo.bar", "v") != TaskReference("foo", "bar", "foo.bar", "v")`, for example when looking up a task plugin. I haven't looked into details of this PR, but may it solves it?

wild-endeavor: We decided to nix the looking up of task plugins and just go with an explicit `@reference_task` decorator.  

The inequality you mention is a bit unfortunate yeah, do you have any recommendations?

Thanks.

flytekit/annotated/launch_plan.py
wild-endeavor: yeah hive follows inputs/outputs.

flytekit/annotated/reference.py
```diff
@@ -0,0 +1,25 @@
...
+
+from typing import Dict, Type
+
+from flytekit.annotated.launch_plan import ReferenceLaunchPlan
+from flytekit.annotated.reference_task import ReferenceTask
+from flytekit.annotated.workflow import ReferenceWorkflow
+from flytekit.models.core import identifier as _identifier_model
+
+
+def get_reference_entity(
```
kumare3, line 11: is this just used for testing?

flytekit/annotated/reference_entity.py
kumare3: I think what @honnix wants to say probably is that the base type should not be instantiable on its own?

flytekit/annotated/reference.py
wild-endeavor: no, this is exported at the flytekit layer.  do you not like it?  I thought it'd be helpful.  I should probably add a raise Exception though.

wild-endeavor: i can delete it if you think it's not helpful.

flytekit/annotated/reference_task.py
```diff
@@ -1,94 +1,57 @@
...
+        tk._id = self.id
         tk._has_registered = True
+        tk.assign_name(self.reference.id.name)
         return tk
 
-    @property
-    def id(self) -> _identifier_model.Identifier:
-        return self._reference._id
 
+def reference_task(
```
kumare3, line 37: I like this, so much cleaner?

flytekit/annotated/testing.py
```diff
@@ -43,25 +44,23 @@ def _log(*args, **kwargs):
...
             logger.warning(f"Invoking mock method for target: '{target.name}'")
             m = MagicMock()
-            if isinstance(target, PythonTask):
-                saved = target.execute
-                target.execute = m
-            else:
-                saved = target._workflow_function
-                target._workflow_function = m
+            saved = target.execute
+            target.execute = m
```
kumare3, line 61: what happened to Workflow?

kumare3: nm, just saw it

wild-endeavor: See comment in pr desc.  I added an execute function to workflow.

kumare3:
Looks good, but lets wait for @honnix to reply?

flytekit/annotated/reference_entity.py
honnix: Yeah that could be a solution if we make Reference abstract. But this is not a big deal and we don't need to block on it.

PR #283, opened 2020-12-17
More mild warning message if no flytekit.conf

honnix:
# TL;DR
More mild and less seemingly harmful warning message is logged when `flytekit.conf` cannot be found.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
When an `flytekit.conf` file with absolute path cannot be found, we log the original configuration as well as point out a different possibility.

## Tracking Issue
https://github.com/lyft/flyte/issues/646

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/configuration/__init__.py
+++ b/flytekit/configuration/__init__.py
@@ -17,9 +17,12 @@ def set_flyte_config_file(config_file_path):
     import flytekit.configuration.internal as _internal
 
     if config_file_path is not None:
+        original_config_file_path = config_file_path
         config_file_path = _os.path.abspath(config_file_path)
         if not _pathlib.Path(config_file_path).is_file():
-            _logging.warning("Invalid flyte config_file_path {} specified.".format(config_file_path))
+            _logging.warning(
+                f"No config file provided or invalid flyte config_file_path {original_config_file_path} specified."
+            )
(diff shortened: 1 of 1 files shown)


flytekit/configuration/__init__.py
```diff
@@ -17,9 +17,12 @@ def set_flyte_config_file(config_file_path):
     import flytekit.configuration.internal as _internal
 
     if config_file_path is not None:
+        original_config_file_path = config_file_path
         config_file_path = _os.path.abspath(config_file_path)
         if not _pathlib.Path(config_file_path).is_file():
-            _logging.warning("Invalid flyte config_file_path {} specified.".format(config_file_path))
+            _logging.warning(
+                f"No config file provided or invalid flyte config_file_path {original_config_file_path} specified."
```
katrogan, line 24: shouldn't we only get to this case if a config file **is** provided?

honnix: This config item has a default value, so even user doesn't config, at runtime this attempt will be made.

katrogan (approved):

honnix:
Thanks @katrogan

PR #284, opened 2020-12-18
Auth, environment and resources

katrogan:


flytekit/annotated/container_task.py
```diff
@@ -63,7 +63,7 @@ def execute(self, **kwargs) -> Any:
         return None
 
     def get_container(self, settings: RegistrationSettings) -> _task_model.Container:
-        env = settings.env
+        env = {**settings.env, **self.environment} if self.environment else settings.env
```
wild-endeavor, line 66: this works?  cool, new to me.  and the second one takes precedence?

wild-endeavor (commented):
can you add one of these tests
https://github.com/lyft/flytekit/blob/annotations/tests/flytekit/unit/annotated/test_type_hints.py#L435

like sdk_task = t.get_registerable_entity() and then confirm it's passed to the sdktask as well?

flytekit/annotated/container_task.py
katrogan: this adds them together. is that undesirable?

wild-endeavor: that is desirable, just wondering what happens if the same key is in both.

katrogan:
@wild-endeavor 

> can you add one of these tests

done

PR #285, opened 2020-12-21
[wip] Add class method for easier Reference construction

honnix:
# TL;DR
Add `from_id` as a class method for easier `Reference` construction.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [w] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
In case users already have an ID, this makes constructing a reference easier and less verbose.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/annotated/reference_task.py
+++ b/flytekit/annotated/reference_task.py
@@ -11,6 +11,10 @@ from flytekit.models.core import identifier as _identifier_model
 
 
 class Reference(object):
+    @classmethod
+    def from_id(cls, id: _identifier_model.Identifier, *args: Any, **kwargs: Any):
+        return cls(id.project, id.domain, id.name, id.version, *args, **kwargs)
+
     def __init__(
         self, type: int, project: str, domain: str, name: str, version: str, *args, **kwargs,
     ):


honnix:
@wild-endeavor Wdyt of this? If this is useful I can go on adding test cases.

wild-endeavor:
Sure, but can we wait til we merge this PR first? https://github.com/lyft/flytekit/pull/281/files  Feel free to take a look at that actually.

And thank you as always!

honnix:
Sure, let's wait.

wild-endeavor:
Can this change be changed to work with the new style? https://github.com/lyft/flytekit/blob/annotations/flytekit/annotated/reference_entity.py#L19

honnix:
@wild-endeavor I am a bit confused. Now defining a reference task and a reference workflow looks different, with one through `@reference_task`, one through `@workflow(reference=WorkflowReference(...))`. What is the thought behind that? Will everything be changed to `@reference_xxx` style eventually?

kumare3:
@wild-endeavor / @honnix I think we should have a uniform style - like `@reference_workflow` and `@reference_task`. I guess the only reason to keep them different is to indicate the difference

honnix:
@kumare3 That makes sense. I guess in the next alpha drop we can have those.

Coming back to this PR, I ended up having a wrapper decorator as following. Not sure how much sense it makes if we accept both optional ID and (p+d+n+v) in flytekit.
```py
def reference_task(
    id: _Identifier,
) -> Callable:
    return _cast(
        Callable,
        _reference_task(
            project=id.project,
            domain=id.domain,
            name=id.name,
            version=id.version,
        ),
    )
```

Nevertheless I think `reference_task` can be simplified because `_task_function` will always be `None` and p+d+n+v should alway be passed in. Maybe something like:
```py
def reference_task(
    project: str,
    domain: str,
    name: str,
    version: str,
) -> Callable:
    def wrapper(fn) -> ReferenceTask:
        interface = transform_signature_to_interface(inspect.signature(fn))
        return ReferenceTask(project, domain, name, version, interface.inputs, interface.outputs)

    return wrapper
```

PR #286, opened 2020-12-21
Mypy

honnix:
# TL;DR
Add mypy as part of linting check.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Enable mypy with strict option which is usually a good default, as well as ignoring missing imports.

## Tracking Issue
https://github.com/lyft/flyte/issues/650

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/types/schema.py
+++ b/flytekit/types/schema.py
@@ -6,7 +6,7 @@ import typing
 from abc import abstractmethod
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any, Type
+from typing import Type
 
 import numpy as _np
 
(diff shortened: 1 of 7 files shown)


flytekit/types/schema.py
```diff
@@ -299,7 +299,7 @@ def as_readonly(self) -> FlyteSchema:
 
 
 class FlyteSchemaTransformer(TypeTransformer[FlyteSchema]):
-    _SUPPORTED_TYPES: typing.Dict[type : SchemaType.SchemaColumn.SchemaColumnType] = {
+    _SUPPORTED_TYPES: typing.Dict[Any, SchemaType.SchemaColumn.SchemaColumnType] = {
```
kumare3, line 302: should this be `Type`

wild-endeavor (approved):

setup.cfg
```diff
@@ -15,6 +15,13 @@ per-file-ignores =
     *:F821
     */__init__.py: F401
 
+[mypy]
+strict_optional = True
+ignore_missing_imports = True
+files = flytekit/annotated/**/*.py,tests/flytekit/unit/annotated/**/*.py
```
kumare3, line 21: can we also add `flytekit/taskplugins/**` and `flytekit/types/**`

kumare3:
Thank you for the quick turnaround @honnix, 2 small nits

flytekit/types/schema.py
honnix: I think here `Any` seems to be correct. `Type` is used to annotate class objects, but here I guess it should just be instance of some type, right?

setup.cfg
honnix: Sure

flytekit/types/schema.py
honnix: Ah wait, you are right. I miss understood the code.

honnix:
PTAL again and feel free to merge. I'm signing off for today.

PR #291, opened 2020-12-23
Annotations - flyte directory

wild-endeavor:
This adds a FlyteDirectory type that maps to multipart Blobs.  It's pretty similar to FlyteFile.

* Found a small issue with FlyteFile, calling `FlyteFile(s3://bucket)` would prepend an os abspath to it.  Added a test.
* Separate out local and mock_remote directories.  If not, when trying to `cp -R` from one to the other, you get exceptions
* Change the `new_execution_context` function to take an existing working dir so that the spark `pre_execute` stuff doesn't nest working dirs


flytekit/types/flyte_directory.py
```diff
@@ -0,0 +1,212 @@
...
+    def __fspath__(self):
+        self._downloader()
+        self._downloaded = True
+        return self._path
+
+    @classmethod
+    def extension(cls) -> str:
+        return ""
+
+    def __class_getitem__(cls, item: typing.Type) -> typing.Type[FlyteDirectory]:
```
kumare3, line 78: TBH when I am using these, I am not loving them to debug. They show weird errors. But, lets move with it, I want to make sure we go through and re-think this interface

flytekit/types/flyte_directory.py
```diff
@@ -0,0 +1,212 @@
...
+        return self._remote_source
+
+    def __repr__(self):
+        return self._path
+
+    def __str__(self):
+        return self._path
+
+
+class FlyteDirectoryTransformer(TypeTransformer[FlyteDirectory]):
```
kumare3, line 123: please add comments

flytekit/types/flyte_directory.py
```diff
@@ -0,0 +1,212 @@
...
+    def __init__(self):
+        super().__init__(name="FlyteDirectory", t=FlyteDirectory)
```
kumare3, line 125: I usually call if FlyteDir->MultipartBlob-Transformer - more descriptive

flytekit/types/flyte_directory.py
```diff
@@ -0,0 +1,212 @@
...
+        def noop():
+            ...
+
+        self._path = path
+        self._downloader = downloader or noop
+        self._downloaded = False
+        self._remote_directory = remote_directory
+        self._remote_source = None
+
+    def __fspath__(self):
```
kumare3, line 69: when is this invoked? For a directory, should we not support list-dir?
And should we download all at once, or allow a streaming download so that you can read as some are downloaded?

kumare3: I see it is invoked for listdir as well. I like that too.
So lets keep this simple, I agree. Something that downloads at once a very simple interface. We can later build a streamingDirectory?

flytekit/types/flyte_directory.py
```diff
@@ -0,0 +1,212 @@
...
+
+        def t1(in1: FlyteDirectory["svg"]):
+            ...
+
+    As outputs:
+
+    The contents of this local directory will be uploaded to the Flyte store.
+        return FlyteDirectory("/path/to/dir/")
+
+        return FlyteDirectory["svg"]("/path/to/dir/", remote_path="s3://special/output/location")
```
katrogan, line 37: what does it mean for a FlyteDirectory to have a type like svg?

kumare3: all files are svg? I deally I think I want to create something called as a flyteDataset, where there is a dataset index file, for every file in the dataset, each with some properties etc

kumare3:
@wild-endeavor can we add a comment / NOTE on this type that it should not be used for very large datasets, as merely listing the dataset will cause the entire dataset to be downloaded. Honestly, listing on S3 and other backend object stores is not consistent and should not need data to be downloaded, but the current structure of FlyteDirectory does not support that

flytekit/types/flyte_directory.py
wild-endeavor: added comment about list dir.  Yeah a streaming directory of some kind in the future sounds good.  I think we'll have a better idea of how to implement things when the streaming story is clearer.

flytekit/types/flyte_directory.py
wild-endeavor: it's not really clear, but I wanted to support it cuz the IDL supports it.

wild-endeavor:
Will add comment... but do you think we can one day implement things correctly changing user code?  Should be able to right?

wild-endeavor:
We can maybe add `list()` and other functions that are a bit smarter in the future at least.  I don't know if we'll ever be able to differentiate between the system open() call and the os.listdir() call though, that sounds too deep in python internals.

flytekit/types/flyte_directory.py
```diff
@@ -0,0 +1,223 @@
...
+        # and just return a literal with a uri equal to the path given
+        if ctx.file_access.is_remote(source_path) or not should_upload:
+            meta = BlobMetadata(type=self._blob_type(format=self.get_format(python_type)))
+            return Literal(scalar=Scalar(blob=Blob(metadata=meta, uri=source_path)))
+
+        # For local paths, we will upload to the Flyte store (note that for local execution, the remote store is just
+        # a subfolder), unless remote_path=False was given
+        else:
+            if remote_directory is None:
+                remote_directory = ctx.file_access.get_random_remote_directory()
```
katrogan, line 193: is get_random_remote_directory deterministic? should you save it once you assign it so to_literal() returns the same value?

wild-endeavor: It is random.  I'm gonna think about that as part of another PR though if that's okay...

PR #293, opened 2020-12-24
Task config

kumare3:


flytekit/annotated/reference_task.py
```diff
@@ -36,10 +36,10 @@ def get_task_structure(self) -> SdkTask:
 
 def reference_task(
```
wild-endeavor, line 37: Ugh, I should change this...  i'll add it to the list.  these shouldn't be optional.

wild-endeavor (approved):
This is so much nicer.

flytekit/taskplugins/hive/task.py
```diff
@@ -10,20 +11,29 @@
 from flytekit.types.schema import FlyteSchema
 
 
-class HiveTask(SQLTask):
+@dataclass
+class HiveConfig(object):
+    cluster_label: str
+    tags: Optional[List[str]] = None
```
katrogan, line 17: Can you add a comment on what these params are for?

katrogan (approved):

flytekit/taskplugins/hive/task.py
kumare3: doing it now

PR #294, opened 2020-12-24
Sidecar tasks are now Pod tasks

katrogan:


flytekit/taskplugins/pod/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .task import Pod
+
+__all__ = [Pod]
```
kumare3, line 3: so the `__all__`` should be a list of string names of the modules you want to expose. 
I think you can just skip it, refer to my new PR

kumare3 (approved):
lgtm, we can wait for @jeevb to comment?

katrogan:
thanks @kumare3 also posted in flytekit slack and will wait until next week to see if anyone has strong opinions

jeevb (approved):
LGTM! To clarify, does this seamlessly work with dynamic tasks?

katrogan:
@jeevb thanks for reviewing! we haven't yet ported over dynamic sidecar tasks (dynamic pod tasks?) to the new flytekit but that is in the pipeline

PR #296, opened 2020-12-29
Annotations minor cleanup and create_node

wild-endeavor:
Ignore the branch name.

I know it doesn't look like it, but I think this is cleaning up.

* If a task returns a named tuple with one output (our convention for naming), then we still return one of our namedtuples with length 1
* Use the name of the user-specified named tuple if there is one.
* Moved some tests around.


flytekit/annotated/base_task.py
```diff
@@ -294,7 +305,14 @@ def dispatch_execute(
 
             expected_output_names = list(self.interface.outputs.keys())
             if len(expected_output_names) == 1:
-                native_outputs_as_map = {expected_output_names[0]: native_outputs}
+                # Here we have to handle the fact that the task could've been declared with a typing.NamedTuple of
+                # length one. That convention is used for naming outputs - and single-length-NamedTuples are
+                # particularly troublesome but elegant handling of them is not a high priority
+                # Again, we're using the custom_interface_name as a proxy.
+                if self.python_interface.custom_interface_name and isinstance(native_outputs, tuple):
```
kumare3, line 312: shouldn't be custom interface name right? should be output tuple name?

flytekit/annotated/interface.py
```diff
@@ -180,7 +189,14 @@ def transform_signature_to_interface(signature: inspect.Signature) -> Interface:
...
 
-    return Interface(inputs, outputs)
+    # This is just for typing.NamedTuples - in those cases, the user can select a name to call the NamedTuple. We
+    # would like to preserve that name in our custom collections.namedtuple.
+    custom_name = None
+    if hasattr(signature.return_annotation, "_field_types"):
+        if hasattr(signature.return_annotation, "__name__") and signature.return_annotation.__name__ != "":
+            custom_name = signature.return_annotation.__name__
+
+    return Interface(inputs, outputs, custom_interface_name=custom_name)
```
kumare3, line 199: I would call it output_name

flytekit/annotated/interface.py
```diff
@@ -258,10 +274,14 @@ def t(a: int, b: str) -> Dict[str, int]: ...
     if hasattr(return_annotation, "__origin__") and return_annotation.__origin__ is tuple:
         # Handle option 3
         logger.debug(f"Task returns unnamed typing.Tuple {return_annotation}")
+        if len(return_annotation.__args__) == 1:
+            raise FlyteValidationException("Please don't use a tuple if you're just returning one thing.")
```
kumare3, line 278:
```suggestion
            raise FlyteValidationException("Tuples should be used to indicate multiple return values, found only one return variable.")
```

flytekit/annotated/promise.py
```diff
@@ -362,23 +362,43 @@ def __str__(self):
...
+) -> Optional[Union[Tuple[Promise], Promise]]:
     if promises is None:
         return None
 
     if isinstance(promises, Promise):
         return promises
 
     if len(promises) == 0:
-        return None
+        raise Exception("jfkdlsjafklsdajlksda")
```
kumare3, line 375: Fix this?

tests/flytekit/unit/annotated/test_realworld_examples.py
```diff
@@ -103,10 +103,10 @@ def split_traintest_dataset(
         # We will fake train test split. Just return the same dataset multiple times
         return x, x, y, y
 
+    nt = typing.NamedTuple("Outputs", model=FlyteFile[MODELSER_JOBLIB])
```
kumare3, line 106: can we name this tuple better, like NamedOutputs or NamedModel?

kumare3: I guess its ok here, but in flytesnacks

tests/flytekit/unit/annotated/test_serialization.py
```diff
@@ -204,20 +204,20 @@ def my_wf(a: int, b: str) -> (int, str):
 def test_serialization_branch():
     @task
     def mimic(a: int) -> typing.NamedTuple("OutputsBC", c=int):
-        return a
+        return (a,)
```
kumare3, line 207: Such a weird syntax, comma and paranthesis

flytekit/annotated/promise.py
```diff
@@ -362,23 +362,43 @@ def __str__(self):
...
-    class Output(collections.namedtuple("TaskOutput", variables)):
+    # These should be OrderedDicts so it should be safe to do this.
+    if entity_interface:
+        variables = [k for k in entity_interface.outputs.keys()]
+
+    named_tuple_name = "DefaultNamedTupleOutput"
+    if entity_interface and entity_interface.custom_interface_name:
+        named_tuple_name = entity_interface.custom_interface_name
+
+    # Should this class be part of the Interface?
```
wild-endeavor, line 400: @kumare3 what do you think about this?

kumare3: what do you mean part? like it is an inner class in interface? It has to be generated on the fly, because of the named tuple semantics

flytekit/annotated/condition.py
```diff
@@ -175,6 +176,7 @@ def _compute_outputs(self, n: Node) -> Union[Promise, Tuple[Promise]]:
             for x in output_var_sets[1:]:
                 curr = curr.intersection(x)
         promises = [Promise(var=x, val=NodeOutput(sdk_node=n, sdk_type=None, var=x)) for x in curr]
+        # TODO: Is there a way to add the Python interface here? Currently, it's an optional arg.
```
wild-endeavor, line 179: also @kumare3

flytekit/annotated/interface.py
```diff
@@ -18,7 +19,10 @@ class Interface(object):
     """
 
     def __init__(
-        self, inputs: typing.Dict[str, Union[Type, Tuple[Type, Any]]] = None, outputs: typing.Dict[str, Type] = None,
+        self,
+        inputs: typing.Dict[str, Union[Type, Tuple[Type, Any]]] = None,
+        outputs: typing.Dict[str, Type] = None,
+        custom_interface_name: Optional[str] = None,
```
katrogan, line 25: can you add a docstring comment for this var?

flytekit/annotated/base_task.py
```diff
@@ -160,7 +164,14 @@ def __call__(self, *args, **kwargs):
             ctx.execution_state is not None and ctx.execution_state.mode == ExecutionState.Mode.LOCAL_WORKFLOW_EXECUTION
         ):
             if ctx.execution_state.branch_eval_mode == BranchEvalMode.BRANCH_SKIPPED:
-                return
+                if self.python_interface.custom_interface_name:
```
katrogan, line 167: should you check to see if `self.python_interface` is not None first?

flytekit/annotated/promise.py
```diff
@@ -362,23 +362,43 @@ def __str__(self):
...
+        # should still return a tuple but it should be one of ours.
+        if not entity_interface.custom_interface_name:
+            return promises[0]
+
+    # More than one promise, let us wrap it into a tuple
+    # Start with just the var names in the promises
     variables = [p.var for p in promises]
 
-    class Output(collections.namedtuple("TaskOutput", variables)):
+    # These should be OrderedDicts so it should be safe to do this.
```
katrogan, line 392: safe to do what?

flytekit/annotated/base_task.py
```diff
@@ -160,7 +164,14 @@ def __call__(self, *args, **kwargs):
...
+                    variables = [k for k in self.python_interface.outputs.keys()]
+                    output_tuple = collections.namedtuple(self.python_interface.custom_interface_name, variables)
+                    nones = [None for _ in self.python_interface.outputs.keys()]
+                    return output_tuple(*nones)
+                else:
```
EngHabu, line 172: Will it always have a Python Interface at this point? and the question is whether the named tuple should have a name or not?

flytekit/annotated/base_task.py
```diff
@@ -230,7 +241,7 @@ def __init__(
 
     # TODO lets call this interface and the other as flyte_interface?
     @property
-    def python_interface(self):
+    def python_interface(self) -> Interface:
```
EngHabu, line 244: This is supposedly the Override of the parent's python_interface? should the return be consistent?

flytekit/annotated/interface.py
```diff
@@ -258,10 +274,14 @@ def t(a: int, b: str) -> Dict[str, int]: ...
     if hasattr(return_annotation, "__origin__") and return_annotation.__origin__ is tuple:
         # Handle option 3
         logger.debug(f"Task returns unnamed typing.Tuple {return_annotation}")
+        if len(return_annotation.__args__) == 1:
```
EngHabu, line 277: I thought this is the only way people can name outputs... or are we saying they can NOT name outputs if they return a single output?

I guess I'm confused because I know we allow named tuples with a single element... so I'm not sure what this is...

flytekit/annotated/promise.py
EngHabu: why, it's already as concise as anyone would wish

tests/flytekit/unit/annotated/test_serialization.py
EngHabu: This is a standard Python syntax for forcing a tuple construction...

flytekit/annotated/promise.py
wild-endeavor: Why does it have to be generated on the fly?

flytekit/annotated/interface.py
wild-endeavor: no these are not typing.NamedTuples, which we do accept.  These are typing.Tuples, which we don't.

flytekit/annotated/base_task.py
wild-endeavor: I think we'll eventually merge PythonTask and Task, so things should make more sense then, but that might not be the case.  Currently yes, it will always have a python interface at this point because the only Tasks are PythonTasks currently but the code technically allows a non-PythonTask.

flytekit/annotated/base_task.py
wild-endeavor: It should but I'd rather it be accurate than consistent with the base class.

flytekit/annotated/promise.py
wild-endeavor: https://github.com/lyft/flytekit/commit/d4c911a290d4ae7ece3e0985cba598eb68667c4a#diff-b4c0ad757d04f70c1eef4ff55d25fd37a9afc669857ac3c0a4ba72f0abacd602R50 is this broken in some way I'm not seeing?

kumare3: wdym, you are generating the class on the Fly right? For every instance `__init__` you will generate a new class, albeit the same name, but that does not matter. 
I was doing it at the create output time, but I guess this is fine too

flytekit/annotated/condition.py
kumare3: there is no interface at this point, what we will need to do is create an interface dynamically here

kumare3: that anyways is the right way

wild-endeavor: but we're in the `self._last_case` condition here, even here we don't have the interface?

PR #298, opened 2020-12-29
Switching from using model Metadata -> TaskMetadata

kumare3:
TaskMetadata will be maintained as a shadow and allows decoupling of
protocol buffer types from contributor code and user code. This allows
more flexiblity

another prime motivation behind this change is that - it allows making the interface less verbose. This makes it trivial to support default values for metadata.

### DIFF
--- a/flytekit/annotated/base_task.py
+++ b/flytekit/annotated/base_task.py
@@ -43,10 +43,11 @@ class TaskMetadata(object):
     Args:
       cache: Boolean that indicates if caching should be enabled
       cache_version: Version string to be used for the cached value
-      interruptible: Boolean that indicates that this task is of for interruptions and can be scheduled on nodes
+      interruptable: Boolean that indicates that this task can be interrupted and/or scheduled on nodes
                            with lower QoS guarantees. This will directly reduce the `$`/`execution cost` associated,
                             at the cost of performance penalties due to potential interruptions
-      deprecated: This should be used to indicate that the task is deprecated
+      deprecated: A string that can be used to provide a warning message for deprecated task. Absence / empty str
+                       indicates that the task is active and not deprecated
       retries: for retries=n; n > 0, on failures of this task, the task will be retried at-least n number of times.
       timeout: the max amount of time for which one execution of this task should be executed for. If the execution
                      will be terminated if the runtime exceeds the given timeout (approximately)
@@ -54,7 +55,7 @@ class TaskMetadata(object):
 
     cache: bool = False
     cache_version: str = ""
-    interruptible: bool = False
+    interruptable: bool = False
     deprecated: str = ""
     retries: int = 0
     timeout: Optional[Union[datetime.timedelta, int]] = None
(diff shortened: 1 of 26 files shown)


wild-endeavor:
I think I'm not seeing the bigger picture.  At least I don't understand how this is a decoupling?  Ultimately, at serialization time, the Python class _will have to_ be converted into the model class right?  I feel like this is more delaying the coupling rather than decoupling.

Which is fine, but I don't understand what this enables.  +1 if you want.  I think it's okay to use the model sometimes though - like currently we use the dynamic job spec model, literal map and parameter map models, all the literal models, auth, labels, annotations, etc.  I don't think we should create parallel classes for all those.

flytekit/annotated/base_task.py
```diff
@@ -33,6 +35,57 @@ def kwtypes(**kwargs) -> Dict[str, Type]:
...
+    """
+    Create Metadata to be associated with this Task
+
+    Args:
+      cache: Boolean that indicates if caching should be enabled
+      cache_version: Version string to be used for the cached value
+      interruptible: Boolean that indicates that this task is of for interruptions and can be scheduled on nodes
+                           with lower QoS guarantees. This will directly reduce the `$`/`execution cost` associated,
+                            at the cost of performance penalties due to potential interruptions
+      deprecated: This should be used to indicate that the task is deprecated
```
katrogan, line 49: nit: can you provide suggestions for what the value should be? at first glance i would wonder why it's not a bool

flytekit/annotated/base_task.py
```diff
@@ -33,6 +35,57 @@ def kwtypes(**kwargs) -> Dict[str, Type]:
...
 
+@dataclass
+class TaskMetadata(object):
+    """
+    Create Metadata to be associated with this Task
+
+    Args:
+      cache: Boolean that indicates if caching should be enabled
+      cache_version: Version string to be used for the cached value
+      interruptible: Boolean that indicates that this task is of for interruptions and can be scheduled on nodes
```
katrogan, line 46: "is of for interruptions" ?

flytekit/annotated/base_task.py
```diff
@@ -33,6 +35,57 @@ def kwtypes(**kwargs) -> Dict[str, Type]:
...
+
+    cache: bool = False
+    cache_version: str = ""
+    interruptible: bool = False
+    deprecated: str = ""
+    retries: int = 0
+    timeout: Optional[Union[datetime.timedelta, int]] = None
+
+    def __post_init__(self):
+        if self.timeout:
```
katrogan, line 63: can you also add a check here if cache is specified but cache_version isn't (and vice-versa)?

katrogan (approved):

flytekit/annotated/base_task.py
kumare3: its the acutal deprecated message, i kept it as is, but will add a comment

flytekit/annotated/base_task.py
kumare3: good idea

flytekit/annotated/base_task.py
kumare3: this is just copy pasta, but will fix it

flytekit/annotated/base_task.py
```diff
@@ -33,6 +35,60 @@ def kwtypes(**kwargs) -> Dict[str, Type]:
...
+            if isinstance(self.timeout, int):
+                self.timeout = datetime.timedelta(seconds=self.timeout)
+            elif not isinstance(self.timeout, datetime.timedelta):
+                raise ValueError("timeout should be duration represented as either a datetime.timedelta or int seconds")
+        if self.cache and not self.cache_version:
```
katrogan, line 69: what if cache version is set but cache isn't?

kumare3: I think this happens a lot of times today (at lyft)

katrogan (approved):

katrogan:
can you take a look at test failures?

PR #299, opened 2020-12-30
Move flyte types to separate folders

katrogan:


katrogan:
@wild-endeavor @kumare3

flytekit/types/__init__.py
```diff
@@ -1,3 +1,3 @@
-from flytekit.types.flyte_file import FlyteFile
-from flytekit.types.pandas_schema import PandasSchemaReader, PandasSchemaWriter
+from flytekit.types.file import FlyteFile
```
wild-endeavor, line 1: we should clear out this file.

katrogan: we need the pandas_schema imports at least to register the pandas types without necessarily importing the package

wild-endeavor: But if this file isn't empty, if someone does
```
```
All the pandas stuff will be loaded as well.

kumare3: ya, for a user to import file, they should need to `from flytekit.types.file import Flytefile`

flytekit/types/pandas/__init__.py
```diff
@@ -0,0 +1 @@
+from .types import PandasSchemaReader, PandasSchemaWriter
```
kumare3, line 1: this is interesting, so after this FlyteSchema will not work? as it needs the PandasSchemaReader/Writer to be present. I actually think PandasSchemaReader/Writer should be in types.schema

katrogan:
PTAL @wild-endeavor @kumare3

flytekit/types/pandas/__init__.py
katrogan: good idea, done

kumare3 (approved):

PR #301, opened 2020-12-30
MyPy errors for constructor args fixed

kumare3:


flytekit/annotated/interface.py
```diff
@@ -33,7 +35,7 @@ def __init__(
                     self._inputs[k] = v
                 else:
                     self._inputs[k] = (v, None)
-        self._outputs = outputs
+        self._outputs = outputs if outputs else {}
```
wild-endeavor, line 38: this won't break anything right?

flytekit/taskplugins/sagemaker/hpo.py
```diff
@@ -62,21 +61,20 @@ def __init__(
 
         self._task_config = task_config
         self._training_task = training_task
-        iface = training_task.python_interface
```
wild-endeavor, line 65: why the move?  did mypy pick up on this?

wild-endeavor (approved):
approving with a couple questions.

flytekit/annotated/base_sql_task.py
```diff
@@ -21,9 +21,8 @@ def __init__(
         task_type="sql_task",
         inputs: Optional[Dict[str, Type]] = None,
         metadata: Optional[TaskMetadata] = None,
-        task_config: T = None,
+        task_config: Optional[T] = None,
         outputs: Dict[str, Type] = None,
```
katrogan, line 25: this doesn't need to be optional too?

katrogan (approved):

flytekit/annotated/interface.py
kumare3: nope

flytekit/taskplugins/sagemaker/hpo.py
kumare3: no, cleaner to have it next to where we use it

flytekit/annotated/base_sql_task.py
kumare3: mypy complains :(

PR #302, opened 2020-12-30
Support for PythonInstanceTask and better constructor args

kumare3:
Python instance task makes it possible to run an execution of a module
variable (like an airflow operator)
Also reduced the constructor args to be minimal

### DIFF
--- a/flytekit/annotated/python_function_task.py
+++ b/flytekit/annotated/python_function_task.py
@@ -1,5 +1,6 @@
 import inspect
 import re
+from abc import abstractmethod, ABC
 from typing import Any, Callable, Dict, List, Optional, TypeVar
 
 from flytekit.annotated.base_task import PythonTask, TaskMetadata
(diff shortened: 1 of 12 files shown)


flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -119,6 +120,7 @@ def serialize_all(pkgs: List[str], folder: str = None, mode: SerializationMode =
                 o.resource_type, _PROJECT_PLACEHOLDER, _DOMAIN_PLACEHOLDER, name, _VERSION_PLACEHOLDER
             )
             loaded_entities.append(o)
+            ctx.registration_settings.add_instance_var(InstanceVar(module=m, name=k, o=o))
```
katrogan, line 123: interesting, we only add these instance vars at serialization?

kumare3: yes

flytekit/annotated/python_function_task.py
```diff
@@ -50,20 +51,20 @@ def get_registerable_container_image(img: Optional[str], cfg: ImageConfig) -> st
 T = TypeVar("T")
 
 
-class PythonFunctionTask(PythonTask[T]):
+class PythonAutoContainerTask(PythonTask[T], ABC):
     """
-    A Python Function task should be used as the base for all extensions that have a python function.
-    This base has an interesting property, where it will auto configure the image and image version to be
-    used for all the derivatives.
+    A Python Function task should be used as the base for all extensions that want the users code to be in the container
```
wild-endeavor, line 56:
```suggestion
    A Python AutoContainer task should be used as the base for all extensions that want the users code to be in the container
```

wild-endeavor (approved):

PR #304, opened 2020-12-31
Annotations flyte execute real

wild-endeavor:


tests/flytekit/unit/annotated/test_node_creation.py
```diff
@@ -0,0 +1,89 @@
...
+
+    @task
+    def t3():
+        ...
+
+    @workflow
+    def empty_wf():
+        t2_node = create_node(t2)
+        t3_node = create_node(t3)
+        t2_node.depends_on(t3_node)
```
kumare3, line 48: also the `>>` syntax?

kumare3:
looks good, need to give a second pass

tests/flytekit/unit/annotated/test_node_creation.py
wild-endeavor: added

PR #306, opened 2020-12-31
Fixes for fast register

katrogan:


flytekit/annotated/base_task.py
```diff
@@ -412,7 +413,7 @@ def get_registerable_entity(self) -> SdkTask:
 
     def get_fast_registerable_entity(self) -> SdkTask:
         entity = self.get_registerable_entity()
-        if entity.container is None:
+        if entity.container is None or entity.type == SdkTaskType.RAW_CONTAINER_TASK:
```
kumare3, line 416: Why are we using this constant? can we use from Annotated?

katrogan: good point! done

katrogan:
ping @wild-endeavor @kumare3

flytekit/annotated/constants.py
```diff
@@ -0,0 +1 @@
+RAW_CONTAINER_TASK_TYPE = "raw-container"
```
wild-endeavor, line 1: don't we have this already somewhere else?  can we use that one @kumare3

katrogan: it was defined inline in container_task.py

wild-endeavor: it's defined in `flytekit/common/constants.py:25`

wild-endeavor: okay apparently this is just a console thing, ignore my last comment.

kumare3: can we not define a constant again :(, it is part of the task

PR #308, opened 2020-12-31
Annotations - workflow metadata

wild-endeavor:
* Add a WorkflowMetadata dataclass
* Add a WorkflowMetadataDefaults dataclass
* Hook up data classes to constuctors and expose those settings in the workflow decorator
* Remove unnecessary Workflow input to construct_input_promises
* Add a docstring to the workflow decorator.

### DIFF
--- a/flytekit/annotated/workflow.py
+++ b/flytekit/annotated/workflow.py
@@ -53,7 +53,29 @@ class WorkflowMetadata(object):
             raise FlyteValidationException(f"Failure policy {self.on_failure} not acceptable")
 
     def to_flyte_model(self):
-        return _workflow_model.WorkflowMetadata(on_failure=self.on_failure)
+        if self.on_failure == WorkflowMetadata.OnFailurePolicy.FAIL_IMMEDIATELY:
+            on_failure = 0
+        else:
+            on_failure = 1
+        return _workflow_model.WorkflowMetadata(on_failure=on_failure)
+
+
+@dataclass
+class WorkflowMetadataDefaults(object):
+    """
+    This class is similarly named to the one above. Please see the IDL for more information but essentially, this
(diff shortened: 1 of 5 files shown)


flytekit/annotated/workflow.py
```diff
@@ -358,7 +389,11 @@ def get_registerable_entity(self) -> _SdkWorkflow:
         return self._registerable_entity
 
 
-def workflow(_workflow_function=None, *args, reference: Optional[WorkflowReference] = None, **kwargs):
+def workflow(
+    _workflow_function=None,
+    reference: Optional[WorkflowReference] = None,
```
kumare3, line 394: the workflow has both - reference and metadata. I guess we will keep it like task. That was one though I had, should we just expose the `WorkflowMetadata` `TaskMetadata` objects to the user?

flytekit/annotated/workflow.py
```diff
@@ -32,6 +35,49 @@
...
+    def to_flyte_model(self):
+        if self.on_failure == WorkflowMetadata.OnFailurePolicy.FAIL_IMMEDIATELY:
+            on_failure = 0
+        else:
+            on_failure = 1
+        return _workflow_model.WorkflowMetadata(on_failure=on_failure)
+
+
+@dataclass
+class WorkflowMetadataDefaults(object):
```
kumare3, line 64: can we call this something else like `WorkflowDefaults` much better than WorkflowMetadataDefaults
And should this be part of WorkflowMetadata?

wild-endeavor: I'm okay with this as it stands but up to you.  They are two separate objects in the IDL so I just kept that pattern, it makes the to_flyte_model function easier to reason about.  Also, these aren't exposed, only the settings themselves are in the decorator.

PR #313, opened 2021-01-06
Out of container serialize

katrogan:


flytekit/tools/module_loader.py
```diff
@@ -118,16 +158,10 @@ def iterate_registerable_entities_in_order(
...
-        for k in dir(m):
-            o = m.__dict__[k]
-            if isinstance(o, _registerable.RegisterableEntity) and not o.has_registered:
-                if o.instantiated_in == m.__name__:
-                    entity_to_module_key[o] = (m, k)
-                    if isinstance(o, _SdkRunnableWorkflow) and o.should_create_default_launch_plan:
-                        # SDK should create a default launch plan for a workflow.  This is a special-case to simplify
-                        # authoring of workflows.
-                        entity_to_module_key[o.create_launch_plan()] = (m, k)
+    if pkgs is not None:
```
wild-endeavor, line 161: Should we use pkgs within directory or somehow try to respect this option?  I'm not sure - I've always found this option a bit weird, but I think people do use it.

Running outside of the container, without this, serialize will just run through everything underneath the given directory right?

wild-endeavor:
Posted one question.  Otherwise I think this looks good, but given the scope of this change can we run serialize on cookbook in the container, and then using this out of the container and shasum all the files and compare what differences there are, if any?

In particular I'm worried about things like the arguments to `pyflyte-execute`.  Would like to make sure they don't change.

flytekit/tools/module_loader.py
katrogan: yeah I waffled on that too. I'm not sure either but I really can't fathom the need where someone only wants to upload specific packages in a directory and exclude others

katrogan:
> I'm worried about things like the arguments to pyflyte-execute.

yeah I was cautious about this case too. essentially the user-specified directory becomes the root directory so that task-module remains the same as well as the name in the identifier

flytekit/tools/module_loader.py
wild-endeavor: And yet people do, they use it all the time for some reason, even Spotify asked about it.  At the very least, let's read in the value of the configuration and write out a very stern WARNING about what's happening.

wild-endeavor:
Let's wait til Monday to merge this?

katrogan:
@wild-endeavor sure thing

flytekit/tools/module_loader.py
kumare3: ohh absolutely, when we are trying to iterate limiting to only a few packages is very useful - no? - the ones that I am specifically workin on as a user

katrogan: sure I guess my question is how often can that not be solved by just using a more specific directory rather than multiple packages? let me add the pkgs option though since the consensus seems to be it's useful to hace

katrogan:
@wild-endeavor @kumare3 mind reviewing?

wild-endeavor:
Couple more things.

So i ran this on cookbook twice, once in the container and once outside, and compared the output.  I think everything now matched (they didn't match exactly but I'm pretty sure because I didn't update the in-container flytekit to use this branch).

Is it possible to make `--directory` something more descriptive like `--local-source-root` or `--local-code-dir` or something?  Also is it possible to make it optional and have it default to the current directory?

Can we make the `--in-container-config-path` help more explicit?  Maybe something like
```
This is where the configuration for your task lives inside the container. The reason it needs to be a separate option is because this pyflyte utility cannot know where the Dockerfile writes the config file to.
```

PR #320, opened 2021-01-08
All python function tasks can now be dynamic

katrogan:


katrogan:
@kumare3 @wild-endeavor mind reviewing? I'd like to leave dynamic tasks as a separate mixin because of the way they overwrite the execute method

wild-endeavor (approved):

katrogan:
ping @kumare3

kumare3:
@katrogan after thinking a little more, I feel we can covert the dynamic logic to be a boolean switch in the base task. this way, truly all tasks can by dynamic without extension. Thus if you use the `@dynamic` decorator - then, the boolean switch is automatically turned on and hence, a dynamic futures file is always expcted? WDYT?

katrogan:
@kumare3 I like it! i will update this PR

katrogan:
PTAL @kumare3 @wild-endeavor not really sure if it's worth keeping the dynamic code as a mixing separate from PythonFunctionTask but the latter is quite clean/simple as is

tested on my local flyte sandbox with dynamic sidecar and vanilla dynamic tasks

wild-endeavor (approved):

wild-endeavor:
we can always decide in the future about the mixin status.

tests/flytekit/unit/taskplugins/pod/test_pod.py
```diff
@@ -62,3 +66,45 @@ def simple_pod_task(i: int):
...
+        registration_settings=context_manager.RegistrationSettings(
+            project="test_proj",
+            domain="test_domain",
+            version="abc",
+            image_config=ImageConfig(Image(name="name", fqn="image", tag="name")),
+            env={},
+        )
+    ) as ctx:
+        with ctx.new_execution_context(mode=ExecutionState.Mode.TASK_EXECUTION) as ctx:
+            dynamic_job_spec = dynamic_pod_task.compile_into_workflow(ctx, dynamic_pod_task._task_function, a=5)
```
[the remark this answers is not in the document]
katrogan: wrt the dynamic mixin? i used the existing dynamic unit test as a model which only verifies this

flytekit/annotated/python_function_task.py
```diff
@@ -167,9 +174,17 @@ def __init__(
...
+
+    @property
+    def execution_mode(self) -> ExecutionMode:
+        return self._execution_mode
 
     def execute(self, **kwargs) -> Any:
-        return self._task_function(**kwargs)
+        if self.execution_mode == self.ExecutionMode.DEFAULT:
+            return self._task_function(**kwargs)
+        elif self.execution_mode == self.ExecutionMode.DYNAMIC:
```
kumare3, line 186: is this a new execution mode?

flytekit/annotated/dynamic_mixin.py
```diff
@@ -0,0 +1,85 @@
+from typing import Any, Callable, Union
+
+from flytekit.annotated.context_manager import ExecutionState, FlyteContext
+from flytekit.annotated.workflow import Workflow, WorkflowFailurePolicy, WorkflowMetadata, WorkflowMetadataDefaults
+from flytekit.loggers import logger
+from flytekit.models import dynamic_job as _dynamic_job
+from flytekit.models import literals as _literal_models
+
+
+class DynamicWorkflowTaskMixin(object):
```
kumare3, line 10: why keep this separate? Whats the use of a mixin? We will never use it elsewhere right?

kumare3: if you want you can just make these has helper methods?

flytekit/annotated/python_function_task.py
```diff
@@ -167,9 +174,17 @@ def __init__(
             **kwargs,
         )
         self._task_function = task_function
+        self._execution_mode = execution_mode if execution_mode is not None else self.ExecutionMode.DEFAULT
```
kumare3, line 177: this has been bugging me to. Can we just start with an ExecutionMode always set to something like DEFAULT?

flytekit/annotated/dynamic_mixin.py
katrogan: nope, see my comment above I wasn't sure whether it was cleaner to leave PythonFunctionTask as is given it's so simple. I'll fold this in.

flytekit/annotated/python_function_task.py
katrogan: ohh yes. should i call it something else?

flytekit/annotated/python_function_task.py
katrogan: done

katrogan:
ping @kumare3

flytekit/annotated/python_function_task.py
```diff
@@ -186,6 +204,82 @@ def get_command(self, settings: RegistrationSettings) -> List[str]:
...
+    ) -> Union[_dynamic_job.DynamicJobSpec, _literal_models.LiteralMap]:
+        with ctx.new_compilation_context(prefix="dynamic"):
+            workflow_metadata = WorkflowMetadata(on_failure=WorkflowFailurePolicy.FAIL_IMMEDIATELY)
+            defaults = WorkflowMetadataDefaults(interruptible=False)
+
+            self._wf = Workflow(task_function, metadata=workflow_metadata, default_metadata=defaults)
+            self._wf.compile(**kwargs)
+
+            wf = self._wf
+            sdk_workflow = wf.get_registerable_entity()
```
kumare3, line 218: this might impact what @wild-endeavor is doing

flytekit/annotated/python_function_task.py
```diff
@@ -167,9 +177,17 @@ def __init__(
...
+    @property
+    def execution_mode(self) -> ExecutionBehavior:
+        return self._execution_mode
 
     def execute(self, **kwargs) -> Any:
-        return self._task_function(**kwargs)
+        if self.execution_mode == self.ExecutionBehavior.DEFAULT:
+            return self._task_function(**kwargs)
+        elif self.execution_mode == self.ExecutionBehavior.DYNAMIC:
+            return self.dynamic_execute(self._task_function, **kwargs)
```
kumare3, line 190: ohh this is interesting. This will break for cases in which execute is overriden right? Which happens quite a few times? Are we sure this is the best place to do it?

kumare3: so the contract is that `execute` is overridable. but dispatch_execute should not be overriden

kumare3: I guess now that I have added pre-execute and post-execute, we can document that `execute` should not be overriden in most cases unless absolutely required and in that case, unless you replicate the dynamic execute call, ability to dynamically execute will be lost! I am ok with that

katrogan: > so the contract is that execute is overridable. but dispatch_execute should not be overriden

yes precisely. added documentation that `execute` can be overwritten if dynamic execute is handled too

flytekit/annotated/python_function_task.py
katrogan: I will wait and rebase after his change is in

wild-endeavor (approved):

flytekit/annotated/python_function_task.py
katrogan: talked offline, i'll merge this as is

PR #321, opened 2021-01-11
Adding default File formats to FlyteFile type

kumare3:


scripts/flytekit_install_spark3.sh
```diff
@@ -42,13 +43,14 @@ rm -rf spark-dist.tgz
 rm -rf spark-dist
 
 # Fetch Hadoop Distribution with AWS Support
-wget http://apache.mirrors.tds.net/hadoop/common/hadoop-2.7.7/hadoop-2.7.7.tar.gz -O hadoop-dist.tgz
-echo 'd129d08a2c9dafec32855a376cbd2ab90c6a42790898cabbac6be4d29f9c2026  hadoop-dist.tgz' | sha256sum --check
+#wget http://apache.mirrors.tds.net/hadoop/common/hadoop-2.7.7/hadoop-2.7.7.tar.gz -O hadoop-dist.tgz
```
wild-endeavor, line 46: remove comments?

flytekit/types/file/__init__.py
```diff
@@ -1 +1,54 @@
+import typing
+
 from .file import FlyteFile
+
+# The following section provides some predefined aliases for commonly used FlyteFile formats.
+# This makes their usage extremely simple for the users. Please keep the list sorted.
+
+
+HDF5EncodedFile = FlyteFile[typing.TypeVar("hdf5")]
```
wild-endeavor, line 9: I still don't like this usage of TypeVar, it feels wrong, but we can always change it in the future without breaking people.

wild-endeavor (approved):

scripts/flytekit_install_spark3.sh
kumare3: ohh this file should not be part of the update, removing

flytekit/types/file/__init__.py
kumare3: ya, it is required today for linter

wild-endeavor (approved):

PR #323, opened 2021-01-12
Annotations extra require - sagemaker

wild-endeavor:


katrogan (approved):

setup.py
```diff
@@ -8,13 +8,15 @@
     "sidecar": ["k8s-proto>=0.0.3,<1.0.0"],
     "schema": ["numpy>=1.14.0,<2.0.0", "pandas>=0.22.0,<2.0.0", "pyarrow>=0.11.0,<1.0.0"],
     "hive_sensor": ["hmsclient>=0.0.1,<1.0.0"],
+    "sagemaker": ["sagemaker-training>=3.6.2,<4.0.0"],
     "all-spark2.4": [
```
EngHabu, line 12: This is very annoying to have... imagine then having `all-spark2.4-pyarrow1.0` ...

EngHabu (approved):

setup.py
kumare3: I think this is done wrongly

wild-endeavor: @akhurana001 when did we say we were gonna deprecate 2.x support entirely?

akhurana001: Have https://github.com/lyft/flyte/issues/483 for the same. I can make a PR if we are ok doing that ?

PR #324, opened 2021-01-12
Create ReferenceTask from task ID

honnix:
# TL;DR
Make it possible to create `ReferenceTask` from task ID directly.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
In case that the task is fetched remotely first, it is simpler just to pass in the task id.

Not super sure how much sense this makes though.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/tests/flytekit/unit/annotated/test_references.py
+++ b/tests/flytekit/unit/annotated/test_references.py
@@ -64,10 +64,13 @@ def test_ref():
 
 def test_ref_task_more():
     @reference_task(
-        project="flytesnacks",
-        domain="development",
-        name="recipes.aaa.simple.join_strings",
-        version="553018f39e519bdb2597b652639c30ce16b99c79",
+        id=_identifier_model.Identifier(
+            _identifier_model.ResourceType.TASK,
+            "flytesnacks",
+            "development",
+            "recipes.aaa.simple.join_strings",
+            "553018f39e519bdb2597b652639c30ce16b99c79",
+        )
     )
     def ref_t1(a: typing.List[str]) -> str:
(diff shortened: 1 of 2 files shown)


honnix:
@wild-endeavor What do you think of this? I myself is not so convinced by mixing id and it's itemized content together, so please feel free to reject and close. I can work on another PR to simplify task_function handling because there is really no point using `reference_task` without argument.

honnix:
The said another PR is https://github.com/lyft/flytekit/pull/330

tests/flytekit/unit/annotated/test_references.py
```diff
@@ -64,10 +64,13 @@ def ref_t1(a: typing.List[str]) -> str:
 
 def test_ref_task_more():
     @reference_task(
```
wild-endeavor, line 66: if someone forgets to specify the ID and just does
```
@reference
def r1():
    ...
```
do we get an error?

wild-endeavor (approved):
+1 but not sure if we need to raise an error

honnix:
I think like #330 better to make things positional. We can come back to this if really needed later. Mixing id and others on the same level hardly convince myself. ;)

tests/flytekit/unit/annotated/test_references.py
honnix: No there won't be an error if they all have default value.

PR #326, opened 2021-01-13
Annotations separate out serializing

wild-endeavor:
* Create a new `NodeOutput` class in the `promise` file not dependent on the old one.  I left an unused `sdk_type` in there for now cuz that's always None but I'll change it.
* Move `create_and_link_node` to the `promise` file.
* Move out all the `get_registerable_entity` logic from everything and put all the logic in one file.
  * Special casing had to be done for Reference entities, raw container tasks, and container-less tasks.
* Changed a lot of tests to not use the `RegistrationSettings` object inside a context object because the new function just expects it explicitly. However I didn't see an easy way to remove it from the `FlyteContext` entirely because of dynamic tasks, which run a compilation step inside of execution and registration settings are not usually passed through during execution.


flytekit/common/translator.py
```diff
@@ -0,0 +1,260 @@
...
+        ret_cases.append(to_registerable_case(settings, c))
+    return ret_cases
+
+
+# Can make this part of the function in the future.
+GLOBAL_CACHE: Dict[FlyteLocalEntity, FlyteControlPlaneEntity] = {}
+
+
+def get_serializable(
+    settings: RegistrationSettings, entity: FlyteLocalEntity, fast: Optional[bool] = False
```
katrogan, line 56: what are your thoughts on renaming RegistrationSettings to SerializationSettings since that's now the primary use-case?

flytekit/common/translator.py
```diff
@@ -0,0 +1,260 @@
...
+                upstream_nodes=upstream_sdk_nodes,
+                bindings=entity._bindings,
+                metadata=entity._metadata,
+                sdk_launch_plan=get_serializable(settings, entity._flyte_entity),
+            )
+        else:
+            raise Exception(f"Node contained non-serializable entity {entity._flyte_entity}")
+
+    elif isinstance(entity, LaunchPlan):
+        if entity._auth_role:
```
katrogan, line 215: nit: do you really need this check? don't you pass it along as is None or otherwise?

flytekit/common/translator.py
```diff
@@ -0,0 +1,260 @@
...
+        # Reset just to make sure it's what we give it
+        cp_entity.id._project = settings.project
+        cp_entity.id._domain = settings.domain
+        cp_entity.id._name = entity.name
+        cp_entity.id._version = settings.version
+
+        if fast:
+            # Containerless tasks are always fast registerable without modification so only operate on tasks that
+            # have a container. Also, raw ContainerTask's should never be touched.
+            if cp_entity.container is not None and not isinstance(entity, ContainerTask):
```
katrogan, line 130: hm it makes me sad we lose the inheritence bit. i wonder if it's worth refactoring `get_container` to accept a bool for whether the context should be regular or fast

katrogan (approved):

flytekit/annotated/condition.py
```diff
@@ -175,7 +141,7 @@ def _compute_outputs(self, n: Node) -> Union[Promise, Tuple[Promise], VoidPromis
         if len(output_var_sets) > 1:
             for x in output_var_sets[1:]:
                 curr = curr.intersection(x)
-        promises = [Promise(var=x, val=NodeOutput(sdk_node=n, sdk_type=None, var=x)) for x in curr]
+        promises = [Promise(var=x, val=NodeOutput(node=n, sdk_type=None, var=x)) for x in curr]
```
kumare3, line 144: can we just default `sdk_type` to `None` in the signature?

flytekit/annotated/promise.py
```diff
@@ -550,3 +558,128 @@ def __str__(self):
 
     def __repr__(self):
         raise AssertionError(f"Task {self._task_name} returns nothing, NoneType return cannot be used")
+
+
+class NodeOutput(type_models.OutputReference):
+    def __init__(self, node: Node, sdk_type, var: str):
```
kumare3, line 564: let us make `sdk_type` `Optional` and default to None?

flytekit/annotated/promise.py
```diff
@@ -550,3 +558,128 @@ def __str__(self):
...
+        """
+        return self._type
+
+    def __repr__(self) -> str:
+        s = f"Node({self.node if self.node.id is not None else None}:{self.var})"
+        return s
+
+
+# TODO we should accept TaskMetadata here and then extract whatever fields we want into NodeMetadata
+def create_and_link_node(
```
kumare3, line 600: maybe the name of this function needs to be renamed to reflect that we are Substituting outputs with promises

flytekit/common/translator.py
kumare3: +1

flytekit/common/translator.py
```diff
@@ -0,0 +1,260 @@
...
+
+def get_serializable(
+    settings: RegistrationSettings, entity: FlyteLocalEntity, fast: Optional[bool] = False
+) -> FlyteControlPlaneEntity:
+    if entity in GLOBAL_CACHE:
+        return GLOBAL_CACHE[entity]
+
+    cp_entity = None
+    # TODO: flytekit.annotated.container_task.ContainerTask needs to be excepted
+    if isinstance(entity, ReferenceTask):
```
kumare3, line 63: my preference each if condition is modeled using a hash-map and we just have functions for each transform. Thus we follow a purely functional approach

kumare3: e.g
map[Type, Callable]?

flytekit/common/translator.py
kumare3: interesting, also, maybe we should do this for any entity that is instance of PythonAutoContainerTask and not ContainerTask. We should assume all container tasks are independent of code?

flytekit/common/translator.py
kumare3: even if we dont do the decomposition by map - i would like us to break this function into smaller functions for ease of testing and easier logic to follow

flytekit/common/translator.py
kumare3: +1, this is a registration time construct anyways

flytekit/common/translator.py
kumare3: as a separate PR is ok

flytekit/common/translator.py
```diff
@@ -0,0 +1,260 @@
...
+            custom={},
+            container=None,
+        )
+        # Reset id to ensure it matches user input
+        cp_entity._id = entity.id
+        cp_entity._has_registered = True
+        cp_entity.assign_name(entity.reference.id.name)
+
+    elif isinstance(entity, ReferenceWorkflow):
+        workflow_metadata = WorkflowMetadata(on_failure=WorkflowFailurePolicy.FAIL_IMMEDIATELY)
```
kumare3, line 77: Why?
Why does a reference workflow not look similar, maybe i am missing something

kumare3:
This looks aaamazing! awesome work

flytekit/annotated/promise.py
wild-endeavor: I'm just gonna remove it entirely.

flytekit/annotated/promise.py
wild-endeavor: suggestions?

flytekit/common/translator.py
wild-endeavor: what do you mean?  similar to what?

flytekit/common/translator.py
katrogan: +1 for separate functions!

flytekit/common/translator.py
```diff
@@ -0,0 +1,301 @@
...
+    # Reset just to make sure it's what we give it
+    cp_entity.id._project = settings.project
+    cp_entity.id._domain = settings.domain
+    cp_entity.id._name = entity.name
+    cp_entity.id._version = settings.version
+
+    if fast:
+        # Containerless tasks are always fast registerable without modification so only operate on tasks that
+        # have a container. Also, raw ContainerTask's should never be touched.
+        if cp_entity.container is not None and not isinstance(entity, ContainerTask):
```
katrogan, line 133: thoughts on https://github.com/lyft/flytekit/pull/326#discussion_r557026522?

flytekit/common/translator.py
```diff
@@ -0,0 +1,301 @@
...
+            nodes=[],  # Fake an empty list for nodes,
+            id=entity.reference.id,
+            metadata=workflow_metadata,
+            metadata_defaults=workflow_model.WorkflowMetadataDefaults(),
+            interface=entity.typed_interface,
+            output_bindings=[],
+        )
+
+    elif isinstance(entity, ReferenceLaunchPlan):
+        wf_id = _identifier_model.Identifier(_identifier_model.ResourceType.WORKFLOW, "", "", "", "")
```
katrogan, line 89: not related to this PR but why do you use an empty identiifer rather than filling out the fields?

katrogan (approved):
thanks for refactoring get_serializable to call smaller methods!

flytekit/common/translator.py
wild-endeavor: Can we do this in a later PR?  As indicated by Miguel's recent question, I want to see if there are additional use cases for changing/overriding the command of a container.  Even if not, I'd rather clean up in a separate PR since it's not user facing.

flytekit/common/translator.py
wild-endeavor: No intelligent reason

wild-endeavor: this will all get deleted i hope

katrogan (approved):

PR #327, opened 2021-01-13
In Container - Sqlite3 Plugin

kumare3:
It is possible to declare sqlite3 statements that can read from a dynamically downloaded sqlite3 database or one that is stored in the container.

TODO:
1. Prevent tasks being written inside the workflow function `<local>`
2. Make it possible that Sqlite3 tasks can be executed on a predefined container - dynamically! No need to build a container. This is not possible today

### DIFF
--- a/flytekit/taskplugins/sqlite3/task.py
+++ b/flytekit/taskplugins/sqlite3/task.py
@@ -5,10 +5,10 @@ import sqlite3
 import tempfile
 import typing
 from dataclasses import dataclass
-from enum import Enum
 
-from flytekit import FlyteContext, TaskMetadata, kwtypes
+from flytekit import FlyteContext, kwtypes
 from flytekit.annotated.base_sql_task import SQLTask
+from flytekit.annotated.python_function_task import PythonInstanceTask
 from flytekit.plugins import pandas as pd
 from flytekit.types.file import FlyteFile
 from flytekit.types.schema import FlyteSchema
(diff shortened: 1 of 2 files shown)


flytekit/taskplugins/sqlite3/task.py
```diff
@@ -0,0 +1,98 @@
...
+class SQLite3Config(object):
+    """
+    Use this configuration to configure if sqlite3 files that should be loaded by the task. The file itself is
+    considered as a database and hence is treated like a configuration
+    The path to a static sqlite3 compatible database file can be
+      - within the container
+      - or from a publicly downloadable source
+
+    Args:
+        file_path: default FlyteFile that will be downloaded on execute
```
katrogan, line 40:
```suggestion
        uri: default FlyteFile that will be downloaded on execute
```

flytekit/taskplugins/sqlite3/task.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @property
+    def output_columns(self) -> typing.Optional[typing.List[str]]:
+        c = self.python_interface.outputs["results"].column_names()
+        return c if c else None
+
+    def execute(self, **kwargs) -> typing.Any:
+        with tempfile.TemporaryDirectory() as temp_dir:
+            ctx = FlyteContext.current_context()
+            file_ext = os.path.basename(self.task_config.uri)
```
katrogan, line 89: nit: basename doesn't just give you just the file extension?

wild-endeavor (approved):

katrogan (approved):

flytekit/taskplugins/sqlite3/task.py
kumare3: i guess you mean, i should rename it to file_with_ext

PR #330, opened 2021-01-14
Make reference_workflow a decorator

honnix:
# TL;DR
Having a dedicated decorator to create reference workflow.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
* Make `@reference_workflow` a decorator
* Move content of `reference_task.py` to `task.py` (to align with `workflow.py`)

## Tracking Issue
https://github.com/lyft/flyte/issues/668

## Follow-up issue
_NA_


flytekit/annotated/task.py
```diff
@@ -134,3 +138,49 @@ def wrapper(fn) -> PythonFunctionTask:
         return wrapper(_task_function)
     else:
         return wrapper
+
+
+class ReferenceTask(ReferenceEntity, PythonFunctionTask):
```
wild-endeavor, line 143: I remember there was a circular import that I couldn't resolve when I tried moving the ReferenceTask into the task file itself.  Did you not run into this??

flytekit/annotated/workflow.py
```diff
@@ -375,6 +375,35 @@ def get_registerable_entity(self) -> _SdkWorkflow:
         return self._registerable_entity
 
 
+def workflow(
+    _workflow_function=None,
+    failure_policy: Optional[WorkflowFailurePolicy] = None,
+    interruptible: Optional[bool] = False,
+):
+    """
+    :param reference: Pass a WorkflowReference object here if you want to make this a pointer to an existing workflow
```
wild-endeavor, line 384: nix

flytekit/annotated/workflow.py
```diff
@@ -422,35 +438,17 @@ def get_registerable_entity(self) -> _SdkWorkflow:
...
-        return workflow_instance
+    def wrapper(fn) -> ReferenceWorkflow:
+        interface = transform_signature_to_interface(inspect.signature(fn))
+        return ReferenceWorkflow(project, domain, name, version, interface.inputs, interface.outputs)
 
-    if _workflow_function:
-        return wrapper(_workflow_function)
-    else:
-        return wrapper
+    return wrapper
```
wild-endeavor, line 454: same question about whether we need to raise an error if the user forgets to fill in the `(...)`

flytekit/annotated/task.py
honnix: No I didn't have a circular import issue.

flytekit/annotated/workflow.py
honnix: `project: str, domain: str, name: str, version: str,` are all mandatory now without default value, so there should be a hard error if only `@reference_workflow()` is used.

honnix: `@reference_workflow` only will also result error like `missing required positional arguments`

flytekit/annotated/task.py
wild-endeavor: ooh i had previously tried to put it into python_function_task.

wild-endeavor (approved):

flytekit/annotated/workflow.py
wild-endeavor: good

PR #338, opened 2021-01-16
Trigger single task or reference entity execution

katrogan:


flytekit/remote/execution.py
```diff
@@ -0,0 +1,105 @@
...
+    auth_role: AuthRole = None
+    # optional name to assign to execution
+    name: str = None
+    # The name of the user launching this execution.
+    principal: str = None
+
+
+def execute(
+    entity: Union[ReferenceEntity, ContainerTask],
+    client: SynchronousFlyteClient,
```
kumare3, line 35: can this be optional?

flytekit/remote/execution.py
```diff
@@ -0,0 +1,105 @@
...
+            try:
+                client.create_task(task_identifer=sdk_task.id, task_spec=TaskSpec(sdk_task))
+            except user_exceptions.FlyteEntityAlreadyExistsException:
+                pass
+    else:
+        reference_identifier = entity.id
+
+    with flyte_context.FlyteContext.current_context():
+        literals = translate_inputs_to_literals(
+            ctx, input_kwargs=inputs, interface=entity.interface, native_input_types=entity.get_input_types()
```
kumare3, line 61: the referenceEntity may not have an interface right? so we should fetch that first?

katrogan: oh I did not know this! good call, yes

flytekit/remote/execution.py
katrogan: sure, we can always initialize using the platform url in the config

flytekit/remote/execution.py
katrogan: actually it looks like inputs and outputs are "required" input arguments for reference entities but i will fetch if they are none

katrogan:
PTAL @kumare3

flytekit/remote/execution.py
```diff
@@ -0,0 +1,128 @@
...
+    name: str = None
+    # The name of the user launching this execution.
+    principal: str = None
+
+
+def execute(
+    entity: Union[Identifier, ContainerTask],
+    project: str,
+    domain: str,
+    inputs: Dict[str, Any] = None,
```
kumare3, line 39: should we inputs as a dictionary or just kwargs? Whats your gut feeling?

katrogan: kwargs gives us consistency with just calling the function in local execute mode but b/c there are required/positional args in this scenario naming inputs isn't what departs from that pattern. what do you think? happy to change or keep as is

PR #339, opened 2021-01-17
Annotations - separate out plugins into top level folder

wild-endeavor:
This is the last remaining item we hope in the beta release, the flytekit plugin system.  This is because this will affect how users install and import plugin tasks (like Spark or Pytorch, or plugin types like Pandera, etc.)

This PR attempts one possible solution, using namespace packages.  For those unfamiliar, please take a look at this blog post first https://medium.com/@jherreras/python-microlibs-5be9461ad979 as this PR is heavily based off of that.  Also see [PEP 420](https://www.python.org/dev/peps/pep-0420) for more info.  Also see the existing [vaex](https://github.com/vaexio/vaex/blob/master/setup.py) setup which is where I found the blog posting.

We rethought the plugin code structure with the following in mind.
* When we remove the alpha/beta qualifiers on the new API, we'll likely release it as 0.16.x or 0.17, it will not be 1.0.  That is, we cannot break existing users (old plugin code imports should continue to work).
* For example, we'd like `pip install flytekit[spark]` to continue to install the existing master-branch `spark` functionality.  But we don't have to install the _new_ spark stuff, based on the `annotations` branch.
* For that, users should install something like `flytekitplugins-spark`.
* Flytekit plugin code that is more split out, perhaps even in another repo, is cleaner because plugin systems should be more delineated and distinct.  This separation ensures that.

This PR
* Moves all the new `annotations` branch `taskplugins` code into separate `flytekitplugins` namespace package folders, under the new top-level folder (outside of the highest level `flytekit` folder).  For all intents and purposes, you can think of this folder as a separate repo.
* Moves all the tasks for each plugin type into a separate tests folder in each plugin.
* Adds a `setup.py` file for each plugin
* Adds a not quite working local-development only (read: not a real Pypi package) `setup.py` for the top level `flytekitplugins` folder.
* Changes the plugin pandas imports to not be from the flytekit lazy loader, but pandas for reals.
* Added a post_init for the `Spark` task config dataclass to instantiate empty Spark and Hadoop configs if missing.

Upsides of doing things this way.
* Enforces separation.
* Promotes ease of new contributions.

Downsides of doing things this way.
* Since separation is enforced, we need a lot more stuff.  Each one will have its own `setup.py`, requirements file, etc.  If you want to bump the minimum `flytekit` version (which now is an explicit dependency), you have to go in to each plugin's setup and bump the version.
* Local development no longer works without installing.  When writing flytekit, often i will `export PYTHONPATH=:$PYTHONPATH` to basically add the current directory to the list of places where Python searches for code. This is no longer possible for the plugins.  This is because the location of the pytorch plugin is now `plugins/pytorch/flytekitplugins/pytorch/blah.py`. When using something defined there, if you just write `from flytekitplugins.pytorch.blah import MyBlah` it won't find it.  It'd only work if you write `from plugins.pytorch.flytekitplugins.pytorch.blah import MyBlah` but of course you can't do that, because that won't work for after it's been deployed.
* This means you have to `pip install -e .` each plugin.  The `-e` means development mode install, and it basically adds an egg link from your `site-packages` to the directory where you ran that command. This is fine, just more steps.
* NB Some weirdness around tests.  Note that the tests for the papermill plugin is called `papermilltests` under the `plugins/tests` folder.  The reason for this is because the tests were failing when the folder was just called `plugins/tests/papermill`.  The reason it failed is because when pytest runs, the `papermill` package that is found ends up being the `papermill` folder in the test folder itself.  When in the real code we write `import papermill as pm` we end up getting this test package instead of the real papermill package.

Additional notes.
* Unclear if we should keep all the plugins as one repo, or have each plugin be its own github repo.  I'm inclined to the former, as is the author of the blog (search for "keep each library in a single repository").
* Even though pandas is large I'd like to propose that we bundle it with flytekit.  In other words, I think the `Schema` type should be included when you do `pip install flytekit`.  I don't think we should make users `pip install flytekitplugins-schema` -- it's just too central.


plugins/hive/setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import setup
+
+PLUGIN_NAME = "hive"
+
+microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
+
+plugin_requires = ["flytekit==0.16.0a2", "hmsclient>=0.0.1,<1.0.0"]
```
kumare3, line 7: this should not be pinned right?

plugins/notebook/setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import setup
+
+PLUGIN_NAME = "notebook"
```
kumare3, line 3: lets call this papermill?

plugins/pytorch/setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import setup
+
+PLUGIN_NAME = "pytorch"
```
kumare3, line 3: lets call this `kfpytorch`?

plugins/sagemaker/setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import setup
+
+PLUGIN_NAME = "sagemaker"
```
kumare3, line 3: what about putting all aws related plugins in one place? this the plugin is `aws` and we can have `aws.sagemaker` as the first thing in there?

plugins/setup.py
```diff
@@ -0,0 +1,72 @@
...
+        install_all_plugins(SOURCES, develop=False)
+        install.run(self)
+
+
+setup(
+    name=PACKAGE_NAME,
+    version="0.1.0",
+    author="flyteorg",
+    author_email="admin@flyte.org",
+    description="Macrolib's description",
```
kumare3, line 62: update the description?

plugins/setup.py
```diff
@@ -0,0 +1,72 @@
...
+    license="TBD",
```
kumare3, line 63: apache2

plugins/sqlite3/setup.py
```diff
@@ -0,0 +1,33 @@
+from setuptools import setup
+
+PLUGIN_NAME = "sqlite3"
```
kumare3, line 3: I think we should ship with this one? as sqlite3 needs nothing as we will ship with pandas, unless you think its better to keep it separate?

plugins/tensorflow/setup.py
```diff
@@ -0,0 +1,33 @@
+from setuptools import setup
+
+PLUGIN_NAME = "tensorflow"
```
kumare3, line 3: `kftensorflow`

kumare3:
@wild-endeavor - WOW! Thank you for doing this. A few comments.
Also we have decided to flytekit as is right - as-in we wont split flytekit into legacy and new right?

honnix:
Some thoughts around downsides:

> Since separation is enforced, we need a lot more stuff...
We can add a boilerplate script to bootstrap a new plugin, and another script to bump versions.

> Local development no longer works without installing...
I guess as long as we well document `pip install -e .`, it should be totally fine. Actually I have been only using this way, as it is more or less a standard/common practice.

honnix:
Regarding repositories, I think keeping all plugins in one repo is sufficient, given most cases the plugins being relatively small. Having too many repos may require a lot more maintenance, e.g. versioning, CI.

I think it is also OK to postpone repo split at this moment, because that still seems like a luxury problem that we don't have to deal with immediately. _Consider this as my Luigi maintenance experience, but of course it has very different setup._ :)

plugins/hive/tests/test_hive_task.py
```diff
@@ -1,13 +1,13 @@
 import pandas
 import pytest
+from plugins.hive.flytekitplugins.hive.task import HiveConfig, HiveSelectTask, HiveTask
```
honnix, line 3: I can understand this absolute import but still a bit strange.

plugins/setup.py
```diff
@@ -0,0 +1,72 @@
...
+    classifiers=[
+        'Private :: Do Not Upload to pypi server',
+    ],
+    install_requires=[],
+    cmdclass={
+        'install': InstallCmd,
+        'develop': DevelopCmd,
```
honnix, line 70: I tried running `pip install -e .` under `plugins` but that didn't seem to install any package. Am I supposed to run `python setup.py develop`?

plugins/pod/setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import setup
+
+PLUGIN_NAME = "notebook"
```
honnix, line 3:
```suggestion
PLUGIN_NAME = "pod"
```

plugins/setup.py
honnix: `python setup.py develop` doesn't work for me either. It logs everything being OK, but nothing gets installed, apart from `flytekitplugins-parent`.

honnix: I guess we need this step? https://github.com/vaexio/vaex/blob/master/setup.py#L33

plugins/hive/tests/test_hive_task.py
wild-endeavor: oh no this is wrong, i'll fix.

plugins/sqlite3/setup.py
wild-endeavor: sure - what folder should we put it in?

plugins/setup.py
honnix: NVM I recreated my virtualenv and it is now working.

wild-endeavor:
So @honnix it sounds like you don't think this PR is a horrible idea?  I'll work on cleaning it up and addressing all the other comments today if that's the case.

plugins/sagemaker/setup.py
wild-endeavor: Yeah sure, but I think this will get pretty big.  In the future we should do `aws[sagemaker]`?  but for now install everything.

katrogan (commented):
how does this work with the lazy loader stuff? can we refactor that too?

plugins/sagemaker/setup.py
kumare3: i guess you are right, just keep sagemaker separate - then?

plugins/kfpytorch/setup.py
```diff
@@ -0,0 +1,32 @@
+from setuptools import setup
+
+PLUGIN_NAME = "kfpytorch"
+
+microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
+
+plugin_requires = ["flytekit>=0.16.0a2"]
```
kumare3, line 7: less than 1.0.0?

plugins/spark/flytekitplugins/spark/task.py
```diff
@@ -28,6 +28,13 @@ class Spark(object):
     spark_conf: Optional[Dict[str, str]] = None
     hadoop_conf: Optional[Dict[str, str]] = None
 
+    def __post_init__(self):
```
kumare3, line 31: thank you

plugins/aws/flytekitplugins/aws/hpo.py
```diff
@@ -4,6 +4,7 @@
 
 from flyteidl.plugins.sagemaker import hyperparameter_tuning_job_pb2 as _pb2_hpo_job
 from flyteidl.plugins.sagemaker import parameter_ranges_pb2 as _pb2_params
+from flytekitplugins.aws.training import SagemakerBuiltinAlgorithmsTask, SagemakerCustomTrainingTask
```
kumare3, line 7: I did not mean rename `sagemaker` -> `aws`, I meant move `sagemaker` -> `aws/sagemaker`

PR #340, opened 2021-01-18
Distributed training

kumare3:
Distributed training is interesting, as all the workers use the same entrypoint, but some may not produce an output.
Flytekit needs a capability of ignoring outputs for some workers.


flytekit/annotated/base_task.py
```diff
@@ -387,7 +406,11 @@ def dispatch_execute(
                 py_type = self.get_type_for_output_var(k, v)
                 if isinstance(v, tuple):
                     raise AssertionError(f"Output({k}) in task{self.name} received a tuple {v}, instead of {py_type}")
-                literals[k] = TypeEngine.to_literal(exec_ctx, v, py_type, literal_type)
+                try:
+                    literals[k] = TypeEngine.to_literal(exec_ctx, v, py_type, literal_type)
+                except Exception as e:
+                    raise AssertionError(f"failed to convert return value for var {k}") from e
```
wild-endeavor, line 412: should we attempt to print the value?

wild-endeavor (approved):
is this the last plugin change or do you have more?  After this is merged, I'll start updating the other PR.

kumare3:
> is this the last plugin change or do you have more? After this is merged, I'll start updating the other PR.

Yes, this is the last one that I can think off and is a little involved. Once merged I will hold of any others

flytekit/annotated/base_task.py
kumare3: i dont like printing the value, what if its a large string?

flytekit/bin/entrypoint.py
```diff
@@ -67,6 +67,137 @@ def _map_job_index_to_child_index(local_input_dir, datadir, index):
...
 
+def _dispatch_execute(ctx: FlyteContext, task_def: PythonTask, inputs_path: str, output_prefix: str):
+    """
+    Dispatches execute to PythonTask
+        Step1: Download inputs and load into a literal map
+        Step2: Invoke task - dispatch_execute
+        Step3:
+            a: [Optional] Record outputs to output_prefix
+            b: OR if IgnoreOutputs is raised, then ignore uploading outputs
+            c: OR if an unhandled exception is retrieved - record it as an errors.pb
```
wild-endeavor, line 78: this is not done yet right?  just making sure, i can do this next.

wild-endeavor (approved):

flytekit/bin/entrypoint.py
kumare3: ya not done yet, it raises the exception

PR #341, opened 2021-01-18
local caching

katrogan:


wild-endeavor:
How do users specify where they want the local cache file to be?  How do they specify that they want to clear it?  Will this automatically detect function signature changes?

katrogan:
> How do users specify where they want the local cache file to be?
We use the working dir which in local execute is just a filepath rather than auto-deleting tmp dir. @kumare3 suggested moving this to a dir in the user's home directory. i can also make the file within the directory configurable as well.

> How do they specify that they want to clear it?
Do we have any semantics around cache busting right now? Suggestions on a good interface? You can clear it currently as is done in tests but that's not exactly documented.

> Will this automatically detect function signature changes?
Yes

flytekit/annotated/base_task.py
```diff
@@ -234,7 +236,11 @@ def __call__(self, *args, **kwargs):
             with ctx.new_execution_context(
                 mode=ExecutionState.Mode.LOCAL_TASK_EXECUTION, execution_params=new_user_params
             ):
-                return self.execute(**kwargs)
+                execute = self.execute
```
kumare3, line 239: couple comments as we discussed

1. does this work for all inputs / outputs (complex types)
2. Should we do this after the translation to Flyte types?
3. Does this workflow for both workflow and single task execution locally?

Eventually I would prefer if we use a common directory like
`~/.flyte/cache`
and for executions
`~/.flyte/executions`

@wild-endeavor ^

flytekit/annotated/base_task.py
```diff
@@ -4,6 +4,8 @@
 from dataclasses import dataclass
 from typing import Any, Dict, Generic, Optional, Tuple, Type, TypeVar, Union
 
+from joblib import Memory
```
kumare3, line 7: we will need to add this to setup.py

katrogan: done as part of this change

katrogan:
closing due to issues with joblib. will put out an alternate proposal for caching

PR #343, opened 2021-01-20
Annotations - Cleanup for Beta

wild-endeavor:
* Bumped version to `0.16.0b0`
* Removed the `[all]` from the base requirements.in file - this shrinks the `requirements.txt` file considerably for flytekit itself.
* Add pandas as an explicit dependency to flytekit.
* Features were added to help plugins identify the root flytekit entrypoint script location in a Docker container.  This helps Spark and potentially other things in the future.
* Ketan's fix to boolean expressions in conditionals.


plugins/spark/scripts/flytekit_install_spark.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
```
kumare3, line 1: delete

PR #344, opened 2021-01-22
Add entrypoint path arg

katrogan:


flytekit/annotated/context_manager.py
```diff
@@ -86,16 +86,42 @@ class InstanceVar(object):
     o: Any
 
 
+class EntrypointSettings(object):
+    def __init__(self, path: str = None, command: str = None, version: int = None):
```
kumare3, line 90: Strings can be defaulted in constructor.

flytekit/annotated/context_manager.py
```diff
@@ -86,16 +86,42 @@ class InstanceVar(object):
     o: Any
 
 
+class EntrypointSettings(object):
```
kumare3, line 89: Data class?

plugins/spark/flytekitplugins/spark/task.py
```diff
@@ -76,11 +76,13 @@ def __init__(self, task_config: Spark, task_function: Callable, **kwargs):
         )
 
     def get_custom(self, settings: SerializationSettings) -> Dict[str, Any]:
-        from flytekit.bin import entrypoint as _entrypoint
+        spark_exec_path = settings.entrypoint_settings.path
+        if spark_exec_path is None:
```
kumare3, line 80: lets drop this? This is the new code, we should just assume this is always set?

kumare3: else actually better maybe we should default to this path? and then if we realize that we are not `in-container-serialize` then we should use `/opt/venv/bin/entrypoint.py`

flytekit/annotated/context_manager.py
katrogan: done, had it originally and then changed when using the env var and reverted back

plugins/spark/flytekitplugins/spark/task.py
katrogan: what do you mean?

katrogan: @kumare3 should I add a mode or some bool to SerializationSettings so we can set whether we are in container or not and whether to use the entrypoint path or the abs path?

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -31,9 +31,15 @@
 _DOMAIN_PLACEHOLDER = "{{ registration.domain }}"
 _VERSION_PLACEHOLDER = "{{ registration.version }}"
 
+# During out of container serialize the absolute path of the entrypoint at serialization time won't match the
+# in-container value at execution time. The following default value is used to provide the in-container entrypoint path
+# but can be optionally overridden at serialization time based on the installation of your flytekit virtualenv.
+_DEFAULT_IN_CONTAINER_ENTRYPOINT = "/opt/venv/bin/entrypoint.py"
```
kumare3, line 37: we probably want to say, default_flytekit_entrypoint_fileloc

plugins/spark/flytekitplugins/spark/task.py
```diff
@@ -76,12 +76,16 @@ def __init__(self, task_config: Spark, task_function: Callable, **kwargs):
         )
 
     def get_custom(self, settings: SerializationSettings) -> Dict[str, Any]:
+
         from flytekit.bin import entrypoint as _entrypoint
```
kumare3, line 80: can we delete this from here and move to the entrypoint path creation location. what i meant was that the defaulting should happen at the top level.

plugins/spark/flytekitplugins/spark/task.py
```diff
@@ -76,12 +76,16 @@ def __init__(self, task_config: Spark, task_function: Callable, **kwargs):
...
         spark_exec_path = os.path.abspath(_entrypoint.__file__)
         if spark_exec_path.endswith(".pyc"):
             spark_exec_path = spark_exec_path[:-1]
 
+        if settings.entrypoint_settings.path is not None:
+            spark_exec_path = settings.entrypoint_settings.path
+
         job = _task_model.SparkJob(
             spark_conf=self.task_config.spark_conf,
             hadoop_conf=self.task_config.hadoop_conf,
```
kumare3, line 91: `executor_path=sys.executable,` is probably wrong

plugins/spark/flytekitplugins/spark/task.py
katrogan: done

plugins/spark/flytekitplugins/spark/task.py
katrogan: done

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -239,6 +260,24 @@ def serialize(ctx, image, local_source_root, in_container_config_path):
...
+        # Set the entrypoint path to the in container default unless a user-specified option exists.
+        ctx.obj[CTX_FLYTEKIT_VIRTUALENV_ROOT] = (
+            in_container_virtualenv_root
+            if in_container_virtualenv_root is not None
+            else _DEFAULT_FLYTEKIT_VIRTUALENV_ROOT
+        )
+    else:
+        # For in container serialize we make sure to never accept an override the entrypoint path and determine it here
+        # instead.
+        import flytekit as _flytekit
```
kumare3, line 274: the import does not need to be here right? it can be at the top of the module?

flytekit/clis/sdk_in_container/serialize.py
katrogan: done

flytekit/clis/sdk_in_container/serialize.py
katrogan: ok done

plugins/spark/flytekitplugins/spark/task.py
```diff
@@ -76,17 +75,11 @@ def __init__(self, task_config: Spark, task_function: Callable, **kwargs):
...
-        if spark_exec_path.endswith(".pyc"):
-            spark_exec_path = spark_exec_path[:-1]
-
         job = _task_model.SparkJob(
             spark_conf=self.task_config.spark_conf,
             hadoop_conf=self.task_config.hadoop_conf,
-            application_file="local://" + spark_exec_path,
-            executor_path=sys.executable,
+            application_file="local://" + settings.entrypoint_settings.path,
+            executor_path=settings.flytekit_virtualenv_root,
```
kumare3, line 82: should this not be 
```suggestion
            executor_path=settings.flytekit_virtualenv_root + "/bin/python",
```

PR #351, opened 2021-01-27
Changes for flytekit serialization

katrogan:
# TL;DR
Running `pyflyte serialize` no longer produces identifier files for each serialized proto. Serialized protos now contain a skeleton identifier instead. This required changing the serialized version for sdk launch plans from `flyteidl.admin.launch_plan_pb2.LaunchPlanSpec` to `flyteidl.admin.launch_plan_pb2.LaunchPlan`. 

Tested serialization and registration on my local flyte sandbox and was able to launch workflows.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue

## Follow-up issue
_NA_


flytekit/clis/flyte_cli/main.py
```diff
@@ -1625,11 +1625,11 @@ def _extract_files(
...
 
     results = []
-    filename_iterator = iter(file_paths)
-    for identifier_file in filename_iterator:
-        object_file = next(filename_iterator)
-        # Serialized proto files are of the form: 12_foo.bar.<resource_type>.pb
-        id, entity = _extract_pair(identifier_file, object_file, project, domain, version, patches or {})
+    for proto_file in file_paths:
+        # Serialized proto files are of the form: 12_foo.bar_1.pb
+        # Where 12 indicates it is the 12 file to process in order and 1 that is of resource type 1, or TASK.
```
wild-endeavor, line 1630: add what 2 and 3 are?  since we don't delete proto fields they shouldn't change, even if we add more in the future.

wild-endeavor (approved):

flytekit/clis/flyte_cli/main.py
katrogan: it's defined above https://github.com/lyft/flytekit/pull/351/files#diff-43826f51b783bcfa61d9f2cba92c1f864167b02b39426746ee1dcf7f8c744436R1600

kumare3:
@katrogan so flytectl cannot blindly override `LaunchPlan.WorkflowId` right, as it may be a remote workflow ID?

katrogan:
@kumare3 exactly, but if it isn't we'll correctly fill it out at registration time

PR #354, opened 2021-01-29
Pandera integration using new plugin system

cosmicBboy:
# Pandera integration using new plugin system

This PR implements a pandera plugin to support `pandera.SchemaModel`s as an alternative way of expressing dataframe types in flyte tasks and workflows.

```python
import flytekitplugins.pandera

import pandera as pa
from pandera.typing import DataFrame, Series

class Schema(pa.SchemaModel):
    col1: Series[int]
    col2: Series[float]


@task
def my_task(df: DataFrame[Schema]):
    ...
```

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [X] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/types/schema/__init__.py
```diff
@@ -1,5 +1,6 @@
 from .types import (
     FlyteSchema,
+    FlyteSchemaTransformer,
```
wild-endeavor, line 3: Is this necessary?  The transformer shouldn't be exposed to the end user.  The idea of this file was to only import things that the user might want to reference/use.

plugins/tests/pandera/test_wf.py
```diff
@@ -0,0 +1,74 @@
+import typing
+
+import pandas
+import pandera
+import pytest
+
+import flytekitplugins.pandera
```
wild-endeavor, line 7: does this line pass lint?  it's not needed in code right?  it's just needed to trigger everything.

tests/flytekit/unit/annotated/test_type_hints.py
```diff
@@ -4,6 +4,7 @@
 from dataclasses import dataclass
 
 import pandas
+import pandera
```
wild-endeavor, line 7: can we move this to plugin tests?

plugins/tests/pandera/test_wf.py
wild-endeavor: we actually have the same problem with the regular pandas library.  Tasks and workflows that take in and return pandas dataframes need to trigger the SchemaTransformer, but unless they explicitly import flytekit.types.schema, the transformer isn't registered and it doesn't run.  Can you think of a way around this?

flytekit/__init__.py
```diff
@@ -13,10 +13,14 @@
 from flytekit.annotated.task import reference_task, task
 from flytekit.annotated.workflow import WorkflowFailurePolicy, reference_workflow, workflow
 from flytekit.loggers import logger
+from flytekit.types import schema
```
cosmicBboy, line 16: > we actually have the same problem with the regular pandas library. Tasks and workflows that take in and return pandas dataframes need to trigger the SchemaTransformer, but unless they explicitly import flytekit.types.schema, the transformer isn't registered and it doesn't run. Can you think of a way around this?

@wild-endeavor this is one workaround: importing the relevant types in the top-level `__init__.py` file such that the pandas dataframe transformer is loaded anytime user imports `flytekit`.

As for plugins, see the `import_plugins` function in `plugins/__init__.py`, which is called in this file. It just tries to import all plugins available in the new plugin system. This won't work for third-party flytekit plugins that aren't part of the plugin microlib.

kumare3: ya, for schema we should be able to do this, as we have now made it a requirement

wild-endeavor:
hey can you give me write access to your fork?  i want to make a PR into your PR.

cosmicBboy:
> hey can you give me write access to your fork? i want to make a PR into your PR.

done!

wild-endeavor:
Thanks.  let me know what you think. https://github.com/cosmicBboy/flytekit/pull/1

wild-endeavor (approved):

PR #355, opened 2021-01-29
Hide remaining implementations

wild-endeavor:
# TL;DR
These were missed!  We need to put these imports at the top layer.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
I just went through the cookbook as it stands on master and looked for imports that still referred to the `annotated` folder (which we're shortly going to rename).

Some of these are only used by extensions but it's still nice to move them.


kumare3 (approved):
I think there is definitely a better way of exposing variables at this level, need to research

katrogan (approved):

PR #364, opened 2021-02-04
Dynamically update lib version and publish all plugins

EngHabu:
# TL;DR
Automatically publishes to pypi when a release is created. Removes the need to update any versions embedded in the files (__init__ and setup.py) for the core library as well as plugins.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/688

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -29,7 +29,7 @@ jobs:
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: |
           python setup.py sdist bdist_wheel
-          twine upload dist/*
+          twine upload -r testpypi dist/*
       - run: cd plugins
       - name: Autobump plugin version
         run: |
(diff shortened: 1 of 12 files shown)


.github/workflows/pythonpublish.yml
```diff
@@ -8,19 +8,40 @@ jobs:
...
-      run: |
-        python setup.py sdist bdist_wheel
-        twine upload dist/*
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: "0"
+      - name: Set up Python
+        uses: actions/setup-python@v1
+        with:
+          python-version: "3.x"
```
katrogan, line 17: do we ever have to worry about this becoming 3.9 given that numpy has incompatibilities with it?

katrogan (approved):
this is so awesome!

.github/workflows/pythonpublish.yml
EngHabu: A fair concern... @wild-endeavor (since I copied this block)

PR #367, opened 2021-02-05
WIP: Workflow closure

kumare3:
It is possible to declare something like this
```python
@workflow
def my_workflow(i: int) -> int:
    # This works, but with a important caveat, my_workflow variables `i` should not be used inside the task. this will
    # cause implicit binding and will fail at runtime. The registration will be wrong! Not sure how to detect this.
    @task
    def foo(j: int) -> int:
        return j

    return foo(j=i)
```
This will automatically allow task foo to be loaded

Which also means it is possible to do
```python
def build():
  @workflow
  def my_workflow(i: int) -> int:
      # This works, but with a important caveat, my_workflow variables `i` should not be used inside the task. this will
      # cause implicit binding and will fail at runtime. The registration will be wrong! Not sure how to detect this.
      @task
      def foo(j: int) -> int:
          return j

      return foo(j=i)
  return my_workflow
```


flytekit/core/python_function_task.py
```diff
@@ -214,6 +277,14 @@ def __init__(
...
             **kwargs,
         )
+        if self.task_resolver is None:
+            if not istestfunction(func=task_function) and isnested(func=task_function):
+                raise ValueError(
+                    "TaskFunction cannot be a nested/inner or local function. "
+                    "It should be accessible at a module level for Flyte to execute it. Test modules with "
+                    "names begining with `test_` are allowed to have nested tasks. If you want to create your own tasks"
+                    "use the TaskResolverMixin"
+                )
```
mahanh, line 287: 1- what happens if task_resolver is None but the task is not nested?
2- what happens if task_resolver is not None but the task is nested (not in a workflow)?

kumare3:
Closing the PR in favor of @wild-endeavor - TaskResolver work

PR #368, opened 2021-02-06
Upload errors.pb in new API

wild-endeavor:
# TL;DR
Currently, new style Python-native typed tasks that raise an exception do not properly write an errors.pb file.  This replicates the existing logic in the flyte engine task code.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Copied code from https://github.com/flyteorg/flytekit/blob/master/flytekit/engines/flyte/engine.py#L320-L376

## Tracking Issue
https://github.com/flyteorg/flyte/issues/739

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -4,6 +4,7 @@ import logging as _logging
 import os as _os
 import pathlib
 import random as _random
+import traceback as _traceback
 
 import click as _click
 from flyteidl.core import literals_pb2 as _literals_pb2
@@ -11,6 +12,7 @@ from flyteidl.core import literals_pb2 as _literals_pb2
 from flytekit.common import constants as _constants
 from flytekit.common import utils as _common_utils
 from flytekit.common import utils as _utils
+from flytekit.common.exceptions import scopes as _scoped_exceptions
 from flytekit.common.exceptions import scopes as _scopes
 from flytekit.common.exceptions import system as _system_exceptions
 from flytekit.common.tasks.sdk_runnable import ExecutionParameters
@@ -29,6 +31,7 @@ from flytekit.interfaces.data.s3 import s3proxy as _s3proxy
 from flytekit.interfaces.stats.taggable import get_stats as _get_stats
 from flytekit.models import dynamic_job as _dynamic_job
 from flytekit.models import literals as _literal_models
+from flytekit.models.core import errors as _error_models
 from flytekit.models.core import identifier as _identifier
 from flytekit.tools.fast_registration import download_distribution as _download_distribution
 
@@ -76,6 +79,7 @@ def _dispatch_execute(ctx: FlyteContext, task_def: PythonTask, inputs_path: str,
             b: OR if IgnoreOutputs is raised, then ignore uploading outputs
             c: OR if an unhandled exception is retrieved - record it as an errors.pb
     """
+    output_file_dict = {}
     try:
         # Step1
         local_inputs_file = _os.path.join(ctx.execution_state.working_dir, "inputs.pb")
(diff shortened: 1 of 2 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -93,23 +98,40 @@ def _dispatch_execute(ctx: FlyteContext, task_def: PythonTask, inputs_path: str,
...
         if isinstance(e, IgnoreOutputs):
             # Step 3b
             _logging.warning(f"IgnoreOutputs received! Outputs.pb will not be uploaded. reason {e}")
             return
         # Step 3c
         _logging.error(f"Exception when executing task {task_def.name}, reason {str(e)}")
-        raise e
+        _logging.error("!! Begin Unknown System Error Captured by Flyte !!")
+        exc_str = _traceback.format_exc()
+        output_file_dict[_constants.ERROR_FILE_NAME] = _error_models.ErrorDocument(
```
kumare3, line 124: So I wanted to do this - in case of error (and this maybe because we are unable to write to S3/GCS etc) we should write the output to https://kubernetes.io/docs/tasks/debug-application-cluster/determine-reason-pod-failure/#customizing-the-termination-message
This would make propeller read the outputs and write it to Flyteadmin.

wild-endeavor: Can we backlog this as it's a bit different?

EngHabu: We kinda do that... [Here](https://github.com/flyteorg/flyteplugins/blob/8a49ee2bbef9e89b6d1cec94b4eef1ee8fe7aca3/go/tasks/pluginmachinery/flytek8s/container_helper.go#L127)
and we do indeed take that message and bubble up to Admin (after shortening it)

flytekit/bin/entrypoint.py
```diff
@@ -93,23 +98,40 @@ def _dispatch_execute(ctx: FlyteContext, task_def: PythonTask, inputs_path: str,
...
+            _error_models.ContainerError("SYSTEM:Unknown", exc_str, _error_models.ContainerError.Kind.RECOVERABLE,)
+        )
+        _logging.error(exc_str)
+        _logging.error("!! End Error Captured by Flyte !!")
+
+    for k, v in output_file_dict.items():
+        _common_utils.write_proto_to_file(v.to_flyte_idl(), _os.path.join(ctx.execution_state.engine_dir, k))
+
+    ctx.file_access.upload_directory(ctx.execution_state.engine_dir, output_prefix)
```
EngHabu, line 133: So what happens if this fails?

EngHabu: This is the issue we talked about earlier, isn't it? anyway we can capture that and write the original error (+ the new) to stdErr?

flytekit/bin/entrypoint.py
kumare3: @EngHabu I know we do that in FlytePropeller, what I meant was do this in flytekit as well?

EngHabu: Hmmm I'm not sure what do we NOT do currently (after this PR)? in the older flytekit, when we fail to write the outputs, we write THAT error into stdErr and that bubbles up... are you saying with this PR we don't yet do that? or are you saying "Also write the real outputs" to stdErr?

flytekit/bin/entrypoint.py
wild-endeavor: The exception gets caught at a higher layer and you see the error message you don't like https://nuclyde.slack.com/archives/D01FD1XGYRZ/p1612578571113500

flytekit/bin/entrypoint.py
kumare3: no, i do not follow, where we write the error to stdErr

kumare3 (approved):

PR #376, opened 2021-02-11
Allow customizing docker file path and context

EngHabu:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/scripts/flytekit_build_image.sh
+++ b/scripts/flytekit_build_image.sh
@@ -16,8 +16,15 @@ echo "           DOCKER BUILD"
 echo "------------------------------------"
 echo ""
 
+if [ -z "$DOCKERFILE_PATH" ]; then
+  DOCKERFILE_PATH=""
+fi
+
+if [ -z "$DOCKER_CONTEXT" ]; then
+  DOCKER_CONTEXT="."
+fi
+
 DIRPATH=""
-DOCKERFILE_PATH=""
 if [ -d "$1" ]; then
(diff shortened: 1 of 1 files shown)


scripts/flytekit_build_image.sh
```diff
@@ -16,8 +16,15 @@ echo "           DOCKER BUILD"
 echo "------------------------------------"
 echo ""
 
+if [ -z "$DOCKERFILE_PATH" ]; then
```
kumare3, line 19: Ohh my god!, we should basically convert this to a python script.

kumare3: Also it might be better to use Flags - https://www.baeldung.com/linux/use-command-line-arguments-in-bash-script#flags

wild-endeavor:
the `-f` switch?

scripts/flytekit_build_image.sh
wild-endeavor: i actually feel python is way more complicated.  once you do that, you tend to want to bring in things like click, have to use subprocess, and then you start seeing python stack traces, etc.  I'd rather get rid of this script entirely if possible, and leave as is if not.

PR #380, opened 2021-02-12
Fix RTD

wild-endeavor:
# TL;DR
Sometime ago, RTD started failing for flytekit. See the earlier successful run https://readthedocs.org/projects/flytekit/builds/12891931/ and the failing run https://readthedocs.org/projects/flytekit/builds/12999973/ both of which are for the same SHA.

Using the published RTD Docker images, I was able to reproduce the issue.  Still not sure of the root cause, but it appears that the behavior differs between:
a) a dependency (like grpcio or cryptography) being installed as part of a `pip install -r` requirements file, and
b) a dependency (like grpcio or cryptography) being installed as part of the `pip install .` that is run on the flytekit library itself.

In both cases, it appears that the latter method would trigger a build from source, invoking the gcc compiler (or the rust compiler in the cryptography case) rather than installing the prebuilt wheel.

In the grpcio case, the compilation would extend the RTD build time by twenty minutes.  In the cryptography case, the compilation failed completely because the RTD container does not have a Rust compiler.

I'm adding here a new requirements file just for docs.  Because RTD installs requirements before the install of `.`, both of these packages now install correctly and quickly.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added


katrogan (approved):
nice investigating

PR #381, opened 2021-02-12
Fix executor path and application file parameters for Spark Tasks

migueltol22:
# TL;DR
Add a new variable `CTX_PYTHON_INTERPRETER` to ctx. This points to the python interpreter that should be used as part of the executor path. 
Revert back to how `CTX_FLYTE_VIRTUALENV_ROOT` was previously calculated and only use the directory. This corrects teh application file from `local:///<some_path>/flytekit/__init__.py/bin/entrypoint.py` to `local:///<some_path>/flytekit/bin/entrypoint.py`

## Type
 - [x ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clis/sdk_in_container/serialize.py
+++ b/flytekit/clis/sdk_in_container/serialize.py
@@ -260,6 +260,7 @@ def serialize(ctx, image, local_source_root, in_container_config_path, in_contai
             if in_container_virtualenv_root is not None
             else _DEFAULT_FLYTEKIT_VIRTUALENV_ROOT
         )
+        ctx.obj[CTX_FLYTEKIT_VIRTUALENV_ROOT] = ctx.obj[CTX_FLYTEKIT_VIRTUALENV_ROOT] + "/bin/python3"
     else:
         # For in container serialize we make sure to never accept an override the entrypoint path and determine it here
         # instead.
--- a/flytekit/clis/sdk_in_container/serialize.py
+++ b/flytekit/clis/sdk_in_container/serialize.py
@@ -100,23 +100,24 @@ def serialize_all(
     python_interpreter: str = None,
 ):
     """
-    In order to register, we have to comply with Admin's endpoints. Those endpoints take the following objects. These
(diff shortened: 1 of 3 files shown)


migueltol22:
@wild-endeavor @katrogan @kumare3 I believe that this may break non-containerized serialization since `/bin/python` was needed for that case, is my understanding correct?

katrogan:
should we add the /bin/python then to the out of container case? https://github.com/flyteorg/flytekit/blob/master/flytekit/clis/sdk_in_container/serialize.py#L258

migueltol22:
@katrogan I'm not too familiar with the new sdk but wouldn't that cause issues elsewhere? Looks like that gets later on assigned as `flytekit_virtualenv_root` and that might be used here https://github.com/flyteorg/flytekit/blob/776e012caabb98b01fdead3d939e8b5ab8552aa6/flytekit/clis/sdk_in_container/serialize.py#L136

katrogan:
@migueltol22 i think you're right, we might want two ctx obj vars: one for the path and one for the flytekit_virtualenv_root or perhaps even just a flag that indicates whether we're in container or not and serialize and append /bin/python as required

katrogan (commented):
I think this looks right! Would you mind testing this out by adding 

`https://github.com/lyft/flytekit/archive/bf09d7b86e97866b5f3585a25a16762650674329.zip#subdirectory=plugins/spark&egg=flytekitplugins-spark` to your project requirements and building that image and running that flyte spark workflow? if that's too annoying let me know and i can test locally too!

migueltol22:
@katrogan @wild-endeavor Tested this with both in container serialize and out of container serialize.

wild-endeavor (approved):

katrogan (approved):

PR #388, opened 2021-02-17
List workflow outputs

wild-endeavor:
# TL;DR
This previously didn't work

```
    @task
    def t1(a: int) -> int:
        a = a + 5
        return a

    @workflow
    def list_output_wf() -> typing.List[int]:
        v = []
        for i in range(2):
            v.append(t1(a=i))
        return v
```

What was happening was that `Promise` objects were getting wrapped twice.  Coming out of the workflow function call, it was already a list of Promises.  What the workflow function should return is a `LiteralCollection` rather than attempting to rewrap. The end result was that `type(result.val.collection.literals[0].scalar.primitive.integer)` was again a `Promise` rather than a proper Python integer. (`result` here is the output of [`_local_execute`](https://github.com/flyteorg/flytekit/pull/388/files#diff-b1a8538044c80e14d7998344a528dccb59cf54c9735de724f39d345e5f8dfcf5R306))

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Removed the `_workflow_fn_outputs_to_promise` function.  I never understood why this function existed.
* Changed the `promise.translate_inputs_to_literals` function to just accept one side of an interface, either inputs or outputs.
* Then `translate_inputs_to_literals` can be used on the outputs of a workflow as well.
* Changed the string template for promises to say "Resolved" instead of "Var".

## Tracking Issue
NA

## Follow-up issue
NA

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -26,11 +26,14 @@ from flytekit.models.literals import Primitive
 
 def translate_inputs_to_literals(
     ctx: FlyteContext,
-    input_kwargs: Dict[str, Any],
-    flyte_interface_inputs: Dict[str, _interface_models.Variable],
-    native_input_types: Dict[str, type],
+    incoming_values: Dict[str, Any],
+    flyte_interface_types: Dict[str, _interface_models.Variable],
+    native_types: Dict[str, type],
 ) -> Dict[str, _literal_models.Literal]:
     """
+    The point of this function is to extract out Literals from a collection of either Python native values (which would
+    be converted into Flyte literals) or Promises (the literals in which would just get extracted).
+
     When calling a task inside a workflow, a user might do something like this.
 
         def my_wf(in1: int) -> int:
@@ -50,6 +53,13 @@ def translate_inputs_to_literals(
 
     Here, in task_2, during execution we'd have a list of Promises. We have to make sure to give task2 a Flyte
     LiteralCollection (Flyte's name for list), not a Python list of Flyte literals.
+
+    This helper function is used both when sorting out inputs to a task, as well as outputs of a function.
+
+    :param ctx: Context needed in case a non-primitive literal needs to be translated to a Flyte literal (like a file)
+    :param incoming_values: This is a map of your task's input kwargs basically
+    :param flyte_interface_types: One side of an :py:class:`flytekit.models.interface.TypedInterface` basically.
+    :param native_types: Map to native Python type.
     """
 
     def extract_value(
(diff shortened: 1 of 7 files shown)


flytekit/core/promise.py
```diff
@@ -50,6 +53,13 @@ def my_wf(in1: int, in2: int) -> int:
 
     Here, in task_2, during execution we'd have a list of Promises. We have to make sure to give task2 a Flyte
     LiteralCollection (Flyte's name for list), not a Python list of Flyte literals.
+
+    This helper function is used both when sorting out inputs to a task, as well as outputs of a function.
+
+    :param ctx: Context needed in case a non-primitive literal needs to be translated to a Flyte literal (like a file)
+    :param incoming_values: This is a map of your task's input kwargs basically
```
katrogan, line 60: or output, right?

katrogan (dismissed):

flytekit/core/promise.py
```diff
@@ -92,11 +102,11 @@ def extract_value(
             return TypeEngine.to_literal(ctx, input_val, val_type, flyte_literal_type)
 
     result = {}  # So as to not overwrite the input_kwargs
-    for k, v in input_kwargs.items():
-        if k not in interface.inputs:
+    for k, v in incoming_values.items():
```
kumare3, line 105: check `incoming_values is not None`

flytekit/core/workflow.py
```diff
@@ -250,28 +215,52 @@ def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], P
...
+
+        # First handle the empty return case.
+        # A workflow function may return a task that doesn't return anything
+        #   def wf():
+        #       return t1()
+        # or it may not return at all
+        #   def wf():
+        #       t1()
+        # In the former case we get the task's VoidPromise, in the latter we get None
         if (
```
kumare3, line 230: so in this case we do not assert that output specification matches the actual outputs? I think we should right?

wild-endeavor: sorry i don't understand... can you elaborate?

flytekit/core/workflow.py
```diff
@@ -250,28 +215,60 @@ def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], P
...
+                    f"{function_outputs} received but interface has {len(self.python_interface.outputs)} outputs.",
+                )
+
             return VoidPromise(self.name)
 
-        # TODO: Can we refactor the task code to be similar to what's in this function?
-        promises = _workflow_fn_outputs_to_promise(
-            ctx, self._native_interface.outputs, self.interface.outputs, function_outputs
+        # Because we should've already returned in the above check, we just raise an Exception here.
+        if len(self.python_interface.outputs) == 0:
```
kumare3, line 240: tests for these cases?

tests/flytekit/unit/core/test_workflows.py
```diff
@@ -48,3 +48,64 @@ def wf(a: int) -> (str, str):
...
+    @workflow
+    def wf(b: int) -> nt:
+        out = subwf(a=b)
+        return t1(a=out.named1)
+
+    x = wf(b=3)
+    assert x == (7,)
+
+
+def test_fdsjaf():
```
kumare3, line 91: can we have a better name please?

kumare3, line 91:
```suggestion
def test_unexpected_outputs():
```

kumare3 (approved):

PR #391, opened 2021-02-19
Instance variable load issue.

kumare3:
# TL;DR
- AFter the refactor in which registrable entities were removed, we
  inadvertently broke instance variable lookup
- This PR resurrects it, creating a separate Path.

One problem with this PR is potentially slower serialization performance
(unnoticeable to humans), because the modules are imported twice to scan
for relevant objects

The second pass scans only for Instance Vars, and topo sort is not
needed, as automatically create a inmemory map of all the tasks loaded.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clis/sdk_in_container/serialize.py
+++ b/flytekit/clis/sdk_in_container/serialize.py
@@ -101,24 +101,24 @@ def serialize_all(
     python_interpreter: str = None,
 ):
     """
-      This function will write to the folder specified the following protobuf types ::
-          flyteidl.admin.launch_plan_pb2.LaunchPlan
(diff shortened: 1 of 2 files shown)


katrogan (dismissed):
sigh, we really need comprehensive regression tests to catch this earlier

katrogan (commented):
is this tested?

wild-endeavor (approved):

PR #394, opened 2021-02-19
Make parsing of Docker image names more robust

sbrunk:
# TL;DR

Add more robust Docker image name parsing via https://github.com/realityone/docker-image-py.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This should fix https://github.com/flyteorg/flyte/issues/747 (Docker registry listening on custom port) and perhaps a few other image name edge cases because https://github.com/realityone/docker-image-py is a Python implementation of the original Docker image name parser (https://github.com/distribution/distribution/blob/main/reference/regexp.go).

Downside is another (albeit quite small) dependency.

I'm not completely sure what needs to be done in case of dependency changes. I ran `make requirements.txt` to update the pinned deps.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/747


sbrunk:
> I'm not completely sure what needs to be done in case of dependency changes. I ran `make requirements.txt` to update the pinned deps.

Ah it's `make requirements` to compile all pinned dependencies. But that fails with the following error:

```
pip-compile dev-requirements.in --upgrade --verbose
Traceback (most recent call last):
  File "/home/usisbrunk/anaconda3/envs/flytekit/bin/pip-compile", line 8, in <module>
    sys.exit(cli())
...
  File "/home/usisbrunk/anaconda3/envs/flytekit/lib/python3.8/site-packages/pip/_vendor/packaging/markers.py", line 232, in _get_env
    raise UndefinedEnvironmentName(
pip._vendor.packaging.markers.UndefinedEnvironmentName: 'extra' does not exist in evaluation environment.
make: *** [Makefile:51: dev-requirements.txt] Error 1
```

Which is probably due to this change
https://github.com/flyteorg/flytekit/blob/1c0559a76cfa6a6361134e7cf7bc9fa286289f8c/requirements.txt#L24 in the generated `requirements.txt`. If I remove the `; extra == "all"` manually it works again.

Any ideas?

doc-requirements.txt
```diff
@@ -41,9 +37,9 @@ black==19.10b0
     #   papermill
 bleach==3.3.0
     # via nbconvert
-boto3==1.17.7
+boto3==1.17.12
```
kumare3, line 40: why is this needed? hmm!

kumare3: ohh no papermill, i guess its for doc-requirements, thats fine

sbrunk:
I just ran `make requirements` again in a fresh virtualenv and this time it does not add the `extra`. Not sure what caused it.

kumare3:
> I just ran `make requirements` again in a fresh virtualenv and this time it does not add the `extra`. Not sure what caused it.

ya its just doc requirements and the legacy flytekit is still hiding underneath

kumare3:
@sbrunk so we use `pip-compile` to create a closure of all the requirements from requirements.in, doc-requirements.in, dev-requirements.in

So the requirements.in is also currently installing the unsupported version of `flytekit[all]` as we do have legacy users using Flytekit, which used `flytekit[spark3]` syntax to install plugins. But this is cumbersome and becomes hard to maintain. But to allow running all tests, we still keep the .all target.

You installation seems to be failing to find some dependencies and that is really weird

sbrunk:
@kumare3 I ran `make requirements` in a fresh virtualenv on another machine. The tests pass now so perhaps you could have another look.

kumare3 (approved):

PR #399, opened 2021-02-24
Protobuf type support

wild-endeavor:
# TL;DR
This adds a transformer to handle general Protobuf message types.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
This feature was missing from the new API.  In the old API, you used to be able to specify a generic custom Protobuf as an interface type. This adds a new transformer to do that.

* There's actually two protobuf types in the old style. https://github.com/flyteorg/flytekit/blob/master/flytekit/common/types/proto.py#L47 and https://github.com/flyteorg/flytekit/blob/master/flytekit/common/types/proto.py#L184.  The first stores a PB object as a `Binary` the second stores it as a `Generic`.  In this PR, we've opted for the latter because it plays nicer with the UI.
* In order to support flexibility of future user-supplied `TypeTransformers`, tweaked the logic that fetches a transformer in the core `TypeEngine` to check for `isinstance` and `issubclass`.

## Tracking Issue
https://flyte-org.slack.com/archives/CREL4QVAQ/p1614099425013900


wild-endeavor:
will wait for some l5 testing first.

flytekit/core/type_engine.py
```diff
@@ -180,6 +184,34 @@ def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type:
...
+
+    def __init__(self):
+        super().__init__("Protobuf-Transformer", _proto_reflection.GeneratedProtocolMessageType)
+
+    @staticmethod
+    def tag(expected_python_type: Type[T]) -> str:
+        return f"{expected_python_type.__module__}.{expected_python_type.__name__}"
+
+    def get_literal_type(self, t: Type[T]) -> LiteralType:
+        return LiteralType(simple=SimpleType.STRUCT, metadata={ProtobufTransformer.PB_FIELD_KEY: self.tag(t)})
```
kumare3, line 198: do you want to model this as binary or struct?

kumare3 (dismissed):

flytekit/core/type_engine.py
wild-endeavor: Haytham said struct was better - something about showing up in the UI.

flytekit/core/type_engine.py
```diff
@@ -180,6 +184,34 @@ def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type:
...
+
+    def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
+        struct = Struct()
+        struct.update(_MessageToDict(python_val))
+        return Literal(scalar=Scalar(generic=struct))
+
+    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> T:
+        if not (lv and lv.scalar and lv.scalar.generic):
+            raise AssertionError("Can only covert a generic literal to a Protobuf")
```
cosmicBboy, line 207:
```suggestion
            raise AssertionError("Can only cover a generic literal to a Protobuf")
```

cosmicBboy: also, should this line be covered in tests?

wild-endeavor: added

cosmicBboy (approved):

PR #404, opened 2021-02-26
Resolver pattern

wild-endeavor:
# TL;DR
This changes the execution side of flytekit. Previously tasks were loaded only with a module and key to look up through an `importlib` call. This introduces a "task resolver" construct which is responsible for two things:

1. Given a task, when serializing be able to produce a series of strings that, at execution time,
2. allow the resolver to reconstruct the task object.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Please see the comments on the `TaskResolverMixin`.
* When a task is serialized, the command that is generated is now the output of the task's resolver.
* The `entrypoint.py` click command has been updated to be able to handle that command.

As part of this change, it was found that it was useful to be able to keep track of the variable that a task resolver was assigned to. Because of this, we've added back (duplicated rather) the instance-tracking metaclass mechanism from the old API (the stuff in `registerable.py`).  See the new tests under the `tracking/` folder for a non-Flyte-based example of how it works.

* To simplify things for legacy API tasks, the `entrypoint.py` execute function has been separated out.
* `PythonAutoContainerTask` has been moved into a separate file.
* Because `PythonAutoContainerTasks` are now a `TrackedInstance` we no longer need the `InstanceVar` construct.

## Tracking Issue

## Follow-up issue
NA


flytekit/bin/entrypoint.py
```diff
@@ -222,98 +224,141 @@ def _handle_annotated_task(task_def: PythonTask, inputs: str, output_prefix: str
...
+    resolver_key = resolver[-1]  # e.g. 'default_task_resolver'
+    resolver_mod = _importlib.import_module(".".join(resolver_mod))
+    return getattr(resolver_mod, resolver_key)
 
 
 @_scopes.system_entry_point
-def _execute_map_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, max_concurrency, test):
-    task_module = _importlib.import_module(task_module)
-    task_def = getattr(task_module, task_name)
+def _execute_task(inputs, output_prefix, raw_output_data_prefix, test, resolver: str, resolver_args: List[str]):
```
kumare3, line 274: i think we need documentation here for each of these methods to explain why they exist

flytekit/core/tracker.py
```diff
@@ -0,0 +1,121 @@
...
+    Please see the notes for the metaclass above first.
+
+    This functionality has two use-cases currently,
+    * Keep track of naming for non-function ``PythonAutoContainerTasks``.  That is, things like the
+      :py:class:`flytekit.extras.sqlite3.task.SQLite3Task` task.
+    * Task resolvers, because task resolvers are instances of :py:class:`flytekit.core.python_auto_container.TaskResolverMixin`
+      classes, not the classes themselves, which means we need to look on the left hand side of them to see how to
+      find them at task execution time.
+    """
+    def __init__(self, *args, **kwargs):
```
kumare3, line 45: thank you for passing them along. This should be ok in the multi-inheritance case

flytekit/bin/entrypoint.py
```diff
@@ -222,98 +224,141 @@ def _handle_annotated_task(task_def: PythonTask, inputs: str, output_prefix: str
 
 
 @_scopes.system_entry_point
-def _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, test):
+def _legacy_execute_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, test):
```
katrogan, line 227: can we add a comment what legacy means in this case?

flytekit/bin/entrypoint.py
```diff
@@ -222,98 +224,141 @@ def _handle_annotated_task(task_def: PythonTask, inputs: str, output_prefix: str
...
 
-            # New core style task
-            elif not test and isinstance(task_def, PythonTask):
-                _handle_annotated_task(task_def, inputs, output_prefix, raw_output_data_prefix)
+
+def _load_resolver(resolver_location: str) -> TaskResolverMixin:
+    # Load the actual resolver - this cannot be a nested thing, whatever kind of resolver it is, it has to be loadable
+    # directly from importlib
+    # TODO: Handle corner cases, like where the first part is [] maybe
+    resolver = resolver_location.split(".")
```
katrogan, line 266: can you add a sample of what the resolver_location/resolver looks like? hard to see the big picture with the indexing below

flytekit/bin/entrypoint.py
```diff
@@ -222,98 +224,141 @@ def _handle_annotated_task(task_def: PythonTask, inputs: str, output_prefix: str
...
+):
     _click.echo(_utils.get_version_message())
     # Backwards compatibility - if Propeller hasn't filled this in, then it'll come through here as the original
     # template string, so let's explicitly set it to None so that the downstream functions will know to fall back
     # to the original shard formatter/prefix config.
     if raw_output_data_prefix == "{{.rawOutputDataPrefix}}":
         raw_output_data_prefix = None
 
-    _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, test)
+    # For new API tasks, we need to call a different function.
```
katrogan, line 352: include which version of the API(so new contributors can understand what's going on)?

flytekit/clis/sdk_in_container/serialize.py
```diff
@@ -158,11 +156,6 @@ def serialize_all(
             )
             old_style_entities.append(o)
 
-        # PythonInstanceTasks will not be picked up by the above, so we need to reiterate
-        for o, v in load_module_object_for_type(pkgs, PythonInstanceTask, additional_path=local_source_root).items():
```
katrogan, line 162: just to double check, we don't need this for https://github.com/flyteorg/flytekit/pull/391?

flytekit/core/python_auto_container.py
```diff
@@ -0,0 +1,274 @@
...
+    """
+    Please see the notes in the TaskResolverMixin as it describes this default behavior.
+    """
+    def name(self) -> str:
+        return "DefaultTaskResolver"
+
+    def load_task(self, loader_args: List[str]) -> PythonAutoContainerTask:
+        task_module = loader_args[1]
+        task_name = loader_args[3]
+        # import ipdb; ipdb.set_trace()
```
katrogan, line 200: rm?

flytekit/core/python_auto_container.py
```diff
@@ -0,0 +1,275 @@
...
+        @task
+        def t1(...) -> ...: ...
+
+    might result in a container with arguments like ::
+
+        pyflyte-execute --inputs s3://path/inputs.pb --output-prefix s3://outputs/location \
+        --raw-output-data-prefix /tmp/data \
+        --resolver flytekit.core.python_auto_container.default_task_resolver \
+        -- \
+        task-module repo_root.workflows task-name t1
```
cosmicBboy, line 135: why is the `example` portion of the module path not necessary?

flytekit/core/python_auto_container.py
```diff
@@ -0,0 +1,275 @@
...
+        task_type="python-task",
+        container_image: Optional[str] = None,
+        requests: Optional[Resources] = None,
+        limits: Optional[Resources] = None,
+        environment: Optional[Dict[str, str]] = None,
+        task_resolver: Optional[TaskResolverMixin] = None,
+        **kwargs,
+    ):
+        """
+        :param task_config: Configuration object for Task. Should be a unique type for that specific Task
```
cosmicBboy, line 49: params in the docstring don't follow the actual function signature

flytekit/clis/sdk_in_container/serialize.py
wild-endeavor: nope, getting rid of it, using the instance tracking metaclass now.

wild-endeavor:
https://github.com/flyteorg/flyte/pull/814 end to end tests.

flytekit/core/map_task.py
```diff
@@ -66,20 +67,23 @@ def __init__(
...
-            self._run_task._task_function.__module__,
-            "--task-name",
-            f"{self._run_task._task_function.__name__}",
             "--inputs",
             "{{.input}}",
             "--output-prefix",
             "{{.outputPrefix}}",
             "--raw-output-data-prefix",
             "{{.rawOutputDataPrefix}}",
+            "--resolver",
```
kumare3, line 78: this implies that the resolver has to be the last arg only?

kumare3: why not use multi-valued args? 
--resolver-arg "x=y" --resolver-arg "..."

flytekit/core/python_function_task.py
```diff
@@ -244,20 +144,23 @@ def execute(self, **kwargs) -> Any:
...
-            self._task_function.__name__,
             "--inputs",
             "{{.input}}",
             "--output-prefix",
             "{{.outputPrefix}}",
             "--raw-output-data-prefix",
             "{{.rawOutputDataPrefix}}",
+            "--resolver",
+            self.task_resolver.location,
+            "--",
```
kumare3, line 157: i think i really prefer --resolver-arg type of usage, why do you want this raw command line?

tests/flytekit/unit/core/tracker/a.py
```diff
@@ -0,0 +1,5 @@
+from flytekit.core.tracker import TrackedInstance
+
+
+class A(TrackedInstance):
```
kumare3, line 4: i think lets try a case in which we have `B(A) -> B subclass of A`

kumare3:
This looks great to me, couple nits

flytekit/core/python_function_task.py
wild-endeavor: Spent half an hour getting it to work with click and gave up?

flytekit/bin/entrypoint.py
```diff
@@ -5,6 +5,7 @@
 import pathlib
 import random as _random
 import traceback as _traceback
+from typing import List
```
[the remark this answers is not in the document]
wild-endeavor: I need to think about this more.  Can I do this as a separate PR?  I'll write up an issue... also because fast-register doesn't work right now with dynamic tasks.

flytekit/core/map_task.py
wild-endeavor: not sure what you mean.  the resolver can be specified anywhere, but the resolver args have to be at the end.

flytekit/core/python_auto_container.py
```diff
@@ -0,0 +1,275 @@
...
+            gpu_request=self.resources.requests.gpu,
+            memory_request=self.resources.requests.mem,
+            storage_limit=self.resources.limits.storage,
+            cpu_limit=self.resources.limits.cpu,
+            gpu_limit=self.resources.limits.gpu,
+            memory_limit=self.resources.limits.mem,
+        )
+
+
+class TaskResolverMixin(object):
```
cosmicBboy, line 116: python 3+: `object` is default parent class

```suggestion
class TaskResolverMixin
```

flytekit/core/context_manager.py
```diff
@@ -158,7 +160,7 @@ def get_instance_var(self, o: Any) -> InstanceVar:
 
 
 class CompilationState(object):
-    def __init__(self, prefix: str):
+    def __init__(self, prefix: str, task_resolver: Optional["TaskResolverMixin"] = None):
```
cosmicBboy, line 163: need docs for `task_resolver` arg?

flytekit/core/map_task.py
cosmicBboy: I think @kumare3 means an alternative syntax is
`--resolver ... --resolver-arg "task-module=<TASK_MODULE>" -- resolver-arg "task-name=<TASK_NAME>"`

Where `--resolver-arg` is a click option instead of an argument.

Will the resolver always have 2 args, `task-module` and `task-name`?

cosmicBboy: just throwing out another idea:
```
--resolver ... --resolver-task-module ... --resolver task-name ...
```

flytekit/core/python_auto_container.py
```diff
@@ -0,0 +1,275 @@
...
+class DefaultTaskResolver(TrackedInstance, TaskResolverMixin):
+    """
+    Please see the notes in the TaskResolverMixin as it describes this default behavior.
+    """
+
+    def name(self) -> str:
+        return "DefaultTaskResolver"
+
+    def load_task(self, loader_args: List[str]) -> PythonAutoContainerTask:
+        task_module = loader_args[1]
```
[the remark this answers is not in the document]
wild-endeavor: done

flytekit/core/python_auto_container.py
wild-endeavor: i like keeping it though, makes it explicit and it's everywhere else.

wild-endeavor:
Just to answer the above - the resolver arg is not an option yeah, it's a click arg.  The number of arguments is unknown except that we require it to be >=1.  it could be just one or someone may make a new resolver a year from now that has a hundred args.

cosmicBboy (approved):

v01dXYZ:
Would it be useful to add a warning about `cloudpickle` ? It's not a  robust way for serializing Python classes/functions compared to textual/bytecode code (or a package). Indeed, only the objects in `__main__` would be serialized, which means if an user wants to gather helper functions in a `utils.py` that won't work anymore.
+ the warnings from `cloudpickle` docs (https://github.com/cloudpipe/cloudpickle)
`cloudpickle` is excellent if we want to send functions from a REPL to a cluster but not with dealing with a modular source tree.

kumare3:
> Would it be useful to add a warning about `cloudpickle` ? It's not a robust way for serializing Python classes/functions compared to textual/bytecode code (or a package). Indeed, only the objects in `__main__` would be serialized, which means if an user wants to gather helper functions in a `utils.py` that won't work anymore + the warnings from `cloudpickle` docs (https://github.com/cloudpipe/cloudpickle)
> 
> `cloudpickle` is excellent if we want to send functions from a REPL to a cluster but not with dealing with a modular source tree.
@v01dXYZ we did not see this, I do agree. that is why Flytekit does not really recommend or support cloudpickle in any meaningful way. but we would love to talk more, would be open to discuss in slack?

PR #405, opened 2021-02-26
Map Tasks in new flytekit

katrogan:
# TL;DR
Re-architecting map tasks (array jobs) for new flytekit.

Instead of determining the array job node structure in a dynamic job spec produced at runtime, this change instead re-architects array task serialization to produce the array job custom at compile time. Execution behavior has also changed, so that individual array job instances now accept the entire input (collection), index into it appropriately and write a single output which the plugin then collects and coalesces into a collection. This output behavior is unchanged. The output interface is left as a collection specifically to support local workflow execution.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/609

## Follow-up issue
_NA_


flytekit/bin/entrypoint.py
```diff
@@ -262,6 +263,31 @@ def _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data
                 _handle_annotated_task(task_def, inputs, output_prefix, raw_output_data_prefix)
 
 
+@_scopes.system_entry_point
+def _execute_map_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, max_concurrency, test):
+    with _TemporaryConfiguration(_internal_config.CONFIGURATION_PATH.get()):
+        with _utils.AutoDeletingTempDir("input_dir"):
```
kumare3, line 269: we do not need the temporary configuration or autodeleting tempdir for the new code right? can we limit this for the old code, that will make it easier to delete in the future

flytekit/bin/entrypoint.py
```diff
@@ -316,5 +342,19 @@ def fast_execute_task_cmd(additional_distribution, dest_dir, task_execute_cmd):
     _os.system(" ".join(task_execute_cmd))
 
 
+@_pass_through.command("pyflyte-map-execute")
+@_click.option("--task-module", required=True)
```
kumare3, line 346: I would make sure this works with the new style that @wild-endeavor is thinking about, that way there will be one lesser modification. I mean if we do not need task-module/task-name then this is the time to get rid of it

flytekit/common/translator.py
```diff
@@ -123,6 +123,7 @@ def get_serializable_task(
         interface=entity.interface,
         custom=entity.get_custom(settings),
         container=entity.get_container(settings),
+        task_type_version=entity.task_type_version,
```
kumare3, line 126: have we tried to serialize any old examples? just to make sure they continue to work? the house price predictor is a good example maybe?

flytekit/core/map_task.py
```diff
@@ -1,72 +1,168 @@
-from typing import Any, Optional
+import os
```
kumare3, line 1: can we add module level docs, those get generated as docs and can be used in reference

flytekit/core/map_task.py
```diff
@@ -1,72 +1,168 @@
...
 from flytekit.core.base_task import PythonTask, TaskMetadata
+from flytekit.core.context_manager import ExecutionState, FlyteContext, SerializationSettings
 from flytekit.core.interface import transform_interface_to_list_interface
+from flytekit.core.python_function_task import PythonFunctionTask, get_registerable_container_image
+from flytekit.models.interface import Variable
+from flytekit.models.task import Container
 
 
 class MapPythonTask(PythonTask):
     """
```
kumare3, line 14: lets have dome docs on the usage and explaining what concurrency, min_success_ratio etc means.

flytekit/core/map_task.py
```diff
@@ -1,72 +1,168 @@
...
-
-    To do this we might have to give up on supporting lambda functions initially
+    TODO: support lambda functions
     """
 
-    def __init__(self, tk: PythonTask, metadata: Optional[TaskMetadata] = None, **kwargs):
+    def __init__(
+        self,
+        tk: PythonFunctionTask,
+        metadata: Optional[TaskMetadata] = None,
```
kumare3, line 21: I think you can skip the taskmetadata right? that is in the base task so will be auto in **kwargs?

flytekit/core/map_task.py
```diff
@@ -1,72 +1,168 @@
...
+
+    @staticmethod
+    def _compute_array_job_index() -> int:
+        """
+        Computes the absolute index of the current array job. This is determined by summing the compute-environment-specific
+        environment variable and the offset (if one's set). The offset will be set and used when the user request that the
+        job runs in a number of slots less than the size of the input.
+        """
+        offset = 0
+        if os.environ.get("BATCH_JOB_ARRAY_INDEX_OFFSET"):
```
kumare3, line 86: is this only for AWS Batch>?

kumare3: if so can we add that to the docs?

flytekit/core/map_task.py
```diff
@@ -1,72 +1,168 @@
...
         if len(outputs) == 1:
             return outputs[0]
 
         return tuple(outputs)
 
 
-def maptask(tk: PythonTask, concurrency="auto", metadata=None):
-    if not isinstance(tk, PythonTask):
-        raise ValueError(f"Only Flyte Task types are supported in maptask currently, received {type(tk)}")
+def maptask(tk: PythonFunctionTask, concurrency=None, min_success_ratio=None, metadata=None):
```
kumare3, line 164: lets have a usage example here

kumare3: what does concurrency=None mean? means full cardinality?
what is concurrency != length of the array?

kumare3:
Ohh Wow, this looks so much simple and nicer! Thank you so much

flytekit/core/map_task.py
katrogan: nope! k8s too. see the double env var lookup below https://github.com/flyteorg/flytekit/pull/405/files#diff-38030c87fd7703a4c95a033167b2ce8efa1d3b61824e891eb063bb1337def271R88 this gets the appropriate index env var name (k8s or batch specific) and then looks that up

tests/flytekit/unit/core/test_type_hints.py
```diff
@@ -329,6 +329,11 @@ def my_wf(a: typing.List[int]) -> (int, str):
     x = my_wf(a=[5, 6])
     assert x == (15, "world-7world-8")
 
+    a = [5, 6]
+    x, y = maptask(t1, metadata=TaskMetadata(retries=1))(a=a)
```
kumare3, line 333: what error do we get in the following cases
- if we try to bind a non-list type so i.e, the input `a` is not a list but an `int` 
- If we try to bind a str to a (not a list)
- if we bind a list of str to `a`

flytekit/common/translator.py
katrogan: the default task type version should still be 0 so existing dynamic array tasks should be unaffected. only map tasks increment the version

flytekit/core/map_task.py
katrogan: done

flytekit/core/map_task.py
katrogan: done in the exported `maptask` below

flytekit/bin/entrypoint.py
katrogan: synced with Yee offline, we'll coordinate to update this

flytekit/bin/entrypoint.py
katrogan: oops copy/pasta. done.

tests/flytekit/unit/core/test_type_hints.py
katrogan: updated with additional test cases

flytekit/core/map_task.py
katrogan: done

flytekit/core/map_task.py
```diff
@@ -1,72 +1,215 @@
...
-        self._run_task = tk
+    def __init__(
+        self,
+        python_function_task: PythonFunctionTask,
+        concurrency: int = None,
+        min_success_ratio: float = None,
+        **kwargs,
+    ):
+        """
+        :param task_function: This argument is implicitly passed and represents the repeatable function
```
cosmicBboy, line 33: should this match with arg `python_function_task`?

flytekit/core/map_task.py
```diff
@@ -1,72 +1,215 @@
...
+        """
+        This is called during locally run executions. Unlike array task execution on the Flyte platform, _raw_execute
+        produces the full output collection.
+        """
         outputs_expected = True
         if not self.interface.outputs:
             outputs_expected = False
         outputs = []
-        for k in self.interface.outputs.keys():
+        for _ in self._outputs_interface.keys():
```
cosmicBboy, line 155: just syntactic 🍬:
```python
outputs = [[] for _ in self._outputs_interface.keys()]
```

tests/flytekit/unit/core/test_type_hints.py
```diff
@@ -323,13 +323,30 @@ def t2(a: typing.List[int], b: typing.List[str]) -> (int, str):
...
     def my_wf(a: typing.List[int]) -> (int, str):
-        x, y = maptask(t1, metadata=TaskMetadata(retries=1))(a=a)
+        x, y = map(t1, metadata=TaskMetadata(retries=1))(a=a)
         return t2(a=x, b=y)
 
     x = my_wf(a=[5, 6])
     assert x == (15, "world-7world-8")
 
 
+def test_map_task():
```
wild-endeavor, line 333: Can we add the following tests?

* some [serialization test](https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/core/test_serialization.py#L15) something where we serialize a task, and look at it to make sure the task type is correct.
* what does serialization look like if the user does
```
mapped_t1 = map(t1, metadata=TaskMetadata(retries=1))

@workflow
def wf1():
    return mapped_t1(a=a)

@workflow
def wf2():
    map(t1, metadata=TaskMetadata(retries=1))(a=a)
```
there's basically two map tasks at this point right?  coming from the same `t1`.  At serialization, will this produce two protobufs?  Will they collide?  Are the names unique?
* a serialization test for the workflow
* test local workflow execution
* also what happens if you map over a launch plan or a sub wf?  should those error? 
* what happens if you map over a task that has multiple inputs?
* what happens if you map over a task that has multiple outputs?

Also can you remind me what restrictions there are here?  Can users do this?
```
@workflow
def wf() -> (something, smth_else):
  results = map(t1, metadata=TaskMetadata(retries=1))(a=a)
  x = t2(list_of_ints=results[0:5])
  y = t2(list_of_ints=results[5:])
  return x, y
```
no right?  What about in a dynamic task?

If we have a dynamic task that `range`s over something, and there ends up being N copies of the same task, does the current implementation preclude future optimization of those N copies into one array task?

flytekit/core/map_task.py
```diff
@@ -1,72 +1,215 @@
...
+        def my_wf(x: typing.List[int]) -> typing.List[str]:
+             return flytekit.map(my_mappable_task, metadata=TaskMetadata(retries=1), requests=Resources(cpu="10M"),
+                                 concurrency=10, min_success_ratio=0.75)(a=x)
+
+    At run time, the underlying map task will be run for every value in the input collection. Task-specific attributes
+    such as :py:class:`flytekit.TaskMetadata` and :py:class:`flytekit.Resources` are applied to individual instances
+    of the mapped task.
+
+    :param task_function: This argument is implicitly passed and represents the repeatable function
+    :param concurrency: If specified, this limits the number of mapped tasks than can run in parallel to the given batch
```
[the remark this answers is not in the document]
katrogan: no 😬 but it's in idl

tests/flytekit/unit/core/test_type_hints.py
```diff
@@ -307,7 +307,7 @@ def test_user_demo_test(mock_sql):
 
 def test_wf1_with_map():
     @task
-    def t1(a: int) -> typing.NamedTuple("OutputsBC", t1_int_output=int, c=str):
```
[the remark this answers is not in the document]
katrogan: because the mapped task would return a list of tuples. this is not currently supported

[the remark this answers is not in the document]
katrogan: +1 thank you for the catch, will add input and output validation

tests/flytekit/unit/core/test_type_hints.py
katrogan: note this test case already tests `test local workflow execution`

katrogan: >  Can users do this?

No this should be unchanged because the results are still a promise.

To answer your last question, we can certainly update dynamic tasks to produce array jobs (that was my initial approach). It's just mildly tedious to implement but certainly doable especially given that the old sdk just does this.

Added test cases for everything else above.

flytekit/core/map_task.py
```diff
@@ -1,72 +1,215 @@
...
-    if not isinstance(tk, PythonTask):
-        raise ValueError(f"Only Flyte Task types are supported in maptask currently, received {type(tk)}")
-    # We could register in a global singleton here?
-    return MapPythonTask(tk, metadata=metadata)
+def map(task_function: PythonFunctionTask, concurrency: int = None, min_success_ratio: float = None, **kwargs):
+    """
+    Use a map task for parallelizable tasks that are run across a List of an input type. A map task can be composed of
+    any individual :py:class:`flytekit.PythonFunctionTask`.
+
+    Invoke a map task with arguments using the :py:class:`list` version of the expected input. TODO this will one day
```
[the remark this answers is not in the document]
katrogan: gladly, done

flytekit/models/task.py
```diff
@@ -291,7 +291,7 @@ def from_flyte_idl(cls, pb2_object):
 
 
 class TaskTemplate(_common.FlyteIdlEntity):
-    def __init__(self, id, type, metadata, interface, custom, container=None):
+    def __init__(self, id, type, metadata, interface, custom, container=None, task_type_version=0):
         """
         A task template represents the full set of information necessary to perform a unit of work in the Flyte system.
         It contains the metadata about what inputs and outputs are consumed or produced.  It also contains the metadata
```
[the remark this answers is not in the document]
katrogan: added.

flytekit/common/tasks/task.py
```diff
@@ -37,7 +37,7 @@ class SdkTask(
     _task_model.TaskTemplate,
     metaclass=_sdk_bases.ExtendedSdkType,
 ):
-    def __init__(self, type, metadata, interface, custom, container=None):
+    def __init__(self, type, metadata, interface, custom, container=None, task_type_version=0):
         """
         :param Text type: This is used to define additional extensions for use by Propeller or SDK.
         :param TaskMetadata metadata: This contains information needed at runtime to determine behavior such as
```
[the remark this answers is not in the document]
katrogan: done

flytekit/core/map_task.py
katrogan: thanks! done

flytekit/models/array_job.py
```diff
@@ -16,9 +16,12 @@ def __init__(self, parallelism, size, min_successes):
         :param int min_successes: An absolute number of the minimum number of successful completions of subtasks. As
             soon as this criteria is met, the array job will be marked as successful and outputs will be computed.
```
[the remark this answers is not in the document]
katrogan: 👍 done

flytekit/core/map_task.py
```diff
@@ -1,72 +1,215 @@
...
+    def _compute_array_job_index() -> int:
+        """
+        Computes the absolute index of the current array job. This is determined by summing the compute-environment-specific
+        environment variable and the offset (if one's set). The offset will be set and used when the user request that the
+        job runs in a number of slots less than the size of the input.
+        """
+        offset = 0
+        if os.environ.get("BATCH_JOB_ARRAY_INDEX_OFFSET"):
+            offset = int(os.environ.get("BATCH_JOB_ARRAY_INDEX_OFFSET"))
+        return offset + int(os.environ.get(os.environ.get("BATCH_JOB_ARRAY_INDEX_VAR_NAME")))
```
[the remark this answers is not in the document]
katrogan: cool, done

flytekit/core/map_task.py
```diff
@@ -1,72 +1,215 @@
...
+        self._run_task = python_function_task
+        self._max_concurrency = concurrency
+        self._min_success_ratio = min_success_ratio
+        self._array_task_interface = python_function_task.python_interface
         super().__init__(
             name=name,
             interface=collection_interface,
-            metadata=metadata,
-            task_type="map_task",
+            task_type="container_array",
```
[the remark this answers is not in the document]
katrogan: updated, thanks

flytekit/core/map_task.py
katrogan: done

katrogan:
PTAL @wild-endeavor @cosmicBboy

katrogan:
PTAL @wild-endeavor @cosmicBboy

cosmicBboy (approved):

flytekit/bin/entrypoint.py
```diff
@@ -262,6 +264,28 @@ def _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data
...
+
+        task_index = _compute_array_job_index()
+        output_prefix = _os.path.join(output_prefix, str(task_index))
+
+        _handle_annotated_task(map_task, inputs, output_prefix, raw_output_data_prefix)
+
+    # Old style task
+    elif not test:
+        raise _system_exceptions.FlyteSystemAssertion(
+            "Map tasks are only supported for task of type `PythonTask` got task {} of type {} instead".format(
```
kumare3, line 283: can you also say, only supported for the typehinted sdk?

flytekit/bin/entrypoint.py
```diff
@@ -262,6 +264,28 @@ def _execute_task(task_module, task_name, inputs, output_prefix, raw_output_data
                 _handle_annotated_task(task_def, inputs, output_prefix, raw_output_data_prefix)
 
 
+@_scopes.system_entry_point
+def _execute_map_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, max_concurrency, test):
+    task_module = _importlib.import_module(task_module)
+    task_def = getattr(task_module, task_name)
+
+    if not test and isinstance(task_def, PythonFunctionTask):
```
kumare3, line 272: does this also mean, we can now use Map task for everything? I dont think that will work. So for example PythonSparkFunctionTask us also derived from this. In the case we should just ensure this in the map-task compilation point

katrogan: sure, that's already done. this check is defensive in case someone mucks with the container args

flytekit/bin/entrypoint.py
katrogan: done

wild-endeavor (commented):
up to you on the map name

wild-endeavor:
just one last comment then +1.  but can you please document all the prefix/path/retry stuff too? 🙏

flytekit/core/map_task.py
```diff
@@ -1,72 +1,221 @@
...
+    ):
+        """
+        :param python_function_task: This argument is implicitly passed and represents the repeatable function
+        :param concurrency: If specified, this limits the number of mapped tasks than can run in parallel to the given
+        batch size
+        :param min_success_ratio: If specified, this determines the minimum fraction of total jobs which can complete
+            successfully before terminating this task and marking it successful.
+        """
+        if len(python_function_task.python_interface.inputs.keys()) > 1:
+            raise ValueError("Map tasks only accept python function tasks with 0 or 1 inputs")
```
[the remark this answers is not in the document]
katrogan: it's kinda stupid but yes it does

flytekit/__init__.py
```diff
@@ -119,7 +119,7 @@
 from flytekit.core.context_manager import ExecutionParameters, FlyteContext
 from flytekit.core.dynamic_workflow_task import dynamic
 from flytekit.core.launch_plan import LaunchPlan
-from flytekit.core.map_task import maptask
+from flytekit.core.map_task import map
```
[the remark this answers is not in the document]
katrogan: sure, done

wild-endeavor (approved):

PR #409, opened 2021-03-04
Revamp pod tasks

katrogan:

# TL;DR
Revamp pod tasks to use the [OpenAPI spec for PodSpec](https://github.com/kubernetes/kubernetes/tree/master/api/openapi-spec)s, increment the task_type_version accordingly and utilize the new config field in the task template to set the primary container name

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/816

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/models/task.py
+++ b/flytekit/models/task.py
@@ -411,7 +411,7 @@ class TaskTemplate(_common.FlyteIdlEntity):
             custom=_json_format.MessageToDict(pb2_object.custom) if pb2_object else None,
             container=Container.from_flyte_idl(pb2_object.container) if pb2_object.HasField("container") else None,
             task_type_version=pb2_object.task_type_version,
-            config={k: v for k, v in _six.iteritems(pb2_object.config)} if pb2_object.config is not None else None,
+            config={k: v for k, v in pb2_object.config.items()} if pb2_object.config is not None else None,
         )
 
 
--- a/flytekit/models/task.py
+++ b/flytekit/models/task.py
@@ -393,8 +393,8 @@ class TaskTemplate(_common.FlyteIdlEntity):
             custom=_json_format.Parse(_json.dumps(self.custom), _struct.Struct()) if self.custom else None,
             container=self.container.to_flyte_idl() if self.container else None,
             task_type_version=self.task_type_version,
-        config={k: v for k, v in self.config.items()} if self.config is not None else None,
-    )
+            config={k: v for k, v in self.config.items()} if self.config is not None else None,
+        )
         return task_template
 
     @classmethod
@@ -411,7 +411,7 @@ class TaskTemplate(_common.FlyteIdlEntity):
             custom=_json_format.MessageToDict(pb2_object.custom) if pb2_object else None,
             container=Container.from_flyte_idl(pb2_object.container) if pb2_object.HasField("container") else None,
             task_type_version=pb2_object.task_type_version,
-            config={k: v for k, v in _six.iteritems(pb2_object.config)} if pb_obj.config is not None else None,
+            config={k: v for k, v in _six.iteritems(pb2_object.config)} if pb2_object.config is not None else None,
         )
 
 
--- a/flytekit/models/task.py
+++ b/flytekit/models/task.py
@@ -291,7 +291,7 @@ class TaskMetadata(_common.FlyteIdlEntity):
 
 
 class TaskTemplate(_common.FlyteIdlEntity):
-    def __init__(self, id, type, metadata, interface, custom, container=None, task_type_version=0):
+    def __init__(self, id, type, metadata, interface, custom, container=None, task_type_version=0, config=None):
         """
         A task template represents the full set of information necessary to perform a unit of work in the Flyte system.
         It contains the metadata about what inputs and outputs are consumed or produced.  It also contains the metadata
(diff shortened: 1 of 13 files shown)


flytekit/models/task.py
```diff
@@ -383,6 +394,10 @@ def to_flyte_idl(self):
             container=self.container.to_flyte_idl() if self.container else None,
             task_type_version=self.task_type_version,
         )
+        if self.config is not None:
+            for k, v in self.config.items():
+                task_template.config[k] = v
```
EngHabu, line 399: is task_template not of type dict? are we just trying to make a copy of it?, can we not do this:
```suggestion
        config={k: v for k, v in self.config.items()} if self.config is not None else None,
    )
```

flytekit/models/task.py
```diff
@@ -398,6 +413,7 @@ def from_flyte_idl(cls, pb2_object):
             custom=_json_format.MessageToDict(pb2_object.custom) if pb2_object else None,
             container=Container.from_flyte_idl(pb2_object.container) if pb2_object.HasField("container") else None,
             task_type_version=pb2_object.task_type_version,
+            config={k: v for k, v in _six.iteritems(pb2_object.config)},
```
EngHabu, line 416: Is it safe to do that if config is None? or should we add 
```suggestion
            config={k: v for k, v in _six.iteritems(pb2_object.config)} if pb_obj.config is not None else None,
```

plugins/pod/setup.py
```diff
@@ -4,7 +4,10 @@
 
 microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
 
-plugin_requires = ["flytekit>=0.16.0b0,<1.0.0", "k8s-proto>=0.0.3,<1.0.0"]
+plugin_requires = [
+    "flytekit>=0.16.0b0,<1.0.0",
+    "kubernetes>=12.0.1",
```
EngHabu, line 9: w0oOt

flytekit/models/task.py
katrogan: oh this doesn't work because the config is a proto map so I can only write key by key

plugins/pod/flytekitplugins/pod/task.py
```diff
@@ -33,7 +32,7 @@ def primary_container_name(self) -> str:
 class PodFunctionTask(PythonFunctionTask[Pod]):
     def __init__(self, task_config: Pod, task_function: Callable, **kwargs):
         super(PodFunctionTask, self).__init__(
-            task_config=task_config, task_type="sidecar", task_function=task_function, **kwargs,
+            task_config=task_config, task_type="sidecar", task_function=task_function, task_type_version=1, **kwargs,
```
kumare3, line 35: This is awesome, but as a follow up can we also create a RawPodTask? (not a PodFunctionTask), where users can just write the pod, spec and we will just run it, where they can say what is the primary container and it may look like the RawContainerTask?

Again not in this PR, maybe in the future

kumare3:
I think this is so cool -
cc @jeevb can you also please review this?

wild-endeavor:
Just a few questions.
1. Do the cookbook examples also need to change?
2. I looked over the IDL PR and read through the comments but I still don't understand the new config field at all.  Can you explain what it is and why it's necessary beyond the custom field?  Is it available to the plugin at execution time?  When should a plugin writer use the custom field versus the config field?
3. What line of code deals with that random question about dictionaries and protos or something that you were asking me today?

katrogan:
@wild-endeavor:

> Do the cookbook examples also need to change?

Yes, after we cut a release this will be a breaking change. In fact now that I think about it we might need that to be the last, last beta release

> Can you explain what it is and why it's necessary beyond the custom field? Is it available to the plugin at execution time? 

Yes it's more like meta-data about the custom. In this case the primary is necessary for monitoring the execution but the custom data produced by the pod is solely the PodSpec.  Another use case could be routing for presto queries where the custom only contains the query itself.

wild-endeavor:
So this is just entirely up to the plugin author?  They can choose to write all data to the metadata field instead right? And leave the custom blank?

flytekit/models/task.py
```diff
@@ -370,19 +373,29 @@ def container(self):
...
-        return _core_task.TaskTemplate(
+        task_template = _core_task.TaskTemplate(
             id=self.id.to_flyte_idl(),
             type=self.type,
             metadata=self.metadata.to_flyte_idl(),
             interface=self.interface.to_flyte_idl(),
             custom=_json_format.Parse(_json.dumps(self.custom), _struct.Struct()) if self.custom else None,
             container=self.container.to_flyte_idl() if self.container else None,
             task_type_version=self.task_type_version,
+            config={k: v for k, v in self.config.items()} if self.config is not None else None,
```
wild-endeavor, line 396: Add a comment here?  This looks like something someone might just revert to `self.config` one day.

katrogan: I added a test that will fail if it's not correctly set

katrogan:
@wild-endeavor 

> So this is just entirely up to the plugin author? They can choose to write all data to the metadata field instead right? And leave the custom blank?

yep

flytekit/models/task.py
```diff
@@ -398,6 +411,7 @@ def from_flyte_idl(cls, pb2_object):
             custom=_json_format.MessageToDict(pb2_object.custom) if pb2_object else None,
             container=Container.from_flyte_idl(pb2_object.container) if pb2_object.HasField("container") else None,
             task_type_version=pb2_object.task_type_version,
+            config={k: v for k, v in _six.iteritems(pb2_object.config)} if pb2_object.config is not None else None,
```
wild-endeavor, line 414: oh missed this sorry, can we get rid of six here and just do `.items()` like above?  we're no longer supporting python 2.

katrogan: done

wild-endeavor (approved):

PR #414, opened 2021-03-09
Fix subworkflow branch skip return

wild-endeavor:
# TL;DR
Please see https://github.com/flyteorg/flyte/issues/807
Basically when a `then()` block ran a task, but wasn't in the true condition case, we'd skip execution but we never did that for subworkflows.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Basically we're copying this section https://github.com/flyteorg/flytekit/blob/master/flytekit/core/base_task.py#L238-L246 which will skip the branches not taken when running locally.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/807

### DIFF
--- a/flytekit/core/workflow.py
+++ b/flytekit/core/workflow.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import collections
 import inspect
 from dataclasses import dataclass
 from enum import Enum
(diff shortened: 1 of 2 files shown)


wild-endeavor:
@cosmicBboy this is an example of what I mean when I said that there's a lot of copied code between workflow and task - they're both 'callable entities' so there's a lot of duplicate behavior.  We'll have to streamline and make all this stuff elegant at some point (but we also don't want to fall into the kinda mire that we found ourselves in wrt the old API - the class inheritance patterns there were so intense it made onboarding new contributors super hard).

EngHabu (approved):

cosmicBboy (approved):

PR #428, opened 2021-03-23
Move configuraiton of a few tools to pyproject.toml

honnix:
# TL;DR
* Move configuraiton of a few tools to pyproject.toml
* Unpinned `black`
* Reformatted by later version of `black`

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
As described by https://github.com/flyteorg/flyte/issues/859, some configuration could be consolidated. Stop pinning `black` to move forward using later version of it.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/859

## Follow-up issue
_NA_

### DIFF
--- a/requirements-spark3.txt
+++ b/requirements-spark3.txt
@@ -14,17 +14,14 @@ async-generator==1.10
     # via nbclient
 attrs==20.3.0
     # via
-    #   black
     #   jsonschema
     #   scantree
 backcall==0.2.0
     # via ipython
 bcrypt==3.2.0
     # via paramiko
-black==19.10b0
-    # via
-    #   flytekit
-    #   papermill
+black==20.8b1
+    # via papermill
 bleach==3.3.0
     # via nbconvert
 boto3==1.17.39
@@ -136,7 +133,9 @@ marshmallow==3.10.0
 mistune==0.8.4
     # via nbconvert
 mypy-extensions==0.4.3
-    # via typing-inspect
+    # via
+    #   black
+    #   typing-inspect
 natsort==7.1.1
     # via flytekit
 nbclient==0.5.3
@@ -304,7 +303,9 @@ traitlets==5.0.5
 typed-ast==1.4.2
     # via black
 typing-extensions==3.7.4.3
-    # via typing-inspect
+    # via
+    #   black
+    #   typing-inspect
 typing-inspect==0.6.0
     # via dataclasses-json
 urllib3==1.25.11
--- a/requirements-spark3.txt
+++ b/requirements-spark3.txt
@@ -10,10 +10,6 @@ ansiwrap==0.8.4
     # via papermill
 appdirs==1.4.4
     # via black
-appnope==0.1.2
-    # via
-    #   ipykernel
-    #   ipython
 async-generator==1.10
     # via nbclient
 attrs==20.3.0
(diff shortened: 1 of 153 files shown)


honnix:
@wild-endeavor PTAL at this minor change. Thanks.

requirements-spark3.txt
```diff
@@ -25,15 +25,15 @@ backcall==0.2.0
     # via ipython
 bcrypt==3.2.0
     # via paramiko
-black==19.10b0
+black==19.10b0; extra == "all-spark3"
```
wild-endeavor, line 28: Oh so you're running into this too.  do you know how to fix this?  I've spent at least a few hours already looking into this.  For me, it usually adds `extra == "all"` in the regular file.  Somehow, our internal qualifier is getting appended onto black's.  It makes no sense to me.

wild-endeavor: maybe @jeevb & @cosmicBboy too if you've seen this... basically running `make requirements` adds this "extra" thing to black (only ever to black, nothing else), which then causes pip to fail because this has nothing to do with black.  "all" and "all-spark3" are flytekit concepts.

honnix: I see. I didn't notice this. Will look into it. Thanks for pointing out.

honnix: It seems to happen "randomly" in any compiled file, and sometimes it doesn't happen at all. :thinking:

wild-endeavor: yeah i've had the same experience.

honnix: After debugging deeply into pip-compile and pip, I think pip-compile got confused by `blacked` introduced by both `papermill` and `flytekit` itself as part of notebook extra. pip-compile does a few rounds of resolution until nothing changed after a round, but the order of things appearing may result differently, so sometimes it choose the one without extra marker but sometimes it doesn't.

It's weird that papermill requires black as installation requirement. Maybe we can ask the maintainer about it.

On the hand, when the extra marker gets added to black, how does pip fail in your case? To me, when there is this marker, it is as if black were not added to requirements.txt when doing `pip install -r` because I'm not providing an extra, so it doesn't fail but may happily install a newer version of black.

Maybe to avoid dealing with black and having to pin the version. Let's just unpin it and do a big reformatting PR? It doesn't change too much as I tried.

honnix: Hmm, it actually has real code dependency on black https://github.com/nteract/papermill/blob/ca46ca6756ff15c1d03857af911f8060e35da8ab/papermill/translators.py#L189

wild-endeavor: Down to do a reformatting yeah.  Thanks for investigating!  Do you mind putting it in?  My failure scenario... I don't remember actually.  I pulled this branch and ran `make requirements` and it ran fine.  If I see it again before the reformat I'll add it here.

wild-endeavor:
Hang on though... why the removal of spark3?  I think we still want to keep that right?  Like for the next few versions we still want the old API to function properly including the old spark tasks.

honnix:
Not removal actually. `all` has been spark3 for some time, so we have been running CI only for spark3. I changed CI to run for both spark3 and 2. I think it's better to do in a separated PR though.

honnix:
@wild-endeavor PTAL again. The failed codecov is unfortunate but doesn't seem reasonable to fix in this PR.

wild-endeavor (approved):

honnix:
Now I feel sorry for all existing PRs. But shouldn't be too messy to resolve.

PR #432, opened 2021-03-25
Accept auth role arg in single task execution

katrogan:
# TL;DR
Accept auth role arg in single task execution. This is a required param for single task executions but currently can only be set from the user flytekit config.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Also bump flyteidl version to pick up the fix for a non-backwards compatible field number change for auth roles.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/860

## Follow-up issue
_NA_

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -35,7 +35,7 @@ flake8==3.9.0
     #   flake8-isort
 iniconfig==1.1.1
     # via pytest
-isort==5.7.0
+isort==5.8.0
     # via
     #   -r dev-requirements.in
     #   flake8-isort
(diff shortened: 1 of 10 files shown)


wild-endeavor:
How was this working in the past?  I know we've tested this before.

katrogan:
you could (and still can) set the auth role in the flytekit config (or with env vars)

wild-endeavor (approved):

PR #442, opened 2021-04-02
Add a lot more user friendly launch plan creating function

wild-endeavor:

# TL;DR
It was decided that the existing functions for creating launch plans were not user-friendly.  The `get_default_launch_plan` function required a context, but it's supposed to be the `FlyteContext` not the user-facing context (`ExecutionParameters`). The regular `create` function requires a name all the time. This PR introduces a new function that is
* cached by default
* Doesn't need the name. If no name is provided, it's assumed the default launch plan is what you're after.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
See code.

## Follow-up issue
https://github.com/flyteorg/flyte/issues/892

### DIFF
--- a/flytekit/core/launch_plan.py
+++ b/flytekit/core/launch_plan.py
@@ -93,6 +93,61 @@ class LaunchPlan(object):
         cls.CACHE[name] = lp
         return lp
 
+    @classmethod
+    def get_or_create(
+        cls,
+        workflow: _annotated_workflow.WorkflowBase,
+        name: Optional[str] = None,
+        default_inputs: Dict[str, Any] = None,
+        fixed_inputs: Dict[str, Any] = None,
+        schedule: _schedule_model.Schedule = None,
+        notifications: List[_common_models.Notification] = None,
+        auth_role: _common_models.AuthRole = None,
+    ) -> LaunchPlan:
+        """
+        This function offers a friendlier interface for creating launch plans. If the name for the launch plan is not
+        supplied, this assumes you are looking for the default launch plan for the workflow. If it is specified, it
+        will be used. If creating the default launch plan, none of the other arguments may be specified.
+
+        The resulting launch plan is also cached and if called again with the same name, the
+        cached version is returned
+
+        :param workflow: The Workflow to create a launch plan for.
+        :param name: If you supply a name, keep it mind it needs to be unique. That is, project, domain, version, and
+          this name form a primary key. If you do not supply a name, this function will assume you want the default
+          launch plan for the given workflow.
+        :param default_inputs: Default inputs, expressed as Python values.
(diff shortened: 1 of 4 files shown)


flytekit/core/launch_plan.py
```diff
@@ -20,6 +20,15 @@ class LaunchPlan(object):
 
     @staticmethod
```
katrogan, line 21: want to mark this deprecated?

katrogan (dismissed):

flytekit/core/launch_plan.py
katrogan: oh nvm, still being used below

cosmicBboy (approved):

kumare3:
haha - slightly more?

flytekit/core/launch_plan.py
```diff
@@ -93,6 +102,61 @@ def create(
...
         return lp
 
+    @classmethod
+    def get_or_create(
+        cls,
+        workflow: _annotated_workflow.WorkflowBase,
+        name: Optional[str] = None,
+        default_inputs: Dict[str, Any] = None,
+        fixed_inputs: Dict[str, Any] = None,
+        schedule: _schedule_model.Schedule = None,
```
kumare3, line 112: should we put all of this 
`schedule, notifications, auth_role` etc in a dataclass? something like meta etc
as we keep adding things here?

flytekit/core/launch_plan.py
```diff
@@ -93,6 +102,61 @@ def create(
...
+        """
+        if name is None and (
+            default_inputs is not None
+            or fixed_inputs is not None
+            or schedule is not None
+            or notifications is not None
+            or auth_role is not None
+        ):
+            raise ValueError(
+                "When get_or_create'ing the default launch plan, no arguments other than the workflow may be specified"
```
kumare3, line 142:
```suggestion
                "Only named launchplans can be created that have other properties. Drop the name if you want to create a default launchplan. Default launchplans cannot have any other associations"
```

Reword better, i added some stuff

flytekit/core/launch_plan.py
wild-endeavor: Maybe, but then the user has to make one more object to make this.  Can we hold off until we have one more thing?  It makes sense I guess if they're creating fifteen launch plans, and set all three settings with the same value - then they can create the meta once and use fifteen times, but i feel like that's not a common use case.

kumare3: if you do later, then it becomes a breaking change. I guess you are right though, this list since its optional can keep on increasing. 

Just that your first check for nil, has to keep evolving as well

kumare3 (approved):

PR #449, opened 2021-04-21
Customized container tasks and Shim tasks/executors

wild-endeavor:
# TL;DR
This PR introduces the notion of task container plugins (aka flytekit only plugins).  That is, you can now write and publish a Flyte Tasks for others to use, that when run, run a Docker container that you (the author of the plugin) specifies, rather than the user container.  That is, typically, when a user writes an `@task` task, at serialization time, the container image that gets associated with that task is the main workflow/task image, the image made by the person writing the `@task` decorator.  This new construct allows the task author to tailor and control the execution container.  The plugin author only has access to the `TaskTemplate` of course, since the user code won't exist in the author's image.

This relies on previous [work done](https://github.com/flyteorg/flyteplugins/pull/166/) to expose the TaskTemplate to the execution end.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

#### As a plugin-task author
Follow the sqlite3 external container example as a template.  Basically you need to
* Create a task type that subclasses `PythonCustomizedContainerTask`
  * Override the `get_custom` function - this object will be serialized into the `TaskTemplate` that the executor ultimately receives so put all the information you need to run that specific instance of the task in here. You can also override any other functions in the base class you need to.
* Create an executor for that task type by subclassing `ShimTaskExecutor` and implementing the `execute_from_model` function.
* In the constructor for your task subclass, point the `executor_type` argument to your executor subclass.  An instance of this class will be created at execution time. 

#### As a user
Simply import the task from the Python library that provides it, and use as normal.  The task will be registered and run as normal, just picking up the task author's container instead of yours at platform run-time.  It is possible that the plugin author did not provide execution behavior in the importable Python task and it is only available in the container.  In that case you'll need to mock it out withe `flytekit.patch` in order to run it locally.

### High Level Changes
* Add the concept of a `ExecutableTemplateShimTask`.
* Add the concept of a `ShimTaskExecutor`.
* Add a `PythonThirdPartyContainerTask` which first uses a `TaskTemplateResolver` implementing the `TaskResolverMixin` interface to load the task, and then the `ShimTaskExecutor` object to run it.
* The `SQLite3Task` has been copied into a new folder under `extras`.  I was going to add it somewhere else, but I actually think having both versions around so that people can compare between them is better.
* `TypeEngine` extended to add a guessing function to go backwards, from Flyte Literal types to Python types.

### Additional Changes
* `TaskResolverMixin` has been moved from the autocontainer file to `base_task.py` file and function signatures updated to reflect lower level classes.
* The `get_command` function in `PythonFunctionTask` and `PythonInstanceTask` were the same so this has been abstracted into the base class.
* `entrypoint.py` cleaned up a bit, setup step was extracted out.

## Tracking Issue
NA


flytekit/extras/sqlite3/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM python:3.8-slim-buster
```
kumare3, line 1: will we keep this dockerfile here? Is the idea that some plugins can ship with their dockerfiles?

flytekit/core/python_third_party_task.py
```diff
@@ -0,0 +1,153 @@
...
+from flytekit.core.resources import Resources, ResourceSpec
+from flytekit.core.tracker import TrackedInstance
+from flytekit.models import task as _task_model
+from flytekit.models.security import Secret, SecurityContext
+
+T = TypeVar("T")
+TC = TypeVar("TC")
+
+
+class TaskTemplateExecutor(TrackedInstance, Generic[T]):
```
kumare3, line 18: can we get some documentation here?

flytekit/extras/sqlite3/task.py
```diff
@@ -84,16 +92,76 @@ def output_columns(self) -> typing.Optional[typing.List[str]]:
...
+            "{{.outputPrefix}}",
+            "--raw-output-data-prefix",
+            "{{.rawOutputDataPrefix}}",
+            "--task_executor",
+            f"{SQLite3TaskExecutor.__module__}.{SQLite3TaskExecutor.__name__}",
+            "--task_template_path",
+            "{{.taskTemplatePath}}",
+        ]
+
+    def serialize_to_template(self, settings: SerializationSettings) -> task_models.TaskTemplate:
```
kumare3, line 119: why? This is confusing right? We have a base serialization method that does this already?

flytekit/extras/sqlite3/task.py
```diff
@@ -84,16 +92,76 @@ def output_columns(self) -> typing.Optional[typing.List[str]]:
...
+        return obj
+
+
+class SQLite3TaskExecutor(TaskTemplateExecutor[SQLite3Task]):
+
+    @classmethod
+    def promote_from_template(cls, tt: task_models.TaskTemplate) -> SQLite3Task:
+        qt = tt.custom["query_template"]
+
+        return SQLite3Task(
```
kumare3, line 143: why do we need this? What this reads as is that TaskTemplateExecutor will always need to be able to create a task from the template?

How about a simpler interface
TaskTemplateExecutor().execute(TaskTemplate, inputs)?

flytekit/core/python_third_party_task.py
```diff
@@ -0,0 +1,153 @@
...
+    @classmethod
+    def execute_from_template(cls, tt: _task_model.TaskTemplate, **kwargs) -> Any:
```
kumare3, line 20: i like this method, probably does not need to be a class method, this could just be a mixin?

I would drop the other 2 methods?

flytekit/extras/sqlite3/Dockerfile
wild-endeavor: I think i'm going to revert all these changes and add another example in the plugins folder.

flytekit/core/shim_task.py
```diff
@@ -0,0 +1,177 @@
...
+        return user_params
+
+    @classmethod
+    def post_execute(cls, user_params: ExecutionParameters, rval: Any) -> Any:
+        """
+        This function is a stub, just here to keep dispatch_execute compatibility between this class and PythonTask.
+        """
+        return rval
+
+    @classmethod
```
kumare3, line 92: why a class method? Do we want it to be static?
I think its better to use class methods when you construct classes from it?

kumare3: nm, i see type-engine uses the same thing

flytekit/core/type_engine.py
```diff
@@ -303,6 +312,24 @@ def get_available_transformers(cls) -> typing.KeysView[Type]:
...
+    def guess_python_types(
+        cls, flyte_variable_dict: typing.Dict[str, _interface_models.Variable]
+    ) -> typing.Dict[str, type]:
+        python_types = {}
+        for k, v in flyte_variable_dict.items():
+            python_types[k] = cls.guess_python_type(v.type)
+        return python_types
+
+    @classmethod
+    def guess_python_type(cls, flyte_type: LiteralType) -> type:
```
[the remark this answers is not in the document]
wild-endeavor: default behavior is to raise a `ValueError` which gets caught and ignored.

flytekit/core/python_third_party_task.py
```diff
@@ -0,0 +1,212 @@
...
+            self.interface,
+            self.get_custom(settings),
+            container=self.get_container(settings),
+            config=self.get_config(settings),
+        )
+        self._task_template = obj
+        return obj
+
+
+class TaskTemplateResolver(TrackedInstance, TaskResolverMixin):
```
kumare3, line 185: docs?

flytekit/core/python_third_party_task.py
```diff
@@ -0,0 +1,212 @@
...
+from flytekit.models.core import identifier as identifier_models
+from flytekit.models.security import Secret, SecurityContext
+from flytekit.tools.module_loader import load_object_from_module
+
+TC = TypeVar("TC")
+
+
+class PythonThirdPartyContainerTask(ExecutableTemplateShimTask, PythonTask[TC]):
+    """
+    Please take a look at the comments for ``ExecutableTemplateShimTask`` as well. This class should be subclassed
```
kumare3, line 26: better documentation, that should explain the rationale behind this class? when to use it and why?

flytekit/extras/external_container_sqlite3/task.py
```diff
@@ -0,0 +1,115 @@
...
+    shutil.unpack_archive(local_path, archive_dir)
+    # file gets uncompressed into to_dir/_arch/*.*
+    files = os.listdir(archive_dir)
+    if not files or len(files) == 0 or len(files) > 1:
+        raise RuntimeError(f"Uncompressed archive should contain only one file - found {files}!")
+    return os.path.join(archive_dir, files[0])
+
+
+@dataclass
+class SQLite3Config(object):
```
kumare3, line 35: How about use [asciiflow](https://asciiflow.com/#/) to explain whats happening in a SQLite3Task or just in general for a ThirdPartContainer task?

flytekit/core/python_third_party_task.py
kumare3: ping?

flytekit/core/python_third_party_task.py
```diff
@@ -0,0 +1,231 @@
...
+from flytekit.loggers import logger
+from flytekit.models import task as _task_model
+from flytekit.models.core import identifier as identifier_models
+from flytekit.models.security import Secret, SecurityContext
+from flytekit.tools.module_loader import load_object_from_module
+
+TC = TypeVar("TC")
+
+
+class PythonThirdPartyContainerTask(ExecutableTemplateShimTask, PythonTask[TC]):
```
kumare3, line 24: PythonShippableTask?

flytekit/core/python_third_party_task.py
kumare3: why this rant
```This class mostly follows the TaskResolverMixin pattern but kinda doesn't as well. The key difference is that
    a task resolver was supposed to
    * Be able to restore the same task when ``load_task`` is called as the object that ``loader_args`` was called on.
      That is, even though at run time it's in a container on a cluster and is obviously a different Python process,
      the Python object in memory should look the same.
    * Offer a one-to-one mapping between the list of strings returned by the ``loader_args`` function, an the task,
      at least within the container.```

Do we have this contract explicitly spelled out somewhere - if not this seems like a not so useful comment

kumare3: what about - special resolver that resolves the ThirdPartyTask at runtime, using only the tasktemplate. Thus for any task, derived from x, this resolver is used and such tasks should be fully rehydrated from the task template alone

wild-endeavor: i didn't think it was a rant.  i'll add this comment and change the wording a bit.

kumare3 (dismissed):

kumare3 (approved):

PR #455, opened 2021-04-27
Populates node names instead of defaulting to `n0`

jaychia:
# TL;DR

https://flyte-org.slack.com/archives/CREL4QVAQ/p1617144631154000

image
Currently, node names appear as `<prefix>-n<N>` on the UI, where N is the index of the node in the workflow. This also makes it difficult if the node name is used in other infrastructure where we rely on it to be human-interpretable (e.g Stackdriver logging)

This PR populates a human-interpretable node name by threading it through from the entity that the node is created from.

**NOTE:** This is my first contribution and I am unfamiliar with the codebase. In particular, I'm unsure if this way of threading the node name through will violate any uniqueness considerations which seem to be why we hacked it in as `prefix-nX` in the first place.

I'm guessing this fix isn't ideal because a simple case like so would probably lead to duplicate node names. However, I'm not sure what the ideal handling here should look like - should we just append a monotonically increasing index to the end of the node name?

```
@workflow
def wf():
    foo()
    foo()
```

## Type
 - [. ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -180,6 +180,9 @@ class CompilationState(object):
         self._branch_nodes: List[Node] = []
         self._task_resolver = task_resolver
 
+        # Count of entities as they are compiled {entity.name: count}
+        self._entity_counts: Dict[str, int] = {}
+
     @property
     def prefix(self) -> str:
         return self._prefix
@@ -195,6 +198,15 @@ class CompilationState(object):
         if self._branch:
             return self._branch_nodes
         return self._nodes
+
+    def add_entity(self, entity_name: str):
+        if entity_name in self._entity_counts:
+            self._entity_counts[entity_name] += 1
+        else:
+            self._entity_counts[entity_name] = 1
+
+    def get_entity_count(self, entity_name: str) -> int:
+        return self._entity_counts.get(entity_name, 0)
 
     def enter_conditional_section(self):
         """
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -687,21 +687,29 @@ def create_and_link_node(
         )
     )
 
-    node_name = f"{entity.__module__}.{entity.name}"
     node_metadata = _workflow_model.NodeMetadata(
-        node_name,
+        f"{entity.__module__}.{entity.name}",
         timeout or datetime.timedelta(),
         retry_strategy or _literal_models.RetryStrategy(0),
     )
 
+    # Node ID should be the entity name, but we need to add a suffix to ensure unique node IDs
+    # when we encounter more than one such entity during compilation
+    node_id = (
+        f"{entity.name}-n{ctx.compilation_state.get_entity_count(entity.name)}"
+        if ctx.compilation_state.get_entity_count(entity.name) != 0
+        else entity.name
+    )
+
     non_sdk_node = Node(
-        id=node_name,
+        id=node_id,
         metadata=node_metadata,
         bindings=sorted(bindings, key=lambda b: b.var),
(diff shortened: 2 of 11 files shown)


flytekit/core/promise.py
```diff
@@ -687,15 +687,15 @@ def create_and_link_node(
         )
     )
 
+    node_name = f"{entity.__module__}.{entity.name}"
```
kumare3, line 690: do we actually need the module in the node name?

kumare3 (changes requested):

jaychia:
PTAL @kumare3 @wild-endeavor

flytekit/core/promise.py
jaychia: Amended the logic, I'm not including the module name here anymore - we'll just handle duplicate entity names naively by incrementing a monotonically increasing count

flytekit/core/promise.py
```diff
@@ -693,15 +693,23 @@ def create_and_link_node(
         retry_strategy or _literal_models.RetryStrategy(0),
     )
 
+    # Node ID should be the entity name, but we need to add a suffix to ensure unique node IDs
+    # when we encounter more than one such entity during compilation
+    node_id = (
+        f"{entity.name}-n{ctx.compilation_state.get_entity_count(entity.name)}"
```
wild-endeavor, line 699: I think we still need the prefix... I added a test https://github.com/flyteorg/flytekit/commit/ff2a5509cc3a34c4dd089ff55fa6064f421e0540#diff-c34bbbaaada631ae99ff9734e7955d79dbc84791c54e5ee177ca796f11878109R153 that might fail weirdly now?

flytekit/core/context_manager.py
```diff
@@ -180,6 +180,9 @@ def __init__(self, prefix: str, task_resolver: Optional["TaskResolverMixin"] = N
         self._branch_nodes: List[Node] = []
         self._task_resolver = task_resolver
 
+        # Count of entities as they are compiled {entity.name: count}
+        self._entity_counts: Dict[str, int] = {}
```
wild-endeavor, line 184: Can we rename to `_entity_name_counts` instead or something?  `_entity_counts` makes it seem like we're storing the pointer location of the object as the key, rather than just the name.

flytekit/core/promise.py
```diff
@@ -693,15 +693,23 @@ def create_and_link_node(
         retry_strategy or _literal_models.RetryStrategy(0),
     )
 
+    # Node ID should be the entity name, but we need to add a suffix to ensure unique node IDs
+    # when we encounter more than one such entity during compilation
+    node_id = (
```
wild-endeavor, line 698: you have a local cluster right?  could you run through an example and add a screenshot of what things look like?  I'm worried that the `entity.name`s are too long in some cases.

Also, I was kinda hoping to gate this entire thing behind a configuration option.  Would that be possible?

flytekit/core/promise.py
wild-endeavor, line 699: What are your thoughts on having a leading `n{count}` where the count is the total count (i.e. the count that shows up today, `len(comp_state.nodes)`?  Would that be too distracting?  If we keep this current PR code, perhaps we can add the total count into the NodeMetadata name?  I feel like a monotonically increasing count index is helpful, esp if we ever have to debug weird topological sort issues.

PR #456, opened 2021-04-27
Validate All Arguments While Using the Cached LaunchPlan

samhita-alla:

# TL;DR
Verify if the cached outputs are the same as the arguments passed in the further repeat calls.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/892

## Complete Description
In flytekit's `core/launch_plan.py::LaunchPlan` object, the `get_or_create` function creates a launch plan. The user can supply things like a schedule or fixed or default inputs to it. The resultant launch plan is cached. On repeat calls, the cached version (based on the name) is returned. We should add a check to make sure that these repeat calls raise an error if any of the other arguments have changed.

## Follow-up issue
_NA_

### DIFF
--- a/tests/flytekit/unit/core/test_launch_plan.py
+++ b/tests/flytekit/unit/core/test_launch_plan.py
@@ -2,9 +2,12 @@ import typing
 
 import pytest
 
-from flytekit.core import launch_plan
+from flytekit.core import launch_plan, notification
+from flytekit.core.schedule import CronSchedule
 from flytekit.core.task import task
 from flytekit.core.workflow import workflow
+from flytekit.models.common import AuthRole
+from flytekit.models.core import execution as _execution_model
 
 
 def test_lp():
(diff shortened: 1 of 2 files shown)


wild-endeavor (commented):
Can we add unit tests for this, something that covers all the parameters?

Thanks!

samhita-alla:
> Can we add unit tests for this, something that covers all the parameters?
> 
> Thanks!

Added a unit test to handle this case.

tests/flytekit/unit/core/test_launch_plan.py
```diff
@@ -37,3 +36,62 @@ def wf(a: int) -> (str, str):
...
+        fixed_inputs={"c": "4"},
+        schedule=obj,
+        notifications=slack_notif,
+        auth_role=auth_role_model,
+    )
+
+    assert lp is lp2
+
+    # Check for assertion error when a different scheduler is used
+    lp3 = launch_plan.LaunchPlan.get_or_create(
```
wild-endeavor, line 86: can we add a test for each item separately?  Like two launch plans with `fixed_inputs` filled in, but all the other optional fields empty.  I'm just trying to make sure we capture any issues around Nones.

samhita-alla: Added `test_lp_each_parameter()` to handle the concerned cases.

PR #461, opened 2021-04-29
DoltTable Plugin

max-hoffman:
Signed-off-by: Max Hoffman <max@dolthub.com>

# TL;DR
Adds a DoltTable -- wrapper around `pandas.DataFrame` that commits changes to versioned database.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

hanging TODO's:
- code coverage
- metadata management?
- flytesnacks (separate PR)

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


plugins/dolt/scripts/flytekit_install_dolt.sh
```diff
@@ -0,0 +1,12 @@
...
+# Fetches and install Dolt. To be invoked by the Dockerfile
+
+# echo commands to the terminal output
+set -eox pipefail
+
+# Install Dolt
+
+apt-get update -y \
+    && apt-get install curl \
+    && sudo bash -c 'curl -L https://github.com/dolthub/dolt/releases/latest/download/install.sh | sudo bash'
```
max-hoffman, line 12: @wild-endeavor Something i was confused about before -- will CI run this script to install Dolt?

wild-endeavor (commented):
Can we rename the plugin from dolt to flytedolt or something?  I need to rename sqlalchemy actually to "flytesqlalchemy".  Under certain installations of pip dependencies, it thinks that the plugin is the real sqlalchemy instead, so it didn't pick up the real library.

max-hoffman:
@wild-endeavor would something like this solve those problems?
image
I'm surprised the other plugins like papermill and pandera haven't had issues

plugins/dolt/scripts/flytekit_install_dolt.sh
wild-endeavor: CI will run this https://github.com/flyteorg/flytekit/blob/master/.github/workflows/pythonbuild.yml#L37 so as long as it gets picked up from that we should be good.

I'm also thinking we'll need to move plugins into a separate repo sooner rather than later.

wild-endeavor:
is that common practice?  using `.` in the name of the plugin?

max-hoffman:
I think this would work too:
```
plugins/
    flytekitplugins.dolt/
        setup.py
        flytekitplugins/
            dolt/
                __init__.py
```
I'm not confident about how different namespacing formats impact pip clashes.

wild-endeavor:
yeah i like the second approach.  i think the top level one is the one messing things up.

wild-endeavor (dismissed):
+1 but unf you have to fix some conflicts.  I'll approve again after. just ping me.

wild-endeavor (approved):

PR #464, opened 2021-05-02
Unit tests

wild-endeavor:

# TL;DR
Just adding a couple unit tests that are a bit more complicated.


kumare3 (approved):
Awesome - these tests are passing!

PR #471, opened 2021-05-06
Make extract_return_annotation compatible with 3.9 NamedTuple

fediazgon:
# TL;DR
Make `extract_return_annotation` compatible with Python 3.9 `NamedTuple`.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
`_field_type` attribute was removed in Python 3.9 (see [typing.NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple)). Instead, check that the type has `tuple` as a base class, and that it has the `_field` attribute (part of [namedtuple API](https://docs.python.org/3/library/collections.html#collections.somenamedtuple._fields)). Then return `__annotations__` which contains the same information as `_field_types`. 

## Tracking Issue
https://github.com/flyteorg/flyte/issues/914

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/interface.py
+++ b/flytekit/core/interface.py
@@ -318,9 +318,10 @@ def extract_return_annotation(return_annotation: Union[Type, Tuple]) -> Dict[str
     if isinstance(return_annotation, Type) or isinstance(return_annotation, TypeVar):
         # isinstance / issubclass does not work for Namedtuple.
         # Options 1 and 2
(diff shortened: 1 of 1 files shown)


wild-endeavor (approved):
Is this the only problem with using Python 3.9 you guys have found?

PR #481, opened 2021-05-19
Serialize directly to model classes

wild-endeavor:
# TL;DR
Gets rid of the Sdk... classes in the `translator.py` module. This should be the last place flytekit is explicitly using the old classes.

Previously the `translator` module converted from the new Python classes (PythonFunctionTask, PythonFunctionWorkflow, LaunchPlan) into the `SdkXyz` classes, and then relied on the `serialize()` call within each of those Sdk classes, invoked by `serialize.py` to convert them into model objects.  Now `translator` will go directly to the model, and the serialization step will just call `to_flyte_idl()`

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
* Remove most of the `common` dependencies in `translator.py` and replace with the model classes.
* Python objects get converted to the same model classes as the output of their respective serialize calls earlier, namely
  * Tasks -> `flytekit.models.task.TaskSpec`
  * Launch Plans -> `flytekit.models.launch_plan.LaunchPlan`
  * Workflows -> `flytekit.models.admin.WorkflowSpec`
* Realized that we were actually still relying on the two `AuthRole` settings in the serialization step. These were being read in the property field for `auth_role` in the `SdkLaunchPlan` class itself.
* Handling for reference tasks/workflows is actually not great when it comes to getting the model representation. I'm currently using `None`.  That is, whenever a reference entity is run through `translator.py`, a `None` is returned and cached in the ordered dict.  This is useful because the translator is used in one other place - in dynamic tasks.  The runtime dynamic task compilation will look for these `None`s and reject them.  Using `None` as a sentinel value is not good, but I think this PR has enough going on so will address that in a future PR.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/822

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clis/flyte_cli/main.py
+++ b/flytekit/clis/flyte_cli/main.py
@@ -1759,7 +1759,6 @@ def _extract_files(
         # Where 12 indicates it is the 12 file to process in order and 1 that is of resource type 1, or TASK.
         resource_type = int(proto_file[-4])
         id, entity = _extract_pair(proto_file, resource_type, project, domain, version, patches or {})
-        print(f"Extracted {proto_file} to {id} and {entity}")
         results.append((id, entity))
 
     return results
--- a/flytekit/clis/flyte_cli/main.py
+++ b/flytekit/clis/flyte_cli/main.py
@@ -1759,6 +1759,7 @@ def _extract_files(
         # Where 12 indicates it is the 12 file to process in order and 1 that is of resource type 1, or TASK.
         resource_type = int(proto_file[-4])
         id, entity = _extract_pair(proto_file, resource_type, project, domain, version, patches or {})
+        print(f"Extracted {proto_file} to {id} and {entity}")
         results.append((id, entity))
 
     return results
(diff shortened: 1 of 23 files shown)


flytekit/clis/flyte_cli/main.py
```diff
@@ -1818,6 +1818,7 @@ def _extract_and_register(
 
     flyte_entities_list = _extract_files(project, domain, version, file_paths, patches)
     for id, flyte_entity in flyte_entities_list:
+        _click.secho(f"Registering {id}", fg="green")
```
kumare3, line 1821:
```suggestion
        _click.secho(f"Registering {id}", fg="yellow")
```

kumare3: green makes you feel that it was successful.
Ideally I would like this

Registering {name}..... [OK] | [FAILED]

kumare3:
this looks awesome

kumare3 (approved):

kumare3 (approved):

kumare3 (approved):

PR #482, opened 2021-05-20
Add gcp auth client

tnsetting:
# TL;DR
Add support for auth in google cloud setup

How to configure it?
In the config file:
```
[credentials]
auth_mode=gcp
```

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clis/auth/auth.py
+++ b/flytekit/clis/auth/auth.py
@@ -9,18 +9,17 @@ from http import HTTPStatus as _StatusCodes
 from multiprocessing import get_context as _mp_get_context
 from urllib.parse import urlencode as _urlencode
 
-import google.auth
-import google.auth.transport.requests
+import google.auth as _google_auth
 import keyring as _keyring
 import requests as _requests
-from google.oauth2 import service_account
+from google.auth.transport import requests as _grequests
+from google.oauth2 import service_account as _service_account
 
 from flytekit.loggers import auth_logger
 
 _code_verifier_length = 64
 _random_seed_length = 40
 _utf_8 = "utf-8"
-
 
 # Identifies the service used for storing passwords in keyring
 _keyring_service_name = "flyteauth"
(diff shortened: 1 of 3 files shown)


EngHabu:
Hey Nian,
As we discussed yesterday. Ideally we want to keep things consistent across providers. Can you explore the idea of generating the token outside of flytekit and just pushing the token as an environment variable?

tnsetting:
It is possible to do that but the user experience will not be as good. In that case, user has to create the environment variable every time.

tnsetting:
> The way I understand it is the target audience here is CI, right? which is something you can automate... please correct me if I'm wrong

There are two reasons for this:

- Since we added a proxy in front of flyteadmin which did not implement the oauth flow, we need to send the token right away for all our users.
- The second reason is that in this case, user(both in CI and real user) does not need to worry about token being timeout which could happen when doing `watch-execution`. If we just run a command to create the token, the token could timeout.

tnsetting:
Hi @EngHabu ! Could you take another look for the PR? Thanks

EngHabu:
@wild-endeavor is thinking if there are other ways we can do this.. I would really prefer not to add this code into flytekit...
Do you happen to know how the refresh token logic work? is it using any standard OAuth2 flow?

tnsetting:
Ok. At least this is a very common way to get credentials and tokens from google. 

We will create another PR which basically call a command and returns a token from the command and set the token. In that way you wont get the token expired problem. Wdyt?

flytekit/clis/auth/auth.py
```diff
@@ -305,3 +307,53 @@ def expired(self):
...
+    @property
+    def has_valid_credentials(self) -> bool:
+        return self._gcp_credentials is not None
+
+    @property
+    def can_id_token(self) -> bool:
+        return self._gcp_credentials is not None
+
+    def refresh_id_token(self):
+        auth_req = _grequests.Request()
```
tnsetting, line 340: @EngHabu and @wild-endeavor Here is where token refresh happens. Basically you send a token refresh request along with client credentials as here: https://github.com/googleapis/google-auth-library-python/blob/82293fe2caaf5258babb5df1cff0a5ddc9e44b38/google/oauth2/credentials.py#L228

wild-endeavor:
Currently, the normal three-legged oauth flow code is inside credentials.py::_refresh_credentials_standard and the client credentials flow is inside credentials.py::_refresh_credentials_basic.  The standard one just creates a full-blown auth.py::AuthorizationClient and delegates refreshing to that.

All of the code in the AuthorizationClient today is for doing the three-legged flow.  The code for doing the basic flow lives in basic_auth.py.  Both the basic flow and the three-legged flow call get_authorization_endpoints. 

If you want, you can move the basic flow into AuthorizationClient:
* Move the basic_auth.py code into the client.
* Add the ability to read tokens from the environment (maybe only when doing basic flow).
* Add client credentials refresh logic to the AuthorizationClient (doesn't currently exist - at Lyft a new token was just requested every time).  I'm not sure if the 3-legged refresh logic is always the same as a client credentials refresh flow.  May need a separate function.  May need a separate function anyways since it's 

If you don't want, you can just add the necessary logic to basic_auth.py
* In _refresh_credentials_basic, detect whether or not a refresh token is in the environment.
* If so, refresh using that, instead of trying to get a new one using client id and client secret. (This latter function may be a mostly copy paste of the AuthorizationClient refresh logic (keyring stuff removed).

The fact that the client credentials logic seems pretty isolated from the three-legged flow is why I decided against merging them when I was looking at this a few weeks ago.  I think it's fine either way though.  If you go the first/more complicated route, just make sure that the AuthorizationClient can spin up with all the three-legged config stuff missing (that is, shouldn't break people at Lyft that only have the client credentials flow settings in their config files).

tnsetting:
Hi @EngHabu and @wild-endeavor ,

Thanks for the detailed explanation. From our point of view, we want to have something simple and hopefully not breaking the flow your guys have already. My initial thought for using the google auth lib was just because it is kind of out-of-box, ready-to-used lib which provide me the tokens needed for authentication.

But as I explained earlier, due to the proxy that we have in front of flyteadmin does not implement the auth flow, we can not use the full 3-legged flow and it looks our flow is some what of similar the basic flow in code base. The difference between the PR and basic flow is that use you use a stored secret to [refresh the token](https://github.com/flyteorg/flytekit/blob/1d6b6842b0ebbcb3567af3882e580824758b44b1/flytekit/clients/raw.py#L72) and there PR use google's credential to refresh the token. I mean at the API call level it is the same. Maybe the part does not fit in the picture is the PR use a google lib which wraps extraction google specific credential.

Back to suggestion you have regarding the refresh token, I don't think we get the refresh token easily and gcloud command line tool does not provide a way to do that and this is no API in google auth api to extract. It doesn't make much sense to write code extract that either. Otherwise why not use the google lib directly and we don't have maintain it.

To me, the other option would be providing flytekit a command which can figure out the tokens for you, such get_id_token. So when flytekit need to refresh token, it just call external binary which will output a token for you. It is very portable since how the token is generated is completely outsourced to the other binary/cmd. So in our case we just use `gcloud print identity-token`. Let me know if this could be an option for you.

sonjaer:
@tnsetting do you mean something like this: https://github.com/sonjaer/flytekit/pull/1 ?

PR #495, opened 2021-05-31
Great Expectations Plugin

samhita-alla:

# TL;DR
Add [great expectations](https://greatexpectations.io/) plugin to validate data.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### TASK
Define `GETask` and validate the dataset by passing it as an argument to the task.

Valid data example:

```python
import os

import pandas as pd
from flytekitplugins.greatexpectations import GETask

from flytekit import kwtypes, task, workflow

task_object = GETask(
    name="test4",
    data_source="data",
    inputs=kwtypes(dataset=str),
    expectation_suite="test.demo",
    data_connector="data_example_data_connector",
)


@task
def my_task(csv_file: str) -> int:
    task_object(dataset=csv_file)
    df = pd.read_csv(os.path.join("data", csv_file))
    return df.shape[0]


@workflow
def valid_wf(dataset: str = "yellow_tripdata_sample_2019-01.csv") -> int:
    return my_task(csv_file=dataset)


if __name__ == "__main__":
    print(f"Running {__file__} main...")
    print(valid_wf())
```

Invalid data example:
```python
@workflow
def invalid_wf(dataset: str = "yellow_tripdata_sample_2019-02.csv") -> int:
    return my_task(csv_file=dataset)


if __name__ == "__main__":
    print(f"Running {__file__} main...")
    invalid_wf()
```

Output:
```bash
...
COLUMN          FAILED EXPECTATION
passenger_count -> expect_column_min_to_be_between
passenger_count -> expect_column_mean_to_be_between
passenger_count -> expect_column_quantile_values_to_be_between
passenger_count -> expect_column_values_to_be_in_set
passenger_count -> expect_column_proportion_of_unique_values_to_be_between
trip_distance -> expect_column_max_to_be_between
trip_distance -> expect_column_mean_to_be_between
trip_distance -> expect_column_median_to_be_between
trip_distance -> expect_column_quantile_values_to_be_between
trip_distance -> expect_column_proportion_of_unique_values_to_be_between
rate_code_id -> expect_column_max_to_be_between
rate_code_id -> expect_column_mean_to_be_between
rate_code_id -> expect_column_proportion_of_unique_values_to_be_between
...
```
### TYPE

Attach `GEType` to your dataset in a `task` which shall then automatically validate your data.

```python
import os

import pandas as pd
import pytest
from flytekitplugins.greatexpectations import GEConfig, GEType
from great_expectations.exceptions import ValidationError

ge_config = GEConfig(
    data_source="data",
    expectation_suite="test.demo",
    data_connector="data_example_data_connector",
)


@task
def my_task(csv_file: GEType[str, ge_config]) -> int:
    df = pd.read_csv(os.path.join("data", csv_file))
    df.drop(5, axis=0, inplace=True)
    return df.shape[0]


@workflow
def valid_wf(dataset: str = "yellow_tripdata_sample_2019-01.csv") -> int:
    return my_task(csv_file=dataset)


valid_wf()
```

**EXAMPLES CONCERNING FlyteSchema & FlyteFile ARE AVAILABLE IN TESTS**

## Tracking Issue

## Follow-up issue
_NA_
OR


plugins/tests/greatexpectations/great_expectations/expectations/.ge_store_backend_id
```diff
@@ -0,0 +1 @@
+store_backend_id = 2f8dcd42-bb6b-4b9f-9ee6-1a7dd1cdb869
```
kumare3, line 1: is this intended to be checked in?

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,166 @@
...
+        )
+
+    def execute(self, **kwargs) -> typing.Any:
+        context = ge.data_context.DataContext(self._data_context)
+
+        if len(self.python_interface.inputs.keys()) != 1:
+            raise RuntimeError(f"Expected one input argument to validate the dataset")
+
+        try:
+            dataset = kwargs[list(self.python_interface.inputs.keys())[0]]
```
kumare3, line 89: you don't have to do this check. this is done automatically by flyte base tasks

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,166 @@
...
+                        + every_result["expectation_config"]["expectation_type"]
+                        + "\n"
+                    )
+
+            # print column names for which the validation failed
+            print(f"\nCOLUMN\t\tFAILED EXPECTATION")
+            print(result_string)
+
+            # raise a GE exception
+            raise ValidationError(f"Validation failed!")
```
kumare3, line 164: Lets raise the validation error and add the entire failure information and `result_string` to the error message. that way. Flytekit will capture it and you will be able to see it in the UI

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,166 @@
...
+        expectation_suite: str,
+        data_connector: str,
+        inputs: typing.Dict[str, typing.Type],
+        checkpoint_params: typing.Optional[typing.Dict[str, typing.Union[str, typing.List[str]]]] = None,
+        task_config: BatchRequestConfig = None,
+        data_context: str = "./great_expectations",
+        **kwargs,
+    ) -> typing.Dict[str, str]:
+
+        _TASK_TYPE = "greatexpectations"
```
kumare3, line 66: make it a class var

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
kumare3: also do you only want one argument?

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,166 @@
...
+                    )
+
+            # print column names for which the validation failed
+            print(f"\nCOLUMN\t\tFAILED EXPECTATION")
+            print(result_string)
+
+            # raise a GE exception
+            raise ValidationError(f"Validation failed!")
+
+        print(f"Validation succeeded!")
```
kumare3, line 166: lets use the logging module

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,166 @@
...
+        # identify every run uniquely
+        run_id = RunIdentifier(
+            **{
+                "run_name": self._data_source + "_run",
+                "run_time": datetime.datetime.utcnow(),
+            }
+        )
+
+        checkpoint_result = checkpoint.run(run_id=run_id)
+        final_result = convert_to_json_serializable(checkpoint_result.list_validation_results())[0]
```
kumare3, line 146: can we not send this as an output of the task?

plugins/tests/greatexpectations/test_task.py
```diff
@@ -0,0 +1,133 @@
...
+                "pipeline_stage": "validation",
+            },
+        ),
+    )
+
+    task_object.execute(dataset="movies.sqlite")
+
+
+def test_ge_validation_within_task():
+    task_object = GETask(
```
kumare3, line 92: better name?

plugins/tests/greatexpectations/test_task.py
```diff
@@ -0,0 +1,133 @@
...
+        name="test4",
+        data_source="data",
+        inputs=kwtypes(dataset=str),
+        expectation_suite="test.demo",
+        data_connector="data_example_data_connector",
+    )
+
+    @task
+    def my_task(csv_file: str) -> int:
+        task_object(dataset=csv_file)
```
kumare3, line 102: isn't this weird? I think to support this, we should add a new type I feel and have a type transformer for this?
That way this can be more integrated into Flyte.
I have an idea for this

What if we create a new Type like this

class GETypeTransformer(GEType):
    def ...
In this type transformer we do the checking etc

and we can create the TypeTransformer for a specific type using a wrapper method

def expect(t: Type, expectation_sute) -> GEType:
     return GEType(t=t, expectation_sute=expectation_sute)

class GEType(Generic[T, SuitePath]):
   ...

kumare3: Thus you can update your examples as follows
```
@task
def my_task(csv_file: GEType[str, "test.demo"]) -> int:
    df = pd.read_csv(os.path.join("data", csv_file))
    return df.shape[0]
```

OR we could also do delayed eval
```
@task
def my_task(csv_file_e: GEType[str, "test.demo"]) -> int:
    # evaluate now
    csv_file.eval()
    # Get data
    csv_file = csv_file_e.get()
    df = pd.read_csv(os.path.join("data", csv_file))
    return df.shape[0]
```

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
samhita-alla: Yes, only one.

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
samhita-alla: We can send it.

plugins/tests/greatexpectations/test_task.py
samhita-alla: Does this look okay?
```python
def test_ge_validation_within_task():
    ge_config = GEConfig(
        data_source="data",
        expectation_suite="test.demo",
        data_connector="data_example_data_connector",
    )

    @task
    def populate_data(csv_file: str) -> GEType:
        return GEType(dataset=csv_file, config=ge_config)

    @task
    def my_task(csv_file: GEType) -> int:
        df = pd.read_csv(os.path.join("data", csv_file))
        return df.shape[0]

    @workflow
    def valid_wf(dataset: str = "yellow_tripdata_sample_2019-01.csv") -> int:
        ge_obj = populate_data(csv_file=dataset)
        return my_task(csv_file=ge_obj)

    valid_result = valid_wf()
    assert valid_result == 10000
```

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
kumare3: is the argument always datasst

samhita-alla: Yes, it is.

plugins/tests/greatexpectations/test_schema.py
```diff
@@ -0,0 +1,176 @@
...
+        expectation_suite="test.demo",
+        data_connector="data_example_data_connector",
+    )
+
+    @task
+    def get_length(dataframe: pd.DataFrame) -> int:
+        return dataframe.shape[0]
+
+    @task
+    def my_task(csv_file: GEType[ge_config]) -> pd.DataFrame:
```
kumare3, line 25: is it that great expectations is limited to csv files or other types of files only?
Can we also use primitive types?

Maybe this is a question for Abe.

samhita-alla: I wouldn't say it's restricted to files. For suppose, take MySQL. Here, GE expects a table name. dataset pertains to `data_asset_name` in GE which is always a string. 

Check this out: https://docs.greatexpectations.io/en/latest/autoapi/great_expectations/core/batch/index.html?highlight=data_asset_name#great_expectations.core.batch.BatchRequest

samhita-alla:
@kumare3 Added support for FlyteSchema and FlyteFile. Please review.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
```diff
@@ -0,0 +1,294 @@
...
+from flytekit.models import types as _type_models
+from flytekit.models.literals import Literal, Primitive, Scalar
+from flytekit.models.types import LiteralType
+from flytekit.types.file.file import FlyteFile, FlyteFilePathTransformer
+from flytekit.types.schema.types import FlyteSchema, FlyteSchemaTransformer, SchemaOpenMode
+
+
+@dataclass_json
+@dataclass
+class BatchConfig(object):
```
talagluck, line 26: Could you explain the difference between this and `BatchRequestConfig` in `task.py`?

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,239 @@
...
+        inputs: inputs to pass to the execute() method
+        local_file_path: dataset file path useful for FlyteFile and FlyteSchema
+        checkpoint_params: optional SimpleCheckpoint parameters
+        task_config: batchrequest config
+        data_context: directory in which GE's configuration resides
+
+    TODO: Connect Data Docs to Flyte Console.
+    """
+
+    _TASK_TYPE = "greatexpectations"
```
talagluck, line 62: Could we amend this to `great_expectations`?

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
```diff
@@ -0,0 +1,239 @@
...
+        **kwargs,
+    ) -> typing.Dict[str, str]:
+
+        self._data_source = data_source
+        self._data_connector = data_connector
+        self._expectation_suite = expectation_suite
+        self._batch_request = task_config
+        self._data_context = data_context
+        """
+        local_file_path is a must in two scenrios:
```
talagluck, line 84:
```suggestion
        local_file_path is a must in two scenarios:
```

plugins/greatexpectations/setup.py
```diff
@@ -0,0 +1,34 @@
+from setuptools import setup
+
+PLUGIN_NAME = "greatexpectations"
```
talagluck, line 3: Could we update this to `great_expectations` as well?

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
```diff
@@ -0,0 +1,294 @@
...
+    data_connector: str
+    """
+    local_file_path is a must in two scenrios:
+    * When using FlyteSchema
+    * When using FlyteFile for remote paths
+    This is because base directory which has the dataset file 'must' be given in GE's config file
+    """
+    local_file_path: str = None
+    checkpoint_params: typing.Optional[typing.Dict[str, typing.Union[str, typing.List[str]]]] = None
+    batchrequest_config: BatchConfig = None
```
talagluck, line 73: Should we rename this to `batch_config` to match the type?

talagluck (commented):
Thanks, @samhita-alla ! Overall this looks awesome. Just a few questions and a few stylistic suggestions from our end right now.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
```diff
@@ -0,0 +1,294 @@
...
+        batch_identifiers: identifiers to identify the data batch
+        batch_spec_passthrough: reader method if your file doesn't have an extension
+        limit: number of data points to fetch
+    """
+
+    data_connector_query: typing.Optional[typing.Dict[str, typing.Any]] = None
+    runtime_parameters: typing.Optional[typing.Dict[str, str]] = None
+    batch_identifiers: typing.Optional[typing.Dict[str, str]] = None
+    batch_spec_passthrough: typing.Optional[typing.Dict[str, typing.Any]] = None
+    limit: typing.Optional[int] = None
```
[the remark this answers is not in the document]
samhita-alla: 1. `limit`: Ah! You're right. I was of the opinion that a user might want to use `limit` without a `data_connector_query`; used it henceforth. But I don't think there's a need to, hence modified it now!
2. I think what you're referring to is this code snippet:
```python
def test_ge_schema_batchrequest_pandas_config():
    @task
    def my_task(
        directory: GEType[
            str,
            GEConfig(
                data_source="data",
                expectation_suite="test.demo",
                data_connector="my_data_connector",
                batchrequest_config=BatchConfig(
                    data_connector_query={
                        "batch_filter_parameters": {
                            "year": "2019",
                            "month": "01",  # noqa: F722
                        },
                        "limit": 10,
                    },
                ),
            ),
        ]
    ) -> str:
        return directory

    @workflow
    def my_wf():
        my_task(directory="my_assets")

    my_wf()
```
This is present in `test_schema.py` file. Will this work?

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+@dataclass_json
+@dataclass
+class GEConfig(object):
```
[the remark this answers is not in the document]
samhita-alla: Made the required changes.
- GEType is now `GreatExpectationsType`.
- GEConfig is now `GreatExpectationsFlyteConfig`.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
```diff
@@ -0,0 +1,294 @@
...
+    """
+    local_file_path is a must in two scenrios:
+    * When using FlyteSchema
+    * When using FlyteFile for remote paths
+    This is because base directory which has the dataset file 'must' be given in GE's config file
+    """
+    local_file_path: str = None
+    checkpoint_params: typing.Optional[typing.Dict[str, typing.Union[str, typing.List[str]]]] = None
+    batchrequest_config: BatchConfig = None
+    data_context: str = "./great_expectations"
```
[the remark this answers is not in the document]
samhita-alla: 1. Modified `data_context` to `context_root_dir`.
2. The `context_root_dir` has to be configurable so as to allow the users to input their own directory, which may not always be present in the current directory. The one that's already present in the code is configurable. I'm just using a default value.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
samhita-alla: My bad! Removed `BatchConfig` in `schema.py`. Now using `BatchRequestConfig` in both the files.

plugins/greatexpectations/flytekitplugins/greatexpectations/task.py
samhita-alla: Done.

plugins/greatexpectations/setup.py
samhita-alla: Done.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
samhita-alla: It's now `BatchRequestConfig`.

samhita-alla:
> Thanks, @samhita-alla ! Overall this looks awesome. Just a few questions and a few stylistic suggestions from our end right now.

@talagluck: Thanks for reviewing it! When you get the time, check the Great Expectations' examples I've written -- https://github.com/flyteorg/flytesnacks/pull/332.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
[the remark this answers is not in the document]
talagluck: This is one of the ways to use it, but it's possible to pass in a `custom_filter_function` to `data_connector_query` instead of keyworded `batch_filter_parameters` as on line 454 in the test shared above. I don't know that this is strictly necessary to include here, but wanted to mention it in case you thought it might be useful.

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
talagluck: Thanks so much for all the changes, @samhita-alla - looks great ! We'll review the Great Expectations example later today or Monday of next week.

samhita-alla: Awesome!

plugins/greatexpectations/flytekitplugins/greatexpectations/schema.py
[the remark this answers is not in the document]
samhita-alla: @talagluck: I tried running this test:
```python
def test_ge_batchrequest_custom_filter():
    def my_custom_batch_selector(batch_identifiers: dict) -> bool:
        return int(batch_identifiers["year"]) == "2019" and int(batch_identifiers["month"]) == "01"

    task_object = GreatExpectationsTask(
        name="test19",
        data_source="data",
        inputs=kwtypes(data=str),
        expectation_suite="test.demo",
        data_connector="my_data_connector",
        task_config=BatchRequestConfig(
            data_connector_query={
                "custom_filter_function": my_custom_batch_selector,
            },
        ),
    )

    task_object(data="my_assets")
```
which resulted in an error: 

```
../../../flytekit/core/base_task.py:298: in __call__
    return self.execute(**kwargs)
../../flytekit-greatexpectations/flytekitplugins/great_expectations/task.py:209: in execute
    checkpoint = SimpleCheckpoint(f"_tmp_checkpoint_{self._expectation_suite}", context, **checkpoint_config)
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/checkpoint/checkpoint.py:664: in __init__
    checkpoint_config: CheckpointConfig = self._configurator_class(
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/checkpoint/configurator.py:114: in build
    return self._build_checkpoint_config()
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/checkpoint/configurator.py:142: in _build_checkpoint_config
    f"SimpleCheckpointConfigurator built this CheckpointConfig:"
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/data_context/types/base.py:116: in to_json_dict
    return convert_to_json_serializable(data=commented_map)
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/core/util.py:144: in convert_to_json_serializable
    new_dict[str(key)] = convert_to_json_serializable(data[key])
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/core/util.py:151: in convert_to_json_serializable
    new_list.append(convert_to_json_serializable(val))
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/core/util.py:144: in convert_to_json_serializable
    new_dict[str(key)] = convert_to_json_serializable(data[key])
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/core/util.py:144: in convert_to_json_serializable
    new_dict[str(key)] = convert_to_json_serializable(data[key])
../../../../../.pyenv/versions/3.8.10/envs/ge/lib/python3.8/site-packages/great_expectations/core/util.py:144: in convert_to_json_serializable
    new_dict[str(key)] = convert_to_json_serializable(data[key])
TypeError: <function test_ge_batchrequest_custom_filter.<locals>.my_custom_batch_selector at 0x108ae9280> is of type function which cannot be serialized.
```
Is `custom_filter_function` not supported within `SimpleCheckpoint` class? Moreover, am I using `custom_filter_function` the right way?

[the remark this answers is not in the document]
talagluck: Hi @samhita-alla, it does appear that you are technically using `custom_filter_function` in the right way, though I suspect that you are right, and `custom_filter_functions` are not supported within `SimpleCheckpoint`.  I'm going to dig a little deeper and open an issue with the team, but in the meantime, I think it's fair to not use this example for right now.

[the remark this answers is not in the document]
samhita-alla: Thank you for the confirmation, @talagluck. We can add the test once the issue is fixed.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
```diff
@@ -0,0 +1,273 @@
...
+    Use this configuration to configure GreatExpectations Plugin.
+
+    Args:
+        data_source: tell where your data lives and how to get it
+        expectation_suite: suite which consists of the data expectations
+        data_connector: connector to identify data batches
+        local_file_path: dataset file path useful for FlyteFile and FlyteSchema
+        checkpoint_params: optional SimpleCheckpoint parameters
+        batchrequest_config: batchrequest config
+        context_root_dir: directory in which GreatExpectations' configuration resides
```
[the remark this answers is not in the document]
samhita-alla: Modified the names.
 
`data_source` to `datasource_name`
`expectation_suite` to `expectation_suite_name`
`data_connector` to `data_connector_name`
`batchrequest_config` to `batch_request_config`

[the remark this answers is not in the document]
talagluck: Great - thank you!

samhita-alla:
https://github.com/flyteorg/flytekit/pull/559 needs to be updated with the GE plugin to change `ctx.file_access.download_directory(x, y)` to `ctx.file_access.get_data(x, y, is_multipart=True)`, if this PR gets merged before #559.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
```diff
@@ -0,0 +1,243 @@
...
+
+        # identify every run uniquely
+        run_id = RunIdentifier(
+            **{
+                "run_name": self._datasource_name + "_run",
+                "run_time": datetime.datetime.utcnow(),
+            }
+        )
+
+        checkpoint_result = checkpoint.run(run_id=run_id)
```
samhita-alla, line 224: @talagluck: When trying to validate data against expectations that's bound to fail, I see the following error at this particular line:
```bash
 File "/Users/samhitaalla/.pyenv/versions/3.8.10/lib/python3.8/site-packages/flytekitplugins/great_expectations/task.py", line 224, in execute
    checkpoint_result = checkpoint.run(run_id=run_id)
  File "/Users/samhitaalla/.pyenv/versions/3.8.10/lib/python3.8/site-packages/great_expectations/checkpoint/checkpoint.py", line 742, in run
    return super().run(
  File "/Users/samhitaalla/.pyenv/versions/3.8.10/lib/python3.8/site-packages/great_expectations/checkpoint/checkpoint.py", line 274, in run
    validator: Validator = self.data_context.get_validator(
  File "/Users/samhitaalla/.pyenv/versions/3.8.10/lib/python3.8/site-packages/great_expectations/data_context/data_context.py", line 1778, in get_validator
    batch_definition = batch_list[-1].batch_definition
IndexError: list index out of range
```

Has anything changed in the recent versions? I remember this particular line not throwing an error even if the validation fails. It returns a dictionary that has "success" equal to True or False (let me know if I'm wrong). 

All in all, is this specific error expected?

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
```diff
@@ -0,0 +1,256 @@
...
+    """
+    This task can be used to validate your data.
+
+    Args:
+        name: name of the task
+        datasource_name: tell where your data lives and how to get it
+        expectation_suite_name: suite which consists of the data expectations
+        data_connector_name: connector to identify data batches
+        inputs: inputs to pass to the execute() method
+        is_runtime: do you want to create data batches at runtime?
```
talagluck, line 52: Hi @samhita-alla - I have mixed feelings about this `is_runtime` argument. I interpreted the main purpose of the `BatchRequestConfig` class as a funnel for taking in both RuntimeBatchRequests and BatchRequests, so that the user wouldn't need to think about this. But if we are requiring the user to specify `is_runtime` anyway, I wonder if it makes sense for the user to pass in a typed GreatExpectations `BatchRequest` or `RuntimeBatchRequest`, and then just test which one it is. I think that might make the interface more straightforward. What do you think?

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
```diff
@@ -0,0 +1,291 @@
...
+            dataset = temp_dataset
+
+        batch_request_conf = ge_conf.batch_request_config
+
+        # fetch the data context
+        context = ge.data_context.DataContext(ge_conf.context_root_dir)
+
+        # minimalistic batch request
+        final_batch_request = {
+            "data_asset_name": "random_string" if ge_conf.is_runtime else dataset,
```
talagluck, line 206: Could we enable the user to pass this in to maintain the utility of `"data_asset_name"` instead of hard-coding it is "random_string"?

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
samhita-alla: Hi, @talagluck. Even now, `BatchRequestConfig` class is a funnel for taking in both `RuntimeBatchRequest` and `BatchRequest`. However, since it isn't possible to determine if the request is run-time or not automatically, we need the user's input. If `is_runtime` is set to `True`, the plugin would consider the user-given dataset as the `batch_data` or the `query`. The user needn't give one separately in such cases. Let me explain the use-case, w.r.t Flyte.

In Flyte, we enclose all our code in tasks and workflows. Great Expectations in Flyte can be a task or a type. If we have to feed input, we do that through the `inputs` argument in the **task** (line 51 in task.py), and the dataset in itself is considered the input in **type**. Say, as you suggested, we use Great Expectations' builtin `RuntimeBatchRequest`, the user then has to pass input (`batch_data` or `query`), which cannot be done this way. It can only be initialized and sent from outside, which wouldn't be within the scope of Flyte's entities, and henceforth, wouldn't respect Flyte's data flow. 

For example, this is how it's being done now wherein the user needn't pay attention to setting the runtime parameter:

```python
runtime_task_obj = GreatExpectationsTask(
    name="greatexpectations.task.runtime",
    datasource_name="my_pandas_datasource",
    inputs=kwtypes(dataframe=FlyteSchema),
    expectation_suite_name="test.demo",
    data_connector_name="my_runtime_data_connector",
    is_runtime=True,
    task_config=BatchRequestConfig(
        batch_identifiers={
            "pipeline_stage": "validation",
        },
    ),
    context_root_dir=CONTEXT_ROOT_DIR,
)

@workflow
def runtime_wf(dataset: str = DATASET_LOCAL) -> None:
    dataframe = ...(csv_file=dataset)
    runtime_task_obj(dataframe=dataframe)
```

The plugin sets the `dataframe` as `batch_data`, which otherwise needs to be initialized outside of the workflow and set within.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
samhita-alla: `data_asset_name` is not shown on the user-facing side. It's palpable only within the plugin code. When the user sends a dataset that needs to be validated, the plugin populates either `data_asset_name` or `batch_data`/`query`. I've used "random_string" because `data_asset_name` doesn't have any real meaning when using a `RuntimeBatchRequest`. Giving a separate `data_asset_name` can be confusing; what do you think?

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
talagluck: Thank you for this explanation, @samhita-alla - I think I get it now. I see now that at the time the user defines the `runtime_task_obj`, they are not ready to create the BatchRequest, so they are creating a sort of pre-BatchRequest that is completed and instantiated when the task is actually run.

My only suggestion here would be to consider parsing the Great Expectations config to determine the type of the `data_connector` instead of requiring the specification of an additional argument. As a very rough example, you could do something like this:

```python
specified_datasource = "my_pandas_datasource"
specified_data_connector = "my_runtime_data_connector"

[my_datasource] = [i for i in context.list_datasources() if i.get("name")==specified_datasource]

data_connector_class_lookup = {data_connector_name: data_connector_class.get("class_name") for data_connector_name, data_connector_class in my_datasource[0].get("data_connectors").items()}

specified_data_connector_class = data_connector_class_lookup.get(specified_data_connector)

if specified_data_connector_class = "RuntimeDataConnector":
...

```

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
talagluck: Hi @samhita-alla - `data_asset_name` is shown when generating a data docs site. This helps users to track what data_asset was used for validation. You can see an example here:
Screen Shot 2021-08-09 at 11 10 05 AM

If this is hard-coded, then all results will look the same, which could be confusing for users. The idea is that when users validate the same data asset multiple times (like monthly sales data for instance), they would use the same `data_asset_name`, and when they validate new data (like monthly user data), they would use a new `data_asset_name`. This isn't something that we can infer, but it is also something that improves the tracking of validation results over time, which is why we require user input.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
samhita-alla: Thank you! Made the required changes.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
samhita-alla: @talagluck: I've added a `data_asset_name` parameter.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
```diff
@@ -144,7 +166,7 @@ def to_python_value(
         if lv.scalar.schema:
 
             # if data batch is to be genearated, skip copying the parquet file
```
talagluck, line 168:
```suggestion
            # if data batch is to be generated, skip copying the parquet file
```

talagluck:
Thanks for these updates, @samhita-alla! This looks great.

talagluck (dismissed):

samhita-alla:
Just a heads up—pinned the Great Expectations version to 0.13.23. If it's the latest, `test_flytekit_sagemaker_runner.py` is throwing an error:

```bash
_______ ERROR collecting tests/scripts/test_flytekit_sagemaker_runner.py _______ 
ImportError while importing test module '/home/runner/work/flytekit/flytekit/tests/scripts/test_flytekit_sagemaker_runner.py'. 
Hint: make sure your test modules/packages have valid Python names. 

Traceback: 
/opt/hostedtoolcache/Python/3.7.11/x64/lib/python3.7/importlib/__init__.py:127: in import_module 
return _bootstrap._gcd_import(name[level:], package, level) 
tests/scripts/test_flytekit_sagemaker_runner.py:5: in <module> 
from scripts.flytekit_sagemaker_runner import run as _flyte_sagemaker_run 
E ModuleNotFoundError: No module named 'scripts.flytekit_sagemaker_runner'
```

talagluck:
Hi @samhita-alla - thanks for letting us know. That error doesn't look like it would be due to Great Expectations. Do you have any ideas about why it might be occurring?

I think pinning will be ok temporarily, but we're in the process of rolling out some important fixes and powerful features, so it would be good to unpin in the near future.

samhita-alla:
> Hi @samhita-alla - thanks for letting us know. That error doesn't look like it would be due to Great Expectations. Do you have any ideas about why it might be occurring?
> 
> I think pinning will be ok temporarily, but we're in the process of rolling out some important fixes and powerful features, so it would be good to unpin in the near future.

I'm not sure as to why it's happening. We will have to dig deeper to find that out. I'm sure the problem is on our end.
@wild-endeavor Any thoughts about the test error?

eapolinario:
I was able to repro this. This was caused by https://github.com/great-expectations/great_expectations/pull/3003/, which ends up defining a package called `scripts` which takes precedence during package resolution.

I opened https://github.com/flyteorg/flytekit/pull/598 to fix the issue.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
```diff
@@ -0,0 +1,310 @@
...
+        expected: LiteralType,
+    ) -> Literal:
+        datatype = GreatExpectationsTypeTransformer.get_config(python_type)[0]
+
+        if issubclass(datatype, FlyteSchema):
+            return FlyteSchemaTransformer().to_literal(ctx, python_val, datatype, expected)
+        elif issubclass(datatype, FlyteFile):
+            return FlyteFilePathTransformer().to_literal(ctx, python_val, datatype, expected)
+
+        return Literal(scalar=Scalar(primitive=Primitive(string_value=python_val)))
```
kumare3, line 121: what about dictionaries or custom objects etc? I think we should raise an error if the type does not match?
Else we should forward the type lookup to the TypeEngine?

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
```diff
@@ -0,0 +1,275 @@
...
+
+    data_connector_query: Optional[Dict[str, Any]] = None
+    runtime_parameters: Optional[Dict[str, Any]] = None
+    batch_identifiers: Optional[Dict[str, str]] = None
+    batch_spec_passthrough: Optional[Dict[str, Any]] = None
+
+
+class GreatExpectationsTask(PythonInstanceTask[BatchRequestConfig]):
+    """
+    This task can be used to validate your data.
```
kumare3, line 44: lets write a little more doc here to explain - usecase or when can you use it

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
```diff
@@ -0,0 +1,275 @@
...
+
+        super(GreatExpectationsTask, self).__init__(
+            name=name,
+            task_config=task_config,
+            task_type=self._TASK_TYPE,
+            interface=Interface(inputs=inputs, outputs=outputs),
+            **kwargs,
+        )
+
+    def execute(self, **kwargs) -> Any:
```
kumare3, line 104: I think the method is too long, can we split the various `if` conditions into sub-functions. Makes it easier to read

plugins/tests/greatexpectations/test_schema.py
```diff
@@ -0,0 +1,355 @@
...
+
+def test_ge_local_flytefile():
+    ge_config = GreatExpectationsFlyteConfig(
+        datasource_name="data",
+        expectation_suite_name="test.demo",
+        data_connector_name="data_flytetype_data_connector",
+    )
+
+    @task
+    def my_task(dataset: GreatExpectationsType[FlyteFile[typing.TypeVar("csv")], ge_config]) -> int:
```
kumare3, line 328: can we add this as a predefined type here - https://github.com/flyteorg/flytekit/blob/master/flytekit/types/file/__init__.py#L31

kumare3:
@samhita-alla this is great, I have a few nits. we can handle them in 2 ways. I am ok with merging and then revisiting the nits, especially refactoring the extremely large functions and also handling other types - either with an error or supporting them

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/schema.py
samhita-alla: Today we support string, FlyteFile, and FlyteSchema datatypes. I don't think any other datatype fits into the Great Expectations ecosystem. Thus, I have added type-checking in this block of code.
I believe there's type checking included everywhere else. Please let me know if that's not the case.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
samhita-alla: I have added two more lines explaining the use case.

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
samhita-alla: I split FlyteFile and FlyteSchema into two methods in both task.py and schema.py files.

plugins/tests/greatexpectations/test_schema.py
samhita-alla: It's available now.

flytekit/types/file/__init__.py
```diff
@@ -79,3 +79,9 @@
     Can be used to receive or return an SVGImage. The underlying type is a FlyteFile, type. This is just a
     decoration and useful for attaching content type information with the file and automatically documenting code.
 """
+
+CSVFile = FlyteFile[typing.TypeVar("csv")]
```
kumare3, line 83: thank you!

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # minimalistic batch request
+        final_batch_request = {
+            "data_asset_name": self._data_asset_name if is_runtime else dataset,
+            "datasource_name": self._datasource_name,
+            "data_connector_name": self._data_connector_name,
+        }
+
+        # Great Expectations' RuntimeBatchRequest
+        if self._batch_request_config and (self._batch_request_config.runtime_parameters or is_runtime):
```
kumare3, line 211: is this always a json? not a typed structure?

kumare3 (approved):

plugins/flytekit-greatexpectations/flytekitplugins/great_expectations/task.py
samhita-alla: `_batch_request_config` is a dataclass, whereas `final_batch_request` is a dictionary.

PR #510, opened 2021-06-08
Support mapping over pod tasks

katrogan:

# TL;DR
Updates map task serialization to support non-container target type tasks (e.g. pod tasks).

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1051

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -15,7 +15,9 @@ from flytekit.core.python_auto_container import get_registerable_container_image
 from flytekit.core.python_function_task import PythonFunctionTask
 from flytekit.models.array_job import ArrayJob
 from flytekit.models.interface import Variable
-from flytekit.models.task import Container
+from flytekit.models.task import Container, K8sObjectMetadata, K8sPod
+
+_K8S_POD_TARGET_TASK_TYPES = ["sidecar"]
 
 
 class MapPythonTask(PythonTask):
@@ -84,6 +86,8 @@ class MapPythonTask(PythonTask):
         return container_args
 
     def get_container(self, settings: SerializationSettings) -> Container:
+        if self._run_task.task_type in _K8S_POD_TARGET_TASK_TYPES:
+            return None
         env = {**settings.env, **self.environment} if self.environment else settings.env
         return _get_container_definition(
             image=get_registerable_container_image(None, settings.image_config),
@@ -92,6 +96,13 @@ class MapPythonTask(PythonTask):
             data_loading_config=None,
             environment=env,
         )
+
+    def get_k8s_pod(self, settings: SerializationSettings) -> K8sPod:
+        if self._run_task.task_type not in _K8S_POD_TARGET_TASK_TYPES:
+            return None
+
+        self._run_task.set_command_fn(self.get_command)
+        return self._run_task.get_k8s_pod(settings)
 
     def get_custom(self, settings: SerializationSettings) -> Dict[str, Any]:
         return ArrayJob(parallelism=self._max_concurrency, min_success_ratio=self._min_success_ratio).to_dict()
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -15,7 +15,7 @@ from flytekit.core.python_auto_container import get_registerable_container_image
 from flytekit.core.python_function_task import PythonFunctionTask
 from flytekit.models.array_job import ArrayJob
 from flytekit.models.interface import Variable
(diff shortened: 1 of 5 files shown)


flytekit/core/map_task.py
```diff
@@ -93,9 +97,19 @@ def get_container(self, settings: SerializationSettings) -> Container:
             environment=env,
         )
 
+    def get_k8s_pod(self, settings: SerializationSettings) -> K8sPod:
+        if self._run_task.task_type not in _K8S_POD_TARGET_TASK_TYPES:
+            return None
+
+        self._run_task.set_command_fn(self.get_command)
```
katrogan, line 104: Struggled with the right way to do this. An alternative was serializing the run_task's k8s pod as is and updating the command with `pyflyte-map-execute` after the fact but then that requires the map task to be familiar with the serialized structure of each non-container target type task, which didn't feel right.

Another option to avoid the _K8S_POD_TARGET_TASK_TYPES set is to call the corresponding method, (e.g. `_run_task.get_k8s_pod(settings)` and then see if it's empty or not before returning in this method - but this seems (marginally) inefficient.

If you can think of a better way to do this, let me know

plugins/tests/pod/test_pod.py
```diff
@@ -278,3 +278,46 @@ def simple_pod_task(i: int):
     assert serialized.template.k8s_pod.metadata.labels == {"label": "foo"}
     assert serialized.template.k8s_pod.metadata.annotations == {"anno": "bar"}
     assert serialized.template.k8s_pod.pod_spec is not None
+
+
+def test_map_pod_task_serialization():
```
katrogan, line 283: it doesn't make complete sense to add this test here as opposed to [test_map_task](https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/core/test_map_task.py) in core/ - but then that introduces a plugin dependency in core tests and I'm not sure what the lesser evil is.

flytekit/core/map_task.py
kumare3: I think it is fair to call https://github.com/flyteorg/flytekit/blob/master/plugins/pod/flytekitplugins/pod/task.py#L103. This is because we now have K8sPodSpec as a target.
So ideally we should have one method per target. Or we could simply call it getTarget

kumare3: the only thing I would say is it should be at the base level. the base should have getContainer, getK8sPodSpec, getSQL, getHTTPAPI etc
and the semantics should be crisp and state that we will only invoke one of them.

Now the question is which one to invoke?

katrogan: I think for python proto serialization we can't just produce one getTarget method for the oneof field, unfortunately. So each task needs to implement each get_foo target option

re: invoking each core target and having each return None except for the applicable target - this sounds good. Refactored, PTAL @kumare3

flytekit/core/map_task.py
```diff
@@ -83,19 +82,31 @@ def get_command(self, settings: SerializationSettings) -> List[str]:
 
         return container_args
 
+    @contextmanager
+    def prepare_target(self):
+        """
+        Alters the underlying run_task command to modify it for map task execution and then resets it after.
```
kumare3, line 88: incorrect doc?

wild-endeavor (approved):

plugins/tests/pod/test_pod.py
kumare3: i completely get it, the thing is K8sPodSpec now is a core plugin.
Maybe the core task should support taking a yaml / json and the k8spod plugin can continue using the python interface to improve it. This way the flytekit remains devoid of k8s dependency

flytekit/core/map_task.py
katrogan: should be correct

plugins/tests/pod/test_pod.py
katrogan: hm not sure if i'm understanding correctly, but you mean have serialize() formulate the k8sPod target from a PodSpec struct? I'm not really a fan of making code more confusing (in this case, splitting up how we populate the k8sPod target across methods) just to make testing easier

kumare3 (approved):
As discussed, a follow up issue to model `tasktemplate.target` in flytekit.core

katrogan:
@kumare3 done: https://github.com/flyteorg/flyte/issues/1122

PR #514, opened 2021-06-11
Fix imperative unbound wf input checking

wild-endeavor:
# TL;DR
The imperative workflow checks for unused inputs, but it should only concern itself with Promise objects, not everything.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
* Also should be able to call `add_entity` on any `PythonTask` not just auto-container tasks.

## Tracking Issue

### DIFF
--- a/flytekit/core/workflow.py
+++ b/flytekit/core/workflow.py
@@ -505,7 +505,9 @@ class ImperativeWorkflow(WorkflowBase):
                 else:
                     return [input_value]
 
-            # Every time an entity is added, mark it as used.
+            # Every time an entity is added, mark it as used. The above function though will gather all the input
+            # values but we're only interested in the ones that are Promises so let's filter for those.
+            # There's probably a way to clean this up, maybe key off of the name instead of value?
             all_input_values = get_input_values(kwargs)
             for input_value in filter(lambda x: isinstance(x, Promise), all_input_values):
                 if input_value in self._unbound_inputs:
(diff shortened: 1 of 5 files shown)


flytekit/core/workflow.py
```diff
@@ -477,7 +478,7 @@ def execute(self, **kwargs):
             return get_promise(self.output_bindings[0].binding, intermediate_node_outputs)
         return tuple([get_promise(b.binding, intermediate_node_outputs) for b in self.output_bindings])
 
-    def add_entity(self, entity: PythonAutoContainerTask, **kwargs) -> Node:
+    def add_entity(self, entity: PythonTask, **kwargs) -> Node:
```
kumare3, line 481: i would have liked the name to be add_task
add_subwf
add_lp
and in turn all of them just call add_entity

flytekit/core/workflow.py
```diff
@@ -504,8 +505,11 @@ def get_input_values(input_value):
                 else:
                     return [input_value]
 
-            # Every time an entity is added, mark it as used.
-            for input_value in get_input_values(kwargs):
+            # Every time an entity is added, mark it as used. The above function though will gather all the input
+            # values but we're only interested in the ones that are Promises so let's filter for those.
+            # There's probably a way to clean this up, maybe key off of the name instead of value?
+            all_input_values = get_input_values(kwargs)
+            for input_value in filter(lambda x: isinstance(x, Promise), all_input_values):
```
kumare3, line 512: for non promises we need to add a binding right? is that what self._unbound_inputs means?

kumare3 (approved):

flytekit/core/workflow.py
wild-endeavor: i don't think so.  let me add a comment to clarify

kumare3 (approved):

PR #518, opened 2021-06-16
Early error for cases in which namedtuple usage causes compile time errors

kumare3:

# TL;DR
The following example is incorrect
```
    nm = typing.NamedTuple("OP", greet=str)

    @task
    def say_hello() -> nm:
        return nm("hello world")

    wf_outputs = typing.NamedTuple("OP2", greet1=str, greet2=str)

    @workflow
    def my_wf() -> wf_outputs:
        # Note only Namedtuples can be created like this
        return wf_outputs(say_hello(), say_hello())
```
This fails at registration time, because Flyte does not support tuples for individual values.

It can be fixed by simply adding `.greet` to the returned value from a task - that returns a namedtuple
```
    @workflow
    def my_wf() -> wf_outputs:
        # Note only Namedtuples can be created like this
        return wf_outputs(say_hello().greet, say_hello().greet)
```
This usually is ok for cases in which you have multiple values returned as the assumption is that you need to de-reference, but for single values, this is not directly obvious. Hence we will cause early failure

## Type
 - [x] Bug Fix - not a bug
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1117
https://github.com/flyteorg/flyte/issues/1139

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -228,6 +228,14 @@ class ComparisonExpression(object):
 
 
 class ConjunctionExpression(object):
+    """
+    A Conjunction Expression is an expression of the form either (A and B) or (A or B).
+    where A, B are two expressions (comparsion or conjunctions) and (and, or) are logical truth operators.
+
+    A conjunctionExpression evaluates to True or False depending on the logical operator and the truth values of
+    each of the expressions A & B
+    """
+
     def __init__(
         self,
         lhs: Union[ComparisonExpression, "ConjunctionExpression"],
@@ -420,7 +428,7 @@ class Promise(object):
 
 def create_native_named_tuple(
     ctx: FlyteContext, promises: Union[Promise, typing.List[Promise]], entity_interface: Interface
-):
+) -> Optional[Tuple]:
     """
     Creates and returns a Named tuple with all variables that match the expected named outputs. this makes
     it possible to run things locally and expect a more native behavior, i.e. address elements of a named tuple
@@ -435,6 +443,9 @@ def create_native_named_tuple(
     if isinstance(promises, Promise):
         v = [v for k, v in entity_interface.outputs.items()][0]  # get output native type
         return TypeEngine.to_python_value(ctx, promises.val, v)
+
+    if len(promises) == 0:
+        return None
 
     named_tuple_name = "DefaultNamedTupleOutput"
     if entity_interface.output_tuple_name:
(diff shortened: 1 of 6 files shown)


tests/flytekit/unit/core/test_serialization.py
```diff
@@ -368,13 +370,51 @@ def parent_wf() -> (int, int, int, int):
 
 def test_serialization_named_outputs_single():
     @task
-    def t1() -> str:
+    def t1() -> typing.NamedTuple("OP", a=str):
         return "Hello"
```
wild-endeavor, line 374: so here we're just wrapping it for them?  not raising an error not forcing them to return `("Hello",)`?

wild-endeavor (approved):

tests/flytekit/unit/core/test_serialization.py
kumare3: Yup

flytekit/core/promise.py
```diff
@@ -411,6 +418,42 @@ def __str__(self):
...
+    for p in promises:
+        if not isinstance(p, Promise):
+            raise AssertionError(
+                "Workflow outputs can only be promises that are returned by tasks. Found a value of"
+                f"type {type(p)}. Workflows cannot return local variables or constants."
+            )
+        outputs[p.var] = TypeEngine.to_python_value(ctx, p.val, entity_interface.outputs[p.var])
+
+    # Should this class be part of the Interface?
+    t = collections.namedtuple(named_tuple_name, list(outputs.keys()))
```
katrogan, line 453: does ordering of the keys matter here?

kumare3: No

kumare3: but it is actually ordered

katrogan (dismissed):

wild-endeavor (approved):

tests/flytekit/unit/core/test_type_hints.py
```diff
@@ -240,11 +240,13 @@ def my_wf(a: int, b: str) -> (int, str):
...
-    def my_wf3(a: int, b: str) -> int:
-        return (a,)
+    with pytest.raises(AssertionError):
+
+        @workflow
+        def my_wf3(a: int, b: str) -> int:
+            return (a,)
 
-    my_wf3(a=10, b="hello")
+    # my_wf3(a=10, b="hello")
```
cosmicBboy, line 249: does this need to be uncommented?

kumare3: aah should be deleted

cosmicBboy (approved):

wild-endeavor (approved):

PR #522, opened 2021-06-24
SQLAlchemy migration

samhita-alla:

# TL;DR
Replace the SQLAlchemy implementation with one that leverages the task template/flytekit-only plugin construct.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


cosmicBboy:
`make fmt` to fix lint error: https://github.com/flyteorg/flytekit/pull/522/checks?check_run_id=2903924944

kumare3:
@samhita-alla you will need to publish a docker container too, that will be default for this task type

cosmicBboy (dismissed):
LGTM, tho should probably get another review from someone more familiar with this code @kumare3 @wild-endeavor @katrogan

samhita-alla:
> @samhita-alla you will need to publish a docker container too, that will be default for this task type

Does this look good -- https://github.com/flyteorg/flytekit/blob/81962db3ddeef0ffee2fa06553d26b09b6a7b787/flytekit/extras/sqlalchemy/Dockerfile?
I'm not really sure about the correctness, let me know if there are any changes required.
For now, I published the docker image on ECR.

plugins/setup.py
```diff
@@ -16,7 +16,6 @@
     "flytekitplugins-kftensorflow": "kftensorflow",
     "flytekitplugins-pandera": "pandera",
     "flytekitplugins-dolt": "flytekit-dolt",
-    "flytekitplugins-sqlalchemy": "flytekitplugins-sqlalchemy",
```
wild-endeavor, line 19: move this back?

flytekit/extras/sqlalchemy/task.py
```diff
@@ -54,13 +57,12 @@ def __init__(
...
         output_schema = output_schema_type if output_schema_type else FlyteSchema
         outputs = kwtypes(results=output_schema)
-        self._uri = task_config.uri
-        self._connect_args = task_config.connect_args or {}
-        self._secret_connect_args = task_config.secret_connect_args
 
         super().__init__(
             name=name,
             task_config=task_config,
+            container_image="public.ecr.aws/nuclyde/sqlalchemy:ce755dd8886cd66af1821051682a5b0181bc3626",
```
wild-endeavor, line 64: Let's use ghcr. Looks much better I think.

samhita-alla: Can you give me access to push an image to ghcr? @kumare3

samhita-alla:
@wild-endeavor Please review it now. I've made the required changes.

plugins/flytekit-sqlalchemy/flytekitplugins/sqlalchemy/task.py
```diff
@@ -54,13 +57,12 @@ def __init__(
...
         output_schema = output_schema_type if output_schema_type else FlyteSchema
         outputs = kwtypes(results=output_schema)
-        self._uri = task_config.uri
-        self._connect_args = task_config.connect_args or {}
-        self._secret_connect_args = task_config.secret_connect_args
 
         super().__init__(
             name=name,
             task_config=task_config,
+            container_image="ghcr.io/flyteorg/flytekit:sqlalchemy-latest",
```
wild-endeavor, line 64: We probably don't want to use `latest` here right?  Let's say someone has version v0.21 of flytekit, and a task registered with it, and then we update code in the sqlalchemy plugin, and then we release it as flytekit v0.22.  Upon publishing that version, GitHub workflows will build a new sqlalchemy image, and all of a sudden the person who had the existing task would pick up the new image, and behavior may change.

I think it's cleaner to hard code the version.  We can try hard coding the image to the tag that we plan on releasing the image under.  Also kind of weird, but I think it's fine for now.

samhita-alla: Modified the version name to v0.1

samhita-alla: @kumare3 / @wild-endeavor

wild-endeavor: Let's not use 0.1.0.  So currently, the thinking is that since plugins are in the same repo as core flytekit code, all the plugins have the same version as core flytekit.  (one GH repo means one version, which I think is quite reasonable.)  So sqlalchemy got published as v0.20.0 recently (see https://pypi.org/search/?q=flytekitplugins).  I think the image version should be the same as the pypi version to avoid confusion.

So basically we should pick the version in the image name to be what we want the release to be.  So if we merge this pr and release into v0.20.1 then that's what the version of the image should be.  I feel like this is a pretty big change to be a patch version, but that's okay.  No one's really using this task anyways, it's just for demo purposes.

samhita-alla: Modified the version to v0.20.0

.github/workflows/sqlachemy_publish.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          fetch-depth: "0"
+      - name: Push SQLAlchemy Image to GitHub Registry
+        uses: whoan/docker-build-with-cache-action@v5
+        with:
+          # https://docs.github.com/en/packages/learn-github-packages/publishing-a-package
+          username: "${{ secrets.FLYTE_BOT_USERNAME }}"
+          password: "${{ secrets.FLYTE_BOT_PAT }}"
+          image_name: ${{ github.repository_owner }}/flytekit
+          image_tag: sqlalchemy-v0.20.1,sqlalchemy-${{ github.sha }},sqlalchemy-${{ github.event.ref }}
```
wild-endeavor, line 27: Sorry i missed this.  we shouldn't hardcode the version here.  We don't want to accidentally override the existing image if we forget to change this value.  The release sha should be good enough right?

wild-endeavor (approved):

PR #527, opened 2021-06-26
Refactor node creation from entities

jeevb:
# TL;DR
Allows for passing of values of `interruptible` into the `NodeMetadata` from `TaskMetadata` and `WorkflowMetadataDefaults` respectively.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
 Created a `SupportsNodeCreation` protocol that if implemented, will allow creation of a node from a given entity. This protocol requires a `construct_node_metadata` method, that will allow tasks, workflows, launch plans, and reference entities to define the appropriate `NodeMetadata` object.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1179

## Follow-up issue
_NA_
OR


jeevb:
Smoke tested with:
1. No values set for `interruptible` at the task or workflow level - task is not interruptible
2. Task has `interruptible=True`, but workflow has no value set - task is interruptible
3. Workflow has `interruptible=True`, but task has no value set - task is interruptible
4. Workflow has `interruptible=False`, but task has `interruptible=True` - task is interruptible
5. Workflow has `interruptible=True`, but task has `interruptible=False` - task is not interruptible

jeevb:
Note that the defaults for `interruptible` were only updated to `None` for new-style classes.

flytekit/core/python_function_task.py
```diff
@@ -146,7 +146,7 @@ def compile_into_workflow(
             from flytekit.common.translator import get_serializable
 
             workflow_metadata = WorkflowMetadata(on_failure=WorkflowFailurePolicy.FAIL_IMMEDIATELY)
-            defaults = WorkflowMetadataDefaults(interruptible=False)
+            defaults = WorkflowMetadataDefaults(interruptible=self.metadata.interruptible)
```
jeevb, line 149: @wild-endeavor: This was a break from the existing logic. Does this look right?

flytekit/core/base_task.py
```diff
@@ -63,7 +64,7 @@ class TaskMetadata(object):
 
     cache: bool = False
     cache_version: str = ""
-    interruptable: bool = False
+    interruptible: Optional[bool] = None
```
katrogan, line 67: thank you for fixing this spelling!

flytekit/core/launch_plan.py
```diff
@@ -305,7 +309,7 @@ def __call__(self, *args, **kwargs):
         if ctx.compilation_state is not None:
             inputs = self.saved_inputs
             inputs.update(kwargs)
-            return create_and_link_node(ctx, entity=self, interface=self.workflow.python_interface, **inputs)
+            return create_and_link_node(ctx, entity=self, **inputs)
```
katrogan, line 312: is it safe to remove the interface arg here? if a launch plan has fixed inputs does it always have the same interface as the reference workflow?

jeevb: Good question @katrogan! `create_and_link_node` currently infers the interface from `entity.python_interface` And the output of `LaunchPlan.python_interface` is just `self.workflow.python_interface`, so at least the logic should be exactly the same. If we foresee a difference between an entity's `python_interface` attribute, and the interface we need to use to create the `Node`, then we can just define a different attribute in the protocol. Afaik, that's not necessary yet, but I'll defer to y'all!

flytekit/core/workflow.py
```diff
@@ -88,11 +88,12 @@ class WorkflowMetadataDefaults(object):
     WorkflowMetadata represents metadata about the workflow itself.
     """
 
-    interruptible: bool
+    interruptible: Optional[bool] = None
 
     def __post_init__(self):
-        if self.interruptible is not True and self.interruptible is not False:
```
wild-endeavor, line 94: Yah I think this we want to keep as this strict check because of the IDL link earlier.  According to @katrogan - the reason the `WorkflowMetadataDefaults`'s `interruptible` is not a `oneof` but `NodeMetadata`'s `interruptible` is is because the former may or may not be set but `NodeMetadata` one should indicate whether it's set locally, or should inherit.  Katrina can clarify more if that's incomplete.

wild-endeavor:
Thanks @jeevb!

Also your comment that starts with "Smoke tested with..." above is helpful - can you copy paste that into a comment somewhere too?

flytekit/core/task.py
```diff
@@ -111,7 +111,7 @@ def my_task(x: int, y: typing.Dict[str, str]) -> str:
            but you can always manually update this field as well to force a cache miss. You should also manually bump
            this version if the function body/business logic has changed, but the signature hasn't.
     :param retries: Number of times to retry this task during a workflow execution.
-    :param interruptable: Boolean that indicates that this task can be interrupted and/or scheduled on nodes
+    :param interruptible: Boolean that indicates that this task can be interrupted and/or scheduled on nodes
```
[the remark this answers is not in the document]
katrogan: yes to more documentation! it's not intuitively understandable why the workflow metadata is a bool but the node one is a oneof (to support our explicit overridable/set value scheme...)

flytekit/core/workflow.py
jeevb: This makes sense to me!

flytekit/core/python_function_task.py
```diff
@@ -146,7 +146,9 @@ def compile_into_workflow(
             from flytekit.common.translator import get_serializable
 
             workflow_metadata = WorkflowMetadata(on_failure=WorkflowFailurePolicy.FAIL_IMMEDIATELY)
-            defaults = WorkflowMetadataDefaults(interruptible=False)
+            defaults = WorkflowMetadataDefaults(
+                interruptible=self.metadata.interruptible if self.metadata.interruptible is not None else False
```
jeevb, line 150: @wild-endeavor: Can you check this logic?

flytekit/core/launch_plan.py
wild-endeavor: yeah it's fine.  i think it'll stay like this

flytekit/core/python_function_task.py
wild-endeavor: +1

flytekit/core/task.py
```diff
@@ -111,10 +111,16 @@ def my_task(x: int, y: typing.Dict[str, str]) -> str:
            but you can always manually update this field as well to force a cache miss. You should also manually bump
            this version if the function body/business logic has changed, but the signature hasn't.
     :param retries: Number of times to retry this task during a workflow execution.
-    :param interruptable: Boolean that indicates that this task can be interrupted and/or scheduled on nodes
+    :param interruptible: [Optional] Boolean that indicates that this task can be interrupted and/or scheduled on nodes
```
jeevb, line 114: @katrogan @wild-endeavor: Does this explanation make sense?

jeevb:
All comments addressed I think, with the exception of the `SupportsNodeCreation` naming.

flytekit/core/task.py
katrogan: I think this sounds right, but I'm wondering what happens when workflow has interruptible=True and task has nothing set. Presumably the task should then inherit the workflow value, but we always default the task interruptible param value to False. So does this always get overwritten?

katrogan: hey @jeevb I'm sorry to pile on more work, but according to the original interruptible change, we defaulted the task interruptible bit to [None, rather than False](https://github.com/flyteorg/flytekit/pull/86/files#diff-4705dca3eaeaae1ac85b8f20caede8a7e8d6561da90c3fb5c7d8d6713df7e0a5R117) - would you mind changing that in [core/task.py](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/task.py#L67)? I think that's all that's required to make sure we don't set the oneof bit for Task/Node metadata at serialization time

cc @migueltol22

jeevb: The task default has been changed to `None`. So the task should inherit from the workflow. I'll add this case to the readme.

katrogan: oh already done, nvm!!

wild-endeavor (approved):

katrogan (approved):
thank you so much!

PR #540, opened 2021-07-07
New-style tasks/workflows use user exception scope

slai:
# TL;DR

New-style tasks/workflows are currently not wrapped in a user exception scope, so all exceptions in user code are treated as unknown system errors which are automatically retried. Further context - https://flyte-org.slack.com/archives/CREL4QVAQ/p1625581856326400

This PR restores the user exception scope to new-style tasks/workflows so exceptions in user code are not retried unless FlyteRecoverableException is raised.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -96,7 +96,11 @@ def _dispatch_execute(
             c: OR if an unhandled exception is retrieved - record it as an errors.pb
     """
     output_file_dict = {}
-    try:
+
+    @_scopes.system_entry_point
+    def do_dispatch_execute():
+        nonlocal output_file_dict
+
         # Step1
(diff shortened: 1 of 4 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -96,7 +96,11 @@ def _dispatch_execute(
             c: OR if an unhandled exception is retrieved - record it as an errors.pb
     """
     output_file_dict = {}
-    try:
+
+    @_scopes.system_entry_point
```
slai, line 100: Although `_dispatch_execute` is already being executed in a system exception scope, that scope is too far up the stack to wrap exceptions raised in this function so the subsequent except blocks below can handle them.

PR #543, opened 2021-07-07
Clean up scopes and exception handling for new tasks

wild-endeavor:
# TL;DR
This cleans up the exception handling of new tasks and workflows.  Currently there is no differentiation of recoverable or non-recoverable errors and all errors were system errors.

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Remove six from the scopes handling, no Python 2 anymore.  This makes the scopes code slightly easier to read.
* Add in [`kind`](https://github.com/flyteorg/flyteidl/blob/f4809c1a52073ec80de41be109bccc6331cdbac3/protos/flyteidl/core/execution.proto#L68) into the `ExecutionError` model.
* Add in [`origin`](https://github.com/flyteorg/flyteidl/blob/f4809c1a52073ec80de41be109bccc6331cdbac3/protos/flyteidl/core/errors.proto#L27) into the `ContainerError`.
* Use the user scope decorator where user-code is invoked (see https://github.com/flyteorg/flytekit/pull/540/files which this PR originally came from).
* Add the system scope decorator to `_dispatch_execute` in the entrypoint.py.

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -42,7 +42,8 @@ from flytekit.interfaces.stats.taggable import get_stats as _get_stats
 from flytekit.models import dynamic_job as _dynamic_job
 from flytekit.models import literals as _literal_models
 from flytekit.models.core import errors as _error_models
-from flytekit.models.core import identifier as _identifier, execution as _execution_models
+from flytekit.models.core import execution as _execution_models
+from flytekit.models.core import identifier as _identifier
 from flytekit.tools.fast_registration import download_distribution as _download_distribution
 from flytekit.tools.module_loader import load_object_from_module
 
@@ -132,8 +133,9 @@ def _dispatch_execute(
             _logging.warning(f"User-scoped IgnoreOutputs received! Outputs.pb will not be uploaded. reason {e}!!")
             return
         output_file_dict[_constants.ERROR_FILE_NAME] = _error_models.ErrorDocument(
-            _error_models.ContainerError(e.error_code, e.verbose_message, e.kind,
-                                         _execution_models.ExecutionError.ErrorKind.USER)
+            _error_models.ContainerError(
+                e.error_code, e.verbose_message, e.kind, _execution_models.ExecutionError.ErrorKind.USER
+            )
         )
         _logging.error("!! Begin User Error Captured by Flyte !!")
         _logging.error(e.verbose_message)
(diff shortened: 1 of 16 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -104,7 +105,9 @@ def _dispatch_execute(
         idl_input_literals = _literal_models.LiteralMap.from_flyte_idl(input_proto)
 
         # Step2
-        outputs = task_def.dispatch_execute(ctx, idl_input_literals)
+        # Decorate the dispatch execute function before calling it, this wraps all exceptions into one
+        # of the FlyteScopedExceptions
+        outputs = _scoped_exceptions.system_entry_point(task_def.dispatch_execute)(ctx, idl_input_literals)
```
kumare3, line 110: omg, this is weird? 
why not a simple
try:
   outputs = ....
except :
  ...

flytekit/common/exceptions/scopes.py
```diff
@@ -138,45 +137,51 @@ def _is_base_context():
...
-    should take action themselves or pass on to the platform owners.  We will dispatch metrics and such appropriately.
+    The reason these two (see the user one below) decorators exist is to categorize non-Flyte exceptions at arbitrary
+    locations. For example, while there is a separate ecosystem of Flyte-defined user and system exceptions
+    (see the FlyteException hierarchy), and we can easily understand and categorize those, if flytekit comes upon
+    a random ``ValueError`` or other non-flytekit defined error, how would we know if it was a bug in flytekit versus an
+    error with user code or something the user called? The purpose of these decorators is to categorize those (see
+    the last case in the nested try/catch below.
+
+    Decorator for wrapping functions that enter a system context. This should decorate every method that may invoke some
+    user code later on down the line. This will allow us to add differentiation between what is a user error and
```
kumare3, line 148: `  user code later on down the line. This will allow us to add differentiation between what is a`
WHY not at the point where the user code is called? i.e. only the last parent of the user code function?

kumare3 (approved):
lgtm, just couple stylistic comments

flytekit/common/exceptions/scopes.py
wild-endeavor: See comment above.

flytekit/bin/entrypoint.py
wild-endeavor: This is effectively what the decorator does.  The reason the decorator exists is mostly for two reasons:
1. We enter and leave user-code in multiple places, 
2. The excpetions that we, flytekit authors, throw do not conform to any hierarchy easily distinguishable from potential user exceptions.  After we complete https://github.com/flyteorg/flyte/issues/1033, and all flytekit exceptions have a base class that users will never use, we can get rid of these scoped exceptions and just use a few try catches to distinguish things.

I think it's preferable to get rid of these scoped exceptions actually, I just don't want to go through everything and update them right now.  Maybe at 1.0

PR #546, opened 2021-07-08
handle Asymmetric return values from tasks within a conditional block

kumare3:

# TL;DR
Flyte conditionals support asymmetric return values for the tasks in each of the case statements. The behavior is to truncate the outputs to only the minimal set that match across all outputs. this mostly affected local execution

```python
    @task
    def square(n: int) -> int:
        return n * n

    @task
    def double(n: int) -> int:
        return 2 * n

    @task
    def coin_toss(seed: int) -> bool:
        """
        Mimic some condition checking to see if something ran correctly
        """
        r = random.Random(seed)
        if r.random() < 0.5:
            return True
        return False

    @task
    def sum_diff(a: int, b: int) -> (int, int):
        return a + b, a - b

    @workflow
    def consume_outputs(my_input: int, seed: int = 5) -> int:
        is_heads = coin_toss(seed=seed)
        res = (
            conditional("double_or_square")
            .if_(is_heads.is_true())
            .then(square(n=my_input))
            .else_()
            .then(sum_diff(a=my_input, b=my_input))
        )

        return double(n=res)
```

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The base task did not return Promises in the case of skipped conditional sections. This also resulted in a case that the Conditional section when executed locally, ignored outputs from other tasks and thus would simply return outputs from the selected block. This output may not match the expected interface. So now just like in compilation, the outputs are matched and truncated in the cases when the task returns more outputs than expected.

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -277,14 +277,12 @@ class Task(object):
             ctx.execution_state is not None and ctx.execution_state.mode == ExecutionState.Mode.LOCAL_WORKFLOW_EXECUTION
         ):
             if ctx.execution_state.branch_eval_mode == BranchEvalMode.BRANCH_SKIPPED:
-                if self.python_interface and self.python_interface.output_tuple_name:
-                    variables = [k for k in self.python_interface.outputs.keys()]
-                    output_tuple = collections.namedtuple(self.python_interface.output_tuple_name, variables)
-                    nones = [None for _ in self.python_interface.outputs.keys()]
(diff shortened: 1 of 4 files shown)


wild-endeavor (commented):
lint so tests run?

wild-endeavor (approved):

PR #571, opened 2021-07-30
Prefix command with SDK_PYTHON_VENV if available

milton0825:
# TL;DR
Prefix the command with `SDK_PYTHON_VENV` if it is present. This ensures feature parity with https://github.com/flyteorg/flytekit/blob/68cd7beb1800bd32e7bb6b7b77fad7fecec01998/flytekit/common/tasks/sdk_runnable.py#L266

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/python_auto_container.py
+++ b/flytekit/core/python_auto_container.py
@@ -119,19 +119,21 @@ class PythonAutoContainerTask(PythonTask[T], metaclass=FlyteTrackedABC):
         if SDK_PYTHON_VENV.get():
             container_args.extend(SDK_PYTHON_VENV.get())
 
-        container_args.extend([
-            "pyflyte-execute",
-            "--inputs",
-            "{{.input}}",
-            "--output-prefix",
-            "{{.outputPrefix}}",
-            "--raw-output-data-prefix",
-            "{{.rawOutputDataPrefix}}",
-            "--resolver",
-            self.task_resolver.location,
-            "--",
-            *self.task_resolver.loader_args(settings, self),
-        ])
+        container_args.extend(
+            [
+                "pyflyte-execute",
+                "--inputs",
+                "{{.input}}",
+                "--output-prefix",
+                "{{.outputPrefix}}",
+                "--raw-output-data-prefix",
+                "{{.rawOutputDataPrefix}}",
+                "--resolver",
+                self.task_resolver.location,
+                "--",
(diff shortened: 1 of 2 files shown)


wild-endeavor:
CC @kumare3 

hmm, interesting.  i feel like we left this off for a reason.  Is there a particular need for this @milton0825?  That is, this is only needed if the virtual environment is not activated in docker image right?  Maybe we should stipulate that the virtual environment be active/on the $PATH when the container starts up.

Also, this was recently added: https://github.com/flyteorg/flytekit/blob/master/flytekit/clis/sdk_in_container/package.py#L95-L102  Currently this option isn't used at all except for spark tasks, but perhaps we should deprecate the `SDK_PYTHON_VENV` config in favor of this argument?  If we want to keep this kind of functionality, should we use this instead?

What do you guys think?

milton0825:
Yeah at Lyft we run `service_venv` to activate the virtual environment and set up some environment variables. I'd vote for adding this support to be more backward compatible and this make the migration easier.

milton0825:
@kumare3 @wild-endeavor ^^

flytekit/core/python_auto_container.py
```diff
@@ -114,19 +115,25 @@ def get_default_command(self, settings: SerializationSettings) -> List[str]:
...
-            "--output-prefix",
-            "{{.outputPrefix}}",
-            "--raw-output-data-prefix",
-            "{{.rawOutputDataPrefix}}",
-            "--resolver",
-            self.task_resolver.location,
-            "--",
-            *self.task_resolver.loader_args(settings, self),
-        ]
+        container_args: List[str] = []
```
kumare3, line 118: why do you need this? you can just activate the virtualenv in the docker container.

https://github.com/flyteorg/flytesnacks/blob/master/cookbook/core/Dockerfile#L16-L19

milton0825: We also do something else (statsd setup) in our service_venv

kumare3: @milton0825 then lets add it to the SerializationSettings. I am trying to get rid of the global config access

PR #572, opened 2021-07-30
Raw output prefix takes precedence for remote proxy creation in entrypoint

wild-endeavor:
# TL;DR
When running in GCP without a config file present, flytekit uses the wrong setting to try to determine what environment it's running in.  We should base the decision on the offloaded data location, not the cloud provider setting, which defaults to AWS.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Add an extra conditional to first check the `raw_output_data_prefix`.  Not too worried about the increased complexity becase this section of the code will soon be removed by #559, which should be going in within a week and change.

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -243,7 +243,9 @@ def setup_execution(
             local_sandbox_dir=_sdk_config.LOCAL_SANDBOX.get(),
             remote_proxy=_gcs_proxy.GCSProxy(raw_output_data_prefix),
         )
-    elif raw_output_data_prefix.startswith("file") or raw_output_data_prefix.startswith("/"):
+    elif raw_output_data_prefix and (
+        raw_output_data_prefix.startswith("file") or raw_output_data_prefix.startswith("/")
+    ):
         # A fake remote using the local disk will automatically be created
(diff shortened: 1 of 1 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -226,7 +226,25 @@ def setup_execution(
         tmp_dir=user_workspace_dir,
     )
 
-    if cloud_provider == _constants.CloudProvider.AWS:
+    # This rather ugly condition will be going away with #559. We first check the raw output prefix, and if missing,
+    # we fall back to the logic of checking the cloud provider. The reason we have to check for the existence of the
+    # raw_output_data_prefix arg first is because it may be set to None by execute_task_cmd. That is there to support
+    # the corner case of a really old propeller that is still not filling in the raw output prefix template.
+    if raw_output_data_prefix:
```
kumare3, line 233: Can we put a comment that points
To the other PR and says this will be deleted

cosmicBboy (approved):

flytekit/bin/entrypoint.py
wild-endeavor: that is already in the comment right above this line.

PR #583, opened 2021-08-09
Fetch node interface from static workflow closure

katrogan:

# TL;DR
Fetch node interface from static workflow closure. This refactor is necessary because in order to fetch recovered node executions they won't actually have any underlying task executions and therefore we need another source of truth for the interface.

Note: we should use the subworkflows in the WorkflowSpec to do the same for workflow nodes, but that's a larger refactor so I'll leave it for a separate change.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1151

## Follow-up issue
_NA_


flytekit/remote/remote.py
```diff
@@ -992,20 +1008,48 @@ def _get_output_literal_map(self, execution_data: ExecutionDataResponse) -> lite
...
+                    # Fetch the workflow and use its interface
+                    sub_workflow_ref = node.workflow_node.sub_workflow_ref
+                    workflow = self.fetch_workflow(
+                        sub_workflow_ref.project,
+                        sub_workflow_ref.domain,
+                        sub_workflow_ref.name,
+                        sub_workflow_ref.version,
+                    )
+                    return workflow.interface
+                elif node.workflow_node is not None and node.workflow_node.launchplan_ref is not None:
```
cosmicBboy, line 1034: add test for this launch plans?

cosmicBboy (dismissed):

flytekit/remote/remote.py
katrogan: @cosmicBboy 
I tried adding an integration test but Yee informs me it's broken, same with subworkflows, see

```
@classmethod
    def promote_from_model(
        cls,
        base_model: _workflow_model.WorkflowNode,
        sub_workflows: Dict[_identifier.Identifier, _workflow_model.WorkflowTemplate],
        tasks: Dict[_identifier.Identifier, _task_model.TaskTemplate],
    ) -> "FlyteWorkflowNode":
        from flytekit.remote import launch_plan as _launch_plan
        from flytekit.remote import workflow as _workflow
    
        fetch_args = (
            base_model.reference.project,
            base_model.reference.domain,
            base_model.reference.name,
            base_model.reference.version,
        )
    
        if base_model.launchplan_ref is not None:
>           return cls(flyte_launch_plan=_launch_plan.FlyteLaunchPlan.fetch(*fetch_args))
E           AttributeError: type object 'FlyteLaunchPlan' has no attribute 'fetch'
```

katrogan: @cosmicBboy can I remove the test and re-add once this is fixed?

wild-endeavor (approved):

flytekit/remote/remote.py
cosmicBboy: yes! I totally missed this method, those `fetch`es aren't meant to be there.

if you want to keep that test code there we can decorate it with `@pytest.mark.skip(reason="some reason")` and delete it with the fix PR

katrogan: thanks @cosmicBboy updated

wild-endeavor (approved):

flytekit/remote/remote.py
wild-endeavor: adding this to https://github.com/flyteorg/flyte/issues/1353

PR #584, opened 2021-08-09
FlyteFile and directory updates

wild-endeavor:

# TL;DR
Just cleaning up the behavior around files and directories a bit, and updating the documentation.  The main changes are

* Add a `trigger_download` function to trigger downloading if you don't want to explicitly call `open()`.
* Removing the transformer in `os.PathLike` from the main type engine file since now we'll be giving the user a `FlyteFile` object even when the signature is just `os.PathLike`.
* Moving support for `os.PathLike` into the `FlyteFile` transformer.

Please see the updated docstrings for more information.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin
 - [x] Refactor

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

* Added an option to the type engine so transformers can declare to flytekit that they support multiple python types.


flytekit/types/file/file.py
```diff
@@ -217,46 +248,78 @@ def to_literal(
...
-                meta = BlobMetadata(type=self._blob_type(format=self.get_format(python_type)))
                 return Literal(scalar=Scalar(blob=Blob(metadata=meta, uri=python_val._remote_source)))
 
-            source_path = python_val.path
-            if python_val.remote_path is False:
-                # If the user specified the remote_path to be False, that means no matter what, do not upload
+            # If the user specified the remote_path to be False, that means no matter what, do not upload. Also if the
+            # path given is already a remote path, say https://www.google.com, the concept of uploading to the Flyte
+            # blob store doesn't make sense.
+            if python_val.remote_path is False or ctx.file_access.is_remote(source_path):
```
kumare3, line 266: should we do the is_remote check? Why not just upload?

flytekit/types/file/file.py
```diff
@@ -184,6 +209,12 @@ def remote_source(self) -> str:
         """
         return self._remote_source
 
+    def trigger_download(self):
```
kumare3, line 212: Should we use this in __fspath__ instead of self._downloader()?

```
def __fspath__(self):
        # This is where a delayed downloading of the file will happen
        if not self._downloaded:
            self._downloader()
            self._downloaded = True
        return self._path
```

Also I think we need to add the `download cache` check here too!

flytekit/types/file/file.py
wild-endeavor: But if a task
```
def t1() -> FlyteFile:
  return FlyteFile("https://www.google.com")
```
what does it mean to upload that?

flytekit/core/type_engine.py
```diff
@@ -290,17 +289,19 @@ class TypeEngine(typing.Generic[T]):
...
         This should be used for all types that respond with the right type annotation when you use type(...) function
         """
-        if transformer.python_type in cls._REGISTRY:
-            existing = cls._REGISTRY[transformer.python_type]
-            raise ValueError(
-                f"Transformer {existing.name} for type {transformer.python_type} is already registered."
-                f" Cannot override with {transformer.name}"
-            )
-        cls._REGISTRY[transformer.python_type] = transformer
+        types = [transformer.python_type, *(additional_types or [])]
```
kumare3, line 296: not needed,,just register multiple times

flytekit/types/directory/types.py
```diff
@@ -128,6 +175,15 @@ def remote_source(self) -> str:
         """
         return self._remote_source
 
+    def trigger_download(self):
```
kumare3, line 178: can we return the path and call the method  `download`

flytekit/core/type_engine.py
wild-endeavor: actually i don't know how to do this cleanly.   Because the transformer specifies the python type when it inherits from the base transformer: `class FlyteFilePathTransformer(TypeTransformer[FlyteFile])`, I'd have to create a base class out of the transformer, and then create two child classes out of it I think.  I think this is actually cleaner.

flytekit/core/type_engine.py
```diff
@@ -290,17 +289,19 @@ class TypeEngine(typing.Generic[T]):
     _DATACLASS_TRANSFORMER: TypeTransformer = DataclassTransformer()
 
     @classmethod
-    def register(cls, transformer: TypeTransformer):
+    def register(cls, transformer: TypeTransformer, additional_types: typing.Optional[typing.List[Type]] = None):
```
cosmicBboy, line 292: consider docstring for `additional_types` (and maybe `transformer` too?)

cosmicBboy (approved):

PR #588, opened 2021-08-10
Streamline __call__ pattern between Tasks and Workflows

wild-endeavor:

## TL;DR
Please read through the description in the [issue](https://github.com/flyteorg/flyte/issues/1032) as it's pretty succinct. 

Note that this PR only touches local execution.  Production runs on a real Flyte backend is not touched.

## Description

This PR basically further streamlines the logic in the `__call__` function of tasks and workflows.  At this point the logic is pretty much identical.  (But premature OOP inheritance pattern has bitten us in the past so we'll refrain from creating a base class out of it for now.)

## Unit Test Changes
We had to make some changes to the unit test to get this new paradigm to work.   You likely will need to make similar changes to your tests.

1.  Dataframe handling: If your task returned an instance of a Pandas dataframe as a `FlyteSchema` type in the past, since the Type Engine wasn't triggered, you just got the raw pandas dataframe.  Now you'll get the `FlyteSchema` object, so you'll need to change how you read that.  (See the change in `tests/flytekit/unit/core/test_type_hints.py`)
1. Sagemaker tasks: Probably nothing needs to happen but if you had modified sagemaker tasks to be tested locally, you'll need to provide real directories to the `train` and `validation` arguments.

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1032


flytekit/core/base_task.py
```diff
@@ -286,16 +294,37 @@ def __call__(self, *args, **kwargs):
...
-                        mode=ExecutionState.Mode.LOCAL_TASK_EXECUTION, user_space_params=new_user_params
-                    )
+                    ctx.new_execution_state().with_params(mode=ExecutionState.Mode.LOCAL_WORKFLOW_EXECUTION)
                 )
-            ):
-                return self.execute(**kwargs)
+            ) as child_ctx:
+                result = self._local_execute(child_ctx, **kwargs)
+
+            expected_outputs = len(self.python_interface.outputs)
```
kumare3, line 305: is the expected outputs check new?

flytekit/core/base_task.py
```diff
@@ -286,16 +294,37 @@ def __call__(self, *args, **kwargs):
...
                 ctx.with_execution_state(
-                    ctx.execution_state.with_params(
-                        mode=ExecutionState.Mode.LOCAL_TASK_EXECUTION, user_space_params=new_user_params
-                    )
+                    ctx.new_execution_state().with_params(mode=ExecutionState.Mode.LOCAL_WORKFLOW_EXECUTION)
                 )
-            ):
-                return self.execute(**kwargs)
+            ) as child_ctx:
+                result = self._local_execute(child_ctx, **kwargs)
```
kumare3, line 303: update  the comment in local execute. It is currently
```
        This code is used only in the case when we want to dispatch_execute with outputs from a previous node
        For regular execution, dispatch_execute is invoked directly.
```

I think it should be should be used when we have native values and want to execute?

kumare3:
And if I did not mention - I think this makes a ton of sense - this makes all flyte transformations always required which is perfect

flytekit/core/base_task.py
wild-endeavor: it is - copied from the workflow `__call__` function - just making things as similar as possible.

wild-endeavor:
@jeevb thoughts?

kumare3 (dismissed):

flytekit/core/promise.py
```diff
@@ -793,7 +793,77 @@ def create_and_link_node(
...
     return create_task_output(node_outputs, interface)
+
+
+class LocallyExecutable(Protocol):
+    def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], Promise, VoidPromise]:
+        ...
+
+
+def executable_artifact_call_handler(entity: Union[SupportsNodeCreation, LocallyExecutable], *args, **kwargs):
+    """
```
kumare3, line 808: the logic is the same for workflows and tasks, but we did not want to create base class, hence a separate method

kumare3 (approved):

PR #591, opened 2021-08-12
Fix resource overrides

jeevb:
# TL;DR
Specifying resource overrides for nodes fails if only one of `requests` or `limits` are specified. This PR fixes this, and allows either `requests` or `limits` overrides to be applied on their own.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Use `kwargs.get(key)` which better handles missing values, instead of `kwargs[key]`.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/node.py
+++ b/flytekit/core/node.py
@@ -86,8 +86,8 @@ class Node(object):
             for k, v in alias_dict.items():
                 self._aliases.append(_workflow_model.Alias(var=k, alias=v))
         if "requests" in kwargs or "limits" in kwargs:
-            requests = _convert_resource_overrides(kwargs["requests"], "requests")
-            limits = _convert_resource_overrides(kwargs["limits"], "limits")
+            requests = _convert_resource_overrides(kwargs.get("requests"), "requests")
+            limits = _convert_resource_overrides(kwargs.get("limits"), "limits")
             self._resources = _resources_model(requests=requests, limits=limits)
         return self
 
--- a/tests/flytekit/unit/core/test_node_creation.py
+++ b/tests/flytekit/unit/core/test_node_creation.py
@@ -166,7 +166,7 @@ def test_runs_before():
     my_wf(a=5, b="hello")
 
 
-def test_resource_overrides():
+def test_resource_request_override():
     @task
     def t1(a: str) -> str:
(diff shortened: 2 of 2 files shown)


tests/flytekit/unit/core/test_node_creation.py
```diff
@@ -166,19 +166,78 @@ def my_wf(a: int, b: str) -> (str, typing.List[str], int):
...
+    def t1(a: str) -> str:
+        return f"*~*~*~{a}*~*~*~"
+
+    @workflow
+    def my_wf(a: typing.List[str]) -> typing.List[str]:
+        mappy = map_task(t1)
+        map_node = mappy(a=a).with_overrides(
             requests=Resources(cpu="1", mem="100", ephemeral_storage="500Mi"),
-            limits=Resources(cpu="2", mem="200", ephemeral_storage="1Gi"),
+            limits=(Resources(cpu="2", mem="200", ephemeral_storage="1Gi")),
```
EngHabu, line 238: Nit: you don't really need the extra `()` do you?

EngHabu (dismissed):

tests/flytekit/unit/core/test_node_creation.py
jeevb: Oof, nice catch. Will fix.

jeevb:
@EngHabu: Ready for re-review.

flytekit/core/node.py
```diff
@@ -86,8 +86,8 @@ def with_overrides(self, *args, **kwargs):
             for k, v in alias_dict.items():
                 self._aliases.append(_workflow_model.Alias(var=k, alias=v))
         if "requests" in kwargs or "limits" in kwargs:
-            requests = _convert_resource_overrides(kwargs["requests"], "requests")
-            limits = _convert_resource_overrides(kwargs["limits"], "limits")
+            requests = _convert_resource_overrides(kwargs.get("requests"), "requests")
```
katrogan, line 89: how is this functionally different? should we be using the limits as the default value if requests aren't specified?

PR #592, opened 2021-08-12
[wip] Add caching to local execution

eapolinario:
# TL;DR
Enable the ability to cache the output of tasks in local executions

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
[`joblib.Memory`](https://joblib.readthedocs.io/en/latest/generated/joblib.Memory.html#joblib.Memory) provides a persistent store to cache the result of function invocations. Similarly to how we do in the hosted case, we use the pair `(task inputs, cache version)` to memoize task outputs across different executions.

We're also exposing a configuration knob to let users decide where the cached objects will be written to disk.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/761

## Follow-up issue
_NA_

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,3 +88,8 @@ update_version:
 
 	grep "$(PLACEHOLDER)" "setup.py"
 	sed -i "s/$(PLACEHOLDER)/__version__ = \"${VERSION}\"/g" "setup.py"
+
+# TODO
+.PHONY: clear-cache
+clear-cache:
+	python -m 'from flytekit.core.local_cache import LocalCache; LocalCache.clear()'
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -253,8 +253,6 @@ class Task(object):
         # TODO: improve comment
         # if metadata.cache is set, check memoized version including cache_version
         if self._metadata.cache:
-            # TODO: do I need to ignore self?
-            # TODO: probably need a wrapper to indicate it's going to pull from the cache. A log message here might be enough?
             dispatch_execute = memory.cache(self._local_dispatch_execute, ignore=['self', 'ctx'])
         else:
             dispatch_execute = self._local_dispatch_execute
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -53,6 +53,12 @@ from flytekit.models.core import workflow as _workflow_model
 from flytekit.models.interface import Variable
 from flytekit.models.security import SecurityContext
 
+from joblib import Memory
+
+# TODO: move the definition of `memory` to a separate file
+CACHE_LOCATION = '/tmp/cache-location'
+memory = Memory(CACHE_LOCATION, verbose=0)
+
 
 def kwtypes(**kwargs) -> Dict[str, Type]:
     """
@@ -217,6 +223,14 @@ class Task(object):
         """
         return None
 
+    def _local_dispatch_execute(
+        self, ctx: FlyteContext, input_literal_map: _literal_models.LiteralMap, cache_version: str
+    ) -> Union[_literal_models.LiteralMap, _dynamic_job.DynamicJobSpec]:
+        """
+        TODO: explain why we need this wrapper.
+        """
+        return self.dispatch_execute(ctx, input_literal_map)
+
     def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], Promise, VoidPromise]:
         """
         This code is used only in the case when we want to dispatch_execute with outputs from a previous node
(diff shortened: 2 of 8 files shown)


flytekit/core/base_task.py
```diff
@@ -236,7 +250,13 @@ def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], P
         )
         input_literal_map = _literal_models.LiteralMap(literals=kwargs)
 
-        outputs_literal_map = self.dispatch_execute(ctx, input_literal_map)
+        # TODO: improve comment
+        # if metadata.cache is set, check memoized version including cache_version
```
kumare3, line 254: cc @wild-endeavor 
FYI, Yee is working on some updates to this part of the code as well

kumare3: nope

Makefile
```diff
@@ -88,3 +88,8 @@ update_version:
 
 	grep "$(PLACEHOLDER)" "setup.py"
 	sed -i "s/$(PLACEHOLDER)/__version__ = \"${VERSION}\"/g" "setup.py"
+
+# TODO
+.PHONY: clear-cache
+clear-cache:
```
kumare3, line 94: can we make this a pyflyte command?

`pyflyte cache clear`

eapolinario: good call, much cleaner.

kumare3 (approved):

flytekit/core/base_task.py
```diff
@@ -236,7 +245,17 @@ def _local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], P
         )
         input_literal_map = _literal_models.LiteralMap(literals=kwargs)
 
-        outputs_literal_map = self.dispatch_execute(ctx, input_literal_map)
+        # if metadata.cache is set, check memoized version
+        if self._metadata.cache:
+            # The cache key is composed only of '(input_literal_map, cache_version)', i.e. all other parameters
```
katrogan, line 250: should we also include the task name as part of the key?

similar to what we do with the identifier on hosted flyte https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/catalog/client.go#L26

tests/flytekit/unit/core/test_local_cache.py
```diff
@@ -0,0 +1,78 @@
+from pytest import fixture
```
katrogan, line 1: can we test some more complex input and output types, like dataclasses, schemas, files?

flytekit/core/local_cache.py
```diff
@@ -0,0 +1,29 @@
...
+CACHE_LOCATION = "~/.flyte/local-cache"
+
+
+class LocalCache(object):
+    _memory: Memory
+    _initialized: bool = False
+
+    @staticmethod
+    def initialize():
+        LocalCache._memory = Memory(CACHE_LOCATION, verbose=5)
```
katrogan, line 16: should we make the verbosity configurable too?

katrogan (commented):
looks great, can we add some more tests!

flytekit/core/base_task.py
eapolinario: The key is already implicitly part of the cache key (in other words, `joblib.Memory` only caches the result of _each_ function call). I'm going to update the comment to reflect that.

flytekit/core/local_cache.py
eapolinario: I don't think we gain a lot from making this configurable, especially given that configuration files are not used at all in local executions. I'm going to leave a TODO as well.

flytekit/core/base_task.py
eapolinario: Actually, this is a really good call. I'm going to add a test case to demonstrate the problem and subsequent fix.

eapolinario: Fixed by https://github.com/flyteorg/flytekit/pull/592/commits/130b4327c7e9b5bad40626c0bdc52669270d97b9.

tests/flytekit/unit/core/test_local_cache.py
```diff
@@ -0,0 +1,252 @@
...
+        # The second and third calls hit the cache
+        assert (my_wf().open().all() == pandas.DataFrame(data={"x": [1, 2], "y": ["3", "4"]})).all().all()
+        assert n_cached_task_calls == 1
+        assert (my_wf().open().all() == pandas.DataFrame(data={"x": [1, 2], "y": ["3", "4"]})).all().all()
+        assert n_cached_task_calls == 1
+
+
+def test_wf_custom_types():
+    @dataclass_json
+    @dataclass
```
katrogan, line 157: awesome, this works? ran into issues with dataclass serialization last time i tried this but maybe joblib has since worked around this. woohoo!

katrogan (approved):

tests/flytekit/unit/core/test_local_cache.py
kumare3: question around this, are we storing `LiteralMap` or are we storing the native `python values`. Also what happens to file refrences, do we store a shallow copy or a deep copy?

PR #595, opened 2021-08-13
Added mpi plugin

evalsocket:
# TL;DR
- Added python mpi plufin

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


scripts/flytekit_mpi_runner.py
```diff
@@ -0,0 +1,94 @@
...
+            if unknowns[i].startswith(FLYTE_ARG_PREFIX) is False:
+                env_vars[processed] = unknowns[i]
+                i += 1
+        else:
+            # To prevent SageMaker from ignoring our __FLYTE_CMD_*__ hyperparameters, we need to set a dummy value
+            # which serves as a placeholder for each of them. The dummy value placeholder `__FLYTE_CMD_DUMMY_VALUE__`
+            # falls into this branch and will be ignored
+            i += 1
+
+    return flyte_cmd, env_vars
```
evalsocket, line 65: @wild-endeavor @EngHabu @kumare3 In the spark example we are directly using flyte command but in my case, I need a command that is very user-specific. 

In MPI we launch python code from mpi binary like this 
```
        command:
            - mpirun
            args:
            - -np
            - "2"
            - --allow-run-as-root
            - -bind-to
            - none
            - -map-by
            - slot
            - -x
            - LD_LIBRARY_PATH
            - -x
            - PATH
            - -mca
            - pml
            - ob1
            - -mca
            - btl
            - ^openib
            - python
            - /examples/tensorflow_mnist.py
```

But we launch a python script from flyte-excute and it is needed for handling flyte stuff....Any suggestion for handling this

kumare3: @evalsocket what are all the various values `slot` -x, etc

kumare3:
@evalsocket i updated, this try it out. I think we should also add Horovod elastic to this plugin itself.

Lets talk about it. but first lets test this one

PR #603, opened 2021-08-19
Insert doesn't return value

wild-endeavor:
# TL;DR
Insert modifies in place, doesn't return the modified value.

* Also removing the `from e` bit, it hides the stack trace which makes it super hard to investigate.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/plugins/flytekit-greatexpectations/setup.py
+++ b/plugins/flytekit-greatexpectations/setup.py
@@ -4,7 +4,7 @@ PLUGIN_NAME = "great_expectations"
 
 microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
 
-plugin_requires = ["flytekit>=0.21.0,<1.0.0", "great-expectations>=0.13.23"]
+plugin_requires = ["flytekit>=0.21.0,<1.0.0", "great-expectations==0.13.28"]
 
 __version__ = "0.0.0+develop"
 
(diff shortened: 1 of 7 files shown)


kumare3 (dismissed):

plugins/flytekit-greatexpectations/setup.py
```diff
@@ -4,7 +4,7 @@
 
 microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
 
-plugin_requires = ["flytekit>=0.21.0,<1.0.0", "great-expectations>=0.13.23"]
+plugin_requires = ["flytekit>=0.21.0,<1.0.0", "great-expectations==0.13.28"]
```
kumare3, line 7: @samhita-alla can we actually change it atleast to be a range, with  upperbound of 0.13.28. Better would be to actually fix it for the new version, since we have not really announced anything

samhita-alla: I believe https://github.com/great-expectations/great_expectations/issues/3290 is the issue.

samhita-alla: We can proceed with 0.13.28 for now. I'll bump the version later.

cosmicBboy (approved):

PR #621, opened 2021-08-30
[WIP] Broader config file support

pingsutw:

# TL;DR
`Setup-config` command in flytekit to use the YAML format so that you can run setup-config with flytekit and then run flytectl.
`flyte-cli` load config in YAML format.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Add an option that lets the user choose the output format.
To support backward compatibility, we could output both `ini` and `yaml` format. 

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1035

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/clis/flyte_cli/main.py
+++ b/flytekit/clis/flyte_cli/main.py
@@ -1,3 +1,4 @@
+import configparser as _configparser
 import importlib as _importlib
 import os
 import os as _os
@@ -8,6 +9,7 @@ from typing import Callable, Dict, List, Tuple, Union
 import click as _click
 import requests as _requests
 import six as _six
+import yaml
 from flyteidl.admin import launch_plan_pb2 as _launch_plan_pb2
 from flyteidl.admin import task_pb2 as _task_pb2
 from flyteidl.admin import workflow_pb2 as _workflow_pb2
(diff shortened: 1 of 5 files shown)


flytekit/clis/flyte_cli/main.py
```diff
@@ -485,6 +487,7 @@ def _render_schedule_expr(lp):
 _output_location_prefix_option = _click.option(
     "-o", "--output-location-prefix", help="Custom output location prefix for offloaded types (files/schemas)"
 )
+_output_config_format = _click.option("-o", "--format", type=str, default="ini", help="Output config format (ini/yaml)")
```
katrogan, line 490: can we use click.Choice here? https://click.palletsprojects.com/en/8.0.x/options/#choice-options

pingsutw: sure, thnaks.

wild-endeavor:
hey sorry should've included more information in the issue.  this won't work unfortunately.  the reason is that not only is flytectl's format different, the keys are entirely different - the names i mean.  for instance
```
admin:
  endpoint: dns:///fakeflyte.com
  insecure: false
  authType: Pkce
logger:
  show-source: true
  level: 1
```

wild-endeavor:
let's think about this in the broader context of updating our configuration support.  The UX definitely needs improvement in a number of places.

pingsutw:
> hey sorry should've included more information in the issue. this won't work unfortunately. the reason is that not only is flytectl's format different, the keys are entirely different - the names i mean. for instance

Okay, I see

wild-endeavor:
closing in favor of https://github.com/flyteorg/flytekit/pull/898 sorry

PR #644, opened 2021-09-09
Refactor plugin tests into individual directories

wild-endeavor:
# TL;DR
Moving tests around for plugins, this is something we've been wanting to do for a while.

**Note**
Not including pandera plugin for now because it depends on `simpleeval` which is experiencing [issues](https://github.com/danthedeckie/simpleeval/issues/90)

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin
 - [x] Refactor

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
* Moving plugin tests to be inside each plugin's top level folder rather than together in one big tests folder.
* One of the issues with having them all run together in the same job is that sometimes we'd notice dependency clashes between flytekit and its plugins.  This came up recently with the debacle around https://github.com/flyteorg/flyte/issues/940.  Basically what happened here was that `make setup` would install a version of flyteidl, but then running the plugin's setup would revert it.
* Plugin names are currently manually listed in the GitHub workflow, we should tweak this to not be the case in the future.  In the future we'll be moving the entire plugins out into a separate repo too, probably a shortly before 1.0 thing.
* Some changes were necessary to the tests to get them to pass.

Developers of flytekit can continue to use `pip install -e .` locally in the `plugins` folder if they so choose, but this is no longer what we'll be doing in CI, which is why the make target to do so has been removed.  Even though we don't have any now, over time, we'll probably have plugins with conflicting dependencies and one environment for all plugins simply doesn't make sense.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1352


.github/workflows/pythonbuild.yml
```diff
@@ -58,6 +56,37 @@ jobs:
...
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        make setup
+        cd plugins/${{ matrix.plugin-names }}
+        pip install -r requirements.txt
+        pip freeze
+    - name: Test with coverage
+      run: |
+        coverage run -m pytest tests
```
eapolinario, line 88:
```suggestion
        coverage run -m pytest plugins/${{ matrix.plugin-names }}/tests
```

.github/workflows/pythonbuild.yml
```diff
@@ -58,6 +56,37 @@ jobs:
       with:
         fail_ci_if_error: true # optional (default = false)
 
+  build-plugins:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.8]
+        # Omitting "flytekit-pandera" for now.
+        plugin-names: ["flytekit-aws-athena", "flytekit-aws-sagemaker", "flytekit-data-fsspec", "flytekit-dolt", "flytekit-greatexpectations", "flytekit-hive", "flytekit-k8s-pod", "flytekit-kf-pytorch", "flytekit-kf-tensorflow", "flytekit-papermill", "flytekit-spark", "flytekit-sqlalchemy"]
```
kumare3, line 65: is it possible to get this using `ls plugins/flytekit-*` or maybe a file `cat plugins/tests.yaml`

.github/workflows/pythonbuild.yml
wild-endeavor: no i tried this, i couldn't get this to work.

.github/workflows/pythonbuild.yml
wild-endeavor: can we do this in the future?  i know this isn't pretty.  i can make another ticket for it.

.github/workflows/pythonbuild.yml
```diff
@@ -58,6 +56,37 @@ jobs:
...
+        path: ~/.cache/pip
+        # Look to see if there is a cache hit for the corresponding requirements files
+        key: ${{ format('{0}-pip-{1}', runner.os, hashFiles('dev-requirements.txt', format('plugins/{0}/requirements.txt', matrix.plugin-names ))) }}
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip setuptools wheel
+        make setup
+        cd plugins/${{ matrix.plugin-names }}
+        pip install -r requirements.txt
+        pip freeze
```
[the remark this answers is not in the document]
wild-endeavor: we always had this, it's nice to see

eapolinario (approved):
remember to leave a TODO for Ketan's suggestion, otherwise, LGTM.

PR #645, opened 2021-09-09
[mypy] Fix class name is not defined error

mayitbeegh:
Signed-off-by: Sean Lin <sean@union.ai>

# TL;DR
Fix class name is not defined error in mypy. Some errors are fixed by importing necessary modules and placing forward reference with normal reference; Others are fixed by introducing forward reference to avoid circular dependency https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1428

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/types/schema/types_pandas.py
+++ b/flytekit/types/schema/types_pandas.py
@@ -10,14 +10,17 @@ from flytekit.models.types import LiteralType, SchemaType
 from flytekit.plugins import pandas
 from flytekit.types.schema import LocalIOSchemaReader, LocalIOSchemaWriter, SchemaEngine, SchemaFormat, SchemaHandler
 
+if typing.TYPE_CHECKING:
+    import pandas as _pandas
+
 
 class ParquetIO(object):
     PARQUET_ENGINE = "pyarrow"
 
-    def _read(self, chunk: os.PathLike, columns: typing.List[str], **kwargs) -> pandas.DataFrame:
+    def _read(self, chunk: os.PathLike, columns: typing.List[str], **kwargs) -> "_pandas.DataFrame":
         return pandas.read_parquet(chunk, columns=columns, engine=self.PARQUET_ENGINE, **kwargs)
 
-    def read(self, *files: os.PathLike, columns: typing.List[str] = None, **kwargs) -> pandas.DataFrame:
+    def read(self, *files: os.PathLike, columns: typing.List[str] = None, **kwargs) -> "_pandas.DataFrame":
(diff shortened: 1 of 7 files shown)


flytekit/types/schema/types_pandas.py
```diff
@@ -10,14 +10,17 @@
 from flytekit.plugins import pandas
```
wild-endeavor, line 10: @kumare3 can you take a look at this one line please?  We should just swap it out right?

flytekit/types/schema/types_pandas.py
```diff
@@ -10,14 +10,17 @@
...
 from flytekit.types.schema import LocalIOSchemaReader, LocalIOSchemaWriter, SchemaEngine, SchemaFormat, SchemaHandler
 
+if typing.TYPE_CHECKING:
+    import pandas as _pandas
```
wild-endeavor, line 14: pending the comment above from ketan, i think we can just move this up.

tests/flytekit/unit/core/test_flyte_directory.py
```diff
@@ -140,8 +140,10 @@ def my_wf() -> FlyteDirectory:
     files = os.listdir(wfd.path)
     assert len(files) == 5
 
+    csv = str
+
     @task
-    def t2(in1: FlyteDirectory["csv"]) -> int:
+    def t2(in1: FlyteDirectory[csv]) -> int:
```
wild-endeavor, line 146: It's a hack but could we use this style instead please https://github.com/flyteorg/flytekit/blob/b8a3c7d34e6b19722a6a968dc05506ad1eb26912/flytekit/types/file/__init__.py#L60?  It's a hack because it just so happens that the string representation is the string that you pass in, but whatever. generics in general are pretty hacky in python.

wild-endeavor (commented):
just a few comments.  thank you!!!

flytekit/types/schema/types_pandas.py
kumare3: yup, flytekit has a direct dependency on pandas.

tests/flytekit/unit/core/test_flyte_directory.py
kumare3: maybe there is an alternative?

wild-endeavor (approved):

PR #650, opened 2021-09-10
Make internal task unique per papermill task

EngHabu:
Signed-off-by: Haytham Abuelfutuh <haytham@afutuh.com>

# TL;DR
Papermill task plugin creates a task with a dummy function. Because it uses the same function definition for every instance of the NotebookTask, that results into a naming conflict. This PR uses the name of the Notebook task when creating the python task to guarantee uniqueness (or at least as unique as Notebook tasks)

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1460

### DIFF
--- a/plugins/flytekit-papermill/flytekitplugins/papermill/task.py
+++ b/plugins/flytekit-papermill/flytekitplugins/papermill/task.py
@@ -22,7 +22,9 @@ T = typing.TypeVar("T")
 def _dummy_task_func():
     return None
 
+
 PAPERMILL_TASK_PREFIX = "pm.nb"
+
 
 class NotebookTask(PythonInstanceTask[T]):
     """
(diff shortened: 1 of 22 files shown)


wild-endeavor (dismissed):

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
```diff
@@ -111,6 +111,8 @@ def __init__(
         outputs: typing.Optional[typing.Dict[str, typing.Type]] = None,
         **kwargs,
     ):
+        _dummy_task_func.__name__ = name
```
eapolinario, line 114: Can you leave comment and explain why we're doing this?

wild-endeavor (approved):

PR #662, opened 2021-09-16
Silence logs

kumare3:
# TL;DR
Silence logs and remove erroneous print


wild-endeavor (approved):
i think we should do a round of cleanup and then move back down to info.

PR #666, opened 2021-09-22
Dataclass should patch int values for nested cases

kumare3:

# TL;DR
Cases in which a Dataclass consists of Generic types of nested types that contain an int, the integer ends up being a Float value. This is because protobuf struct does not have ints only a number.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
fixes https://github.com/lyft/flyte/issues/1510

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -234,6 +234,47 @@ class DataclassTransformer(TypeTransformer[object]):
             scalar=Scalar(generic=_json_format.Parse(cast(DataClassJsonMixin, python_val).to_json(), _struct.Struct()))
         )
 
+    def _fix_val_int(self, t: typing.Type, val: typing.Any) -> typing.Any:
+        if t == int:
+            return int(val)
+
+        if isinstance(val, list):
+            try:
+                if ListTransformer.get_sub_type(f.type) == int:
+                    return [int(i) for i in val]
+            except:
+                return val
+
+        if isinstance(val, dict):
+            ktype, vtype = DictTransformer.get_dict_types(t)
+            if (ktype is not None and ktype == int) or (vtype is not None and vtype == int):
+                d = {}
+                for k, v in val.items():
+                    if ktype is not None and ktype == int:
+                        k = int(k)
+                    if vtype is not None and vtype == int:
+                        v = int(v)
+                    d[k] = v
+                return d
+
+        if dataclasses.is_dataclass(t):
+            return self._fix_dataclass_int(t, val)
+
+        return val
+
+    def _fix_dataclass_int(self, dc_type: Type[DataClassJsonMixin], dc: DataClassJsonMixin) -> DataClassJsonMixin:
+        """
+        This is a performance penalty to convert to the right types, but this is expected by the user and hence
+        needs to be done
+        """
+        # NOTE: Protobuf Struct does not support explicit int types, int types are upconverted to a double value
+        # https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value
+        # Thus we will have to walk the given dataclass and typecast values to int, where expected.
+        for f in dataclasses.fields(dc_type):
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
```diff
@@ -234,6 +234,47 @@ def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], exp
...
         )
 
+    def _fix_val_int(self, t: typing.Type, val: typing.Any) -> typing.Any:
+        if t == int:
+            return int(val)
+
+        if isinstance(val, list):
+            try:
+                if ListTransformer.get_sub_type(t) == int:
+                    return [int(i) for i in val]
```
samhita-alla, line 244: How about we replace `[int(i) for i in val]` with `list(map(int, val))`?

pingsutw:
Few updates
- Handle nested dict and list
- Update test
    - Add nested dict and list
    - Check element type is `int`. 
    - `assert ot.s.a == 5`, `==` is not enough since 1.0 is equal to 1 in python

flytekit/core/type_engine.py
pingsutw: Updated it, Thanks.

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -375,6 +399,78 @@ def test_dataclass_transformer():
...
+
+
+def test_dataclass_int_preserving():
+    ctx = FlyteContext.current_context()
+
+    i = InnerStruct(a=5, b=None, c=[1, 2, 3])
+    tf = DataclassTransformer()
+    lv = tf.to_literal(ctx, i, InnerStruct, tf.get_literal_type(InnerStruct))
+    ot = tf.to_python_value(ctx, lv=lv, expected_python_type=InnerStruct)
+    assert ot.a == 5
```
eapolinario, line 422: why can't we rely on the equality comparison method generated for the dataclass?

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -353,13 +376,14 @@ def test_dataclass_transformer():
...
     }
+
     tf = DataclassTransformer()
     t = tf.get_literal_type(TestStruct)
     assert t is not None
     assert t.simple is not None
     assert t.simple == SimpleType.STRUCT
     assert t.metadata is not None
-    assert t.metadata == schema
+    # assert t.metadata == schema
```
eapolinario, line 386: why is this commented out?

tests/flytekit/unit/core/test_type_engine.py
kumare3: Aah we should

flytekit/core/type_engine.py
```diff
@@ -234,6 +234,37 @@ def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], exp
             scalar=Scalar(generic=_json_format.Parse(cast(DataClassJsonMixin, python_val).to_json(), _struct.Struct()))
         )
 
+    def _fix_val_int(self, t: typing.Type, val: typing.Any) -> typing.Any:
+        if t == int:
+            return int(val)
+
+        if isinstance(val, list):
+            # Handle nested List. e.g. [[1, 2], [3, 4]]
+            return list(map(lambda x: self._fix_val_int(ListTransformer.get_sub_type(t), x), val))
```
kumare3, line 243: thank you @pingsutw

tests/flytekit/unit/core/test_type_engine.py
pingsutw: I've updated it, thank you @eapolinario

eapolinario (approved):
That's awesome.

PR #667, opened 2021-09-22
[WIP] Add support for python pickle type in flytekit/flyte

pingsutw:

# TL;DR
Add a new type `PythonPickle = FlyteFile[typing.TypeVar(PYTHON_PICKLE_FORMAT)]`. 
Any type that flyte can't recognize will become `PythonPickle`. e.g. `typing.Any`, `typing.List[int]` -> `PythonPickle`
Function output (native python type) will save in PythonPickle

Let's assume that we have two tasks "A" and "B", We have several cases that need to handle.

### Case 1:  A.output (Pickle) -> B.input (Pickle)
1. Run the Task A
2. Serialize task output into the pickle
3. Task B find the pickle path 
4. Deserialize the pickle and feed it into Task B input 

### Case 2:  A.output (str) -> B.input (Pickle) 
_Seems like it only happens in local execution, we will fail to register this kind of workflow_
1. Run the Task A
2. Native python type transform to literal and pass to Task B
3. Extract python value from literals.scalar.primitive.value.
4. feed python value into Task B input

We can't use a transformer to convert Literal to python value since we can't know literal's python type
Task A output type is str, but Task B expect input type is Any which will fall back to PythonPickle
In this scenario, We directly extract value from scalar, collection, or map.

### Case 3:  A.output (Pickle) -> B.input (str)
_Seems like it only happens in local execution, we will fail to register this kind of workflow_
1. Task A serialize the output into a pickle
2. Task A pass the FlyteFile which is just a pickle path in IDL to Task B
3. We can just deserialize the data in the pickle file and use it. We don't need to use the transformer to translate Literal to python value here.

### Case 4: Should not write the data to the same Pickle file
Each task output should write to a unique file path.
To solve this problem, the file name will be combined with the module name, function name, and output index. 
e.g. `__main__.add_question.o0`

### Case 5: Flytekit without type annotations
```
def _contains_explicit_return(f):
            return any(isinstance(node, ast.Return) for node in ast.walk(ast.parse(inspect.getsource(f))))
```
We can use the `ast` package to find whether a task or workflow has a return value.
if task output doesn't have type annotations and has a return value, then we will use PythonPickle by default.
It can make users more quickly run a flyte workflow on Kubernetes without handling the type that flyte doesn't support.

Here is an example without type annotations, and feel free to try it.
Workflow input needs type annotations to make Flyte console work. Otherwise, Flyte Console will expect the user to have a FlyteFile input.
```

@task
def greet(name):
    return f"Hi {name}"

@task
def add_question(greeting):
    return f"{greeting} How are you?"

@workflow
def welcome(name: str):
    greeting = greet(name=name)
    return add_question(greeting=greeting)


if __name__ == "__main__":
    print(welcome(name="Kevin"))
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?
Still work in progress, will add test after I finish

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1362

## Follow-up issue
_NA_


kumare3:
Each task output should write to a unique file path.
To solve this problem, the file name will be combined with the module name, function name, and output index.
e.g. __main__.add_question.o0

You do not need this. You are using the Flyte raw output path, so just simply use the output name like o0

wild-endeavor:
Can we talk about this?  I think I may be missing something.  the original issue was to support arbitrary types, not a new type right?  so that if a user does

```
from lib.types import CustomClass

@task
def tn(in: CustomClass):
  ...
```
it will still work.  I haven't looked at the code change but this description seems to imply that the user would have to change it to

```
@task
def tn(in: PythonPickle):
  ...
```

Am I understanding it correctly?  we don't want to make the user do this, the code shouldn't have to change.

wild-endeavor:
Also let's leave the ast package out of it for now.  I think we want the user to make things explicit.

kumare3:
@wild-endeavor we already talked about this, @pingsutw is making either a new PR or updating it

pingsutw:
> Am I understanding it correctly? we don't want to make the user do this, the code shouldn't have to change.

We don't need to change anything now, and user-defined class will become pickle automatically

>  Also let's leave the ast package out of it for now. I think we want the user to make things explicit.

yeah, we've planned to remove it after meeting last week.

> Regarding "Seems like it only happens in local execution, we will fail to register this kind of workflow"... if we're going to fail at registration time, we should fail earlier during local execution.

okay, I will make a new PR to address this issue.

PR #673, opened 2021-09-27
Add cache_serialize parameter to tasks

hamersaw:
# TL;DR
Add cache_serialize parameter to tasks to enable the cache serialization API, where one a single instance of an identical (same inputs and version) task is executable at a time. This aids in reducing duplicate work for long running cachable tasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
 Populating the flyteidl TaskMetadata discoverySerializable field with cache_serialize parameter for each task type that supports it.

RFC - https://github.com/flyteorg/flyte/pull/1461

## Follow-up issue
_NA_


EngHabu (commented):
This PR needs to be updated with latest merged versions.. then I think we can merge it...

EngHabu (approved):

EngHabu:
@wild-endeavor  @eapolinario

eapolinario (commented):
@hamersaw , the change looks good, but it's missing a few unit tests to confirm the different behaviors (e.g. to confirm that the field is set in the different task types, also to cover the case of someone forgetting to set `cache`, etc). Can you add them?

eapolinario (approved):

PR #676, opened 2021-09-28
Deprecate oauth_scopes in flyte-cli setup-config in favor of new scopes field 

eapolinario:
# TL;DR
Configuration generated via `flytekit setup-config` should use the emit the credentials scopes under a new field called `scopes`. 
## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
We create a new field called `scopes` in the `flyte-cli` config file and deprecate the old one (i.e. `oauth_scopes`), while at the same time fixing the typo that prevented valid configurations from being generated.
We also ensure that we stringify all values before passing them to `ConfigParser`. 

## Tracking Issue

## Follow-up issue
_NA_
OR


wild-endeavor (approved):

flytekit/clients/raw.py
```diff
@@ -47,11 +47,11 @@ def _get_basic_flow_scopes() -> List[str]:
     deprecated_single_scope = _DEPRECATED_SCOPE.get()
     if deprecated_single_scope:
         return [deprecated_single_scope]
-    oauth_scopes = OAUTH_SCOPES.get()
-    if "openid" in oauth_scopes:
+    scopes = SCOPES.get()
```
wild-endeavor, line 50: i think we may have to do something uglier like

```suggestion
    scopes = SCOPES.get() or OAUTH_SCOPES.get()
    if scopes and "openid" in scopes:
```

wild-endeavor (approved):

PR #681, opened 2021-09-29
Remove sync singledispatch

wild-endeavor:

# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -927,7 +927,7 @@ class FlyteRemote(object):
 
     def wait(
         self,
-        execution: typing.Union[FlyteWorkflowExecution, FlyteNodeExecution, FlyteTaskExecution],
+        execution: FlyteWorkflowExecution,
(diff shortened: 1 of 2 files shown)


kumare3 (commented):
Why can't we still keep the sync method and just use an if else case to switch dispatch?

eapolinario:
> Why can't we still keep the sync method and just use an if else case to switch dispatch?

I'd argue that this makes the code less clear, I'd much rather keep `sync` in the API for backwards compatibility (people in the wild are already relying on this) and make the library code more clear. In the future we might deprecate `sync` altogether if we can reach a point where it's not needed (i.e. we'll sync whenever necessary, still TBD if we can do it in all cases without taking a huge perf hit).

PR #705, opened 2021-10-14
Concise dynamic node ids and names

kumare3:
# TL;DR
PART 1:
 - Dynamic node ids currently have additional context, namely
a shortened name of the dynamic task function
 - This causes a tremendous bloat and is completely dependent on the
name of the task
 - This PR standardizes the name and affixes a simple `d` in the
begining

PART 2:
Reduce node names, which are purely for decorative purposes

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
fixes https://github.com/lyft/flyte/issues/1660

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/common/utils.py
+++ b/flytekit/common/utils.py
@@ -11,7 +11,7 @@ from flytekit.configuration import sdk as _sdk_config
 from flytekit.models.core import identifier as _identifier
 
 
-def _dnsify(value):  # type: (str) -> str
+def _dnsify(value: str) -> str:
     """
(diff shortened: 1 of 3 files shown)


wild-endeavor:
This won't work in certain dynamic cases if users are on the old version of Admin right?  before the proper node to node relationship was done.  Should we call that out?

PR #708, opened 2021-10-18
Change `dirhash` package to `checksumdir`

martinlyra:
# TL;DR
This is a proposed fix to a compatibility issue with Windows as described in this issue: https://github.com/flyteorg/flyte/issues/1561 (with additional details in the Slack thread). In short the issue stems from an package called `dirtree` which depends on an other package by the same author; `scantree`. Since the packages are no longer maintained, `checksumdir` is proposed to replace it.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
The package `dirhash` which depends on the culprit `scantree` package is replaced by `checksumdir` in `fast_registration.py`

Unfortunately, `scantree` uses `posix` directly instead of `os` which becomes an issue with Windows. After somewhat thorough deliberation in Slack, we reached an impasse that we should replace `dirhash` with `checksumdir` and see if it would help resolve the compatibility issue.

In order to improve compatibility on Windows. An explicit `encoding="utf-8"` is added to an `open()` in `setup.py` as well.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1561

## Follow-up issue
https://github.com/flyteorg/flyte/issues/1561

### DIFF
--- a/flytekit/tools/fast_registration.py
+++ b/flytekit/tools/fast_registration.py
@@ -18,7 +18,7 @@ def compute_digest(source_dir: _os.PathLike) -> str:
     :param _os.PathLike source_dir:
     :return Text:
     """
-    return f"fast{checksumdir.dirhash(source_dir, 'md5')}"
+    return f"fast{checksumdir.dirhash(source_dir, 'md5', include_paths=True)}"
 
 
 def _write_marker(marker: _os.PathLike):


samhita-alla:
@martinlyra, thanks for creating this PR. Can we have `dirhash` removed and `checksumdir` added to the requirements.txt, dev-requirements.txt, doc-requirements.txt, setup.py and requirements-spark2.txt files?

cc: @wild-endeavor

PR #725, opened 2021-10-28
Add support FlyteFile in dataclass

pingsutw:

# TL;DR
Another way to support FlyteFile in dataclass
https://github.com/flyteorg/flytesnacks/pull/491

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Add additional metadata to serialize/deserialize FlyteFile.

## Tracking Issue

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -280,57 +280,66 @@ class DataclassTransformer(TypeTransformer[object]):
             elif dataclasses.is_dataclass(f.type):
                 self._serialize_flyte_type(v, f.type)
 
-    def _deserialize_flyte_type(self, python_val: T, expected_python_type: Type["FlyteSchema"]):
+    def _deserialize_flyte_type(self, python_val: T, expected_python_type: Type) -> T:
         from flytekit.types.directory.types import FlyteDirectory, FlyteDirToMultipartBlobTransformer
         from flytekit.types.file.file import FlyteFile, FlyteFilePathTransformer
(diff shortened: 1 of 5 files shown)


flytekit/core/type_engine.py
```diff
@@ -258,27 +267,68 @@ def _serialize_flyte_type(self, python_val: T, python_type: Type[T]):
...
-                t = FlyteSchemaTransformer()
-                t.to_python_value(
-                    FlyteContext.current_context(),
-                    Literal(scalar=Scalar(schema=Schema(v.remote_path, t._get_schema_type(f.type)))),
-                    f.type,
-                )
+            if inspect.isclass(f.type):
+                if issubclass(f.type, FlyteSchema):
+                    t = FlyteSchemaTransformer()
+                    t.to_python_value(
```
wild-endeavor, line 293: I'm actually a bit confused now.  I didn't see this before.  We just throw away the output of the to_python_value call?  Why don't we use it?

My concern is that the value is going to be an object that's hooked up to some downloading function.  https://github.com/flyteorg/flytekit/blob/742f40eb718777887043db3842b076b7f87a8427/flytekit/types/schema/types.py#L230  Doesn't this function need to be invoked for the content to actually be there?  Same for file/directory.

wild-endeavor (approved):

PR #733, opened 2021-11-02
Remote entrypoint serialize

YmirKhang:
# TL;DR
Added default EntrypointSettings to flyte remote to eliminate the RuntimeError with spark tasks. 

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description
Adapted the settings from `pyflyte package` to remote as in: https://github.com/flyteorg/flytekit/blob/751425d573d7ef7d2faa20de772b82b147083c5b/flytekit/clis/sdk_in_container/package.py#L137-L139

## Tracking Issue
https://github.com/lyft/flyte/issues/1755

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -37,13 +37,20 @@ except ImportError:
 
 from flytekit.clients.helpers import iterate_node_executions, iterate_task_executions
 from flytekit.clis.flyte_cli.main import _detect_default_config_file
+from flytekit.clis.sdk_in_container import serialize
 from flytekit.common import constants
 from flytekit.common.exceptions import user as user_exceptions
 from flytekit.common.translator import FlyteControlPlaneEntity, FlyteLocalEntity, get_serializable
 from flytekit.configuration import auth as auth_config
 from flytekit.configuration.internal import DOMAIN, PROJECT
 from flytekit.core.base_task import PythonTask
-from flytekit.core.context_manager import FlyteContextManager, ImageConfig, SerializationSettings, get_image_config
+from flytekit.core.context_manager import (
+    EntrypointSettings,
+    FlyteContextManager,
+    ImageConfig,
+    SerializationSettings,
+    get_image_config,
+)
 from flytekit.core.data_persistence import FileAccessProvider
 from flytekit.core.launch_plan import LaunchPlan
 from flytekit.core.type_engine import TypeEngine
@@ -527,6 +534,11 @@ class FlyteRemote(object):
                 self.image_config,
                 # https://github.com/flyteorg/flyte/issues/1359
                 env={internal.IMAGE.env_var: self.image_config.default_image.full},
+                entrypoint_settings=EntrypointSettings(
+                    path=os.path.join(
+                        serialize._DEFAULT_FLYTEKIT_VIRTUALENV_ROOT, serialize._DEFAULT_FLYTEKIT_RELATIVE_ENTRYPOINT_LOC
+                    )
+                ),
             ),
             entity=entity,
         )
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -137,6 +137,7 @@ class FlyteRemote(object):
         default_domain: typing.Optional[str] = None,
         config_file_path: typing.Optional[str] = None,
         grpc_credentials: typing.Optional[grpc.ChannelCredentials] = None,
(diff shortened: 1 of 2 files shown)


flytekit/remote/remote.py
```diff
@@ -527,6 +534,11 @@ def _serialize(
                 self.image_config,
                 # https://github.com/flyteorg/flyte/issues/1359
                 env={internal.IMAGE.env_var: self.image_config.default_image.full},
+                entrypoint_settings=EntrypointSettings(
+                    path=os.path.join(
+                        serialize._DEFAULT_FLYTEKIT_VIRTUALENV_ROOT, serialize._DEFAULT_FLYTEKIT_RELATIVE_ENTRYPOINT_LOC
```
wild-endeavor, line 539: I'm not too familiar with Spark - this is supposed to be where we find the python executable within the spark task image right?  should we make this configurable when we instantiate FlyteRemote?

YmirKhang: Yes, you are right. It might be a good idea to make it configurable having the same default values as pyflyte package. Other arguments may also be included:  

https://github.com/flyteorg/flytekit/blob/751425d573d7ef7d2faa20de772b82b147083c5b/flytekit/clis/sdk_in_container/package.py#L48-L110

YmirKhang:
@fg91

flytekit/remote/remote.py
YmirKhang: @wild-endeavor do you have any plans to make this work? Or any plans with the PR?

wild-endeavor: Oh I thought @fg91 was going to work on adding the option.  I can do it too if you'd like.

pingsutw: @wild-endeavor I think it's fine, we don't make it configurable in the pyflyte package as well.
https://github.com/flyteorg/flytekit/blob/2c9fab495b7daf4504a927a4077b2e6799752a4e/flytekit/clis/sdk_in_container/package.py#L137-L140
or maybe make it configurable in separate PR. There are some other variables that also need to be configurable. 
https://github.com/flyteorg/flytekit/blob/673e076a6a5cd33be281d962e3da8204d7dca6cd/flytekit/clis/sdk_in_container/serialize.py#L39-L48

wild-endeavor (dismissed):

flytekit/remote/remote.py
kumare3: @pingsutw / @YmirKhang / @wild-endeavor so the idea is that we should have reasonable defaults, with the user having the capability to override. But, when they override, they should understand what they are overriding.

Here is the way `pyflyte package` works
```python
_DEFAULT_RUNTIME_PYTHON_INTERPRETER = "/opt/venv/bin/python3"
...
    "--python-interpreter",
    default=_DEFAULT_RUNTIME_PYTHON_INTERPRETER,
...
 env_binary_path = os.path.dirname(python_interpreter)
    venv_root = os.path.dirname(env_binary_path)
...
 flytekit_virtualenv_root=venv_root,
        python_interpreter=python_interpreter,
        entrypoint_settings=context_manager.EntrypointSettings(
            path=os.path.join(venv_root, serialize._DEFAULT_FLYTEKIT_RELATIVE_ENTRYPOINT_LOC)
        ),
```

So here is my recommendation.
`FlytekitRemote` can optionally take `python_interpreter` path in the container. If not provided then we should default to
`_DEFAULT_RUNTIME_PYTHON_INTERPRETER = "/opt/venv/bin/python3"` from pyflyte package
and then use the same logic to derive the relative path to the entrypoint

kumare3:
@YmirKhang this is very close, I think this does need a small change in `Flytekit remote` constructors, to add support for in_container_python_interpreter_path, refer to my comment. 
Its derived from here - https://github.com/flyteorg/flytekit/blob/master/flytekit/clis/sdk_in_container/package.py#L16

wild-endeavor:
@YmirKhang if you don't mind giving me write access to your fork I can make this change

YmirKhang:
@wild-endeavor done

wild-endeavor:
sorry will do this today.

eapolinario (approved):

eapolinario (approved):

PR #734, opened 2021-11-02
XGBoost Plugin

samhita-alla:

# TL;DR
XGBoost Plugin for ODSC

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


plugins/flytekit-xgboost/flytekitplugins/xgboost/task.py
```diff
@@ -0,0 +1,148 @@
...
+        super(XGBoostTask, self).__init__(
+            name,
+            task_type=self._TASK_TYPE,
+            task_config=task_config,
+            interface=Interface(inputs=inputs, outputs=outputs),
+            **kwargs,
+        )
+
+    # Train method
+    def train(self, dtrain, dvalid, **kwargs) -> Tuple[str, Dict[str, Dict[str, List[float]]]]:
```
kumare3, line 74: can we use type hints please?

plugins/flytekit-xgboost/flytekitplugins/xgboost/task.py
```diff
@@ -0,0 +1,148 @@
...
+        # if validation data is provided, then populate evals and evals_result
+        booster_model = (
+            xgboost.train(
+                params=self._hyperparameters,
+                dtrain=dtrain,
+                **asdict(self._train_parameters if self._train_parameters else TrainParameters()),
+                evals=[(dvalid, "validation")],
+                evals_result=evals_result,
+            )
+            if dvalid
```
kumare3, line 85: lets make it
if dvalid:
   ....
else:
   ...

Much easier to read

plugins/flytekit-xgboost/flytekitplugins/xgboost/task.py
```diff
@@ -0,0 +1,148 @@
...
+        if not (train_key and test_key):
+            raise ValueError("Must have train and test inputs")
+
+        dataset_vars = {}
+
+        for each_key in [train_key, test_key, validation_key]:
+            if each_key:
+                dataset = kwargs[each_key]
+                if issubclass(self.python_interface.inputs[each_key], str):
+                    if "csv" in dataset:
```
kumare3, line 125: can we make this FlyteFile, and this this should be simpler to use, you will not need to do `csv` in dataset?

plugins/flytekit-xgboost/flytekitplugins/xgboost/task.py
```diff
@@ -0,0 +1,148 @@
...
+                        dataset_vars[each_key] = xgboost.DMatrix(
+                            dataset + "?format=csv&label_column=" + str(self._label_column)
+                        )
+                    else:
+                        dataset_vars[each_key] = xgboost.DMatrix(dataset)
+                elif issubclass(self.python_interface.inputs[each_key], FlyteFile):
+                    if FlyteContext.current_context().file_access.is_remote(dataset):
```
kumare3, line 132: i dont think you should do this, you should simply 
```python
with open(filename) as f:
    ....
```

plugins/flytekit-xgboost/tests/test_task.py
```diff
@@ -0,0 +1,176 @@
...
+    assert task_obj.python_interface.inputs == {"train": FlyteFile, "test": FlyteFile}
+    train_test_wf()
+
+
+def test_csv_data():
+    """
+    Note: For CSV training, the algorithm assumes that the CSV does not have a header record.
+    To set the target variable, set the label_column parameter, which by default is 0.
+    """
+    task_obj = XGBoostTask(
```
kumare3, line 63:
```suggestion
    xgboost_trainer = XGBoostTask(
```

plugins/flytekit-xgboost/tests/test_task.py
```diff
@@ -0,0 +1,176 @@
...
+        name="test2",
+        hyperparameters=asdict(CSVHyperparameters()),
+        inputs=kwtypes(train_data=CSVFile, test_data=CSVFile),
+        label_column=0,
+    )
+
+    @task
+    def partition_data(dataset: str) -> Tuple[CSVFile, CSVFile]:
```
kumare3, line 71: why CSV, why not pandas dataframes or Better FlyteSchema

kumare3:
I am not sure why we are using CSV Files. My preference is that we use FlyteSchema and then load CSV into FlyteSchema or have a task to transform it. That being said cc @wild-endeavor we should probably make CSV a valid type of schema?

plugins/flytekit-xgboost/flytekitplugins/xgboost/task.py
samhita-alla: I have included both string and FlyteFile as valid types for train and test data. Do you want me to remove the string data type?

plugins/flytekit-xgboost/tests/test_task.py
samhita-alla: We can also use FlyteSchema; that way, we could support FlyteFile and FlyteSchema as valid data types. But why do we want to convert CSV to Schema internally if we have CSVFile as the supported Flyte type? XGBoost supports CSV as well.

plugins/flytekit-xgboost/flytekitplugins/xgboost/task.py
samhita-alla: Since XGBoost accepts CSV, I think we can directly send the CSV file path. The same check is implemented in the [FlyteFile code](https://github.com/flyteorg/flytekit/blob/17201ce9725558f7c88e1f07ec4c97bc0ea31b6b/flytekit/types/file/file.py#L342) as well.

kumare3: Even if it is implemented in Flytefile, we may changed it. Can we simply use - https://github.com/flyteorg/flytekit/blob/17201ce9725558f7c88e1f07ec4c97bc0ea31b6b/flytekit/types/file/file.py#L208

dataset.download()? and that will give the path

samhita-alla: I have updated the code. I'm using `with open(dataset):` instead of download() because download() doesn't work with locally available FlyteFiles.

samhita-alla:
@kumare3, let me know if everything looks okay. I can then copy this to the ODSC repo.

PR #738, opened 2021-11-04
Cookie cutter initial template

eapolinario:
# TL;DR
Add a command to generate the boilerplate for flyte-ready repos.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
This PR defines a new `pyflyte` command called `init` that takes a template and a project name as arguments. These are used to call the template engine provided by [cookiecutter](https://cookiecutter.readthedocs.io/en/1.7.2/).

In terms of where templates are stored, we defined a cookiecutter template in https://github.com/flyteorg/flytekit-python-template/pull/16. Note how we have the ability to keep multiple templates in the same repo, the idea being that we're going to expose them using different templates under the `pyflyte init` command.

The `pyflyte init` command defaults to "simple-example" cookiecutter template and this template prompts the user for 2 additional inputs: 
1. app: which is the top-level directory under `project-name` that will store the actual workflow
2. workflow: the name of the example workflow.

All inputs have default values.

Also worth noting that the command does not clobber existing user code, i.e. the command fails if the user attempts to overwrite a directory by re-running the command.

Here's how the help provided by `click` looks like:

```
❯ pyflyte init --help
...
Usage: pyflyte init [OPTIONS] PROJECT_NAME

  Create flyte-ready projects.

Options:
  --template TEXT  cookiecutter template to be used
  --help           Show this message and exit.
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1799

## Follow-up issue
_NA_


flytekit/clis/sdk_in_container/init.py
```diff
@@ -0,0 +1,60 @@
...
+def init():
+    """
+    Create flyte-ready projects.
+    """
+    # TODO: Add `interactive` boolean flag to allow for programmatic access to this subcommand.
+    pass
+
+
+@click.command("example")
+def generate_simple_example():
```
kumare3, line 16: Can we not create dynamic
Commands that automatically create different examples

eapolinario: Can you say more? Maybe give an example.

eapolinario:
Unit test is failing. Fix in https://github.com/flyteorg/flytekit/pull/739.

flytekit/clis/sdk_in_container/init.py
```diff
@@ -0,0 +1,60 @@
...
+        checkout="cookie-cutter",
+        no_input=True,
+        # We do not want to clobber existing files/directories.
+        overwrite_if_exists=False,
+        extra_context=config,
+        # By specifying directory we can have multiple templates in the same repository,
+        # as described in https://cookiecutter.readthedocs.io/en/1.7.2/advanced/directories.html.
+        # The idea is to extend the number of templates, each in their own subdirectory, for example
+        # a tensorflow-based example.
+        directory="simple-example",
```
kumare3, line 56: This is what I mean. This is hard coded. Or are you thinking of parameterizing this? I was suggesting that you could create
`pyflyte init tensorflow` `pyflyte init xyz`
Or otherwise we should call it
`pyflyte init <project-name> --template=tensorflow`
and then --template can be defaulted to `simple-example`?

eapolinario: I was thinking of parametrizing this going forward, but I really liked the suggestion of making `project-name` part of the invocation. I'll go ahead and implement both suggestions.

eapolinario: I also updated the PR description.

flytekit/clis/sdk_in_container/init.py
```diff
@@ -0,0 +1,38 @@
...
+def init(template, project_name):
+    """
+    Create flyte-ready projects.
+    """
+    click.echo(
+        "What should we call your application? This serves as the top level package where your workflows will live."
+    )
+    app = read_user_variable("app", "myapp")
+    click.echo("What should be the name of your example workflow?")
+    workflow_name = read_user_variable("workflow", "workflow_example")
```
kumare3, line 18: Should we just use click for consistency http://pallets-click.readthedocs.io/en/stable/prompts.html

eapolinario: using cookiecutter just for the templating feels less cluttered. Thanks for the suggestion!

eapolinario:
Only the code coverage tasks are failing.

flytekit/clis/sdk_in_container/init.py
```diff
@@ -0,0 +1,32 @@
+import click
+from cookiecutter.main import cookiecutter
+
+
+@click.command("init")
+@click.option("--template", default="simple-example", help="cookiecutter template to be used")
```
kumare3, line 6:
```suggestion
@click.option("--template", default="simple-example", help="cookiecutter template folder name to be used in the repo - https://github.com/flyteorg/flytekit-python-template.git")
```

Also eventually I think the repo itself should be parameterizable

kumare3:
Small comment then +1

flytekit/clis/sdk_in_container/init.py
eapolinario: this idea of exposing the repo is interesting. So your expectation is that people will have their own private (or public forks) of the repo? In any case, if the need arises we can definitely go that way, but not in this PR, ok?

kumare3 (dismissed):

flytekit/clis/sdk_in_container/init.py
```diff
@@ -0,0 +1,36 @@
...
+        overwrite_if_exists=False,
+        extra_context=config,
+        # By specifying directory we can have multiple templates in the same repository,
+        # as described in https://cookiecutter.readthedocs.io/en/1.7.2/advanced/directories.html.
+        # The idea is to extend the number of templates, each in their own subdirectory, for example
+        # a tensorflow-based example.
+        directory=template,
+    )
+
+    click.echo(f"Visit the {project_name} directory and follow the next steps in the Getting started guide.")
```
kumare3, line 36: docs.flyte.org/...

wild-endeavor (approved):

PR #745, opened 2021-11-10
Add module name parameter for workflows and tasks

YmirKhang:
# TL;DR
This PR adds an optional parameter to workflow and task types to make the module name explicitly available.

```python
@task(
    cache_version="1",
    container_image="{{.images.default.fqn}}:{{.images.default.version}}",
    module_name="<app.workflows.workflow_file>",
)
def my_task() -> None:
...

...
@workflow(module_name="<app.workflows.workflow_file>")
def my_workflow() -> None:
```

So that it can be avoided that the workflows and tasks default to `__main__.my_task` and `__main__.my_workflow`, by overriding module name in tasks and workflows

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Added extra optional parameter `module_name` to workflow and task definitions.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1813

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/python_function_task.py
+++ b/flytekit/core/python_function_task.py
@@ -112,6 +112,8 @@ class PythonFunctionTask(PythonAutoContainerTask[T]):
         :param Optional[ExecutionBehavior] execution_mode: Defines how the execution should behave, for example
             executing normally or specially handling a dynamic case.
         :param Optional[TaskResolverMixin] task_type: String task type to be associated with this Task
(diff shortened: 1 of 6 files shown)


YmirKhang:
This is just an implementation to make it work with a self contained file. This was not tested, also other side effects are not considered. But since the parameter is optional it should be backward compatible.

PR #749, opened 2021-11-11
[WIP] Schema overhaul

pingsutw:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
image

## Tracking Issue

## Follow-up issue
_NA_
OR


dataset_example.py
```diff
@@ -0,0 +1,162 @@
...
+    # bq -> pandas -> s3 (parquet)
+    df = dataset.open_as(pd.DataFrame)
+    return df
+
+
+@task
+def t7(df1: pd.DataFrame, df2: pd.DataFrame) -> (FlyteDataset[column_type], FlyteDataset[column_type]):
+    # df1: pandas -> bq
+    # df2: pandas -> s3 (parquet)
+    return FlyteDataset(dataframe=df1, remote_path=BQ_PATH, file_format=SchemaFormat.BIGQUERY), df2
```
pingsutw, line 87: BIGQUERY is a storage location, but we still need to set a format for `StructuredDatasetMetadata.format`. 
The name need to be discussed

dataset_example.py
```diff
@@ -0,0 +1,161 @@
...
+    StructuredDataset,
+)
+
+PANDAS_PATH = "/tmp/pandas.pq"
+NUMPY_PATH = "/tmp/numpy.pq"
+BQ_PATH = "bq://photo-313016:flyte.new_table3"
+
+# https://github.com/flyteorg/flyte/issues/523
+# We should also support map and list in schema column
+column_type = kwtypes(w=typing.Dict[str, typing.Dict[str, int]], x=typing.List[typing.List[int]], y=int, z=str)
```
wild-endeavor, line 23:
```suggestion
my_cols = kwtypes(w=typing.Dict[str, typing.Dict[str, int]], x=typing.List[typing.List[int]], y=int, z=str)
```
just to make it clear

dataset_example.py
```diff
@@ -0,0 +1,161 @@
...
+    # S3 (parquet) -> Pandas -> S3 (parquet)
+    return dataframe
+
+
+@task
+def t3(dataset: StructuredDataset[column_type]) -> StructuredDataset[column_type]:
+    # s3 (parquet) -> pandas -> s3 (parquet)
+    print("Pandas dataframe")
+    print(dataset.open_as(pd.DataFrame))
+    return dataset
```
wild-endeavor, line 54: let's add a comment saying that this should not upload anything, since we're returning just the input object.

flytekit/models/core/types.py
```diff
@@ -1,3 +1,6 @@
+from __future__ import annotations
+
+import json as _json
```
wild-endeavor, line 3: why do we need to import json if it's not used.  also let's move away from the `_` prefix.  i don't think it's ever prevented me from making a bug.

flytekit/types/structured/structured_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+
+    @property
+    def file_format(self) -> DatasetFormat:
+        return self._file_format
+
+    def open_as(self, df_type: Type) -> typing.Any:
+        return FLYTE_DATASET_TRANSFORMER.download(self.file_format, df_type, self.remote_path)
+
+
+class DatasetEncodingHandler:
```
wild-endeavor, line 110: these should be abstractbaseclasses

flytekit/types/structured/arrow.py
```diff
@@ -0,0 +1,68 @@
...
+class ArrowToParquetPersistenceHandlers(DatasetPersistenceHandler, ABC):
+    def persist(
+        self,
+        table: pa.Table,
+        to_file: os.PathLike,
+    ):
+        pq.write_table(table, to_file)
+
+
+class ArrowToBQPersistenceHandlers(DatasetPersistenceHandler, ABC):
```
wild-endeavor, line 30: These should not be ABCs - these are the concrete classes.

flytekit/types/structured/arrow.py
```diff
@@ -0,0 +1,68 @@
...
+    def persist(self, table: pa.Table, path: str, job_config: typing.Optional[LoadJobConfig] = None) -> LoadJob:
+        table_id = path.split("://", 1)[1].replace(":", ".")
+        client = bigquery.Client()
+        return client.load_table_from_dataframe(table.to_pandas(), table_id, job_config=job_config)
+
+
+class ParquetToArrowRetrievalHandler(DatasetRetrievalHandler, ABC):
```
wild-endeavor, line 37: same here, not an ABC.

flytekit/types/structured/arrow.py
```diff
@@ -0,0 +1,68 @@
...
+    def retrieve(self, path: str) -> pa.Table:
+        return pq.read_table(path)
+
+
+class BQToArrowRetrievalHandler(DatasetRetrievalHandler, ABC):
```
wild-endeavor, line 42: and here.

flytekit/types/structured/structured_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+            self._REGISTER_TYPES.append(from_type)
+            TypeEngine.override_transformer(self, from_type)
+
+        if to_type not in self._REGISTER_TYPES:
+            self._REGISTER_TYPES.append(to_type)
+            TypeEngine.override_transformer(self, to_type)
+
+        registry: dict
+
+        if isinstance(h, DatasetRetrievalHandler):
```
wild-endeavor, line 274: i think we should make these issubclass.  See the other comment about registering classes instead of instances of the class.

flytekit/types/structured/arrow.py
```diff
@@ -0,0 +1,68 @@
...
+
+        stream = read_session.streams[0]
+        reader = client.read_rows(stream.name)
+        frames = []
+        for message in reader.rows().pages:
+            frames.append(message.to_dataframe())
+        return pa.Table.from_pandas(pd.concat(frames))
+
+
+FLYTE_DATASET_TRANSFORMER.register_handler(DatasetFormat.PARQUET, pa.Table, ParquetToArrowRetrievalHandler())
```
wild-endeavor, line 64: I think we should follow the same pattern as in the TypeEngine itself, unless there's a reason for doing it this way.  In the TypeEngine, plugin authors register the class, not an instance of the class.  So I feel like ParquetToArrowRetrievalHandler should be without the `()`.

But kinda your call - ignore the other comment if you want to keep it this way.  But if we keep it this way we should add a comment explaining why or at least making it clear to plugin contributors that this is different than the main TypeEngine.

flytekit/types/structured/arrow.py
wild-endeavor, line 42: after we clean this up a bit more and have a working example, let's send this to dylan and see what he thinks.

flytekit/types/structured/structured_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+    return a flyte Literal.
+
+    Flytekit ships with default handlers that know how to:
+
+    - Write Parquet files to AWS/GCS
+    - Write pandas DataFrame objects directly to BigQuery
+    - Write Arrow objects/files to AWS/GCS/BigQuery
+    """
+
+    def persist(self, *args, **kwargs):
```
wild-endeavor, line 142: `@abstractmethod`

flytekit/types/structured/structured_dataset.py
```diff
@@ -0,0 +1,380 @@
...
+    def _get_dataset_column_literal_type(self, t: Type):
+        if t in self._SUPPORTED_TYPES:
+            return self._SUPPORTED_TYPES[t]
+        if hasattr(t, "__origin__") and t.__origin__ == list:
+            return type_models.LiteralType(collection_type=self._get_dataset_column_literal_type(t.__args__[0]))
+        if hasattr(t, "__origin__") and t.__origin__ == dict:
+            return type_models.LiteralType(map_value_type=self._get_dataset_column_literal_type(t.__args__[1]))
+        raise AssertionError(f"type {t} is currently not supported by StructuredDataset")
+
+    def _get_dataset_type(self, t: Type[StructuredDataset]) -> StructuredDatasetType:
```
wild-endeavor, line 245: The type signature here isn't right though.  It should be `Union[Type[StructuredDataset], typing.Any]`.  The goal of this function, and of the caller (`get_literal_type`) is that it takes in a Python type, and returns the Flyte type.  The python type can be a `StructuredDataset` (with or without columns, (or perhaps Annotated)) but it can also be a `pd.DataFrame` or a `some_future_lib.DataFrame`.  The idea here is that the translation from the Python type to the Flyte type needs to be redirected to the registered handlers somehow - this is the part that I didn't think through.

But like,
```
TypeEngine.to_literal_type(my_custom.DataFrame)
```
should fail with the stock flytekit.  But after the user calls
```
FLYTE_DATASET_TRANSFORMER.register_handler(MyCustomEncodingHandler)
# and, or
FLYTE_DATASET_TRANSFORMER.register_handler(MyCustomPersistenceHandler)
```
then the call to the type engine should succeed.

flytekit/models/core/types.py
pingsutw: sorry, forgot to remove it.

flytekit/types/structured/structured_dataset.py
pingsutw: For custom dataframe, I think we have a few ways to translate it to FlyteType.
1. Users must use StructuredDataset with the column in the input/output signature.
2. Users must tell us how to convert the dataframe to arrow or parquet, and then we can convert arrow or parquet data column type to Flyte Type
3. Users implement `to_literal_type` by themself

dataset_example.py
```diff
@@ -0,0 +1,178 @@
...
+def t6(dataset: StructuredDataset[my_cols]) -> pd.DataFrame:
+    # bq -> pandas -> s3 (parquet)
+    df = dataset.open_as(pd.DataFrame)
+    return df
+
+
+@task
+def t7(df1: pd.DataFrame, df2: pd.DataFrame) -> (StructuredDataset[my_cols], StructuredDataset[my_cols]):
+    # df1: pandas -> bq
+    # df2: pandas -> s3 (parquet)
```
wild-endeavor, line 81: _maybe_ another one
```
    # df1: pandas -> bq (streamed)
    # df2: pandas -> bq (downloaded to parquet file first)
```

.github/workflows/pythonbuild.yml
```diff
@@ -93,6 +95,8 @@ jobs:
         cd plugins/${{ matrix.plugin-names }}
         pip install -e .
         pip install --no-deps -U https://github.com/flyteorg/flytekit/archive/${{ github.sha }}.zip#egg=flytekit
+        git clone https://github.com/pingsutw/flyteidl.git && cd flyteidl && git checkout new-schema && pip install . && cd ..
+        pip install fsspec
```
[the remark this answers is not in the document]
pingsutw: removed it

flytekit/types/schema/types.py
```diff
@@ -24,6 +24,7 @@ class SchemaFormat(Enum):
     """
 
     PARQUET = "parquet"
+    BIGQUERY = "bigquery"
```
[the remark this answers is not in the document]
pingsutw: Updated it.

PR #751, opened 2021-11-16
Fix subworkflow and launch plan FlyteRemote behavior

wild-endeavor:
# TL;DR
Update the `FlyteRemote` experience when working with subworkflows and launch plans, including object navigation when dealing with subworkflows and launch plans run through dynamic tasks.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

### Logical changes
See the changes to the `sync_node_execution` - most of the logic is in there.

* Change the way we get each node's interface - we now pull it from the associated node's `flyte_entity` instead of traversing a `FlyteWorkflow`.
* Removed the `get_node_execution_interface` function from FlyteRemote.
* Added `promote_from_closure` to `FlyteWorkflow` since that's what we get from Admin.  Expects launch plans to be passed in as a separate argument though (which aren't included in the Admin response) - didn't want to have to pass a client.
* Dynamic workflows are "promoted" from the `CompiledWorkflowClosure` returned in the `DynamicWorkflowNodeMetadata`.

### Refactoring changes
* Moved all the execution classes into one file.  `FlyteTaskExecution`, `FlyteWorkflowExecution`, `FlyteNodeExecution`
* Deleted all the identifier remote wrapper classes.  It's easier to just use base model classes for now.  The wrapper classes weren't really buying us anything.
* Removed the separate `tasks` folder. Moved the task.py file up one level.
* Removed the `get_sub_workflows` function from FlyteWorkflow. This was only used in the past to assemble subworkflows for registration to Admin.  This was copied over from `SdkWorkflow`.  Since admin should return us all subworkflows, there's no need for this anymore.
* Refactored constructor for `FlyteNode` and removed some old copied functions from the class.

### Model changes
There were model file updates that were needed as well.
* `CatalogArtifactTag`, `CatalogMetadata` added.
* `dynamic_workflow` field added to `NodeExecutionGetDataResponse ` pointing to the new `DynamicWorkflowNodeMetadata`.
* `TaskNodeMetadata` and `WorkflowNodeMetadata` added for the `target_metadata` in the `NodeExecutionClosure`

Changes are also in https://github.com/flyteorg/flytekit/pull/756/files in an attempt to try to keep the model cleanup up to date.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1482


flytekit/remote/workflow.py
```diff
@@ -1,16 +1,17 @@
+from __future__ import annotations
```
eapolinario, line 1: we don't need this, right?

flytekit/remote/nodes.py
```diff
@@ -50,15 +43,20 @@ def __init__(
         elif flyte_launch_plan is not None:
             workflow_node = _component_nodes.FlyteWorkflowNode(flyte_launch_plan=flyte_launch_plan)
 
+        task_node = None
+        if flyte_task:
+            task_node = _component_nodes.FlyteTaskNode(flyte_task)
+        branch_node = None
+
         super(FlyteNode, self).__init__(
-            id=_dnsify(id) if id else None,
```
eapolinario, line 54: why we don't need to `_dnsify` anymore?

eapolinario (commented):
This is pretty awesome. I have just a few minor questions.

flytekit/remote/nodes.py
```diff
@@ -1,24 +1,18 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
wild-endeavor: why not?  it looks better

flytekit/remote/workflow.py
wild-endeavor: nope, but i like it.  so that i don't have to do -> "MyClass"

wild-endeavor: https://www.python.org/dev/peps/pep-0563 search for "Annotations using nested classes and their respective state are still valid. They"

flytekit/remote/nodes.py
wild-endeavor: a lot of this code was copied from the SdkXyz classes, which were used on both read and write.  remove cuz before the id came from user input basically, now we're reading from admin, so the id should be the id, we shouldn't mess with it.

eapolinario (approved):

PR #752, opened 2021-11-16
Add support for architecture

akhurana001:
SDK changes to support architecture for container/images : Ref: https://github.com/flyteorg/flyte/issues/1539


pingsutw:
@akhurana001 Thank you for working on this.
Could you sign the commit? thanks!
https://github.com/flyteorg/flytekit/pull/752/checks?check_run_id=4231839088

eapolinario:
The snyk error seems to be an issue on their side. According to https://support.snyk.io/hc/en-us/articles/360004358517-Unknown-PR-test-error it should be temporary and we can force a re-run by closing and opening the PR again.

eapolinario:
Retrying snyk tests just once more.

eapolinario:
Retrying snyk tests again.

flytekit/models/task.py
```diff
@@ -854,10 +854,12 @@ def __init__(self, image, command, args, resources, env, config, data_loading_co
         :param Resources resources: A definition of requisite compute resources.
         :param dict[Text, Text] env: A definition of key-value pairs for environment variables.
         :param dict[Text, Text] config: A definition of configuration key-value pairs.
+        : param Architecture: Architecture supported by this container's image.
```
wild-endeavor, line 857:
```suggestion
        :param architecture: Architecture supported by this container's image.
```

flytekit/models/task.py
```diff
@@ -843,7 +843,7 @@ def from_flyte_idl(cls, pb2: _core_task.DataLoadingConfig) -> "DataLoadingConfig
 
 
 class Container(_common.FlyteIdlEntity):
-    def __init__(self, image, command, args, resources, env, config, data_loading_config=None):
+    def __init__(self, image, command, args, resources, env, config, architecture=None, data_loading_config=None):
```
wild-endeavor, line 846: add type hint here? 
```suggestion
    def __init__(self, image, command, args, resources, env, config, architecture: typing.Optional[int] = None, data_loading_config=None):
```

flytekit/models/task.py
```diff
@@ -872,6 +874,14 @@ def image(self):
         """
         return self._image
 
+    @property
+    def architecture(self):
```
wild-endeavor, line 878:
```suggestion
    def architecture(self) -> int:
```
or `typing.Optional[int]`

flytekit/models/task.py
```diff
@@ -872,6 +874,14 @@ def image(self):
...
+        """
+        Architecures supported by the container's image.
+        :rtype: Text
```
wild-endeavor, line 881: shouldn't this be int?  Also we can just remove this line now that we have mypy style hints.  https://github.com/flyteorg/flyteidl/pull/226/files#diff-8bd2613bf781125f77247c29a2f5502658c63cccf2b597a9ec3695ef1e7b2ca9R188

flytekit/models/task.py
wild-endeavor: or maybe we should just make the default 0/unknown and always fill it in.  either way as long as it works with the backend.

flytekit/models/task.py
reverson: Done

flytekit/models/task.py
reverson: Updated with int

flytekit/models/task.py
reverson: Defaulted to 0, the backend changes check for 0/Unknown and don't do anything in that case, so this should work.

flytekit/models/task.py
reverson: Done

PR #754, opened 2021-11-17
Add support union type

pingsutw:

# TL;DR
Add union type transformer, and it can handle both union and optional type

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1349

## Follow-up issue
No

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -626,11 +626,11 @@ class UnionTransformer(TypeTransformer[T]):
         return TypeEngine.to_literal(ctx, python_val, t, expected)
 
     def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> Optional[typing.Any]:
+        if lv is None:
+            return None
         st = self.get_sub_type(expected_python_type)
         for v in st:
             try:
-                if v == type(None) and lv is None:
-                    return None
                 val = TypeEngine.to_python_value(ctx, lv, v)
                 if val:
                     return val
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -444,6 +444,8 @@ class TypeEngine(typing.Generic[T]):
         """
         Converts a python value of a given type and expected ``LiteralType`` into a resolved ``Literal`` value.
         """
+        if python_val is None and expected.union_type is None:
+            raise AssertionError(f"Python value cannot be None, expected {python_type}/{expected}")
         transformer = cls.get_transformer(python_type)
         if transformer.type_assertions_enabled:
             transformer.assert_type(python_type, python_val)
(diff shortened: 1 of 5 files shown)


flytekit/core/type_engine.py
```diff
@@ -596,6 +596,58 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+class UnionTransformer(TypeTransformer[T]):
+    """
+    Transformer that handles a univariate typing.Union[T]
+    """
+
+    def __init__(self):
+        super().__init__("Typed Union", typing.Union)
+
+    @staticmethod
+    def get_sub_type(t: Type[T]) -> Type[T]:
```
maximsmol, line 608: this method does not really make sense since a Union type only ever has multiple "sub types"

i.e. `Union[x]` is by definition equivalent to `x`

I think the way this method is used in `get_literal_type` implies that the type signature should actually be `(t: Type[T]) -> List[Type]` instead

flytekit/core/type_engine.py
```diff
@@ -596,6 +596,58 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+        """
+        Return the generic Type T of the Union
+        """
+        if hasattr(t, "__origin__") and t.__origin__ is typing.Union:  # type: ignore
+            if hasattr(t, "__args__"):
+                return t.__args__  # type: ignore
+        raise ValueError("Only generic univariate typing.Union[T] type is supported.")
```
maximsmol, line 615: see above: only *poly*variate Union types are meaningful

flytekit/core/type_engine.py
```diff
@@ -596,6 +596,58 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+        raise TypeError(f"Cannot convert from {python_val} to {python_type}")
+
+    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> Optional[typing.Any]:
+        if lv is None:
+            return None
+        st = self.get_sub_type(expected_python_type)
+        for v in st:
+            try:
+                val = TypeEngine.to_python_value(ctx, lv, v)
+                if val:
```
maximsmol, line 639: `val` could be `None` if this is an Optional. `None` should not indicate failure

maximsmol (changes requested):

flytekit/core/type_engine.py
pingsutw: `val`  will be `None` only when `Literal` is `None`. if `Literal` is None, we will return None in line 634.
So we should return `val` if and only if `val` has value here, otherwise, we should raise an error.

flytekit/core/type_engine.py
pingsutw: will update it, thanks

EngHabu:
Closing in favor of https://github.com/flyteorg/flytekit/pull/763

PR #759, opened 2021-12-01
TypeAnnotation

kennyworkman:
# TL;DR

Use new flyte core object to saturate an Annotation field on flyte literal types

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added
 - [] Code documentation added

## Complete description

## Tracking Issue

https://github.com/flyteorg/flyte/pull/1855

## Follow-up issue
_NA_


flytekit/core/type_engine.py
```diff
@@ -437,7 +447,29 @@ def to_literal_type(cls, python_type: Type) -> LiteralType:
         Converts a python type into a flyte specific ``LiteralType``
         """
         transformer = cls.get_transformer(python_type)
-        return transformer.get_literal_type(python_type)
+        res = transformer.get_literal_type(python_type)
+        data = None
+        if hasattr(python_type, "__metadata__"):
```
eapolinario, line 452: instead, let's use [get_args](https://docs.python.org/3/library/typing.html#typing.get_args) to list the annotations.

flytekit/core/type_engine.py
```diff
@@ -401,14 +403,22 @@ def get_transformer(cls, python_type: Type) -> TypeTransformer[T]:
...
         # Step 1
         if python_type in cls._REGISTRY:
             return cls._REGISTRY[python_type]
 
         # Step 2
         if hasattr(python_type, "__origin__"):
+
+            # Handling of annotated generics, eg:
+            # typing.Annotated[typing.List[int], 'foo']
+            if isinstance(python_type, typing._AnnotatedAlias):
```
eapolinario, line 416: we should use the [get_origin](https://docs.python.org/3/library/typing.html#typing.get_origin) and check if the resulting value is `typing.Annotated`. For example:

```
>>> get_origin(Annotated[List[int], 'annotation_0', 'annotation_1']) is Annotated
True
>>> get_origin(List[int]) is Annotated
False
>>> get_origin(int) is Annotated
False
>>> get_origin(Annotated[int, 'a', 42]) is Annotated
True
>>> get_origin(Annotated[Dict[str, int], 'dict_annotation']) is Annotated
True
```

flytekit/core/type_engine.py
```diff
@@ -401,14 +403,22 @@ def get_transformer(cls, python_type: Type) -> TypeTransformer[T]:
...
+                return cls.get_transformer(python_type.__origin__)
```
eapolinario, line 417: I know we've been using the `__origin__` dunder attribute in a few places, but let's take this opportunity to use [get_args](https://docs.python.org/3/library/typing.html#typing.get_args) to access the annotations.

eapolinario (commented):
Thanks for making this contribution! I left a few comments, but nothing too major. Overall this is shaping up to be really amazing.

flytekit/core/type_engine.py
kennyworkman: is there a reason? is there a case where checking for `typing._AnnotatedAlias` would fail but this approach would catch it? or conversely any false positives that this approach would avoid? or is strictly style

flytekit/core/type_engine.py
kennyworkman: same here, scanning the documentation its unclear what is more "correct" and both do the same thing, so would like to understand why ( for my own understanding )

flytekit/core/type_engine.py
kennyworkman: sure but why

eapolinario: just to bullet-proof this code. dunder methods/attributes are not guaranteed to stay consistent between versions, as per https://docs.python.org/3/reference/lexical_analysis.html#reserved-classes-of-identifiers.

flytekit/core/type_engine.py
eapolinario: again, the idea is to bullet-proof this code. If you look at the [list of symbols exported](https://github.com/python/cpython/blob/3.9/Lib/typing.py#L32-L121) by the typing module, `_AnnotatedAlias` is not present.

flytekit/core/type_engine.py
eapolinario: This one I'm not too particularly passionate about since we use in a lot of other places already.

flytekit/core/type_engine.py
kennyworkman: ok that makes sense

flytekit/core/type_engine.py
kennyworkman: also makes sense

flytekit/core/type_engine.py
kennyworkman: will change because its dunder and didn't know those could break

kennyworkman:
@eapolinario comments have been addressed

eapolinario:
Will take another look after we merge https://github.com/flyteorg/flyteidl/pull/232 and incorporate it to this PR.

flytekit/core/annotation.py
```diff
@@ -0,0 +1,27 @@
...
+    into fields on the flyteidl type literals. This data is not accessible at
+    runtime but rather can be retrieved from flyteadmin for custom presentation
+    of typed parameters.
+
+    For a task definition:
+
+    .. code-block:: python
+
+        @task
+        def x(a: typing.Annotated[int, FlyteAnnotation({"foo": {"bar": 1}})]):
```
wild-endeavor, line 17: can we clarify in this docstring and in code if there can be more than one of these?  Should we cap it to one?

flytekit/core/annotation.py
```diff
@@ -0,0 +1,27 @@
...
+            return
+
+    """
+
+    def __init__(self, data: Dict[str, Any]):
+        self._data = data
+
+    @property
+    def data(self):
```
wild-endeavor, line 26: should we call this `annotations` to match the IDL?  Or leave it `data`?

flytekit/core/type_engine.py
```diff
@@ -437,7 +447,29 @@ def to_literal_type(cls, python_type: Type) -> LiteralType:
...
         transformer = cls.get_transformer(python_type)
-        return transformer.get_literal_type(python_type)
+        res = transformer.get_literal_type(python_type)
+        data = None
+        if typing.get_origin(python_type) is typing.Annotated:
+            for x in typing.get_args(python_type)[1:]:
+                if not isinstance(x, FlyteAnnotation):
+                    continue
+                if x.data.get("__consumed", False):
+                    continue
```
wild-endeavor, line 457: can we add a comment as to why we need a check at all?  If we're just doing a straight loop, why would we ever be in danger of reading something more than once?

flytekit/core/type_engine.py
```diff
@@ -437,7 +447,29 @@ def to_literal_type(cls, python_type: Type) -> LiteralType:
...
+                data = x.data
+                x.data["__consumed"] = True
+        if data is not None:
+            idl_type_annotation = _annotation_model(annotations=data)
+            return LiteralType(
```
wild-endeavor, line 462: these are python objects - it should be safe to just attach the annotation result to the existing `res` object right?

flytekit/models/annotation.py
```diff
@@ -0,0 +1,45 @@
+import json as _json
+from typing import Any, Dict
+
+from flyteidl.core import types_pb2 as _types_pb2
+from google.protobuf import json_format as _json_format
+from google.protobuf import struct_pb2 as _struct
+
+
+class TypeAnnotation:
```
wild-endeavor, line 9: could we add the kinda dumb but _surprisingly_ helpful there-and-back-again unit test for this class?  https://github.com/flyteorg/flytekit/blob/742f40eb718777887043db3842b076b7f87a8427/tests/flytekit/unit/models/core/test_types.py#L19 as an example

tests/flytekit/unit/core/test_typing_annotation.py
```diff
@@ -0,0 +1,61 @@
...
+    project="project",
+    domain="domain",
+    version="version",
+    env=None,
+    image_config=ImageConfig(default_image=default_img, images=[default_img]),
+)
+entity_mapping = OrderedDict()
+
+
+@task
```
wild-endeavor, line 22: can we add some unit tests to `test_type_engine.py` that just does the type conversion without the @task part?  Could we do one of a primitive, one of a typing.List, one of a typing.Dict, and one of a dataclass or something complex?

flytekit/core/type_engine.py
```diff
@@ -437,7 +447,29 @@ def to_literal_type(cls, python_type: Type) -> LiteralType:
         Converts a python type into a flyte specific ``LiteralType``
         """
         transformer = cls.get_transformer(python_type)
-        return transformer.get_literal_type(python_type)
+        res = transformer.get_literal_type(python_type)
+        data = None
+        if typing.get_origin(python_type) is typing.Annotated:
```
wild-endeavor, line 452: get_origin behavior is different between python 3.7/8 and 3.9.

In python3.8
```
In [1]: from typing_extensions import Annotated

In [2]: class A():
   ...:     ...
   ...:

In [3]: ii = Annotated[int, A()]

In [4]: import typing

In [5]: typing.get_origin(ii)
Out[5]: int
```

In 3.9
```
>>> import typing
>>> class A():
...   ...
...
>>> ii = typing.Annotated[int, A()]
>>> typing.get_origin(ii)
<class 'typing.Annotated'>
```
we need to handle both.

flytekit/core/type_engine.py
```diff
@@ -565,9 +597,17 @@ def get_sub_type(t: Type[T]) -> Type[T]:
         """
         Return the generic Type T of the List
         """
-        if hasattr(t, "__origin__") and t.__origin__ is list:  # type: ignore
-            if hasattr(t, "__args__"):
-                return t.__args__[0]  # type: ignore
+
+        if hasattr(t, "__origin__"):
```
wild-endeavor, line 601: The same changes need to be applied to the `DictTransformer` right?

flytekit/core/type_engine.py
```diff
@@ -565,9 +597,17 @@ def get_sub_type(t: Type[T]) -> Type[T]:
...
+
+            # Handle annotation on list generic, eg:
+            # typing.Annotated[typing.List[int, 'foo']]
+            if isinstance(t, typing._AnnotatedAlias) and hasattr(t.__origin__, "__origin__"):
```
wild-endeavor, line 605: not entirely sure i understand - why is the checking here using isinstance of the annotated alias and using get origin above?  Could you add a comment please?

wild-endeavor (commented):
Thank you for doing this!  Made some comments... not sure that they all make sense.  I haven't looked very deeply at all into this Annotated stuff.  Didn't even know it existed until a couple months ago.  Apologies if I'm off the mark.

flytekit/core/type_engine.py
kennyworkman: wasn't planning on supporting it in this pr

flytekit/core/type_engine.py
kennyworkman: typing.Annotated is only supported by 3.9
developers will only be able to use the core.FlyteAnnotation object with this version of python in my eye
https://www.python.org/dev/peps/pep-0593/

tests/flytekit/unit/core/test_typing_annotation.py
kennyworkman: only planning on supporting primitive + List
but sure

flytekit/models/annotation.py
kennyworkman: sure

flytekit/core/type_engine.py
eapolinario: Two things:
1. That PEP was implemented in the [typing_extensions](https://github.com/python/typing/tree/master/typing_extensions) module, which essentially means this is available to all python versions starting from 3.6.
2. In python 3.8 and below, the implementation of `get_origin` present in the typing module is not aware of `Annotated`, instead, `get_origin` present in the `typing_extensions` module does the right thing in all versions (including 3.9 and onwards).

tldrl: let's standardize on `typing_extensions` and not worry about compatibility between different versions.

flytekit/core/type_engine.py
```diff
@@ -437,7 +447,29 @@ def to_literal_type(cls, python_type: Type) -> LiteralType:
...
         """
         transformer = cls.get_transformer(python_type)
-        return transformer.get_literal_type(python_type)
+        res = transformer.get_literal_type(python_type)
+        data = None
+        if typing.get_origin(python_type) is typing.Annotated:
+            for x in typing.get_args(python_type)[1:]:
+                if not isinstance(x, FlyteAnnotation):
+                    continue
+                if x.data.get("__consumed", False):
```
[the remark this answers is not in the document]
kennyworkman: apologize that this made it in - it was used in our codebase internally and does not belong in flyte src

flytekit/core/type_engine.py
kennyworkman: well no, they need to be serialized into proto correctly given the new message created to suppport annotations
https://github.com/flyteorg/flyteidl/pull/232/files#diff-da0cd41ae99f181280058bc79cbc3f3565a22039fbefcbe5f191bc2780d72735R68

flytekit/core/annotation.py
kennyworkman: i named it `data` to reinforce that this is the client-side, python object and so it wouldn't get confused with the idl structure and the python model that represents it (eg. `FlyteAnnotation` vs. `TypeAnnotation`)

flytekit/core/annotation.py
kennyworkman: did so and added new test

flytekit/core/type_engine.py
kennyworkman: got it
makes sense and will push shortly

tests/flytekit/unit/core/test_typing_annotation.py
kennyworkman: added tests for primitive + lists in type engine

flytekit/core/type_engine.py
kennyworkman: sloppy (made the change for @eapolinario comment in one spot but not other)

flytekit/core/type_engine.py
```diff
@@ -19,14 +19,17 @@
...
 
 from flytekit.common.exceptions import user as user_exceptions
 from flytekit.common.types import primitives as _primitives
+from flytekit.core.annotation import FlyteAnnotation
 from flytekit.core.context_manager import FlyteContext
 from flytekit.core.type_helpers import load_type_from_tag
 from flytekit.loggers import logger
 from flytekit.models import interface as _interface_models
 from flytekit.models import types as _type_models
+from flytekit.models.annotation import TypeAnnotation as _annotation_model
```
eapolinario, line 32: how about we do `from flytekit.models.annotation import TypeAnnotation as TypeAnnotationModel` instead? It feels a bit weird to import a specific model and use the underscore (to indicate that it's private).

flytekit/core/type_engine.py
```diff
@@ -439,14 +442,22 @@ def get_transformer(cls, python_type: Type) -> TypeTransformer[T]:
...
 
         """
+
         # Step 1
         if python_type in cls._REGISTRY:
             return cls._REGISTRY[python_type]
 
         # Step 2
         if hasattr(python_type, "__origin__"):
+
```
eapolinario, line 452: nit: remove blank line.

flytekit/core/type_engine.py
```diff
@@ -603,9 +637,17 @@ def get_sub_type(t: Type[T]) -> Type[T]:
         """
         Return the generic Type T of the List
         """
-        if hasattr(t, "__origin__") and t.__origin__ is list:  # type: ignore
-            if hasattr(t, "__args__"):
-                return t.__args__[0]  # type: ignore
+
```
eapolinario, line 640: nit: remove blank line.

flytekit/models/annotation.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def to_flyte_idl(self) -> _types_pb2.TypeAnnotation:
+        """
+        :rtype: flyteidl.core.types_pb2.TypeAnnotation
+        """
+
+        import inspect
+
+        print("\nrosa: ", self.annotations, inspect.stack()[1])
+
```
eapolinario, line 30: delete.

tests/flytekit/unit/models/test_types.py
```diff
@@ -81,6 +83,16 @@ def test_literal_types():
     assert obj == _types.LiteralType.from_flyte_idl(obj.to_flyte_idl())
 
 
+def test_annotated_literal_types():
```
eapolinario, line 86: Can you add a few tests for more complex literal types as well?

eapolinario (changes requested):
This is looking good, I have just a few questions and a concern about the composition of annotated types, so I asked for more unit tests. Otherwise, this is shaping up to be really nice.

Don't worry too much about the "nit:" comments as the pre-commit hook runs [black](https://github.com/psf/black) and other linters (so those stylistic comments should not be a concern).

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -662,6 +664,31 @@ def test_dict_to_literal_map_with_wrong_input_type():
         TypeEngine.dict_to_literal_map(ctx, input, guessed_python_types)
 
 
+def test_annotated_simple_types():
+
+    t = typing.Annotated[int, FlyteAnnotation({"foo": "bar"})]
+    lt = TypeEngine.to_literal_type(t)
+    assert isinstance(lt.annotation, TypeAnnotation)
+    assert lt.annotation.annotations == {"foo": "bar"}
```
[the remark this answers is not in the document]
kennyworkman: sure can add dicts and lists
I guess dataclass as well, see no reason that any object could not be supported

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -662,6 +664,31 @@ def test_dict_to_literal_map_with_wrong_input_type():
...
+
+    t = typing.List[typing.Annotated[int, FlyteAnnotation({"foo": "bar"})]]
+    lt = TypeEngine.to_literal_type(t)
+    assert isinstance(lt.collection_type.annotation, TypeAnnotation)
+    assert lt.collection_type.annotation.annotations == {"foo": "bar"}
+
+
+def test_multiple_annotations():
+    t = typing.Annotated[int, FlyteAnnotation({"foo": "bar"}), FlyteAnnotation({"anotha": "one"})]
```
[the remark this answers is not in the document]
kennyworkman: I dont think that the order of the nesting should matter, or that we should consider clobber
will make error more verbose and disallow such use of the annotation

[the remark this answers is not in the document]
kennyworkman: Ok, so in retrospect, I will clearly document that we should just take the first argument and allow the user to use the  `typing.Annotated` freely 

In the case you presented:
```
>>> class TypingAnnotation:
...     data = None
...
>>> a1, a2 = TypingAnnotation(), TypingAnnotation()
>>> a1.data, a2.data = ("foo", "bar")
>>> inner_t = typing.Annotated[int, a1]
>>> t = typing.Annotated[inner_t, a2]
>>> for a in typing.get_args(t)[1:]:
...     print(a.data)
...
foo
bar
```
We just take the first argument, believe this error is already instructive to a user that is nesting types --
https://github.com/flyteorg/flytekit/pull/759/files#diff-68624e201f644896b015de1a5716eabef9b8b5a106caba1d9f21c51c0708684bR497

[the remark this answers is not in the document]
kennyworkman: adding a case for this

tests/flytekit/unit/models/test_types.py
kennyworkman: added for:
* enum
* dict
* dataclass

flytekit/core/type_engine.py
```diff
@@ -221,6 +224,13 @@ def get_literal_type(self, t: Type[T]) -> LiteralType:
         Extracts the Literal type definition for a Dataclass and returns a type Struct.
         If possible also extracts the JSONSchema for the dataclass.
         """
+        if get_origin(t) is typing.Annotated:
+            raise ValueError(
+                f"Flytekit does not currently have support \
+                    for FlyteAnnotations applied to Dataclass. {t} cannot be \
+                    parsed."
+            )
+
```
eapolinario, line 233: Split this multi-line string in multiple strings instead, e.g.:
```
raise ValueError(
    "Flytekit does not currently have support for FlyteAnnotations applied to Dataclass. "
    f"Type {t} cannot be parsed."
)
```

This will allow for people to grep / skip these log lines more easily.

That said, we should probably be more prescriptive here (i.e. add a max column number in our linters). Not in this PR obviously.

flytekit/models/types.py
```diff
@@ -5,6 +5,7 @@
 from google.protobuf import struct_pb2 as _struct
 
 from flytekit.models import common as _common
+from flytekit.models.annotation import TypeAnnotation as _annotation_model
```
eapolinario, line 8: Let's also use `from flytekit.models.annotation import TypeAnnotation as TypeAnnotationModel` instead.

flytekit/models/types.py
```diff
@@ -120,6 +122,8 @@ def __init__(
         :param flytekit.models.core.types.BlobType blob: For blob objects, this describes the type.
         :param flytekit.models.core.types.EnumType enum_type: For enum objects, describes an enum
         :param dict[Text, T] metadata: Additional data describing the type
+        :param flytekit.models.annotation.FlyteAnnotation annotation: Additional data
```
eapolinario, line 125: this type should be flytekit.models.annotation.TypeAnnotation, rigth?

flytekit/models/types.py
```diff
@@ -166,18 +171,31 @@ def metadata(self):
         """
         return self._metadata
 
+    @property
+    def annotation(self) -> _annotation_model:
+        """
+        :rtype: flytekit.models.annotation.FlyteAnnotation
```
eapolinario, line 177: ditto.

eapolinario (commented):
Left a few more comments, one comment especially is going to unblock unit tests. Also, once https://github.com/flyteorg/flytekit/pull/759/files#r766346591 is handled we should be close.

flytekit/core/type_engine.py
kennyworkman: made the change to avoid another iteration, but can you describe why making the log multiple strings makes it more "greppable"?

tests/flytekit/unit/core/test_type_engine.py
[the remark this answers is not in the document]
kennyworkman: added

flytekit/core/type_engine.py
eapolinario: Sure, what I mean is that it'll be easier to come up with a regex if you don't have to worry about counting spaces.

For example, as it was before, the string present in the `ValueError` would show up in the logs as:
```
'Flytekit does not currently have support             for FlyteAnnotations applied to Dataclass. Annotated[SomeDataClass, FlyteAnnotation()] cannot be             parsed.
``` 

Notice how one have to worry about coming up with a regex that takes into account the extra spaces, as opposed to the more straightforward:
```
'Flytekit does not currently have support for FlyteAnnotations applied to Dataclass. Annotated[SomeDataClass, FlyteAnnotation()] cannot be parsed.
```

flytekit/core/annotation.py
```diff
@@ -0,0 +1,30 @@
+from typing import Any, Dict
+
+
+class FlyteAnnotation:
+    """A core object to add arbitrary annotations to flyte types.
+
+    This metadata is ingested as a python dictionary and will be serialized
+    into fields on the flyteidl type literals. This data is not accessible at
+    runtime but rather can be retrieved from flyteadmin for custom presentation
+    of typed parameters.
```
eapolinario, line 10: What happens if someone annotates the return of a task with a `FlyteAnnotation`? Should we detect and raise an error in this case?

eapolinario:
@kennyworkman , can you merge master? Integration tests are passing in master again.

eapolinario (approved):
This is amazing work, @kennyworkman! Thanks for pushing through a lengthy review process, I hope we can make this smoother in the future.

PR #760, opened 2021-12-01
Support for delayed annotations

bethebunny:
# TL;DR
Support delayed annotations through `from __future__ import annotations` in python 3.8/3.9 task and workflow definitions.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description
`inspect.signature` behavior changes in an awkward way in python 3.8/3.9 when applying `from __future__ import annotations`, where it returns the string values of the delayed annotations for type hints rather than the values themselves. This is fixed in 3.10. The recommended solution is to use `typing.get_type_hints()` for all annotations, and then join in `inspect.signature` for default values where necessary.

I modified `transform_signature_to_interface` to `transform_function_to_interface`, which was sensible anyway because 100% of the usages of this function called `inspect.signature` on a function before passing it into `transform`. `transform_function_to_interface` then owns the logic of correctly resolving type hints and default values through `typing.get_type_hints` and `inspect.signature`.

Relevant read: https://www.mail-archive.com/python-dev@python.org/msg111220.html

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1028

## Follow-up issue
_NA_


wild-endeavor:
Thanks for the fix @bethebunny!!  Let me play around with this later today/tonight.  This is super useful, been on our plate forever. 🙏

flytekit/core/interface.py
```diff
@@ -267,27 +267,34 @@ def _change_unrecognized_type_to_pickle(t: Type[T]) -> Type[T]:
...
+    }
+
+    # Inputs with default values are currently ignored, we may want to look into that in the future
+    inputs[k] = (_change_unrecognized_type_to_pickle(annotation), default)
+
+    outputs = extract_return_annotation(type_hints.get("return", None))
+    inputs = {
+        arg: (_change_unrecognized_type_to_pickle(type_hints.get(arg, inspect.Parameter.empty)), default)
+        for arg, default in defaults.items()
+    }
```
eapolinario, line 292: can we keep the original two loops after we set `type_hints` and `signature` instead?

bethebunny: Ah yeah I had removed some important behavior on outputs (sorry, merged this in from a change I was doing on 0.22.2); I updated, but if you're also suggesting using a for loop instead of a dict comprehension happy to make that change specifically also :)

eapolinario: nah, it's pretty clear what's happening after the last update!

bethebunny:
Tests should pass now :)

flytekit/core/interface.py
bethebunny: Okay, made the change a bit less intrusive and kept as much of the previous loops as possible :) resolving this.

eapolinario:
@bethebunny , presumably you're force-pushing because of the DCO warnings, but can you instead follow the instructions to fix those warnings instead? It's harder to see what's changing from commit to commit if we only have a single commit.

bethebunny:
Ah yes, I only got as far in the doc as option 1 and assumed that was happy path for contributors :) I'll re-read and stop squashing commits for future changes.

wild-endeavor (dismissed):

wild-endeavor:
as is, this PR has a breaking change - let's toy around with this a bit more as a beta release before committing to it (perhaps merge it, perhaps not)
See the deleted unit test for an example of the break, basically json dataclasses with improper schemas were previously accepted as a schemaless generic and now will not be.

the current thinking is to request users to remove the `dataclass_json` in this case, and patch the typeengine code so that the Flyte pickle hander picks them up.  This still is a breaking change and since users will have to change code, but at least it's a workaround.

Another alternative is to continue to [swallow the error](https://github.com/flyteorg/flytekit/blob/b08d388017f292f1a5d7190dce489d79e9ddf91c/flytekit/core/type_engine.py#L247).

wild-endeavor:
Hmm - never mind.  I think too many things rely on this behavior.  Let me see about continuing to hide the error.

wild-endeavor (approved):

PR #762, opened 2021-12-03
Caching of offloaded objects

eapolinario:
# TL;DR
flytekit portion of proposal described in https://github.com/flyteorg/flyte/pull/1893

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
We define a new field in Type Transformers that indicate if hashes are overridable. Such hashes are produced for annotated objects returned by tasks.

In case they are set in a literal, hashes are taken into consideration during cache key calculation and we do that by using them as opposed to a string representation of the literal. 

As part of this feature we revamped the calculation of the cache key to contemplate the entire literal (to cover the cases of collections and maps). It also brings the calculation of the cache key more inline with what happens in [flytepropeller](https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/nodes/task/catalog/datacatalog/transformer.go#L105).
 

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1581

## Follow-up issue
_NA_


flytekit/models/literals.py
```diff
@@ -794,4 +798,9 @@ def from_flyte_idl(cls, pb2_object):
             scalar=Scalar.from_flyte_idl(pb2_object.scalar) if pb2_object.HasField("scalar") else None,
             collection=collection,
             map=LiteralMap.from_flyte_idl(pb2_object.map) if pb2_object.HasField("map") else None,
+            # TODO: explain that string always have a value set in protobufs, which means
```
wild-endeavor, line 801: why is this a todo?  you just want to add more text to the comment?  or document it in flyte cookbook or something?

wild-endeavor: you can explain with this example:
```
def test_jcdskl():
    from flyteidl.core.literals_pb2 import KeyValuePair
    from io import BytesIO

    k1 = KeyValuePair(key="fjdksl", value=None)
    bytIO = BytesIO()
    bytIO.write(k1.SerializeToString())

    print(bytIO.getvalue())
    bytIO.seek(0)

    k2 = KeyValuePair()
    k2.ParseFromString(bytIO.getvalue())
    print(k2)
    assert k2.value == ""
```

flytekit/core/type_engine.py
```diff
@@ -487,7 +504,25 @@ def to_literal(cls, ctx: FlyteContext, python_val: typing.Any, python_type: Type
         transformer = cls.get_transformer(python_type)
         if transformer.type_assertions_enabled:
             transformer.assert_type(python_type, python_val)
+
+        # In case the value is an annotated type we inspect the annotations and look for hash-related annotations.
+        hash = None
+        if transformer.hash_overridable and get_origin(python_type) is Annotated:
```
wild-endeavor, line 510: get_origin works in all 3.7-3.10?

flytekit/core/type_engine.py
```diff
@@ -487,7 +504,25 @@ def to_literal(cls, ctx: FlyteContext, python_val: typing.Any, python_type: Type
...
+            # 2. The annotated type is being used for a different purpose other than calculating hash values, in which case
+            #    we should just continue.
+            for annotation in get_args(python_type)[1:]:
+                if not isinstance(annotation, HashMethod):
+                    continue
+                hash = annotation.calculate(python_val)
+                break
+
         lv = transformer.to_literal(ctx, python_val, python_type, expected)
+        # TODO: should we worry about empty string?
```
wild-endeavor, line 523: no

wild-endeavor (commented):
https://github.com/flyteorg/flyteidl/pull/237/files#diff-e9c7f9528aac9670ee8251a31dd27a04b07bd1afbb195f053ad0182e39eb2119R80 why is this a todo?  should we just add some comments now?

flytekit/models/literals.py
```diff
@@ -737,6 +739,7 @@ def __init__(self, scalar: Scalar = None, collection: LiteralCollection = None,
         self._scalar = scalar
         self._collection = collection
         self._map = map
+        self.hash = hash
```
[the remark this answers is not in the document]
eapolinario: sure thing. I'll keep the pattern, it's useful.

flytekit/core/type_engine.py
eapolinario: yes, it goes even further, it's 3.6 and newer.

flytekit/models/literals.py
eapolinario: This was a note to self as I was debating if there was any value in having wrapping the value of hash in another type (as opposed to a string). The original problem was the impossibility of differentiating between `None` and empty string, which is a quirk of how strings work in protobufs. 

We could use https://github.com/envoyproxy/protoc-gen-validate#protoc-gen-validate-pgv to handle the case of empty strings (by setting `min_bytes`), but definitely not in this PR.

eapolinario:
> https://github.com/flyteorg/flyteidl/pull/237/files#diff-e9c7f9528aac9670ee8251a31dd27a04b07bd1afbb195f053ad0182e39eb2119R80 why is this a todo? should we just add some comments now?

yes, I'm going to update that PR to include a proper description of what the hash is and how to use it.

tests/flytekit/unit/core/test_local_cache.py
```diff
@@ -252,6 +255,136 @@ def my_wf(a: int, b: str) -> (int, str):
...
+    assert n_cached_task_calls == 0
+    my_workflow()
+    assert n_cached_task_calls == 1
+
+    # Confirm that we see a cache hit in the case of annotated dataframes.
+    my_workflow()
+    assert n_cached_task_calls == 1
+
+
 """
```
eapolinario, line 388: why is this docstring here? Can we remove it, @pingsutw , @wild-endeavor ?

pingsutw: It should not be here, we can remove it.

wild-endeavor (approved):

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #763, opened 2021-12-03
Union Types

maximsmol:
## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
https://github.com/maximsmol/flyte/blob/master/rfc/core%20language/sum-types.md

## Tracking Issue
[flyteorg/flyte#1349](https://github.com/flyteorg/flyte/issues/1349)

## Follow-up issue
_NA_


.github/workflows/pythonbuild.yml
```diff
@@ -36,6 +36,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip==21.2.4 setuptools wheel
         make setup${{ matrix.spark-version-suffix }}
+        git clone https://github.com/flyteorg/flyteidl.git && cd flyteidl && git checkout union_type && pip install . && cd ..
```
pingsutw, line 39:
```suggestion
        git clone https://github.com/maximsmol/flyteidl.git && cd flyteidl && git checkout union_type && pip install . && cd ..
```

flytekit/core/type_engine.py
```diff
@@ -604,47 +604,50 @@ class UnionTransformer(TypeTransformer[T]):
...
         raise TypeError(f"Cannot convert from {python_val} to {python_type}")
 
     def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> Optional[typing.Any]:
-        if lv is None:
-            return None
-        st = self.get_sub_type(expected_python_type)
-        for v in st:
+        for x in typing.get_args(expected_python_type):
+            if x is type(None):
+                if lv is None: # todo(maximsmol): why would this ever be None and not Void?
```
pingsutw, line 629: Because the `None` transformer didn't convert the `None` value to Literal(Void). https://github.com/flyteorg/flytekit/blob/742f40eb718777887043db3842b076b7f87a8427/flytekit/core/type_engine.py#L946
I think we should update that transformer.

maximsmol: agreed, will update

flytekit/core/type_engine.py
```diff
@@ -595,6 +617,99 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+            except TypeTransformerFailedError as e:
+                logger.debug(f"Failed to convert from {python_val} to {t}", e)
+                continue
+
+        if found_res:
+            return Literal(scalar=Scalar(union=Union(value=res, stored_type=res_type)))
+
+        raise TypeTransformerFailedError(f"Cannot convert from {python_val} to {python_type}")
+
+    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> Optional[typing.Any]:
```
wild-endeavor, line 664: Sorry got kinda confused reading this code.  We're going to go with the changes in [gist](https://gist.github.com/EngHabu/8c7534ab218c5f40393c45decd994340) right?  If so, can't we just extract the literal type in the union, and then match it to exactly one literal type inside the Union?  Why do we need to fetch the transformer and then inspect the tag?  What if we have one transformer that handles both `int` and `MyInt` (which are distinct LiteralType's because of the new structure field)?

tests/flytekit/unit/core/test_type_hints.py
```diff
@@ -1427,3 +1431,177 @@ def foo3(a: typing.Dict) -> typing.Dict:
...
+
+    @workflow
+    def wf(a: int) -> typing.Union[int, str]:
+        return t1(a=a)
+
+    assert wf(a=2) == 2
+    assert wf(a=-10) == "-10"
+
+
+def test_union_type_ambiguity_checking():
```
wild-endeavor, line 1529: Don't entirely understand the last two tests here... not quite sure why one is ambiguous and the other is not?  Following the changes in the gist, `int` and `MyInt` are distinct Literal Types right?

flytekit/core/type_engine.py
```diff
@@ -595,6 +617,99 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+        res_tag = None
+        for v in typing.get_args(expected_python_type):
+            try:
+                trans = TypeEngine.get_transformer(v)
+                if union_tag is not None:
+                    if trans.name != union_tag:
+                        continue
+
+                    assert lv.scalar is not None # type checker
+                    assert lv.scalar.union is not None # type checker
```
eapolinario, line 682: Assertions can be [turned off](https://docs.python.org/3/reference/simple_stmts.html#the-assert-statement).

flytekit/core/type_engine.py
```diff
@@ -595,6 +617,99 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
             return typing.List[ct]
         raise ValueError(f"List transformer cannot reverse {literal_type}")
 
+def _add_tag_to_type(x: LiteralType, tag: str) -> LiteralType:
+    x._structure = TypeStructure(tag=tag)
+    return x
+
+class UnionTransformer(TypeTransformer[T]):
+    """
+    Transformer that handles a univariate typing.Union[T]
```
eapolinario, line 626: Update comment.

flytekit/core/type_engine.py
```diff
@@ -595,6 +617,99 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+                    found_res = True
+                else:
+                    res = trans.to_python_value(ctx, lv, v)
+                    if found_res:
+                        raise TypeError(
+                            "Ambiguous choice of variant for union type. " +
+                            f"Both {res_tag} and {trans.name} transformers match")
+                    res_tag = trans.name
+                    found_res = True
+            except TypeTransformerFailedError as e: # todo(maximsmol): separate programmer errors from type conversion failures
```
eapolinario, line 699: Can you expand a bit on this todo?

eapolinario (commented):
First of all, thanks for working on this. We've come a long way in our understanding of sum types and how they can be implemented in a language like python. This PR is taking most of the way there as discussed in the RFCs<sup>1,</sup><sup>2</sup> and also the [gist](https://gist.github.com/EngHabu/8c7534ab218c5f40393c45decd994340).

A few overall comments:
- The idea of having a custom exception, i.e.`TypeTransfomerFailedError`, to denote type-transformer-specific errors is nice (and in this case necessary since we need to handle that specific class of errors in the Union transformer). 
- we need a few more unit tests, for example, around how the order of the types in a union declaration does not matter.

<sup>1,</sup> https://github.com/maximsmol/flyte/blob/master/rfc/core%20language/sum-types.md
<sup>2</sup> https://github.com/flyteorg/flyte/pull/1926

flytekit/core/type_engine.py
maximsmol: This was before `TypeTransformerFailedError` was implemented

flytekit/core/type_engine.py
maximsmol: These assertions are already provably true since `union_tag` is not None iff `lv.scalar` and `lv.scalar.union` are not None

The type checker cannot make this inference so these are here to silence the warnings

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -712,6 +717,220 @@ def test_enum_type():
...
+        LiteralType(
+            collection_type=LiteralType(simple=SimpleType.INTEGER, structure=TypeStructure(tag="int")),
+            structure=TypeStructure(tag="Typed List")
+        ),
+        LiteralType(
+            collection_type=LiteralType(simple=SimpleType.STRING, structure=TypeStructure(tag="str")),
+            structure=TypeStructure(tag="Typed List")
+        ),
+    ]
+    assert not union_type_tags_unique(lt) # tags are deliberately NOT unique
```
eapolinario, line 903: the tags are not unique, but the literal types are, correct?

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -712,6 +717,220 @@ def test_enum_type():
     with pytest.raises(AssertionError):
         TypeEngine.to_literal_type(UnsupportedEnumValues)
 
+def union_type_tags_unique(t: LiteralType):
```
eapolinario, line 720: Can you also add a few unit tests where annotated types are present? Including `Annotated[Union[A, B, C], ...]` and `Union[Annotated[A, ...], B, Annotated[C, ...], ...]`.

flytekit/types/schema/types.py
```diff
@@ -368,14 +368,20 @@ def to_literal(
             local_path=ctx.file_access.get_random_local_directory(),
             remote_path=ctx.file_access.get_random_remote_directory(),
         )
+        try:
+            h = SchemaEngine.get_handler(type(python_val))
+        except ValueError as e:
+            raise TypeTransformerFailedError(f"Could not convert {type(python_val)} to flyte schema") from e
```
eapolinario, line 374: nice catch moving this up.

flytekit/models/literals.py
```diff
@@ -548,6 +549,45 @@ def from_flyte_idl(cls, pb2_object):
...
+
+    @property
+    def value(self):
+        """
+        :rtype: flytekit.models.literals.Literal
+        """
+        return self._value
+
+    @property
+    def stored_type(self):
```
eapolinario, line 571: what do you think of renaming this property to `literal_type`?

flytekit/core/type_engine.py
```diff
@@ -1119,13 +1256,15 @@ def _register_default_type_transformers():
...
-            None,
+            type(None),
             _type_models.LiteralType(simple=_type_models.SimpleType.NONE),
-            lambda x: None,
-            lambda x: None,
-        )
+            lambda x: Literal(scalar=Scalar(none_type=Void())),
+            lambda x: _check_and_convert_void(x)
+        ),
+        [None],
```
eapolinario, line 1264: Can we add a few unit tests to cover this change? I'm particularly interested in demonstrating that the old way of specifying void literals will also be handled correctly (i.e. users will be able to update flytekit independently from flytepropeller).

eapolinario (commented):
Overall this is shaping up to be an amazing change. I left some comments where I deemed appropriate. We need a few more things besides the comments for the next iteration:
1. a change to install the correct version of `flyteidl` before running tests in CI. https://github.com/flyteorg/flytekit/pull/763/commits/21163c914eb4068869b865035e72091604cb8302 was a previous attempt at this. Please see https://github.com/flyteorg/flytekit/pull/762/commits/09e736d8847c4e447ae42cf0e2c5292f2c9f8fb9 for a simpler way.
2.  Can you add a high-level description of the change? Mention the union transformer, the change to the void transformer and the inclusion of a new type-transformer-specific exception (and why we need it, i.e. the need to handle specific type transformer exceptions in the union type transformer).
3. A couple of unit tests to demonstrate that the order in which variants are defined does not influence the results coming from the union transformer. 

Thanks for pushing through with this, @maximsmol !

flytekit/models/literals.py
maximsmol: given that `Union` is itself a `LiteralType` I think the current name makes more sense

flytekit/core/type_engine.py
```diff
@@ -793,6 +826,98 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[list]:
...
+
+        found_res = False
+        res = None
+        res_tag = None
+        for v in _get_args(expected_python_type):
+            try:
+                trans = TypeEngine.get_transformer(v)
+                if union_tag is not None:
+                    if trans.name != union_tag:
+                        continue
```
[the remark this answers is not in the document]
maximsmol: some notes, for the record:

a value of type `List[int]` should be assignable to a value of type `Union[List[Union[int, str]], List[bool]]` since the types `List[int]` and `List[Union[int, str]]` are compatible
this means that the stored type of the literal does not necessarily match either of the union variants 1-for-1

instead the types must be "castable" (see `AreTypesCastable` from propeller) or, really, the `Literal` must be compatible with the type transformer for the target type

the only way to check whether a type transformer accepts a `Literal` is to try running it and look for exceptions (generic exception previously, `TypeTransformerFailedError` in this PR). this is exactly what the current implementation does

I wil implement `AreTypesCastable` in flytekit and rely on it instead. note that this introduces a discrepancy between whether a type transformer is supposed to succeed and whether it actually succeeds, but I think it's fine because I cannot come up with a counter-example. my reasoning relies on the failure of a type transformer that matches the literal type being non-fatal i.e. causing a `continue` instead of an exception in the union type transformer loop

the main dangerous interaction here is that the `LiteralType` -> `TypeTransformer` mapping is non-injective. what saves us, however, is that the Python Type -> `TypeTransformer` mapping *is* injective and, as a consequence, the `[TypeEngine.to](http://TypeEngine.to)_literal_type(x: Type)` function is also injective. if it was not injective, comparing `LiteralType`s would be impossible because we would not have enough information to get the actual expected `LiteralType`

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -367,7 +371,23 @@ def test_guessing_basic():
...
+    pt = TypeEngine.guess_python_type(lt)
+    assert pt is FlytePickle
+
+    lt = model_types.LiteralType(
+        blob=BlobType(
+            format=FlytePickleTransformer.PYTHON_PICKLE_FORMAT, dimensionality=BlobType.BlobDimensionality.SINGLE
+        )
+    )
+    pt = TypeEngine.guess_python_type(lt)
+    assert pt is FlytePickle
```
eapolinario, line 390: these are duplicate of lines 376-382, no?

tests/flytekit/unit/core/test_type_hints.py
eapolinario: @wild-endeavor , in `test_union_type_ambiguity_checking` we don't have enough type information in the incoming literal (i.e. the `int` that's passed to `t1`), whereas in `test_union_type_ambiguity_resolution` `t1` takes a `Union[int, MyInt]`, but notice how that comes from the result of `t2`, which will produce a literal that contains the necessary type information to disambiguate.

eapolinario (commented):
Amazing stuff. Just a few questions and a couple of nits.

Can you merge master? We recently disabled integration tests, which greatly sped up CI tests.

flytekit/core/type_engine.py
```diff
@@ -147,18 +157,36 @@ def __init__(
...
     ):
         super().__init__(name, t)
+        self._type = t
         self._lt = lt
         self._to_literal_transformer = to_literal_transformer
         self._from_literal_transformer = from_literal_transformer
 
     def get_literal_type(self, t: Type[T] = None) -> LiteralType:
-        return self._lt
+        return LiteralType.from_flyte_idl(self._lt.to_flyte_idl())
```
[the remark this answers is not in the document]
maximsmol: it should either be here or the `TypeTransformer` for `Union` must copy all the types before adding the `_structure` field. For most other transformers this is not an issue because they construct a new `LiteralType` instance on each invocation, but these ones reuse the same instance so modifying it poisons all future invocations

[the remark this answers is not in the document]
eapolinario: got it. Let's leave it here.

eapolinario (approved):

wild-endeavor:
thank you @maximsmol! 🙏💪🖖

PR #770, opened 2021-12-09
Schema arrow ping interfaces 2

wild-endeavor:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/types/structured/structured_dataset.py
```diff
@@ -279,14 +299,17 @@ def to_literal(
         # StructuredDataset instance.
         if issubclass(python_type, StructuredDataset):
             assert isinstance(python_val, StructuredDataset)
-            df_type = type(python_val.dataframe)
+            python_val.uri = python_val.uri or ctx.file_access.get_random_remote_path()
```
wild-endeavor, line 302: I don't think we should be using `file_access` in this file.  The determination of what the uri should be should be left entirely up to the individual encoders/decoders.  Is there a reason this was needed?

flytekit/types/structured/structured_dataset.py
```diff
@@ -279,14 +299,17 @@ def to_literal(
...
+            protocol = protocol_prefix(python_val.uri)
             format = python_val.file_format
-            protocol = self.DEFAULT_PROTOCOLS[df_type]
-            try:
-                protocol = protocol_prefix(python_val.uri)
-            except ValueError:
-                ...
-
+            if protocol != StructuredDatasetProtocol.BIGQUERY:
```
wild-endeavor, line 305: we shouldn't reference either parquet, or bigquery in this file.

flytekit/common/constants.py
```diff
@@ -38,3 +38,13 @@ class CloudProvider(object):
     AWS = "aws"
     GCP = "gcp"
     LOCAL = "local"
+
+
+class StructuredDatasetProtocol(object):
```
wild-endeavor, line 43: we should delete these two classes.  The point of making the format/protocols just plain strings is that they're easily extensible.

We can have parquet be the default format, but we should just have a string constant in the structured_dataset.py file.  This file (and everything else under `flytekit/common`) is getting deleted in january.

flytekit/models/literals.py
```diff
@@ -548,7 +549,7 @@ def from_flyte_idl(cls, pb2_object):
 
 
 class StructuredDatasetMetadata(_common.FlyteIdlEntity):
-    def __init__(self, format: str, structured_dataset_type: StructuredDatasetType):
+    def __init__(self, format: typing.Optional[str], structured_dataset_type: StructuredDatasetType = None):
```
wild-endeavor, line 552: the type I'm not sure about, but the format shouldn't be optional.  We should always have a default, even if it's an "".

If we want to make structured_dataset_type optional, the `to_flyte_idl` and `from_flyte_idl` methods will also have to change to check for none-ness

flytekit/models/literals.py
```diff
@@ -574,7 +575,7 @@ def from_flyte_idl(cls, pb2_object: _literals_pb2.StructuredDatasetMetadata) ->
 
 
 class StructuredDataset(_common.FlyteIdlEntity):
-    def __init__(self, uri: str, metadata: StructuredDatasetMetadata):
+    def __init__(self, uri: str, metadata: typing.Optional[StructuredDatasetMetadata] = None):
```
wild-endeavor, line 578: same here - if metadata is made optional, then to_flyte_idl and from_flyte_idl will have to change.  Otherwise, you'll end up calling:

self.metadata.to_flyte_idl(), which will resolve to `None.to_flyte_idl()` which will obviously fail.

flytekit/types/structured/structured_dataset.py
```diff
@@ -1,17 +1,15 @@
...
-from enum import Enum
-from typing import Any, Dict, Generator, List, Optional, Type, Union
+from typing import Dict, Generator, Optional, Type, Union
 
+import _datetime
 import numpy as _np
-import pandas as pd
-import pyarrow as pa
 
+from flytekit.common.constants import StructuredDatasetFileFormat, StructuredDatasetProtocol
```
wild-endeavor, line 12: delete these

flytekit/types/structured/structured_dataset.py
```diff
@@ -279,14 +299,17 @@ def to_literal(
...
             format = python_val.file_format
-            protocol = self.DEFAULT_PROTOCOLS[df_type]
-            try:
-                protocol = protocol_prefix(python_val.uri)
-            except ValueError:
-                ...
-
+            if protocol != StructuredDatasetProtocol.BIGQUERY:
+                format = python_val.file_format or StructuredDatasetFileFormat.PARQUET
+            if python_val.dataframe is None:
```
wild-endeavor, line 307: this is still under the `if issubclass(python_type, StructuredDataset)` block right?  If that is true, and the python_val is an instance of `StructuredDataset` - shouldn't we just raise an exception?

flytekit/types/structured/structured_dataset.py
```diff
@@ -279,14 +299,17 @@ def to_literal(
...
+                sd_model = literals.StructuredDataset(
+                    uri=python_val.uri, metadata=StructuredDatasetMetadata(format=format)
+                )
+                return Literal(scalar=Scalar(structured_dataset=sd_model))
```
wild-endeavor, line 311: Can you explain why this is necessary?  This function should not return a Literal directly - that should always be redirected to one of the encoders.

flytekit/types/structured/structured_dataset.py
```diff
@@ -297,34 +320,33 @@ def to_literal(
 
         # Otherwise assume it's a dataframe instance. Wrap it with some defaults
         fmt = self.DEFAULT_FORMATS[python_type]
-        protocol = self.DEFAULT_PROTOCOLS[python_type]
+        protocol = protocol_prefix(ctx.file_access.get_random_remote_path())
```
wild-endeavor, line 323: Remove - we shouldn't have to call file_access in this file.  If we do then I'm missing something and the interface is wrong.

flytekit/common/constants.py
pingsutw: got it, will move to structured_dataset.py

flytekit/types/structured/structured_dataset.py
pingsutw: yeah

flytekit/types/structured/structured_dataset.py
pingsutw: if we move to encoders, every encoder will have `file_access` because `python_val.uri` might be empty. People will not have good UX if they want to extend new encodes because they still need to handle empty URI or use file_access in the encoder.

flytekit/types/structured/structured_dataset.py
pingsutw: when people do something below, `StructuredDataset.dataframe` will be empty here. so I directly return Literal here instead of invoking the encoder or raising an exception.
```
@task
def t3(dataset: StructuredDataset[my_cols]) -> StructuredDataset[my_cols]:
    return dataset
```

flytekit/types/structured/structured_dataset.py
pingsutw: ^

PR #775, opened 2021-12-13
Logging updates

wild-endeavor:
# TL;DR
Cleaning up logging a bit.  Existing behavior should remain unchanged.

Continue to use the `FLYTE_SDK_LOGGING_LEVEL` var to control logging across flytekit.  This level will be used to set the log handler level, so it affects all loggers.

By default it's set to warning and higher.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Also added some new environment variables that introduce additional granularity, but this api should be considered to be beta, and is subject to change.

The hello world example now produces these logs:
```
{"asctime": "2021-12-13 19:40:20,113", "name": "flytekit.entrypoint", "levelname": "INFO", "message": "Welcome to Flyte! Version: 0.0.0+develop"}
{"asctime": "2021-12-13 19:40:20,113", "name": "flytekit.entrypoint", "levelname": "DEBUG", "message": "Running task execution with resolver flytekit.core.python_auto_container.default_task_resolver..."}
{"asctime": "2021-12-13 19:40:20,114", "name": "flytekit.entrypoint", "levelname": "INFO", "message": "Using user directory /tmp/flyte/20211213_194019/sandbox/local_flytekit/69004e6eb271ffd5ca43bc6128899eda"}
{"asctime": "2021-12-13 19:40:20,121", "name": "flytekit.entrypoint", "levelname": "DEBUG", "message": "Starting _dispatch_execute for core.flyte_basics.hello_world.say_hello"}
{"asctime": "2021-12-13 19:40:20,349", "name": "flytekit.entrypoint", "levelname": "INFO", "message": "Engine folder written successfully to the output prefix s3://my-s3-bucket/metadata/propeller/flytesnacks-development-rx2abgp0eg/n0/data/0"}
{"asctime": "2021-12-13 19:40:20,349", "name": "flytekit.entrypoint", "levelname": "DEBUG", "message": "Finished _dispatch_execute"}
```
This is with the following set:
```
ENV FLYTE_SDK_LOGGING_LEVEL 10
ENV FLYTE_SDK_LOGGING_LEVEL_ROOT 30
ENV FLYTE_SDK_LOGGING_LEVEL_ENTRYPOINT 10
```

but again, users should not rely on the latter two remaining as is forever.  Configuration of granularity, assuming it continues to be exposed (which is not certain) may change.


flytekit/bin/entrypoint.py
```diff
@@ -219,7 +218,7 @@ def setup_execution(
                 "api_version": _api_version,
             },
         ),
-        logging=_logging,
+        logging=logger,
```
eapolinario, line 221: just a curiosity: it looks like this is not used?

flytekit/bin/entrypoint.py
```diff
@@ -458,7 +456,7 @@ def execute_task_cmd(
         _click.echo("No resolver found, assuming legacy API task...")
         _legacy_execute_task(task_module, task_name, inputs, output_prefix, raw_output_data_prefix, test)
     else:
-        _click.echo(f"Attempting to run with {resolver}...")
+        logger.debug(f"Running task execution with resolver {resolver}...")
```
eapolinario, line 459: why not replace all uses of `_click.echo` in this file?

flytekit/loggers.py
```diff
@@ -1,29 +1,60 @@
...
+# The ux/api may change in the future.
+# At time of writing, the code was written to preserve existing default behavior
+# For now, assume this is the environment variable whose usage will remain unchanged and controls output for all
+# loggers defined in this file.
+LOGGING_ENV_VAR = "FLYTE_SDK_LOGGING_LEVEL"
+
+# By default, the root flytekit logger to debug so everything is logged, but enable fine-tuning
+logger = logging.getLogger("flytekit")
+# Root logger control
+flytekit_root_env_var = f"{LOGGING_ENV_VAR}_ROOT"
```
eapolinario, line 17: why introduce this env var? What's the use case for having `FLYTE_SDK_LOGGING_LEVEL` different than `FLYTE_SDK_LOGGING_LEVEL_ROOT`?

wild-endeavor: yeah the former controls the handler (which serves all loggers), the latter controls the root flytekit logger itself.

flytekit/bin/entrypoint.py
wild-endeavor: actually i'm going to revert this.  this is the logging handler passed to the user.

flytekit/bin/entrypoint.py
wild-endeavor: not sure, it's still a click cli at the end of the day.  should we?

wild-endeavor: changed a couple more.

flytekit/bin/entrypoint.py
```diff
@@ -219,7 +219,7 @@ def setup_execution(
                 "api_version": _api_version,
             },
         ),
-        logging=_logging,
+        logging=python_logging,
```
wild-endeavor, line 222: this is the handler passed to the user - will preserve the old behavior so that users don't see changes.

eapolinario (approved):

PR #780, opened 2021-12-15
When using the `task` and `workflow` decorator, correctly wrap the fu…

bstadlbauer:
# TL;DR
This enables tooling such as docstring search tools to unwrap the object and show the correct docstring. The method sets the `__wrapped__` attribute of the created instance.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
We're using [mkdocs](https://www.mkdocs.org/) together with [mkdocstrings](https://mkdocstrings.github.io/usage/) to render our Python documentation. We noticed, that documentation for Flyte tasks didn't show up, and upon further investigation found that `mkdocstrings` unwraps all decorated elements to show the correct documentation. See also: https://mkdocstrings.github.io/troubleshooting/#my-wrapped-function-shows-documentationcode-for-its-wrapper-instead-of-its-own

For this unwrapping to work, wrapper elements need to have `__wrapped__` set to a reference of the wrapped (decorated) function. With functions one could use [`functools.wraps()`](https://docs.python.org/3/library/functools.html#functools.wraps), but as Flyte is returning instances of Flyte objects, [`functools.update_wrapper()`](https://docs.python.org/3/library/functools.html#functools.update_wrapper) is what works.

### DIFF
--- a/tests/flytekit/unit/core/test_wrapping.py
+++ b/tests/flytekit/unit/core/test_wrapping.py
@@ -1,4 +1,3 @@
-# Copyright (C) 2015-2021 Blackshark.ai GmbH. All Rights reserved. www.blackshark.ai
 from flytekit import task, workflow
 
 
--- a/tests/flytekit/unit/core/test_wrapping.py
+++ b/tests/flytekit/unit/core/test_wrapping.py
@@ -1,0 +1,18 @@
+# Copyright (C) 2015-2021 Blackshark.ai GmbH. All Rights reserved. www.blackshark.ai
+from flytekit import task, workflow
+
+
+def test_task_correctly_wrapped():
+    @task
+    def my_task(a: int) -> int:
+        return a
+
+    assert my_task.__wrapped__ == my_task._task_function
+
+
+def test_wf_correctly_wrapped():
+    @workflow
+    def my_workflow(a: int) -> int:
+        return a
+
+    assert my_workflow.__wrapped__ == my_workflow._workflow_function
--- a/tests/flytekit/unit/core/test_wrapping.py
+++ b/tests/flytekit/unit/core/test_wrapping.py
@@ -1,3 +1,5 @@
+from functools import wraps
+
 from flytekit import task, workflow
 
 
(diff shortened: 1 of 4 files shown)


eapolinario:
@bstadlbauer , can you investigate the unit test failure? This [test](https://github.com/flyteorg/flytekit/blob/f500af0dd7d8a1310f8819ebc2152e21fb458f4f/plugins/flytekit-greatexpectations/tests/test_task.py#L133) is failing and it seems that it might be a weird interaction with the wrapped function and the pytest.mark.xfail. We should be able replace that use with the more idiomatic `with pytest.raises` (e.g. https://github.com/flyteorg/flytekit/blob/8528268a29a07fe7e9ce9f7f08fea68c41b6a60b/tests/flytekit/unit/common_tests/exceptions/test_scopes.py#L18-L19).

bstadlbauer:
Hi @eapolinario! 

Thank you for the fast reply. I've checked, and you're right. Seems like what actually should happen is a `with pytest.raises()` instead of the `pytest.mark.xfail`. I've updated the test accordingly

tests/flytekit/unit/core/test_wrapping.py
```diff
@@ -0,0 +1,17 @@
+from flytekit import task, workflow
+
+
+def test_task_correctly_wrapped():
```
eapolinario, line 4: @bstadlbauer , we allow for stacking of @task decorators (e.g. https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/core/functools/stacked_decorators.py). Can you add a unit test to confirm the behavior too?

eapolinario (commented):
@bstadlbauer , thanks for making this change, it looks good and should be ok to merge after we have that extra unit test I mentioned in the comment.

tests/flytekit/unit/core/test_wrapping.py
bstadlbauer: @eapolinario I've added the corresponding test, seems to work as expected. 

Also, I wasn't sure where to place the tests, so if you feel like there would be a better spot, just let me know.

eapolinario: leaving them here is fine.

eapolinario (approved):

PR #782, opened 2021-12-16
[Core feature] Add Raw AWS Batch Task

pingsutw:

# TL;DR
Run regular tasks on AWS batch service

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Add AWS batch job plugin

Example:
```python
import typing
from flytekitplugins.awsbatch import AWSBatch
from flytekit import TaskMetadata, map_task, task, workflow

config = AWSBatch(
    parameters={"codec": "mp4"},
    PlatformCapabilities=["EC2"],
    PropagateTags=True,
    RetryStrategy={"attempts": 10},
    Tags={"hello": "world"},
    Timeout={"attemptDurationSeconds": 60},
)

@task(task_config=AWSBatch(Tags={"hello": "world"}))
def t1(a: typing.List[int]) -> str:
    return str(a[0])


@task(task_config=AWSBatch(Tags={"hello": "flyte"}))
def t3(a: str, b: str) -> str:
    return b + "".join(a)


@task(task_config=config)
def create_list(n: int) -> typing.List[int]:
    res = []
    for i in range(n):
        res.append(i)
    return res


@workflow
def my_wf(a: int, b: str) -> str:
    l = create_list(n=a)
    y = t1(a=l)
    d = t3(a=y, b=b)
    return y


if __name__ == "__main__":
    x = my_wf(a=3, b="hello")
    print(f"Workflow output: {x}")
```

image

image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1792


plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
```diff
@@ -0,0 +1,71 @@
...
+class AWSBatch(object):
+    """
+    Use this to configure a job definition for a AWS batch job. Task's marked with this will automatically execute
+    natively onto AWS batch service.
+    Refer to AWS job definition template for more detail: https://docs.aws.amazon.com/batch/latest/userguide/job-definition-template.html
+    """
+
+    parameters: Optional[Dict[str, str]] = None
+    schedulingPriority: Optional[int] = None
+    PlatformCapabilities: Optional[List[str]] = None
```
kumare3, line 23: can we use the same case for everything? 
Else can we define a custom to_dict?

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
```diff
@@ -0,0 +1,71 @@
...
+from flytekit.extend import SerializationSettings, TaskPlugins
+
+
+@dataclass_json
+@dataclass
+class AWSBatch(object):
+    """
+    Use this to configure a job definition for a AWS batch job. Task's marked with this will automatically execute
+    natively onto AWS batch service.
+    Refer to AWS job definition template for more detail: https://docs.aws.amazon.com/batch/latest/userguide/job-definition-template.html
```
kumare3, line 18: also link to here - https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/batch.html#Batch.Client.register_job_definition

kumare3 (approved):

EngHabu:
Should we allow people to define the [job definition](https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html) or the [submitjob input](https://docs.aws.amazon.com/sdk-for-go/api/service/batch/#SubmitJobInput) structs? 

"most" of the properties set on the job definition can be overridden in the SubmitJobInput, the benefit of reusing a job definition is 1) there is a limit on the number of job definitions and the rate of creation and 2) reducing the number of API calls needed to launch a job is always a good thing for reliability standpoint....

So please go through both links and let me know if there are properties that you need users to set that are only available in job definitions...

EngHabu:
@kumare3 @pingsutw ^

kumare3:
@pingsutw I agree with @EngHabu, we should try and minimize job definition creations

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
```diff
@@ -0,0 +1,77 @@
...
+@dataclass_json
+@dataclass
+class AWSBatchConfig(object):
+    """
+    Use this to configure SubmitJobInput for a AWS batch job. Task's marked with this will automatically execute
+    natively onto AWS batch service.
+    Refer to AWS SubmitJobInput for more detail: https://docs.aws.amazon.com/sdk-for-go/api/service/batch/#SubmitJobInput
+    """
+
+    parameters: Optional[Dict[str, str]] = None
```
wild-endeavor, line 21: are all five of these in the SubmitJobInput documentation?  Or are some of these Flyte concepts?

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
```diff
@@ -0,0 +1,77 @@
...
+        return self._task_config.to_dict()
+
+    def get_config(self, settings: SerializationSettings) -> Dict[str, str]:
+        # Parameters in taskTemplate config will be used to create aws job definition.
+        # More detail about job definition: https://docs.aws.amazon.com/batch/latest/userguide/job_definition_parameters.html
+        return {"platformCapabilities": self._task_config.platformCapabilities}
+
+    def get_command(self, settings: SerializationSettings) -> List[str]:
+        container_args = [
+            "pyflyte-map-execute",
```
wild-endeavor, line 59: Can you explain this to me?  If I have one Python task, running one container, once - it still uses `pyflyte-map-execute`?  Why is that?  Seems kinda messy...

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, task_config: AWSBatchConfig, task_function: Callable, **kwargs):
+        if task_config is None:
+            task_config = AWSBatchConfig()
+        super(AWSBatchFunctionTask, self).__init__(
+            task_config=task_config, task_type=self._AWS_BATCH_TASK_TYPE, task_function=task_function, **kwargs
+        )
+        self._task_config = task_config
+
+    def get_custom(self, settings: SerializationSettings) -> Dict[str, Any]:
+        # task_config will be used to create SubmitJobInput in propeller except platformCapabilities.
```
wild-endeavor, line 49: how do you make it so that `platformCapabilities` doesn't get included?  Or are you saying propeller will do that?

flytekit/bin/entrypoint.py
```diff
@@ -356,8 +381,12 @@ def _execute_map_task(
             # Use the resolver to load the actual task object
             _task_def = resolver_obj.load_task(loader_args=resolver_args)
             if not isinstance(_task_def, PythonFunctionTask):
-                raise Exception("Map tasks cannot be run with instance tasks.")
-            map_task = MapPythonTask(_task_def, max_concurrency)
+                raise Exception("Map tasks cannot be run with instance tasks.", _task_def)
+
+            if is_aws_batch_single_job:
```
wild-endeavor, line 386: hmmm... seems weird.  why?

kumare3: yes, why can this not be part of the default way of handling things?

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
kumare3: I agree, can we not make a AWS batch task and then have a map-task override ride the command?

kumare3: or if needed have a separate task

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
pingsutw: all in the SubmitJobInput documentation

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
pingsutw: propeller will do that

plugins/flytekit-aws-batch/flytekitplugins/awsbatch/task.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def get_command(self, settings: SerializationSettings) -> List[str]:
+        container_args = [
+            "pyflyte-execute",
+            "--inputs",
+            "{{.input}}",
+            "--output-prefix",
+            # FlytePropeller will always read the output from this directory (outputPrefix/0)
+            # More detail, see https://github.com/flyteorg/flyteplugins/blob/0dd93c23ed2edeca65d58e89b0edb613f88120e0/go/tasks/plugins/array/catalog.go#L501.
+            "{{.outputPrefix}}/0",
```
pingsutw, line 65: I think this is the reason why I used pyflyte-map-execute before. map task will reconstruct output_prefix to `output_prefix/0`. check https://github.com/flyteorg/flytekit/blob/b09a9c76aac65c65af1c43d32060bf01585bf812/flytekit/bin/entrypoint.py#L386

Agree with you, we should use `pyflyte-execute` instead. it will be more concise.

wild-endeavor: Can you add another comment here on why the single batch task is calling the array batch plugin?

wild-endeavor (approved):
just need a comment.  will +1 again after

wild-endeavor:
oh and end to end testing worked on demo right?

wild-endeavor (approved):

PR #785, opened 2021-12-20
Schema overhaul

pingsutw:
# TL;DR
This PR introduces a new type and user experience for working with tabular data in Flyte. Please see the [design doc](https://docs.google.com/document/d/1-t8C4vW_RhrJoC5Ju16fmqpyOvsCRb3XmulLLo5VX0E/) for the complete background.

* The new type is the `StructuredDataset`. (see the [IDL PR](https://github.com/flyteorg/flyteidl/pull/227)) We thought the name sufficiently generic and differentiated from `FlyteSchema`, which we're also keeping around for the time being (see more in the Migration section below). The main difference between the `FlyteSchema` and `StructuredDataset` types themselves is that columns in the latter are just named `LiteralType`s. That is, columns are no longer constrained to be purely primitives.
* The contributor-experience has also changed substantially from `FlyteSchema` to something we hope is more straightforward and easier to understand, but we've tried to keep the user-experience as similar as possible.

**Note:**
The `[kwtype(...)]` syntax used for specifying columns and storage formats will be changing before the next non-beta release of flytekit, so if you try this out now, keep in mind you'll have some minor minor updates to make in a couple weeks.

**Note:**
To handle the new type, you will need to be on Propeller version [v0.16.14](https://github.com/flyteorg/flytepropeller/releases/tag/v0.16.14) and Admin version [v0.6.78](https://github.com/flyteorg/flyteadmin/releases/tag/v0.6.78) or later.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

### Type Handling

We found the double-interaction between the core `TypeEngine` and the `SchemaEngine` too confusing.
Screen Shot 2022-01-12 at 1 02 34 PM

In the new model, contributors will only be dealing with the `StructuredDatasetTransformerEngine`. You register your encoders/decoders with that only, and it in turn will handle interaction with the core `TypeEngine`.

Screen Shot 2022-01-12 at 1 17 23 PM

If you have a new dataframe type that you want to be able to use in Flytekit, you'll just have to subclass `StructuredDatasetEncoder` and `StructuredDatasetDecoder` and implement the logic. Encoders are responsible for taking a dataframe in Python memory and turning it into a Flyte literal (replete with any uploading/writing of data) and decoders do the opposite (and should handle downloading of data).

Encoders and decoders are now distinct entities rather than two functions within one class because we felt the the asymmetry increases flexibility. One might choose to write a custom decoder for pandas but use the default encoder. 

### Encoder/Decoder (aka Handler) Selection
This PR ships with handlers to encode and decode pandas and spark dataframes. And users can of course add their own. How does the transformer engine pick the encoder/decoder?

The handlers in the new transformer-engine are keyed off of three things:
1. Python dataframe type
1. Storage protocol (`s3`, `bq`, `gs`, etc.)
1. Storage format (a string like "parquet", "csv", etc.)

Each en/decoder is required to have all three. When the transformer engine needs to convert a Python dataframe in memory to a Flyte literal, it will pick an encoder based on exact match of all three, _except_ for the `format`. For the format, if an exact match cannot be found, it will then search for an encoder with the format `""`. The same pattern is used for decoders/when converting from a Flyte literal to a Python dataframe.

#### Default protocols and formats
When registering an en/decoder with the `FLYTE_DATASET_TRANSFORMER` singleton, you have the option of specifying whether the protocol and storage format that that handler handles, should be taken as the default format for that dataframe type.

For example, if you

```python
from flytekit.types.structured.structured_dataset import FLYTE_DATASET_TRANSFORMER

FLYTE_DATASET_TRANSFORMER.register(MyOwnEncoder(my.DataFrame, "s3", "parquet"), default_for_type=True)
FLYTE_DATASET_TRANSFORMER.register(MyOtherEncoder(my.DataFrame, "gs", "avro"), default_for_type=False)
```

this will tell flytekit that whenever it sees

```python
def t1() -> my.DataFrame:
    ...
    return my.DataFrame(...)
```

that it will assume that the format should be "parquet" and the protocol should be "s3", and it will look for the appropriate en/decoder. 

### Usage
Please see the `tests/flytekit/unit/type_engines/structured_dataset/test_structured_dataset_workflow.py` test for user examples. The basic rules are though that if you

```python
def t1() -> some.DataFrame:
  return some.DataFrame(...)
```
then you'll get the default behavior (as determined by the default protocol and format). If you want to customize behavior (for example, specifying the uri that the data will be stored in), then you have to change the return signature and wrap the object returned inside the Python `StructuredDataset` class (as opposed to the `literals.StructuredDataset` model class).

```python
def t1() -> StructuredDataset:
  return StructuredDataset(dataframe=some.DataFrame(...), uri="bq://some/path")
```

Reading is done much the same as before...

If you
```python
def t1(a: pd.DataFrame): ...
```
then the default handler for `pd.DataFrame` will be used and whatever object the decoder returns will be what the task function is called with. Alternatively you can

```python
def t1(a: StructuredDataset):
    a.open(pd.DataFrame).all()  # or
    # a.open(pd.DataFrame).iter  # if the decoder returns an iterator.
```
## Migration and Compatibility
We'll be keeping the existing `FlyteSchema` object around for some time. However, there are still some backwards compatibility issues you may come across.

With this PR, `pd.DataFrame` and `pyspark.sql.DataFrame` will henceforth be stored in the Flyte backend as a `StructuredDataset` literal. Previously they were stored as `Schema` literals. We recommend that if you have any cached data frame tasks, that you bump the cache version and recompute.

This can also lead to compatibility issues between tasks. For example if you have:

```python
@task
def t1() -> FlyteSchema: ...

@task
def t2(a: pd.DataFrame):  ...
```
where `t2(a=t1())`, once you upgrade to this version of flytekit, when `t2` runs, it will try to use the new code, to read the old `Schema` literal. The reverse is also true

```python
@task
def t1() -> pd.DataFrame: ...

@task
def t2(a: FlyteSchema):  ...
```

To handle this, we've added code to both the existing SchemaEngine and the new StructuredDataset engine to handle the other. Let us know if you have problems.

### Plugins
We update the spark plugin made a few fixes to the great expectations plugin but most of the dataframe-related plugins remain untouched. We will be updating these in the weeks to come.

## Upcoming Changes
There's a couple changes that we just felt would be better to do after we had a beta release to play with but they should really be thought of as part of this PR.
* The change to the column/format specifying syntax `StructuredDataset[kwtypes(col_a=int), "my_parquet"]`. This is going to be replaced within the next week or two.
* Column subsetting - if a task returns two columns but the next task only needs one of them, the dataframe provided should only have that one column.

Separately,

* The handlers shipped with this preserve the existing behavior of using aws/gcp clis to download and upload data. Faster handlers that use fsspec will be coming as well.
* Aforementioned plugin updates.

### Core changes
* Add override to TypeEngine.
* Strip Annotated in TypeEngine.

#### Plugins
* Moved Spark encoder and decoder into the spark plugin.
  * Changed encoder to use a randomly generated remote directory instead of a file.
* `flytekit-sqlalchemy`, `flytekit-snowflake`, `flytekit-hive`, `flytekit-aws-athena` need backend changes to handle the new StructuredDataset type in addition to FlyteSchema.  Probably lower priority unless someone asks for it. After backend changes are done, we can update the Python type hints.
* `flytekit-modin` seems to operate on its own type, no interoperability present with the normal `pandas.DataFrame` type, so we don't have to worry about `pandas` -> `modin` compatibility.
* `flytekit-dolt` seems to be the same as `modin`.

## Tracking Issues
User submitted, not the main issue (not sure if there is one). https://github.com/flyteorg/flyte/issues/523

## Follow-up issue
_NA_


flytekit/types/structured/utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+from flytekit.configuration import aws as _aws_config
+from flytekit.core.data_persistence import split_protocol
+
+
+def get_filesystem(path: typing.Union[str, os.PathLike]) -> typing.Optional["fsspec.AbstractFileSystem"]:
+    protocol, _ = split_protocol(path)
+    if protocol == "s3":
+        kwargs = _get_s3_config()
+        import fsspec
```
wild-endeavor, line 12: let's get rid of this for the time being now also, and bring it back as a plugin after?

flytekit/types/structured/utils.py
```diff
@@ -0,0 +1,37 @@
...
+
+        return fsspec.filesystem(protocol, **kwargs)  # type: ignore
+    return None
+
+
+def get_storage_config(path: typing.Union[str, os.PathLike]) -> dict:
```
wild-endeavor, line 18: is this still used?  can we kill for now?

flytekit/types/structured/utils.py
```diff
@@ -0,0 +1,37 @@
...
+    protocol, _ = split_protocol(path)
+    if protocol == "s3":
+        return _get_s3_config()
+    return {}
+
+
+def _get_s3_config() -> dict:
```
wild-endeavor, line 25: and this

.github/workflows/pythonbuild.yml
```diff
@@ -38,6 +38,7 @@ jobs:
       run: |
         python -m pip install --upgrade pip==21.2.4 setuptools wheel
         make setup${{ matrix.spark-version-suffix }}
+        pip install fsspec
```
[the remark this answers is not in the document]
pingsutw: ok

.github/workflows/pythonbuild.yml
```diff
@@ -146,7 +148,7 @@ jobs:
           python-version: 3.8
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip==21.2.4 setuptools wheel
+          python -m pip install --upgrade pip setuptools wheel
```
pingsutw, line 151: Note: revert it before this pr is merged

dev-requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with python 3.9
 # To update, run:
 #
-#    make dev-requirements.txt
+#    pip-compile dev-requirements.in
```
pingsutw, line 5: Note: revert it before this pr is merged

flytekit/core/data_persistence.py
```diff
@@ -298,6 +298,8 @@ def __init__(self, local_sandbox_dir: Union[str, os.PathLike], raw_output_prefix
 
         self._default_remote = DataPersistencePlugins.find_plugin(raw_output_prefix)(default_prefix=raw_output_prefix)
         self._raw_output_prefix = raw_output_prefix
+        if not self.is_remote(raw_output_prefix) and not os.path.exists(raw_output_prefix):
```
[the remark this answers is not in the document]
pingsutw: `fssepc` won't create dir if not exist, so I create a new dir here.
will remove it since we're going to remove `fsspec`.

wild-endeavor (approved):

wild-endeavor (approved):

PR #792, opened 2021-12-23
Add support for string-format-like sytax for shell task

palchicz:
# TL;DR
Use pythonic string formatting like syntax for shell task variable interpolation. This allows using nested input types
such as dataclasses.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

This commit adds string-formatting like syntax (ie `"{x}".format(x=17)` for
shell_tasks. This supports using nested types for script inputs, such
as data classes. This change was motivated by the desire to combine
shell_tasks that have multiple inputs with map_tasks which only support
tasks with a single input.

This commit modifies the previous behavior of throwing an error when there
are unused input arguments. It also breaks with the convention of using the `.input` and `.output` prefixes for template
variable names. Finally, as a consequence of dropping the `input` `output` prefixes, it also makes it a value error to use the same variable name as both an input and output


tests/flytekit/unit/extras/tasks/test_shell.py
```diff
@@ -99,62 +140,144 @@ def test_input_output_substitution_files():
...
-        cat {{ .inputs.f }} >> {{ .outputs.y }}
-        echo "Hello World {{ .inputs.y }} on  {{ .inputs.j }}"
-        """
+@pytest.mark.parametrize(
+    "template_style,script,output_location",
+    [
+        (
+            "default",
+            """
+            cat {{ .inputs.f }} >> {{ .outputs.z }}
```
palchicz, line 149: example where I had to change the output variable `y` since it was the same as in the input. Did this for ease of implementation in this POC

tests/flytekit/unit/extras/tasks/test_shell.py
```diff
@@ -99,62 +140,144 @@ def test_input_output_substitution_files():
...
+            OutputLocation(var="z", var_type=FlyteFile, location=output_location_z),
         ],
     )
 
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="missing"):
         t(f=test_csv, y=testdata, j=datetime.datetime(2021, 11, 10, 12, 15, 0))
 
 
+def test_can_use_complex_types_for_inputs_to_f_string_template():
```
palchicz, line 259: main goodness. You can use nested inputs

kumare3:
@palchicz this is so much better than what I did. And this allows one to leverage the entire power of python. I think my assumption to make the variable templates look similar to how we do it in the backend is ill-advised. This is fantastic and would love to productionize this.

cc @samhita-alla / @eapolinario. 
@samhita-alla we should wait on your BLAST PR as this is better

flytekit/extras/tasks/shell.py
```diff
@@ -80,18 +59,81 @@ def _dummy_task_func():
...
+        def format_field(self, value, format_spec):
+            if isinstance(value, FlyteFile):
+                value.download()
+                return value.path
+            if isinstance(value, FlyteDirectory):
+                value.download()
+                return value.path
+            if isinstance(value, datetime.datetime):
+                return value.isoformat()
+            return super().format_field(value, format_spec)
```
samhita-alla, line 113: Can we not call `stringify()` because the function definition looks the same?

flytekit/extras/tasks/shell.py
```diff
@@ -80,18 +59,81 @@ def _dummy_task_func():
...
+
+    def interpolate(self, tmpl: str, inputs=None, outputs=None) -> str:
+        inputs = inputs or {}
+        outputs = outputs or {}
+        consolidated_args = collections.ChainMap(inputs, outputs)
```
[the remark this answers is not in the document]
palchicz: added an error check for this case

flytekit/extras/tasks/shell.py
palchicz: great question! After discussing with @kumare3 out-of-band, he suggested I get rid of the Regex based way of doing this entirely, so the stringify logic is no longer needed in two places (here and in the Regex based Interpolizer that showed up a few commits ago).  We _could_ delegate out to the stringify function within here, but since this logic is now only used in one place I think it's probably OK to just inline it here.

tests/flytekit/unit/extras/tasks/test_shell.py
```diff
@@ -100,70 +100,104 @@ def test_input_output_substitution_files():
...
-    with pytest.raises(ValueError):
+    with pytest.raises(ValueError, match="Variables {'y'} in Query"):
         t(f=test_csv, y=testdata, j=datetime.datetime(2021, 11, 10, 12, 15, 0))
 
 
+def test_can_use_complex_types_for_inputs_to_f_string_template():
+    @dataclass_json
+    @dataclass
+    class InputArgs:
+        in_file: CSVFile
```
kumare3, line 175: interesting, since this is using Flytefile, it will be automatically downloaded too?

kumare3:
This looks really cool, can you fix the build please?

wild-endeavor (approved):

kumare3:
The error has been fixed in master - this change is orthogonal. +1 and merging

PR #794, opened 2021-12-29
add `with_overrides` to map task

samhita-alla:

# TL;DR
`with_overrides` is not covered in the map task API. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1574

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -220,8 +220,8 @@ def map_task(task_function: PythonFunctionTask, concurrency: int = None, min_suc
        :language: python
        :dedent: 4
 
-    At run time, the underlying map task will be run for every value in the input collection. Task-specific attributes
-    such as :py:class:`flytekit.TaskMetadata` and :py:class:`flytekit.Resources` are applied to individual instances
+    At run time, the underlying map task will be run for every value in the input collection. Attributes
+    such as :py:class:`flytekit.TaskMetadata` and ``with_overrides`` are applied to individual instances
     of the mapped task.
 
     :param task_function: This argument is implicitly passed and represents the repeatable function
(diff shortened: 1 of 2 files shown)


flytekit/core/map_task.py
```diff
@@ -230,6 +230,11 @@ def map_task(task_function: PythonFunctionTask, concurrency: int = None, min_suc
         all inputs are processed.
     :param min_success_ratio: If specified, this determines the minimum fraction of total jobs which can complete
         successfully before terminating this task and marking it successful.
+
+    ``with_overrides`` on a map task can be used to set individual map task resource assigment.
```
samhita-alla, line 234: @katrogan, what's the preferred way of assigning resourced to individual map tasks–`TaskMetadata`/`Resources` or `with_overrides`?

katrogan: I believe it's with `with_overrides` but I haven't worked on flytekit in so long I'm not sure if it's changed! cc @wild-endeavor @eapolinario

wild-endeavor: could you try this out? I believe this should work.   calling `map_task(t1)` produces the map task, and then `()` on that should produce a `Node` which should hit https://github.com/flyteorg/flytekit/blob/master/flytekit/core/node.py#L79

wild-endeavor: but note that this is on the node... not the map task itself.  if you use the map task twice, you have to call with overrides twice.

wild-endeavor: let me know when you want a +1

samhita-alla: `TaskMetadata` works, whereas `Resources` doesn't. I replaced `Resources` with `with_overrides` in the docstring and test case.

wild-endeavor (approved):

PR #804, opened 2022-01-06
ONNX Plugin

samhita-alla:

# TL;DR
Available as:
- ScikitLearn ONNX TypeTransformer
- TensorFlow ONNX TypeTransformer
- PyTorch ONNX TypeTransformer

These plugins can help generate serialized versions of the models.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Converting an ML model to ONNX format isn't well suited to be a _task plugin_ because having so will require users to first fetch the model from a task whose type needs to be pickle or joblib. This creates friction in the way the code's structured due to serializing the model twice — pickle/joblib & ONNX. Hence, having it as a _type_ is the rational thing to do; that way, users can return the model as an ONNX file.

PyTorch example:
```python
@task
def train() -> Annotated[
    PyTorch2ONNX,
    PyTorch2ONNXConfig(
        args=torch.randn(1, 1, 224, 224, requires_grad=True),
        export_params=True,  # store the trained parameter weights inside
        opset_version=10,  # the ONNX version to export the model to
        do_constant_folding=True,  # whether to execute constant folding for optimization
        input_names=["input"],  # the model's input names
        output_names=["output"],  # the model's output names
        dynamic_axes={"input": {0: "batch_size"}, "output": {0: "batch_size"}},  # variable length axes
    ),
]:
    ...

@task
def onnx_predict(model_file: FlyteFile[TypeVar("onnx")]) -> JPEGImageFile:
    ...

@workflow
def wf() -> JPEGImageFile:
    model = train()
    return onnx_predict(model_file=model)
```

## Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/2553

## Follow-up issue
_NA_
OR

## TODO
- [ ] Add examples to docs


samhita-alla:
@kumare3 / @cosmicBboy
1. We have three ONNX TypeTransformers currently: ScikitLearn, TensorFlow, PyTorch; all of them seem the same except for the ONNX generator function call. Is there a way to reduce the redundancy? (I don't think it's a good idea to have all of them in a single plugin as it installs unnecessary dependencies)
2. The plugins don't really seem to solve a significant problem. They take a trained model and produce the ONNX version. Does this suffice the core idea of having a plugin, or is there a scope to add anything else to this?

kumare3:
I agree, they don't solve a problem. That's why maybe it's not a plugin. I wanted to look at optimizing these models. Also I don't think this should be type transformers

plugins/flytekit-onnx-scikitlearn/tests/test_onnx_scikitlearn.py
```diff
@@ -0,0 +1,112 @@
...
+        X, y = iris.data, iris.target
+        X_train, X_test, y_train, _ = train_test_split(X, y)
+        model = RandomForestClassifier()
+        model.fit(X_train, y_train)
+
+        return TrainOutput(test=X_test, model=ScikitLearn2ONNX(model))
+
+    @task
+    def predict(
+        model: FlyteFile[TypeVar("onnx")],
```
pingsutw, line 49: I think we can add `TypeVar("onnx")` to this [file](https://github.com/flyteorg/flytekit/blob/master/flytekit/types/file/__init__.py), and here we can update to `model: ONNXFile`

pingsutw (commented):
We need to update [pythonbuild.yml](https://github.com/flyteorg/flytekit/blob/a968a5aa8338adbec10a38f79c5fdddb72c7c161/.github/workflows/pythonbuild.yml#L66-L85)

plugins/flytekit-onnx-scikitlearn/flytekitplugins/onnxscikitlearn/schema.py
```diff
@@ -0,0 +1,140 @@
...
+    def to_python_value(
+        self,
+        ctx: FlyteContext,
+        lv: Literal,
+        expected_python_type: Type[FlyteFile],
+    ) -> FlyteFile:
+        if not lv.scalar.blob.uri:
+            raise TypeTransformerFailedError(f"ONNX format isn't of the expected type {expected_python_type}")
+
+        return FlyteFile[self.ONNX_FORMAT](path=lv.scalar.blob.uri)
```
pingsutw, line 127: I think here we should return `ScikitLearn2ONNX(model=...)`

pingsutw: I see, sorry, ignore me

samhita-alla:
@pingsutw, updated `pythonbuild.yaml` and the tests are passing. Also, added `ONNXFile` as a FlyteFile type.

kumare3 (approved):

PR #811, opened 2022-01-11
[pr into #785] Remove fsspec

pingsutw:
# TL;DR
- Remove fsspec
- remote `util.py` 
- Add e2e test to `test_workflows.py` 
- Revert `data_persistence.py `


tests/flytekit/unit/core/test_workflows.py
```diff
@@ -1,15 +1,24 @@
...
 from flytekit.core.condition import conditional
 from flytekit.core.context_manager import Image, ImageConfig
 from flytekit.core.task import task
 from flytekit.core.workflow import WorkflowFailurePolicy, WorkflowMetadata, WorkflowMetadataDefaults, workflow
+from flytekit.types.schema import FlyteSchema
+
+try:
+    from typing import Annotated
+except ImportError:
+    from typing_extensions import Annotated
```
eapolinario, line 21: Do we need to do this? We're already pulling `typing_extensions` as part of a regular `flytekit` installation, right?

tests/flytekit/unit/core/test_workflows.py
```diff
@@ -1,15 +1,24 @@
 import typing
 from collections import OrderedDict
 
+import pandas as pd
 import pytest
+from pandas._testing import assert_frame_equal
```
eapolinario, line 6: This is more of a nit, but can we import that function from `pandas.testing` instead? It's present in the [docs](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html#pandas-testing-assert-frame-equal).

flytekit/types/structured/basic_dfs.py
```diff
@@ -76,7 +75,10 @@ def encode(
     ) -> literals.StructuredDataset:
         path = typing.cast(str, structured_dataset.uri) or ctx.file_access.get_random_remote_path()
         df = structured_dataset.dataframe
-        pq.write_table(df, path, filesystem=get_filesystem(path))
+        local_dir = ctx.file_access.get_random_local_directory()
+        local_path = os.path.join(local_dir, f"{0:05}")
```
eapolinario, line 79: Can you leave a comment explaining why we need a file named `00000`? This is already present in `dataset_example.py` (I'm not sure why), but that doesn't have a comment either.

wild-endeavor: will add a comment... this is only to keep the semantics the same as existing behavior.

tests/flytekit/unit/core/test_workflows.py
wild-endeavor: hmm?  no we need this.  this is for compatibility between 3.7/8 and 3.9

wild-endeavor:
i think one of the integration tests timed out too, next run will probably be fine.

tests/flytekit/unit/core/test_workflows.py
eapolinario: Why? We already have typing_extensions, so we could rely on that for all versions, including 3.9-10.

pingsutw: From the user side, they may use `from typing import Annotated ` or `from typing_extensions import Annotated`. Therefore, we need to make sure both of them can work here.

wild-endeavor (approved):

PR #829, opened 2022-01-25
Parse duration field from flyteidl to `flytekit.models.execution.ExecutionClosure`

bstadlbauer:
# TL;DR
This passes on the duration field of the `ExecutionClosure` `fltyteidl` object so that it is available from Python. It converts it to 
a `datetime.timedelta` upon deserializing.

This is a first draft, please let us know what you think and whether this is useful in general. We would add a few tests after this.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
This passes on the duration field of the `ExecutionClosure` `fltyteidl` object so that it is available from Python. It converts it to 
a `datetime.timedelta` upon deserializing.

## Tracking Issue
-

## Follow-up issue
-

### DIFF
--- a/tests/flytekit/unit/models/test_execution.py
+++ b/tests/flytekit/unit/models/test_execution.py
@@ -18,33 +18,52 @@ _OUTPUT_MAP = _literals.LiteralMap(
 )
 
 
-def test_execution_closure():
+def test_execution_closure_with_output():
     test_datetime = datetime.datetime(year=2022, month=1, day=1, tzinfo=pytz.UTC)
     test_timedelta = datetime.timedelta(seconds=10)
-    test_error = _core_exec.ExecutionError(
-        code="foo", message="bar", error_uri="http://foobar", kind=_core_exec.ExecutionError.ErrorKind.USER
-    )
     test_outputs = _execution.LiteralMapBlob(values=_OUTPUT_MAP, uri="http://foo/")
 
     obj = _execution.ExecutionClosure(
         phase=_core_exec.WorkflowExecutionPhase.SUCCEEDED,
         started_at=test_datetime,
         duration=test_timedelta,
-        error=test_error,
         outputs=test_outputs,
     )
     assert obj.phase == _core_exec.WorkflowExecutionPhase.SUCCEEDED
     assert obj.started_at == test_datetime
     assert obj.duration == test_timedelta
-    assert obj.error == test_error
     assert obj.outputs == test_outputs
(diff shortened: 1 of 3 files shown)


eapolinario:
@bstadlbauer , this looks reasonable. 

Let's add some unit tests, including tests that exercise the flyteremote paths.

bstadlbauer:
@eapolinario I've added some tests. I could not get the integration test setup to run on my local machine, which is why I've only added a very short test to it. Let me know if you'd like to have more there.

tests/flytekit/unit/models/test_execution.py
```diff
@@ -15,6 +18,35 @@
...
+    assert obj.started_at == test_datetime
+    assert obj.duration == test_timedelta
+    assert obj.error == test_error
+    assert obj.outputs == test_outputs
+    obj2 = _execution.ExecutionClosure.from_flyte_idl(obj.to_flyte_idl())
+    assert obj2 == obj
+    assert obj2.phase == _core_exec.WorkflowExecutionPhase.SUCCEEDED
+    assert obj2.started_at == test_datetime
+    assert obj2.duration == test_timedelta
+    # assert obj2.error == test_error  # FIXME: This won't work for some reason?
```
eapolinario, line 46: what error are you seeing here?

eapolinario (commented):
@bstadlbauer , thanks for adding the tests. Don't worry about adding more integration tests, that single one is enough.

I have a minor comment about the `error` field in the unit test, otherwise this LGTM.

tests/flytekit/unit/models/test_execution.py
mdjong1: I believe the issue is that after serializing and deserializing the `obj` the error field goes blank
image

eapolinario: I did some digging and this is a separate bug. If you check the [protobuf definition](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/admin/execution.proto#L127-L144), `ExecutionClosure` has a `one_of` field that should cover `outputs` and `error`, i.e. those two fields should not be set at the same time. Unfortunately the protobuf python library [does not raise any error in this case](https://github.com/flyteorg/flytekit/blob/master/flytekit/models/execution.py#L341-L345) (essentially the last write wins, in this case `outputs`). 

tldr: this is a 3 year old bug in the use of the python protobuf library and should not impact your change.

eapolinario (approved):

PR #831, opened 2022-01-26
Gate new Structured Dataset feature

wild-endeavor:

# TL;DR
The Structured Dataset type will not work if users have not upgraded their backend. This introduces a feature gate that will disable it by default.

We'll revert this PR by 1.0.0 if not earlier.  At that time, if you have not upgraded your Flyte backend, compilation of new tasks will fail if they use dataframes since Admin/Propeller won't recognize the new type.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
* Add a `conftest.py` file to all tests directories.
* Add a new test directory that doesn't have the `conftest.py` file so that we can test the older case.
* Set the environment variable to use SDs in all Dockerfiles
* Remove some old configs while we're at it.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2074


tests/flytekit/conftest.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+
+def pytest_sessionstart():
+    os.environ["FLYTE_SDK_USE_STRUCTURED_DATASET_OLD"] = os.environ.get("FLYTE_SDK_USE_STRUCTURED_DATASET_OLD", "FALSE")
```
eapolinario, line 5: no need to have 2 fixtures, you can instead use `yield` to restore the old values, like so:
```
def pytest_session():
    old_value = os.environ["FLYTE_SDK_USE_STRUCTURED_DATASET"]
    # Use Structured Datasets. This should get picked up automatically by the loading of structured_dataset.py
    os.environ["FLYTE_SDK_USE_STRUCTURED_DATASET"] = "TRUE"
    yield
    os.environ["FLYTE_SDK_USE_STRUCTURED_DATASET"] = old_value

File: Makefile [hunk 2]
```diff
@@ -47,11 +47,11 @@ spellcheck:  ## Runs a spellchecker over all code and documentation
 
 .PHONY: test
 test: lint ## Run tests
-	pytest tests/flytekit/unit
+	pytest tests/flytekit/unit tests/flytekit_compatibility
 
 .PHONY: unit_test
 unit_test:
-	pytest tests/flytekit/unit
+	pytest tests/flytekit/unit tests/flytekit_compatibility
```
eapolinario, line 54: just to be safe, should we run the tests twice (with the flag set to false and true)?

Makefile
```diff
@@ -47,11 +47,11 @@ spellcheck:  ## Runs a spellchecker over all code and documentation
 
 .PHONY: test
 test: lint ## Run tests
-	pytest tests/flytekit/unit
+	pytest tests/flytekit/unit tests/flytekit_compatibility
```
eapolinario, line 50: how about we make `unit_test` a pre-requirement of the `test` target instead?

flytekit/configuration/sdk.py
```diff
@@ -32,47 +18,6 @@
...
-This is a Python format string which the SDK will use to generate names for workflows. Any template portion of the
-string can only include 'module' or 'name'.  So '{name}' is valid, but '{key}' is not. If not specified,
-we fall back to the configuration for :py:attr:`flytekit.configuration.sdk.NAME_FORMAT`
-"""
-
-LAUNCH_PLAN_NAME_FORMAT = _config_common.FlyteStringConfigurationEntry(
-    "sdk", "launch_plan_name_format", fallback=NAME_FORMAT
-)
-"""
-This is a Python format string which the SDK will use to generate names for launch plans. Any template portion of the
```
eapolinario, line 71: are these leftovers from the legacy code removal?

flytekit/configuration/platform.py
```diff
@@ -14,10 +13,6 @@
 
 INSECURE = _config_common.FlyteBoolConfigurationEntry("platform", "insecure", default=False)
 
-CLOUD_PROVIDER = _config_common.FlyteStringConfigurationEntry(
```
eapolinario, line 17: ditto.

flytekit/configuration/sdk.py
wild-endeavor: yup

flytekit/configuration/platform.py
wild-endeavor: yup

Makefile
wild-endeavor: not sure i understand... what do you mean?

eapolinario: sorry, I mean run the tests once where the env var `FLYTE_SDK_USE_STRUCTURED_DATASET` is set to `TRUE` and again with it set to false (or possibly unset?)

wild-endeavor: this is pretty hard to do...  i can copy/paste some tests into the compatibility folder.

eapolinario (approved):

PR #832, opened 2022-01-27
`ctx` Context can be used within shell tasks - to access context vars and secrets

kumare3:

# TL;DR
- Allows users to use `ctx.*` in shell tasks
- For secrets, it is necessary to use the attribute accessor api - `ctx.secrets.group.key`. The group and key are the actual group and key values to look for
- Also all inputs and outputs are now namesapced. So to access inputs it is necessary to use `inputs.x`

Example:
```python
    t = ShellTask(
        name="test",
        script="""
            export EXEC_ID={ctx.execution_id}
            export SEC={ctx.secrets.group.key}
            cat {inputs.f}
            echo "Hello World {inputs.y} on  {inputs.j}"
            """,
        inputs=kwtypes(f=CSVFile, y=FlyteDirectory, j=datetime.datetime),
    )
```

all variables supported by flytekit.current_context are replaceable.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2110

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -325,23 +325,6 @@ class ExecutionParameters(object):
         return self.__getattr__(attr_name=key)
 
 
-class _GroupSecrets(object):
-    """
-    This is a dummy class whose sole purpose is to support "attribute" style lookup for secrets
-    """
-
-    def __init__(self, group: str, sm: typing.Any):
-        self._group = group
-        self._sm = sm
-
-    def __getattr__(self, item: str) -> str:
-        """
-        Returns the secret that matches "group"."key"
-        the key, here is the item
-        """
-        return self._sm.get(self._group, item)
-
-
 class SecretsManager(object):
     """
     This provides a secrets resolution logic at runtime.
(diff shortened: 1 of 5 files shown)


flytekit/core/context_manager.py
```diff
@@ -325,6 +325,23 @@ def get(self, key: str) -> typing.Any:
...
+    def __init__(self, group: str, sm: typing.Any):
+        self._group = group
+        self._sm = sm
+
+    def __getattr__(self, item: str) -> str:
+        """
+        Returns the secret that matches "group"."key"
+        the key, here is the item
+        """
+        return self._sm.get(self._group, item)
```
eapolinario, line 342: can we make this an inner class to `SecretsManager`?

kumare3: can do

eapolinario (approved):

PR #835, opened 2022-01-27
Update argument setting for in fast registered, dynamically generated, pod tasks

wild-endeavor:

# TL;DR
Also updates non-pod (regular container) tasks.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Add the second fast register setting (`distribution_location`) to the `FastSerializationSettings` object.
This also changes the behavior for non-pod tasks.  Basically to get the arguments to be filled in, you have to call `dispatch_execute` instead of just `compile_to_workflow` since the former is where the fast serialization settings are set. This should be okay though since that function isn't really exposed to the public but that is why there's a unit test that had to change.

Example of how the Pod spec of the child task looks now in local sandbox:
```
  containers:
  - args:
    - pyflyte-fast-execute
    - --additional-distribution
    - s3://my-s3-bucket/fast/fastb103ce16298e09cdb56426aa6a1f14439b82ea15-fast3545d54894176ff1583873bd88b200f5.tar.gz
    - --dest-dir
    - .
    - --
    - pyflyte-execute
    - --inputs
    - s3://my-s3-bucket/metadata/propeller/flytesnacks-development-s9yz7st064/podpodwithdynamicdynamictaskwithpodsubtask/data/0/dn0/inputs.pb
    - --output-prefix
    - s3://my-s3-bucket/metadata/propeller/flytesnacks-development-s9yz7st064/podpodwithdynamicdynamictaskwithpodsubtask/data/0/dn0/0
    - --raw-output-data-prefix
    - s3://my-s3-bucket/v2/s9yz7st064-fwfrds3q-0-dn0-0
    - --checkpoint-path
    - s3://my-s3-bucket/v2/s9yz7st064-fwfrds3q-0-dn0-0/_flytecheckpoints
    - --prev-checkpoint
    - '""'
    - --resolver
    - flytekit.core.python_auto_container.default_task_resolver
    - --
    - task-module
    - pod.pod_with_dynamic
    - task-name
    - pod_task_with_resources
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2072

### DIFF
--- a/flytekit/tools/translator.py
+++ b/flytekit/tools/translator.py
@@ -68,19 +68,17 @@ def _fast_serialize_command_fn(
 ) -> Callable[[SerializationSettings], List[str]]:
     default_command = task.get_default_command(settings)
 
-    dest_dir = (
-        settings.fast_serialization_settings.destination_dir if settings.fast_serialization_settings is not None else ""
-    )
-    if dest_dir is None or dest_dir == "":
-        dest_dir = "{{ .dest_dir }}"
-
     def fn(settings: SerializationSettings) -> List[str]:
         return [
             "pyflyte-fast-execute",
             "--additional-distribution",
-            "{{ .remote_package_path }}",
+            settings.fast_serialization_settings.distribution_location
+            if settings.fast_serialization_settings and settings.fast_serialization_settings.distribution_location
+            else "{{ .remote_package_path }}",
             "--dest-dir",
-            dest_dir,
+            settings.fast_serialization_settings.destination_dir
+            if settings.fast_serialization_settings and settings.fast_serialization_settings.destination_dir
+            else "{{ .dest_dir }}",
             "--",
             *default_command,
         ]
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -400,7 +400,11 @@ class FastSerializationSettings(object):
     """
 
     enabled: bool = False
+    # This is the location that the code should be copied into.
     destination_dir: Optional[str] = None
+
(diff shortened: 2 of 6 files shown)


flytekit/tools/translator.py
```diff
@@ -68,19 +68,17 @@ def _fast_serialize_command_fn(
...
-    if dest_dir is None or dest_dir == "":
-        dest_dir = "{{ .dest_dir }}"
-
     def fn(settings: SerializationSettings) -> List[str]:
         return [
             "pyflyte-fast-execute",
             "--additional-distribution",
-            "{{ .remote_package_path }}",
+            settings.fast_serialization_settings.distribution_location
+            if settings.fast_serialization_settings and settings.fast_serialization_settings.distribution_location
```
katrogan, line 76: should you also check if `settings.fast_serialization_settings.enabled`?

katrogan (approved):
can we please add an integration test for this too somewhere?

flytekit/tools/translator.py
wild-endeavor: no i don't think so this is the fast execute command, if it's not enabled by the time we get here something else is wrong.

wild-endeavor:
I don't know how to do an integration test - cc @eapolinario let's think about this in the future

katrogan (approved):

eapolinario (approved):

PR #841, opened 2022-02-01
Set default values to map task template

eapolinario:
# TL;DR
Set sane defaults in the map task template, matching the defaults in the [backend](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/array/core/state.go#L154-L160).

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Map tasks containing custom fields would always succeed. That happened because in such cases we were not serializing `min_success_ratio`. In this PR we set defaults to the custom field.

I'm leaving support for `min_successes` intact just in case we have some client out there that's building these array jobs manually using flytekit.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -209,7 +209,7 @@ class MapPythonTask(PythonTask):
         return outputs
 
 
-def map_task(task_function: PythonFunctionTask, concurrency: int = 1, min_success_ratio: float = 1.0, **kwargs):
+def map_task(task_function: PythonFunctionTask, concurrency: int = 0, min_success_ratio: float = 1.0, **kwargs):
     """
     Use a map task for parallelizable tasks that run across a list of an input type. A map task can be composed of
     any individual :py:class:`flytekit.PythonFunctionTask`.
@@ -231,7 +231,7 @@ def map_task(task_function: PythonFunctionTask, concurrency: int = 1, min_succes
     :param task_function: This argument is implicitly passed and represents the repeatable function
     :param concurrency: If specified, this limits the number of mapped tasks than can run in parallel to the given batch
         size. If the size of the input exceeds the concurrency value, then multiple batches will be run serially until
-        all inputs are processed.
+        all inputs are processed. If left unspecified, this means unbounded concurrency.
     :param min_success_ratio: If specified, this determines the minimum fraction of total jobs which can complete
         successfully before terminating this task and marking it successful.
 
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -209,7 +209,7 @@ class MapPythonTask(PythonTask):
         return outputs
 
 
-def map_task(task_function: PythonFunctionTask, concurrency: int = None, min_success_ratio: float = None, **kwargs):
+def map_task(task_function: PythonFunctionTask, concurrency: int = 1, min_success_ratio: float = 1.0, **kwargs):
     """
     Use a map task for parallelizable tasks that run across a list of an input type. A map task can be composed of
     any individual :py:class:`flytekit.PythonFunctionTask`.
--- a/flytekit/models/array_job.py
+++ b/flytekit/models/array_job.py
@@ -85,3 +85,24 @@ class ArrayJob(_common.FlyteCustomIdlEntity):
             )
 
         return _json_format.MessageToDict(array_job)
+
+    @classmethod
+    def from_dict(cls, idl_dict):
+        """
+        :param dict[T, Text] idl_dict:
+        :rtype: ArrayJob
+        """
+        pb2_object = _json_format.Parse(_json.dumps(idl_dict), _array_job.ArrayJob())
+
+        if pb2_object.HasField("min_successes"):
+            return cls(
+                parallelism=pb2_object.parallelism,
+                size=pb2_object.size,
+                min_successes=pb2_object.min_successes,
+            )
+        else:
+            return cls(
+                parallelism=pb2_object.parallelism,
+                size=pb2_object.size,
+                min_success_ratio=pb2_object.min_success_ratio,
+            )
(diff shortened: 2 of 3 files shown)


hamersaw:
Do we want to set the default concurrency / parallelism to 1? It seems that by default we would want to execute all subtasks concurrently? We could use 0 (or a negative number) to denote that parallelism is not applicable. These would be changes reflected in plugins as well.

This is a similar conversation in setting the default min_success_ratio to set to 1.0 rather than 0.0. I think in this case there is no conceivable scenario where a user would schedule a map task and not allow any tasks to be executed, so 0 is not a valid parallelism.

eapolinario:
> Do we want to set the default concurrency / parallelism to 1? It seems that by default we would want to execute all subtasks concurrently? We could use 0 (or a negative number) to denote that parallelism is not applicable. These would be changes reflected in the flyteplugins as well.

This suggestion looks reasonable. We can leverage the special value `0` to do this and add a check in the client code to ensure that value is never set.

flytekit/models/array_job.py
```diff
@@ -70,13 +70,21 @@ def to_dict(self):
...
         :rtype: dict[T, Text]
         """
-        return _json_format.MessageToDict(
-            _array_job.ArrayJob(
+        array_job = None
+        if self.min_successes is not None:
+            array_job = _array_job.ArrayJob(
                 parallelism=self.parallelism,
                 size=self.size,
                 min_successes=self.min_successes,
```
kumare3, line 78: is it one-of min_successes and min_success_ratio?

flytekit/core/map_task.py
```diff
@@ -209,7 +209,7 @@ def _raw_execute(self, **kwargs) -> Any:
         return outputs
 
 
-def map_task(task_function: PythonFunctionTask, concurrency: int = None, min_success_ratio: float = None, **kwargs):
+def map_task(task_function: PythonFunctionTask, concurrency: int = 1, min_success_ratio: float = 1.0, **kwargs):
```
kumare3, line 212: I do not think concurrency should be `1` it should be `0`, which means unbounded?

kumare3 (commented):
i think the concurrency should be 0 -> unbounded

flytekit/models/array_job.py
eapolinario: yes:  https://github.com/flyteorg/flyteidl/blob/091505aae78d935ae55e4d0e46deee5a1e79a22f/protos/flyteidl/plugins/array_job.proto#L20-L29

flytekit/core/map_task.py
eapolinario: sure thing.

flytekit/models/array_job.py
```diff
@@ -70,13 +70,21 @@ def to_dict(self):
...
             )
-        )
+        elif self.min_success_ratio is not None:
```
hamersaw, line 80: quick thought, should this be changed to check for 1.0 instead of None? If we default to 1.0 in both flytekit and flyteplugins then there is no reason to write the custom arrayjob if it is 1.0.

eapolinario: That is a great alternative IMO.

flytekit/models/array_job.py
```diff
@@ -70,13 +70,24 @@ def to_dict(self):
...
+            array_job = _array_job.ArrayJob(
+                parallelism=self.parallelism,
+                size=self.size,
+                min_success_ratio=self.min_success_ratio,
+            )
+
+        array_job_dict = _json_format.MessageToDict(array_job)
+        if self.parallelism is not None and self.parallelism == 0:
+            array_job_dict['parallelism'] = '0'
```
eapolinario, line 89: I dislike the fact that we have to write this code to handle the case of `0`. How horrible would it be to use `-1` to represent unbounded instead?

flytekit/models/array_job.py
hamersaw: I suspect I don't have enough context for where this function is used. It seems to check if the provided values are different then the default to populate the array_job only if necessary. If they are the default values, then we don't write the custom array job parameters and the backend used the default.

I think the 0 actually doesn't mean 0 only applies for parallelism. Though probably unlikely, having a 0.0 min_success_ratio means the map task should succeed even if all subtasks fail, which could be valid. However, a parallelism of 0, meaning no tasks can execute, doesn't make any sense. Do we need to add another check here if parallelism != 0?

eapolinario: > I suspect I don't have enough context for where this function is used. It seems to check if the provided values are different then the default to populate the array_job only if necessary. If they are the default values, then we don't write the custom array job parameters and the backend used the default.

This function translates the python values to protobuf, which ends up hitting the issue of [default values](https://developers.google.com/protocol-buffers/docs/proto3#default), specifically how the default value for numeric values is 0. There's no way to set the default value to a different value in a numeric field. 

> 
> I think the 0 actually doesn't mean 0 only applies for parallelism. Though probably unlikely, having a 0.0 min_success_ratio means the map task should succeed even if all subtasks fail, which could be valid. However, a parallelism of 0, meaning no tasks can execute, doesn't make any sense. 

The key here is exactly what you said. In other words, a parallelism of 0 makes sense as a magic value (and not as a user-input). However a min_success_ratio of 0 might make sense as a user-input and since we use a numeric field to represent this in protobuf we cannot distinguish between the two cases in the case of min_success_ratio.

> Do we need to add another check here if parallelism != 0?
IMO, to simplify usage we should allow users to set parallelism to 0 with the caveat tha this means unbounded concurrency. The scenario I'm thinking is someone generating these configurations programmatically, if we we don't allow for parallelism to be set to 0 they will have to special case that in their configuration (which is very annoying).

hamersaw (approved):

PR #838, opened 2022-01-28
Regenerate requirements files - dependencies dropping support for python 3.7

eapolinario:
Signed-off-by: Eduardo Apolinario <eapolinario@users.noreply.github.com>

# TL;DR
Some flytekit dependencies are already dropping support for python 3.7, including `numpy` and `pandas`. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
In their latest release both pandas (starting on [version 1.4.0](https://github.com/pandas-dev/pandas/releases/tag/v1.4.0)) and numpy (starting on [version 1.22.0](https://github.com/numpy/numpy/releases/tag/v1.22.0)) dropped support for python 3.7, which is supposed to reach EOL only on June-2023 (according to https://www.python.org/dev/peps/pep-0537/#id27).

This PR supersedes https://github.com/flyteorg/flytekit/pull/845.

## Tracking Issue
https://github.com/lyft/flyte/issues/<number>

## Follow-up issue
_NA_
OR
_https://github.com/lyft/flyte/issues/<number>_

PR #845, opened 2022-02-04
Loosen up click constraint

eapolinario:

# TL;DR
`click` stable version reached 8.0 recently. Time to bump that requirement to 9.0

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
The upper bound constraint in the version of `click` has been set to `8.0` for a [long time](https://github.com/tnsetting/flytekit/commit/60737e306d72f05d07de2d4598b8e6a281176c0e#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7), but `click` has recently launched version `8.0`. We're starting to see some of our dependencies require at least 8.0 (for example, `papermill` requires `black` which just recently set the lower bound to ['>=8.0.0'](https://github.com/psf/black/blame/main/setup.py#L100)).
 

## Tracking Issue

## Follow-up issue
_NA_
OR


setup.py
```diff
@@ -39,7 +39,7 @@
         "wheel>=0.30.0,<1.0.0",
         "pandas>=1.0.0,<2.0.0",
         "pyarrow>=4.0.0,<7.0.0",
-        "click>=6.6,<8.0",
+        "click>=6.6,<9.0",
```
wild-endeavor, line 42: can we just remove the upper bound?

eapolinario: we could. I guess the idea for upper-bounds is to catch the case where the API changes between major versions, but I'm not sure if we need this kind of protection for `click`.

wild-endeavor (approved):

eapolinario:
Superseded by https://github.com/flyteorg/flytekit/pull/838.

PR #849, opened 2022-02-08
Add structured dataset encoder/decoder in fsspec plugin

pingsutw:

# TL;DR
We can override default structured encoder/decoder by installing fsspec plugin
## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2074

## Follow-up issue
_NA_


wild-endeavor:
What do you think about doing the following?  (I can take over this if you don't want to work on it anymore... just leaving some thoughts here for whoever does end up working on it.)

* Can you confirm that these plugins get loaded after the default ones?  This is more of a python packaging question.  Basically I'm not sure how the entrypoints stuff works.
* Is overloading the `LOCAL` protocol really necessary?
* Should we make it so that `pip install flytekitplugins-fsspec[aws]` registers the s3 protocol, and `[gcp]` registers the gs protocol.  And if the answer to the previous question is yes, the a plain `pip install flytekitplugins-fsspec` should register the local protocol.  If there happens to be code shared between s3 and gcs, we can factor that out to a common place.
  * I know this isn't how it's done right now.  In `persist.py` at the top level, we have s3 handling logic, so the correct answer to this question might be 'no'.  But having everything together, especially once we add azure, will be pretty confusing.

pingsutw:
1. yeah, this plugin gets loaded after the default ones
2. It's not necessary, but the local encoder/decoder will have different behavior. The encoder in fsspec-plugin will not write dataframe to the intermediate directory like the [default encoder](https://github.com/flyteorg/flytekit/blob/7bc0d4bd8898a30cafd5f68458f893e87bf14a79/flytekit/types/structured/basic_dfs.py#L82-L85)
3. just added `extras_require` in `setup.py`, so now we only override the s3 encoder/decoder when users install `s3fs`

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/basic_dfs.py
```diff
@@ -0,0 +1,131 @@
...
+        self,
+        ctx: FlyteContext,
+        flyte_value: literals.StructuredDataset,
+    ) -> pd.DataFrame:
+        uri = flyte_value.uri
+        if flyte_value.metadata.structured_dataset_type.columns:
+            columns = []
+            for c in flyte_value.metadata.structured_dataset_type.columns:
+                columns.append(c.name)
+            return pd.read_parquet(uri, columns=columns, storage_options=get_storage_options(uri))
```
wild-endeavor, line 77: do you know anything about this @pingsutw https://unionai.slack.com/archives/C01H0FN1NJX/p1644861917030349.  I can test this tomorrow when I'm working on it, but something to be aware of.  It seems that if it's a public bucket, then the `aws s3` cli command needs to be different.  Not sure if the to_parquet logic is smart enough to handle that.  (only affects reading for now, don't think we'll be allowing public writes for a while)

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/basic_dfs.py
```diff
@@ -0,0 +1,131 @@
...
+from flytekit.types.structured.structured_dataset import (
+    PARQUET,
+    S3,
+    StructuredDataset,
+    StructuredDatasetDecoder,
+    StructuredDatasetEncoder,
+    StructuredDatasetTransformerEngine,
+)
+
+T = TypeVar("T")
```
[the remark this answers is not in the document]
pingsutw: Agreed to break up to multiple files. I like to do the registration in the init file.

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/basic_dfs.py
pingsutw: I just tested it. both pandas and arrow can read from private or public buckets without change.
Here is the sample code. https://gist.github.com/pingsutw/c663070ae4413f980d0a24cf4e7cd091

wild-endeavor: Nice, thank you!  We'll just have to change the normal one then.

flytekit/clients/friendly.py
```diff
@@ -588,11 +588,12 @@ def get_execution_data(self, id):
...
         :rtype: flytekit.models.execution.WorkflowExecutionGetDataResponse
         """
-        return _execution.WorkflowExecutionGetDataResponse.from_flyte_idl(
-            super(SynchronousFlyteClient, self).get_execution_data(
-                _execution_pb2.WorkflowExecutionGetDataRequest(id=id.to_flyte_idl())
-            )
+        t = super(SynchronousFlyteClient, self).get_execution_data(
+            _execution_pb2.WorkflowExecutionGetDataRequest(id=id.to_flyte_idl())
         )
+        print("id = ", id)
```
wild-endeavor, line 594: kill

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
```diff
@@ -0,0 +1,69 @@
...
+from flytekit.types.structured.structured_dataset import (
+    PARQUET,
+    S3,
+    StructuredDataset,
+    StructuredDatasetDecoder,
+    StructuredDatasetEncoder,
+)
+
+
+def get_storage_options(uri: str) -> typing.Optional[typing.Dict]:
```
wild-endeavor, line 22: you're sure this function does what you want right?  it feels kinda weird to me.  cuz it calls s3_setup_args but there's like some duplicate logic there.  as long as you're happy with it.   (there's some funniness in the unit test i didn't catch the first time, sorry about that.  see the comment there)

plugins/flytekit-data-fsspec/tests/test_basic_dfs.py
```diff
@@ -0,0 +1,46 @@
...
+
+try:
+    from typing import Annotated
+except ImportError:
+    from typing_extensions import Annotated
+
+
+def test_get_storage_options():
+    uri = "s3://bucket/somewhere"
+    with aws.S3_ENDPOINT.get_patcher(uri):
```
wild-endeavor, line 16: this option isn't supposed to be the bucket.  We added this to support minio in local execution.  See https://github.com/flyteorg/flyte/blob/master/deployment/sandbox/flyte_helm_generated.yaml#L303

wild-endeavor (commented):
did you test this on local sandbox btw?  i think we should

flytekit/clients/friendly.py
pingsutw: sorry, pushed it by accident

plugins/flytekit-data-fsspec/tests/test_basic_dfs.py
pingsutw: Updated it, thanks.

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
pingsutw: just revised it

flytekit/clients/friendly.py
pingsutw: > did you test this on local sandbox btw? 

I did

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/arrow.py
```diff
@@ -0,0 +1,60 @@
...
+        uri = flyte_value.uri
+        if not ctx.file_access.is_remote(uri):
+            Path(uri).parent.mkdir(parents=True, exist_ok=True)
+        _, path = split_protocol(uri)
+        filesystem = FSSpecPersistence._get_filesystem(uri)
+        if flyte_value.metadata.structured_dataset_type.columns:
+            columns = []
+            for c in flyte_value.metadata.structured_dataset_type.columns:
+                columns.append(c.name)
+            return pq.read_table(path, filesystem=filesystem, columns=columns)
```
wild-endeavor, line 59: do we need to add the stupid `"anon": True` bit here too and a try/catch?

pingsutw: added it

plugins/flytekit-data-fsspec/setup.py
```diff
@@ -4,7 +4,7 @@
 
 microlib_name = f"flytekitplugins-data-{PLUGIN_NAME}"
 
-plugin_requires = ["flytekit>=0.21.3,<1.0.0", "fsspec>=2021.7.0"]
+plugin_requires = ["flytekit>=0.21.3,<1.0.0", "fsspec>=2021.7.0", "botocore==1.23.24"]
```
wild-endeavor, line 7: remove the pin?

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
```diff
@@ -0,0 +1,75 @@
...
+        if flyte_value.metadata.structured_dataset_type.columns:
+            columns = []
+            for c in flyte_value.metadata.structured_dataset_type.columns:
+                columns.append(c.name)
+        try:
+            return pd.read_parquet(uri, columns=columns, storage_options=kwargs)
+        except NoCredentialsError as e:
+            logger.debug("S3 source detected, attempting anonymous S3 access")
+            kwargs["anon"] = True
+            pd.read_parquet(uri, columns=columns, storage_options=kwargs)
```
wild-endeavor, line 74: missing return?

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
```diff
@@ -0,0 +1,75 @@
...
+            raise e
```
wild-endeavor, line 75: we want to raise here?

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
wild-endeavor: also shouldn't we condition this on the uri being an S3 uri?

pingsutw: added it. 
we catch this error only when we try to access the AWS bucket, so the URI here will always be an s3 URI.

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
```diff
@@ -0,0 +1,74 @@
...
+        uri = flyte_value.uri
+        columns = None
+        kwargs = get_storage_options(uri)
+        if flyte_value.metadata.structured_dataset_type.columns:
+            columns = []
+            for c in flyte_value.metadata.structured_dataset_type.columns:
+                columns.append(c.name)
+        try:
+            return pd.read_parquet(uri, columns=columns, storage_options=kwargs)
+        except NoCredentialsError as e:
```
wild-endeavor, line 71: can we add a conditional here?  or are we sure this will always be an S3 link?  If this class will only ever be used for s3, can we just hardcode the protocol in the constructor?

plugins/flytekit-data-fsspec/setup.py
```diff
@@ -4,7 +4,7 @@
 
 microlib_name = f"flytekitplugins-data-{PLUGIN_NAME}"
 
-plugin_requires = ["flytekit>=0.21.3,<1.0.0", "fsspec>=2021.7.0"]
+plugin_requires = ["flytekit>=0.21.3,<1.0.0", "fsspec>=2021.7.0", "botocore>=1.23.24"]
```
wild-endeavor, line 7: can we just use `>=1.0?`?  or at least lower it a bit?  the exception probably has been there for a while.  1.23.24 is the latest/very recent version.  some people might not want to upgrade.

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/pandas.py
pingsutw: we will support gs in the future, so the URI at line 70 could be `gs://` or `s3//`. However, the URI at line 74 will always be `s3://`. 
we will catch another error for the gs bucket. something like:
```
except NoGoogleCredentialsError as e
    ....
    return pd.read_parquet(...)
```

wild-endeavor: oh got it, cuz it's an S3 error.

wild-endeavor (approved):

PR #856, opened 2022-02-18
Run unit tests on windows

eapolinario:

# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


eapolinario:
We should confirm that this fixes the issue described in https://github.com/flyteorg/flyte/issues/2174.

flytekit/core/data_persistence.py
```diff
@@ -438,7 +447,13 @@ def put_data(self, local_path: Union[str, os.PathLike], remote_path: str, is_mul
 DataPersistencePlugins.register_plugin("/", DiskPersistence)
 
 # TODO make this use tmpdir
```
eapolinario, line 449: remove TODO.

flytekit/tools/fast_registration.py
```diff
@@ -50,7 +51,7 @@ def get_additional_distribution_loc(remote_location: str, identifier: str) -> st
     :param Text identifier:
     :return Text:
     """
-    return _os.path.join(remote_location, "{}.{}".format(identifier, "tar.gz"))
+    return posixpath.join(remote_location, "{}.{}".format(identifier, "tar.gz"))
```
eapolinario, line 54: why do we need this?

flytekit/core/data_persistence.py
```diff
@@ -438,7 +447,13 @@ def put_data(self, local_path: Union[str, os.PathLike], remote_path: str, is_mul
...
-tmp_dir = os.path.join("/tmp/flyte", datetime.datetime.now().strftime("%Y%m%d_%H%M%S"))
+tmp_dir_prefix = "/tmp/flyte"
+
+# Update tmp_dir_prefix if running on Windows
+if os.name == "nt":
+    tmp_dir_prefix = os.path.join(tempfile.gettempdir(), "flyte")
+
+tmp_dir = os.path.join(tmp_dir_prefix, datetime.datetime.now().strftime("%Y%m%d_%H%M%S"))
```
eapolinario, line 456: This is essentially creating a unique temp dir, right? We could do something like:
```
tmp_dir = tempfile.mkdtemp(prefix="/tmp/flyte")
```
So we wouldn't need to special-case windows (or any other OS fwiw).

flytekit/core/data_persistence.py
```diff
@@ -136,13 +138,20 @@ def register_plugin(cls, protocol: str, plugin: typing.Type[DataPersistence], fo
 
         cls._PLUGINS[protocol] = plugin
 
+    @staticmethod
+    def get_protocol(url: str):
+        parts = re.split(r"(\:\:|\://)", url, 1)
+        if len(parts) > 1:
+            return parts[0]
+        return "file"
```
eapolinario, line 146: sorry for being dense, but how does this work with windows paths?

flytekit/tools/fast_registration.py
pingsutw: when we use `os.path.join` on Windows, it will use `\\` to join these two paths. However, It doesn't work for the remote path. For example, the s3 path will become something like `s3://bucket/test\\key`

flytekit/core/data_persistence.py
pingsutw: it will not generate a protocol like `c:\\` if path is `c:\\app\\test`. The part will become `["c:\\app\\test"]` instead, and will use dick persistence.

eapolinario: ok, I see it, thanks for explaining.

Why can't we import `get_protocol` from `fsspec` directly?

tests/flytekit/unit/core/test_shim_task.py
```diff
@@ -38,14 +38,14 @@ def test_resolver_load_task():
 
     resolver = TaskTemplateResolver()
     ts = get_serializable(OrderedDict(), serialization_settings, square)
-    with tempfile.NamedTemporaryFile() as f:
-        write_proto_to_file(ts.template.to_flyte_idl(), f.name)
```
pingsutw, line 42: we will open the temporary file without closing it at lines 41 and 42 respectively, and it doesn't work on Windows.
```python
def write_proto_to_file(proto, path):
    Path(_os.path.dirname(path)).mkdir(parents=True, exist_ok=True)
    with open(path, "wb") as writer:
        writer.write(proto.SerializeToString())
```

Refer https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

flytekit/core/data_persistence.py
pingsutw: done, thanks

flytekit/core/data_persistence.py
pingsutw: we don't have the fsspec package in flytekit dependency now.

pingsutw:
- Fixed the issue described in https://github.com/flyteorg/flyte/issues/2174.
- Failed to install some packages in python3.10 on Windows, could we just ignore the tests in python3.10?

eapolinario:
> Failed to install some packages in python3.10 on Windows, could we just ignore the tests in python3.10?

Yes, we should ignore that combination (windows + python3.10). The reason is that docker-py is installing a version of pywin that doesn't support python 3.10 yet. More specifically, `docker-py` [is installing pywin32==227](https://github.com/docker/docker-py/issues/2902), whereas pywin only added support for python 3.10 in version 301 (note the https://github.com/mhammond/pywin32/blob/main/CHANGES.txt#L26-L38). Can you leave a comment?

eapolinario (commented):
Oh, I can't approve a PR I opened. 😂

Can you approve and merge, @pingsutw ?

pingsutw (approved):

PR #858, opened 2022-02-21
Delete unnecessary auth configuration

wild-endeavor:

# TL;DR
Make auth flows cleaner by removing options now that Admin [publishes them publicly](https://github.com/flyteorg/flyteidl/blob/08a958c085c939101893f40fb1582e92b31c9a40/protos/flyteidl/service/auth.proto#L64).

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Move contents of `basic_auth.py` into `raw.py`. Not sure why it was there originally.
* Triggering `GetPublicClientConfig` and `GetOAuth2Metadata` on construction of the raw client.
* Change `AuthorizationClient`'s `get_client` function to take in explicitly everything it needs to trigger pkce correctly.
* Don't require pkce flow to always return a refresh token.
* Fixed an issue where if tokens had been stored in keyring, we'd still refresh tokens instead of trying the saved ones first.
* Removing the now-unused config objects.
* Removing the `DiscoveryClient`.
* Adding the `protoc_gen_swagger` library to setup.py because for some reason `auth_pb2` wouldn't import without it.  Not sure what flyteidl is doing.  We should probably move the dependency there if we can't get rid of it.
* Remove client secret field from `AuthorizationClient`

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/clients/raw.py
```diff
@@ -29,31 +34,26 @@ def _refresh_credentials_standard(flyte_client):
...
-    """
-    deprecated_single_scope = _DEPRECATED_SCOPE.get()
-    if deprecated_single_scope:
-        return [deprecated_single_scope]
-    scopes = DEPRECATED_OAUTH_SCOPES.get() or SCOPES.get()
-    if "openid" in scopes:
-        cli_logger.warning("Basic flow authentication should never use openid.")
-
-    return scopes
+    authorization_header_key = flyte_client.public_client_config.authorization_metadata_key or None
```
EngHabu, line 52: What happens in case of `or None` ? will it default to `Authorization` header?

flytekit/clients/raw.py
```diff
@@ -63,16 +63,24 @@ def _refresh_credentials_basic(flyte_client):
...
-    authorization_header = _basic_auth.get_basic_authorization_header(_CLIENT_ID.get(), client_secret)
-    token, expires_in = _basic_auth.get_token(token_endpoint, authorization_header, _get_basic_flow_scopes())
+    if not flyte_client.oauth2_metadata or not flyte_client.public_client_config:
+        raise ValueError(
+            "Raw Flyte client attempting client credentials flow but no response from Admin detected. "
+            "Check your Admin server's .well-known endpoints to make sure they're working as expected."
+        )
+
+    token_endpoint = flyte_client.oauth2_metadata.token_endpoint
+    scopes = flyte_client.oauth2_metadata.scopes_supported
```
EngHabu, line 73: For public client case, we should request scopes listed [here](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/service/auth.proto#L55) instead... the supported scopes can be a huge list...

For client_credentials flow, users should be able to set that just like they set client id and client_secret... if not passed, we default to the public client's scopes [here](https://github.com/flyteorg/flyteidl/blob/0ff9b16ad4c37971dfaa1e51b5fe970bae40e356/clients/go/admin/token_source_provider.go#L140).

EngHabu (commented):
I _think_ it all looks fine.. have we tested PKCE and client_cred flows yet?

flytekit/clients/raw.py
wild-endeavor: I put it in the constructor downstream `def set_access_token(self, access_token: str, authorization_header_key: Optional[str] = "authorization")`

flytekit/configuration/creds.py
```diff
@@ -16,71 +11,19 @@
...
-"""
-This is the callback uri registered with the app which handles authorization for a Flyte deployment.
-Please note the hardcoded port number. Ideally we would not do this, but some IDPs do not allow wildcards for
-the URL, which means we have to use the same port every time. This is the only reason this is a configuration option,
-otherwise, we'd just hardcode the callback path as a constant.
-FYI, to see if a given port is already in use, run `sudo lsof -i :<port>` if on a Linux system.
-More details here: https://www.oauth.com/oauth2-servers/redirect-uris/.
-"""
-
-SCOPES = _config_common.FlyteStringListConfigurationEntry("credentials", "scopes", default=["openid"])
```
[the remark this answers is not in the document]
wild-endeavor: @kumare3 we're gonna keep this one.

EngHabu (approved):
🙇🏽

PR #859, opened 2022-02-22
Introduce flyte Decks into flytekit

pingsutw:

# TL;DR
Allow user to visualize dataframe, preview markdown file on flyteconsole
Note: now It only works for local execution

To run an example: 
1. `pip install git+https://github.com/flyteorg/flytekit@deck && python example_deck.py`
2. every task will output a HTML file like below contains input, output tab. In addition, Users can add custom tab if they want. For example, they can add a deck to see the markdown file or profile their panda dataframes.

image

output will be like

image

We should add some metadata in flyteidl to make flyteconsole be able to find those HTML files
```protobuf
message DeckMetadata {

  // Tab name.
  string name = 1;

  // HTML file.
 string path= 2;
}
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
- Add a `renderer` interface
- Add some default renderer such as `FrameRenderer`, `MarkdownRenderer`, `FrameProfilingRenderer`
- folks can easily add new renderer by extending `renderer`

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2175

## Follow-up issue
_NA_


example.py
```diff
@@ -0,0 +1,33 @@
...
+from flytekit.visualize import FrameProfilingRenderer, FrameRenderer, MarkdownRenderer, ScatterRenderer
+
+
+@task
+def t1() -> str:
+    md_text = "#Hello Flyte\n##Hello Flyte\n###Hello Flyte"
+    m = MarkdownRenderer(md_text)
+    s = ScatterRenderer(range(10), range(10))
+    flytekit.visualize.Deck("Tab1", s).append(m)
+    flytekit.visualize.Deck("Tab2", m)
```
wild-endeavor, line 15: is the object usable more than once?  should we have an example where it's saved to a variable and called again?

wild-endeavor (commented):
I know it's still wip but just at first glance...
* If we're going to have templating in the html, should we consider something like jinja2?
* Could you summarize the user-facing API somewhere. 
* How will this work for dynamic tasks?  Dynamic tasks are compiled at run-time - will these Deck objects interfere with compilation?
* How does this work with cached stuff?

@jsonporter could you take a look at the html page in the PR?  just in case you have any thoughts.

flytekit/visualize/deck.py
```diff
@@ -0,0 +1,30 @@
...
+        self.name = name
+        self.renderers = renderers if isinstance(renderers, list) else [renderers]
+        ctx = FlyteContextManager.current_context()
+        ctx.user_space_params.decks.append(self)
+
+    def append(self, r: Renderer) -> "Deck":
+        self.renderers.append(r)
+        return self
+
+    def upload(self, path):
```
kumare3, line 19: what is upload? it should not have this method in the user facing api. It should be in dispatch_execute right?

example.py
pingsutw: yes, updated it

pingsutw:
Few updates, 
- input and output tab to see task's input/output data
- Use jinja to dynamic create new html file, and each task will output one html file.
- html files are generated in run time, so it will not affect compilation.
- The task will not be executed when a cache hit occurs, so html files will not be generated.

setup.py
```diff
@@ -68,6 +68,9 @@
         "cloudpickle>=2.0.0",
         "cookiecutter>=1.7.3",
         "numpy<=1.22.1; python_version < '3.8.0'",
+        "markdown",
+        "plotly",
```
kumare3, line 72: Also move

kumare3:
This is looking much better now

flytekit/core/type_engine.py
```diff
@@ -126,6 +126,10 @@ def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type:
             f"Conversion to python value expected type {expected_python_type} from literal not implemented"
         )
 
+    @abstractmethod
+    def to_html(self, ctx: FlyteContext, python_val: T):
```
kumare3, line 130: is to_html, not going to return a div? it can be a simple string?
then maybe html string

kumare3: can we write docs for this

flytekit/core/context_manager.py
```diff
@@ -303,6 +320,10 @@ def checkpoint(self) -> Checkpoint:
             raise NotImplementedError("Checkpointing is not available, please check the version of the platform.")
         return self._checkpoint
 
+    @property
```
kumare3, line 323: lets add docs for this method

flytekit/core/base_task.py
```diff
@@ -525,7 +529,8 @@ def dispatch_execute(
                     raise TypeError(
                         f"Failed to convert return value for var {k} for function {self.name} with error {type(e)}: {e}"
                     ) from e
-
+            if ENABLE_DECK.get():
```
kumare3, line 532: why a flag? also can you use the new config system

flytekit/configuration/sdk.py
```diff
@@ -34,5 +34,7 @@
 USE_STRUCTURED_DATASET = _config_common.FlyteBoolConfigurationEntry("sdk", "use_structured_dataset", default=False)
 """
 Note: This gate will be switched to True at some point in the future. Definitely by 1.0, if not v0.31.0.
-
 """
+
+# Feature Gate
+ENABLE_DECK = _config_common.FlyteBoolConfigurationEntry("sdk", "enable_deck", default=False)
```
kumare3, line 40: this is deprecated

flytekit/deck/deck.py
```diff
@@ -0,0 +1,79 @@
...
+    # Each task has a least three decks (input, output, default). Input/output decks are
+    # used to render tasks' input/output data, and the default deck is used to render line plots,
+    # scatter plots or markdown text. In addition, users can create new decks to render
+    # their data with custom renderers.
+    def __init__(self, name: str, renderers: Union[List[Renderer], Renderer]):
+        self.name = name
+        self.renderers = renderers if isinstance(renderers, list) else [renderers]
+        FlyteContextManager.current_context().user_space_params.decks.append(self)
+
+    def append(self, r: Renderer) -> "Deck":
```
kumare3, line 32: should we just take the html here? what do you think?

flytekit/types/schema/types_pandas.py
```diff
@@ -147,6 +147,9 @@ def to_python_value(
         r = PandasSchemaReader(local_dir=local_dir, cols=None, fmt=SchemaFormat.PARQUET)
         return r.all()
 
+    def to_html(self, ctx: FlyteContext, python_val: pandas.DataFrame):
+        return python_val.to_html()
```
kumare3, line 151: we just want to print the column names by default. Dumping the entire pandas dataframe is not a great idea

flytekit/types/structured/structured_dataset.py
```diff
@@ -630,6 +631,10 @@ def t2(in_a: Annotated[StructuredDataset, kwtypes(col_b=float)]): ...
         # we should do the opening/downloading and whatever else it might entail right now. No iteration option here.
         return self.open_as(ctx, lv.scalar.structured_dataset, df_type=expected_python_type, updated_metadata=metad)
 
+    def to_html(self, ctx: FlyteContext, python_val: typing.Any):
+        if issubclass(type(python_val), pandas.DataFrame):
+            return python_val.to_html()
```
kumare3, line 636: same thing

plugins/flytekit-deck/flytekitplugins/deck/renderer.py
```diff
@@ -0,0 +1,37 @@
+from typing import List, Union
```
kumare3, line 1: lets rename the plugin
```
pip install flytekit-deck-standard
```

maybe we can make all deck plugins be
```
flytekit-deck-*
```

plugins/flytekit-deck/flytekitplugins/deck/renderer.py
```diff
@@ -0,0 +1,37 @@
...
+
+import markdown
+import pandas
+import plotly.express as px
+from pandas_profiling import ProfileReport
+
+from flytekit.deck.renderer import Renderer
+
+
+class FrameProfilingRenderer(Renderer):
```
kumare3, line 11: docs for all of these

kumare3:
Overall I love the PR, this is simple. But, I think we should support a simple `html` append also?

kumare3:
cc @pradithya / @eapolinario

plugins/flytekit-deck/flytekitplugins/deck/renderer.py
```diff
@@ -0,0 +1,37 @@
...
+
+class MarkdownRenderer(Renderer):
+    def __init__(self, text: str):
+        self._text = text
+
+    def render(self) -> str:
+        return markdown.markdown(self._text)
+
+
+class ScatterRenderer(Renderer):
```
kumare3, line 29: I guess we need a simple plotly to html?

flytekit/deck/renderer.py
```diff
@@ -0,0 +1,22 @@
+import typing
+from abc import ABC
+
+import pandas as pd
+
+
+class Renderer(ABC):
+    def render(self) -> str:
```
kumare3, line 8: should we call it to_html, that way any pandas-dataframe etc that have to_html method can become Renderable.

```suggestion
    def to_html(self) -> str:
```

flytekit/deck/renderer.py
```diff
@@ -0,0 +1,22 @@
+import typing
+from abc import ABC
+
+import pandas as pd
+
+
+class Renderer(ABC):
```
kumare3, line 7: Let's rename
```suggestion
class Renderable(Protocol):
```

Also I think Renderable should be a protocol

flytekit/deck/renderer.py
```diff
@@ -0,0 +1,22 @@
...
+    def render(self) -> str:
+        """Convert a object(markdown, pandas.dataframe) to HTML and return HTML as a unicode string.
+        Returns: An HTML document as a string.
+        """
+        raise NotImplementedError
+
+
+class FrameRenderer(Renderer):
```
kumare3, line 15: After we make the RenderableProtocol, we do not need to extend it.
```suggestion
class FrameRenderer(Renderer):
```

We can create a new Renderable called
```python
class TopFrameRenderable():
    ....
```

flytekit/deck/deck.py
```diff
@@ -0,0 +1,79 @@
...
+def _output_deck(task_name: str, new_user_params: ExecutionParameters, task_input: dict, task_output: dict):
+    deck_map = {}
+    INPUT = "input"
+    OUTPUT = "output"
+    ctx = FlyteContext.current_context()
+    output_dir = ctx.file_access.get_random_local_directory()
+
+    deck_map[INPUT] = []
+    for k, v in task_input.items():
+        _deck_to_html_file(deck_map, INPUT, output_dir, k, TypeEngine.to_html(ctx, v, type(v)))
```
ggydush-fn, line 46: So this also works for inputs, meaning if an input type has `to_html` implemented, it will be rendered in the deck, correct? We often have large specs passed into input fields, and rendering them in the task view would be a huge QoL improvement!

pradithya:
This is awesome!! I have few feedback/question:

1. The task input and output will be rendered by default using a compatible renderer, currently it's in best effort basis, correct?
2. I think it's important to be able to give title to a rendered object and display it in the html. It will be helpful for user especially when they have multiple object being rendered from a single task. For input and/or output it should happen automatically. 

e.g. 
```
flytekit.Deck("custom", [ProfileReport(df, title="Iris Dataset")])
```
will replace the title of the profiler from "Pandas Profiling Report" --> "Iris Dataset"

Screenshot 2022-03-09 at 6 30 52 PM



3. I see that deck is implemented as plugin and behind feature toggle what's the reason behind it?

Anyway, really love it ❤️

flytekit/types/schema/types_pandas.py
pradithya: I think information in DataFrame.info() or DataFrame.describe() is sufficient for the built-in renderer

flytekit/deck/deck.py
```diff
@@ -0,0 +1,79 @@
...
+    # generate a html file. For example, FrameRenderer can render a DataFrame as an HTML table,
+    # MarkdownRenderer can convert Markdown string to HTML
+    #
+    # Flyte context saves a list of deck objects, and we use renderers in those decks to render
+    # the data and create an HTML file when those tasks are executed
+    #
+    # Each task has a least three decks (input, output, default). Input/output decks are
+    # used to render tasks' input/output data, and the default deck is used to render line plots,
+    # scatter plots or markdown text. In addition, users can create new decks to render
+    # their data with custom renderers.
```
cosmicBboy, line 26: nit: can we use triple-quote string literal here?

flytekit/core/context_manager.py
```diff
@@ -171,6 +174,7 @@ def __init__(self, current: typing.Optional[ExecutionParameters] = None):
             self.execution_id = current.execution_id if current else None
             self.logging = current.logging if current else None
             self.checkpoint = current._checkpoint if current else None
+            self.decks = current._decks if current else None
```
cosmicBboy, line 177:
```suggestion
            self.decks = current._decks if current else []
```

For type consistency

flytekit/core/type_engine.py
cosmicBboy, line 130:
```suggestion
    def to_html(self, ctx: FlyteContext, python_val: T) -> str:
```

flytekit/core/type_engine.py
```diff
@@ -673,6 +677,11 @@ def to_python_value(cls, ctx: FlyteContext, lv: Literal, expected_python_type: T
         transformer = cls.get_transformer(expected_python_type)
         return transformer.to_python_value(ctx, lv, expected_python_type)
 
+    @classmethod
+    def to_html(cls, ctx: FlyteContext, python_val: typing.Any, python_type: Type):
```
cosmicBboy, line 681:
```suggestion
    def to_html(cls, ctx: FlyteContext, python_val: typing.Any, python_type: Type) -> str:
```

flytekit/deck/renderer.py
cosmicBboy: but don't we want the pandas dataframe rendered to _not_ output actual dataframe values? I'd imagine we want some control over how a particular datatype is rendered.

flytekit/types/schema/types_pandas.py
eapolinario: What if we used [typing.Annotated](https://docs.python.org/3/library/typing.html#typing.Annotated) to define what is going to be rendered? By default it only does `df.describe()` but it could potentially run other kinds of aggregation (e.g. list of unique values) or maybe parameters to [`df.describe()`](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.describe.html).

example_deck.py
```diff
@@ -0,0 +1,47 @@
...
+except ImportError:
+    from typing_extensions import Annotated
+
+import pandas as pd
+import plotly.express as px
+from flytekitplugins.deck.renderer import BoxRenderer, MarkdownRenderer
+
+import flytekit
+from flytekit import dynamic, task, workflow
+from flytekit.deck import default_deck
```
kumare3, line 12: I do not think we should make this global This is a problem in multi-threaded scenarios.

kumare3: I guess we can use contextvars?
cc @eapolinario / @wild-endeavor 

I did not even know about contextvars. Probably should replace FlyteContext?

wild-endeavor: yeah can we just move this into ExecutionParameters for now? `flytekit.current_context().deck.

Why do we need multiple?  Can we just have the default one?  Or have people asked for multiple?

flytekit/configuration/internal.py
```diff
@@ -31,6 +31,11 @@ def get_specified_images(cfg: ConfigFile) -> typing.Dict[str, str]:
         return images
 
 
+class Deck(object):
+    SECTION = "deck"
+    ENABLE_DECK = ConfigEntry(LegacyConfigEntry(SECTION, "enable_deck", bool))
```
wild-endeavor, line 36: Why do we need to gate this?  Can't we just enable it for everyone?  Or

flytekit/types/structured/structured_dataset.py
```diff
@@ -625,6 +623,24 @@ def t2(in_a: Annotated[StructuredDataset, kwtypes(col_b=float)]): ...
         # we should do the opening/downloading and whatever else it might entail right now. No iteration option here.
         return self.open_as(ctx, lv.scalar.structured_dataset, df_type=expected_python_type, updated_metadata=metad)
 
+    def to_html(self, ctx: FlyteContext, python_val: typing.Any, expected_python_type: Type[T]) -> str:
+        if isinstance(python_val, StructuredDataset):
+            if python_val.dataframe is not None:
+                df = python_val.dataframe
+            else:
+                df = typing.cast(StructuredDataset, python_val).open(pd.DataFrame).all()
```
wild-endeavor, line 631: This will have the downside of always downloading the data right?

tests/flytekit/unit/types/structured_dataset/test_structured_dataset_workflow.py
```diff
@@ -183,7 +189,7 @@ def t9(dataframe: np.ndarray) -> Annotated[StructuredDataset, my_cols]:
 
 
 @task
-def t10(dataset: Annotated[StructuredDataset, my_cols]) -> np.ndarray:
+def t10(dataset: StructuredDataset) -> np.ndarray:
```
wild-endeavor, line 192: Did this need to change for tests to pass for some reason?

flytekit/configuration/internal.py
kumare3: I think it should be possible to disable maybe as a perf thing

kumare3: But let's do it later

flytekit/types/structured/structured_dataset.py
kumare3: Great catch. I think the default should only put column names or something

pingsutw: That makes sense, updated it to render column names and type only, thanks.

example_deck.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+@task()
+def t1() -> str:
+    md_text = "#Hello Flyte\n##Hello Flyte\n###Hello Flyte"
+    m = MarkdownRenderer()
+    s = BoxRenderer("sepal_length")
+    deck = flytekit.Deck("demo", s.to_html(iris_df))
+    deck.append(m.to_html(md_text))
+    default_deck = FlyteContextManager.current_context().user_space_params.default_deck
```
wild-endeavor, line 20:
```suggestion
    default_deck = flytekit.current_context().default_deck
```

wild-endeavor: should be the same object as what you have.

pingsutw: okay, thanks. will remove this file `example.py` before we merge this file

flytekit/core/base_task.py
```diff
@@ -521,6 +524,23 @@ def dispatch_execute(
                         f"Failed to convert return value for var {k} for function {self.name} with error {type(e)}: {e}"
                     ) from e
 
+            INPUT = "input"
+            OUTPUT = "output"
+            from flytekit import Deck
```
wild-endeavor, line 529: same here import from the original place where this is defined.  I don't think it needs to be here instead of the top of the file...

flytekit/types/structured/structured_dataset.py
```diff
@@ -9,19 +10,20 @@
...
-from dataclasses_json import config, dataclass_json
-from marshmallow import fields
-
-try:
-    from typing import Annotated, TypeAlias, get_args, get_origin
-except ImportError:
-    from typing_extensions import Annotated, get_origin, get_args, TypeAlias
-
 import _datetime
+import numpy as np
```
wild-endeavor, line 14: can we get rid of these duplicates?

kumare3:
cc @eapolinario we discussed 2 things right?
1. Move the feature to `flytekit.experimental` or add beta note to the api
2. on every `@task(disable_deck=True)` so that you can optionally disable the deck?

pingsutw:
Few updates, 
- Add `disable_deck` arg in `@task`
- beta note to the deck interface
- remove` example_deck.py`

wild-endeavor (approved):
nit

wild-endeavor (approved):

PR #867, opened 2022-02-25
Make fetched entities callable within workflows

wild-endeavor:
# TL;DR
Make all Flyte entities (FlyteTask, FlyteWorkflow, FlyteLaunchPlan) callable.  This PR replaces what we were doing for Task and LaunchPlan earlier where we redirected it to a reference entity.  The reason is because reference entities require a Python native interface which we don't have (and we don't want to rely on guessing).

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?
 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
* Remove the reference entity stuff from `FlyteTask` and `FlyteLaunchPlan`
* Create a `RemoteEntity` abc that satisfies the new `HasFlyteInterface` interface.
  * This has node creation logic which means you can `__call__` (aka `()`) it and get a node out of it.
* In dynamic compile, make sure to add any TaskTemplates to the dj spec.
* Change translator to handle nodes that point to fetched entities.
* Make the python type in `binding_data_from_python_std` optional so that we can use the function for flyte fetched entities.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2157

## Follow-up
https://github.com/flyteorg/flyte/issues/2215

### DIFF
--- a/.github/workflows/pythonbuild.yml
+++ b/.github/workflows/pythonbuild.yml
@@ -49,7 +49,6 @@ jobs:
       shell: 'script -q -e -c "bash {0}"'
       run: |
         python -m pip install awscli
-        FLYTE_SDK_USE_STRUCTURED_DATASET=TRUE coverage run --append -m pytest tests/flytekit/integration || true
         FLYTE_SDK_USE_STRUCTURED_DATASET=TRUE coverage run --append -m pytest tests/flytekit/integration
     - name: Codecov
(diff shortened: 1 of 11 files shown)


wild-endeavor:
cc @ckiosidis @sonjaer

PR #886, opened 2022-03-10
flytekit-dbt plugin

ariefrahmansyah:
* Supports dbt run and dbt test tasks
* The plugin includes integration test that need local PostgreSQL database

Signed-off-by: ariefrahmansyah <ariefrahmansyah@hotmail.com>

# TL;DR

Add flytekit-dbt plugin. Fixes https://github.com/flyteorg/flyte/issues/2202

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [X] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description

For integration testing, flytekit-dbt needs a running PostgreSQL database. To achieve that, it assumes that the machine running the test already have PostgreSQL installed. Then, the pytest.fixtures executes `setup_db.sh` to create dbttest database. This database is used for dbt tasks to execute the transformation of dbt's [jaffle shop example](https://github.com/dbt-labs/jaffle_shop).

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2202

## Follow-up issue
_NA_


plugins/flytekit-dbt/tests/test_task.py
```diff
@@ -0,0 +1,149 @@
...
+@pytest.fixture(scope="module", autouse=True)
+def prepare_db():
+    dir_path = os.path.dirname(os.path.realpath(__file__))
+    setup_db_script = os.path.join(dir_path, "..", "..", "flytekit-dbt", "tests", "setup_db.sh")
+
+    print("preparing db")
+    subprocess.run(setup_db_script)
+    print("db is ready")
+
+    yield
```
ariefrahmansyah, line 26: Need your input on this part: setting up the integration test environment. In Gojek, our flytekit-dbt transformation test is targeting our BQ dataset which cannot be reproduced in open-source space. So, a more suitable choice is to use PostgreSQL. This approach assumes that PostgreSQL is already running - whether via binary or container.

Another approach that I've considered:
1. Mark these tests as integration tests, entirely separated from unit tests. It doesn't break existing tests but introduces more complexity on how to do testing. But if in the future there's another plugin that needs an integration test, it can inherit this setup

plugins/flytekit-dbt/flytekitplugins/dbt/task.py
```diff
@@ -0,0 +1,245 @@
...
+        DBTHandledError
+            If the ``dbt run`` command returns ``exit code 1``.
+        DBTUnhandledError
+            If the ``dbt run`` command returns ``exit code 2``.
+        """
+
+        task_input: DBTRunInput = kwargs["input"]
+
+        args = task_input.to_args()
+        cmd = ["dbt", "--log-format", "json", "run"] + args
```
kumare3, line 92: Is this like a blocking call? Can we get a handle to the underlying call so that if we interrupt can resume?

plugins/flytekit-dbt/tests/test_task.py
kumare3: how about sqlite3?

plugins/flytekit-dbt/flytekitplugins/dbt/util.py
```diff
@@ -0,0 +1,41 @@
+import json
+import logging
+import subprocess
+from typing import List
+
+
+def run_cli(cmd: List[str]) -> (int, List[str]):
```
kumare3, line 7: you can use - 
https://github.com/flyteorg/flytekit/blob/2ca5607d6b313fa9d6aeb88a807febcefcea7cdb/flytekit/tools/subprocess.py

kumare3:
thank you so much!

.github/workflows/pythonbuild.yml
```diff
@@ -79,35 +80,45 @@ jobs:
...
-        pip install -e .
-        if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
-        pip install --no-deps -U https://github.com/flyteorg/flytekit/archive/${{ github.sha }}.zip#egg=flytekit
-        pip freeze
-    - name: Test with coverage
-      run: |
-        cd plugins/${{ matrix.plugin-names }}
-        FLYTE_SDK_USE_STRUCTURED_DATASET=TRUE coverage run -m pytest tests
+      - uses: actions/checkout@v2
+      - name: Set up postgres (linux)
```
wild-endeavor, line 93: @eapolinario can you take a look at this?  I'm okay with this for now assuming we're committed to coming up with a proper integration testing framework in the future (this is currently used for just unit tests I think? but we can move them around in the future).

.github/workflows/pythonbuild.yml
```diff
@@ -79,35 +80,45 @@ jobs:
...
+        if: runner.os == 'Linux' && matrix.plugin-names == 'flytekit-dbt'
+        shell: bash
+        run: |
+          sudo systemctl start postgresql.service
+          pg_isready
+      - name: Set up Python ${{ matrix.python-version }}
```
wild-endeavor, line 99: Is there a material change in any of the steps below?  The diff isn't rendering well given the whitespace changes.

wild-endeavor (commented):
Thanks!  I think it looks good, leaving a couple comments/questions.

Also, could we get a screenshot of how this looks in the UI?  Always nice to see a finished run.

And @kumare3 this could've also been implemented as a Shell task?  (no need to change, just wondering).

plugins/flytekit-dbt/tests/test_task.py
ariefrahmansyah: * How about sqlite3?
Currently, sqlite is not officially supported by DBT. There's a community SQL adapter plugin (https://github.com/codeforkjeff/dbt-sqlite), but using it for integration testing is a different scenario than using the official adapter, we are ended up testing the adapter too.

* All the csv files are for integration testing right?
These CSV files are only for integration testing.

wild-endeavor: Can we mark the integration tests as skipped for now?  We're going to be soon revisiting integration testing in flytekit in general.  Maybe in late April.  What do you think @eapolinario?

wild-endeavor: mark as skipped and then comment out the setup step for now.

.github/workflows/pythonbuild.yml
eapolinario: Can we do the equivalent of this in code? We can skip tests based on platform as exemplified in https://docs.pytest.org/en/latest/how-to/skipping.html#skipping-test-functions

plugins/flytekit-dbt/tests/test_task.py
eapolinario: Setting up postgres on the github CI boxes is pretty quick (the [initial run](https://github.com/flyteorg/flytekit/runs/5490579965?check_suite_focus=true#step:3:1) for this took 3s), so we should be ok leaving that enabled assuming that (1) we can make this work with postgres, and (2) the tests themselves are not flaky.

kumare3:
@wild-endeavor - this could have been implemented by shell,  but have a built in plugin helps a lot

kumare3:
cc @ariefrahmansyah thank you for the PR, do you need some help from anyone in the community? I think this feature would really help many folks

kumare3:
@ariefrahmansyah hows it going with this?

esadler-hbo:
@ariefrahmansyah this might be useful for a project that I am working in. I'm about to do a write a spike and if dbt is accepted then I could help with this plugin

eapolinario:
Subsumed by https://github.com/flyteorg/flytekit/pull/1150.

PR #887, opened 2022-03-10
Add back propagating of fast register state

wild-endeavor:

# TL;DR
This was inadvertently removed when we cleaned up configuration.  This code is necessary to ensure that if a dynamic task is fast registered, that calls a task that is also not in the base image, that those child tasks will also pull the same code layer before running. 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Copy back.


kumare3 (dismissed):
lgtm

kumare3 (approved):

PR #891, opened 2022-03-18
Register the task first when executing the workflow

pingsutw:

# TL;DR
Failed to execute a workflow when we run `remote.execute(chain_tasks.workflows.example.chain_tasks_wf, inputs={}, version=version, wait=False)`  because we try to register a Luanch Plan before registering tasks

- Register the task first when registering the workflow
- Add `image_config` in Options, so we can set up the image when creating a remote client or executing the workflow
- Remove `from_project` and `from_domain` because we already have `project` and `domain` in args

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/configuration/__init__.py
+++ b/flytekit/configuration/__init__.py
@@ -459,6 +459,7 @@ class Config(object):
     secrets: SecretsConfig = SecretsConfig()
     stats: StatsConfig = StatsConfig()
     data_config: DataConfig = DataConfig()
+    image_config: ImageConfig = ImageConfig()
     local_sandbox_path: str = tempfile.mkdtemp(prefix="flyte")
 
     def with_params(
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -439,16 +439,18 @@ class FlyteRemote(object):
         m = OrderedDict()
         workflow_spec = get_serializable_workflow(m, serialization_settings, entity)
         ident = self._resolve_identifier(ResourceType.WORKFLOW, entity.name, version, serialization_settings)
-        self.client.create_workflow(ident, workflow_spec=workflow_spec)
-        if default_launch_plan:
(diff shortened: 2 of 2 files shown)


flytekit/configuration/__init__.py
```diff
@@ -459,6 +459,7 @@ class Config(object):
     secrets: SecretsConfig = SecretsConfig()
     stats: StatsConfig = StatsConfig()
     data_config: DataConfig = DataConfig()
+    image_config: ImageConfig = ImageConfig()
```
wild-endeavor, line 462: I think this was intentionally left out of here, and put into just the `SerializationSettings` object.  We'll have to add it to FlyteRemote separately.

flytekit/remote/remote.py
```diff
@@ -473,7 +475,7 @@ def register_launch_plan(
 
         if serialization_settings is None:
             serialization_settings = SerializationSettings(
-                image_config=None, project=self.default_project, domain=self.default_domain
+                image_config=self._config.image_config, project=self.default_project, domain=self.default_domain
```
wild-endeavor, line 478: We can leave it None for now... we're only registering a Launch Plan, which we don't need image for, at least for now.

flytekit/remote/remote.py
```diff
@@ -790,7 +779,7 @@ def _(
                     f" image cannot be automatically deducted. Please register and then execute."
                 )
             ss = SerializationSettings(
-                image_config=None,
+                image_config=self._config.image_config,
```
wild-endeavor, line 782: i'm confused, so this never worked in the past?  oh boy

PR #899, opened 2022-03-24
#1437 Add .gitignore and .dockerignore behavior to packaging

bimtauer:
# TL;DR
Implements flyteorg/flyte#1437 by refactoring the tarfile filter logic to use a group of filter classes. Previous filtering logic was preserved in `StandardIgnore`. 

Decided to not go with a separate .flyteignore. Covering .gitignore and .dockerignore if present should work well with a users intentions and avoids having to manage yet another file. If its in .gitignore, you probably dont want it in source control (from where you probably build at least your production images) and if its in .dockerignore you probably dont want it in the task-running container anyways. So following these two should be good :)
And well if you like it messy, then as before, the in-built filtering saves you from packaging cache and pyc stuff.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The `GitIgnore` checks for the presence of git cli and then uses that to check whether a path is ignored. Given the complexity of gitignores with the possibility of nesting and various wildcards, this was the safest way to reliable reproduce actual gitignore behavior. 
The `DockerIgnore` uses the `PatternMatcher` from docker-py to match file paths against patterns read from a .dockerignore if present. Docker only takes into accout .dockerignore files at the root of the build context, not in subdirectories, so we only check for that. The behavior should thus also be identical to what `docker build` would include. 

The archiving code was refactored into `create_archive` for less repetition and separate unit testing.
The previous filter function and test was removed since the new feature covers that functionality. 

## Tracking Issue
flyteorg/flyte#1437

## Follow-up issue
_NA_


wild-endeavor:
Thank you for the PR!  I will review this weekend.

kumare3:
> Are all the subprocess calls okay? Some of these repos might get pretty large - subprocess calling git on every single file, is that UX okay?

@bimtauer I like this code. But @wild-endeavor is right, I see you have implemented pattern matcher, why not use the same for git?

kumare3:
> > Are all the subprocess calls okay? Some of these repos might get pretty large - subprocess calling git on every single file, is that UX okay?
> 
> @bimtauer I like this code. But @wild-endeavor is right, I see you have implemented pattern matcher, why not use the same for git?

Also the patterns are listed [here](https://git-scm.com/docs/gitignore#_pattern_format). Only weird one is `!`

flytekit/clis/sdk_in_container/package.py
```diff
@@ -125,8 +126,7 @@ def package(ctx, image_config, source, output, force, fast, in_container_source_
                     archive_fname = os.path.join(output_tmpdir, f"{digest}.tar.gz")
```
kumare3, line 126: not your code, but i think compute_digest is wrong!. It is computing on the entire source and not the filtered objects only

bimtauer:
> > Are all the subprocess calls okay? Some of these repos might get pretty large - subprocess calling git on every single file, is that UX okay?
> 
> @bimtauer I like this code. But @wild-endeavor is right, I see you have implemented pattern matcher, why not use the same for git?

Cannot use that for git as gitignore pattern matching is slightly different than docker ignore (some examples [here](https://zzz.buzz/2018/05/23/differences-of-rules-between-gitignore-and-dockerignore/)) and also git allows for multiple nested gitignores. 
But I followed @wild-endeavor's advice and switched from many subprocess calls to a single one with `git ls-files -io`. That accumulates all ignored files once. Only small issue: this does not list entirely ignored directories, only files within them. As a result those directories turn up in the tarball but empty. Should be an okay compromise? 

I'll look at the digest asap, good catch!

bimtauer:
Developed update to include only not ignored files in checksum. But unfortunately the `exclude_files` of `checksumdir`[only checks for file basenames, not for relative path](https://github.com/cakepietoast/checksumdir/blob/master/checksumdir/__init__.py#L52-L67)s 🤦🏼 I'm hesitant to strip relative path from the paths to check, since `foo/bar.py`might be ignored but `bar.py` might not be. I'm tempted to write my own quick checksum func and eliminate this dependency. Wdyt?

Also having a look at the failing windows tests. I dont have access to a windows machine 🤢 but I'll try my best 🐧

bimtauer:
@wild-endeavor @kumare3 Bigger refactoring:
- eliminated subprocesses (only one left) but keep ignoring empty dirs 💪🏼 
- performance improvement through dict key lookup in gitignore - its still marginally slower than old filter but overall usually faster because fewer files to package
- Integrated with latest master changes
- Began fixing requirements (pinning websocket-client) but grpcio io yank is still making waves through dep tree

flytekit/tools/ignore.py
```diff
@@ -0,0 +1,124 @@
...
+
+class GitIgnore(Ignore):
+    """Uses git cli (if available) to list all ignored files and compare with those."""
+
+    def __init__(self, root: Path):
+        super().__init__(root)
+        self.has_git = which("git") is not None
+        self.ignored = self._list_ignored()
+
+    def _list_ignored(self) -> Dict:
```
kumare3, line 46: this is great

kumare3 (approved):

kumare3:
@bimtauer some tests are failing and merge conflicts. But this otherwise looks awesome 🙏🏽

bimtauer:
@kumare3 I need help debugging the windows tests. I don't have access to a windows dev environment and cannot figure out why the tests fail there. Otherwise this should be ready.

eapolinario:
@bimtauer , https://github.com/flyteorg/flytekit/pull/967 was supposed to be a stacked PR on top of this PR, but I botched it. In any case, feel free to cherry-pick the specific commits that fix the unit failures on windows. I also took the opportunity to resolve the conflict and merge master.

eapolinario:
This change ended up being merged in https://github.com/flyteorg/flytekit/pull/967. 

@bimtauer , thanks for your contribution!

bimtauer:
@eapolinario Thank you for pushing it over the finish line!

PR #902, opened 2022-03-29
FullyQualified package names

kumare3:

# TL;DR
This PR makes it possible to always get consistent package names.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
by default the package name for a workflow or a task is relative from where the `package` or `serialize` commands are run. After this change the package name will always be relative to the root. 

The package name is deduced using hierarchical parsing of the file-system tree, to identify the root. The root is identified if the the directory does not contain a __init__.py file. For cases like flytekitplugins, it works as the module normalizer, assumes that the newly constructed path is always larger than the assumed path. If not, then the original path is returned.

Furthermore this feature can be turned of by setting the following environment variable. 

```
export FLYTE_PYTHON_PACKAGE_ROOT="."
OR
export FLYTE_PYTHON_PACKAGE_ROOT="/acutal-path/to/root/"
``` 

The default value is `FLYTE_PYTHON_PACKAGE_ROOT="auto"` which turns it on.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2241

## Follow-up issue
_NA_


wild-endeavor (approved):

convexquad:
Just an FYI that if someone later looks at this PR - at Woven Planet we are indeed setting `export FLYTE_PYTHON_PACKAGE_ROOT="."` before serializing our workflows for compatibility with our Bazel build system (which installs Python dependencies like `flytekit` at a deeply nested path in order to provide isolation from system dependencies).

PR #928, opened 2022-04-11
Extend ShellTask to support wrapping pure shell scripts

myz540:
# TL;DR

**NOTE**: I had to rebase commits to sign them off so they are all clustered together.

This PR aims to better support existing shell scripts which need to be ported over into a `flyte` task as a `ShellTask`.

The current `ShellTask` implementation encourages users to in-line their shell scripts, or pass them in via `script_file`, but the interpolation of f-string step can cause syntax collisions with the curly brace (`{}`) and its usage in shell scripts. Additionally, environment variables cannot be set at runtime (only when the task is instantiated), and the script cannot accept any arguments, positional or otherwise (`$@`)

This PR introduces a minimal in-line template which can wrap a pure shell script, allowing users to port existing shell scripts that need:
1. runtime environment variable setting
2. arguments to the script

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Per reviewer comments, I moved the implementation into a new class, though it is minimally different from the existing `ShellTask`. I created a `_RawShellTask` which is currently not intended to be instantiated on its own, but rather an instance with a fixed template is returned by a function call. This moves the new functionality into a new class and does not modify the existing `ShellTask` in anyway.

Usage:
```python
from flytekit.extras.tasks.shell import get_raw_shell_task

pst = get_portable_shell_task()
pst(
        script_file="/path/to/your/shell_script.sh",
        env={"A": "AAAA", "B": "BBBB"},
        script_args="first_arg second_arg"
)
```

Executing the above task is equivalent to exporting the environment variables specified, and calling the `"/path/to/your/shell_script.sh"` with the args "first_arg second_arg".

## Tracking Issue
_NA_

## Follow-up issue
_NA_


myz540:
@kumare3 @wild-endeavor @eapolinario 

Quick nudge, could I get a workflow approval on the pending commits? I believe I've resolved all the issues and would like to get this reviewed. It is a feature we are currently using on our fork

kumare3:
@myz540 my apologies, I missed your ping. I am reviewing this. Thank you for the update

flytekit/extras/tasks/shell.py
```diff
@@ -184,6 +184,20 @@ def script(self) -> typing.Optional[str]:
     def script_file(self) -> typing.Optional[os.PathLike]:
         return self._script_file
 
+    def make_export_string_from_env_dict(self, d) -> str:
```
kumare3, line 187: can we have type for d
```suggestion
    def make_export_string_from_env_dict(self, d: str) -> str:
```

flytekit/extras/tasks/shell.py
```diff
@@ -204,6 +218,11 @@ def execute(self, **kwargs) -> typing.Any:
         if os.name == "nt":
             self._script = self._script.lstrip().rstrip().replace("\n", "&&")
 
+        if "env" in kwargs and isinstance(kwargs["env"], dict):
```
kumare3, line 221: how about extending ShellTask to make a new tasktype called RawShellTask?
this could automaticaly export all vars as env?
And the helper method that you have can be added to that.
This way, there are no side-effects, like un-intended use of env etc and keeps shell task exactly what it is today. 

Also cc @palchicz for comments on this?

flytekit/extras/tasks/shell.py
```diff
@@ -237,3 +256,32 @@ def execute(self, **kwargs) -> typing.Any:
...
+# This utility function allows for the specification of env variables, arguments, and the actual script within the
+# workflow definition rather than at `ShellTask` instantiation
+def get_portable_shell_task() -> ShellTask:
+    return ShellTask(
+    name="portable_shell_task_instance",
+    debug=True,
+    inputs=flytekit.kwtypes(env=typing.Dict[str, str], script_args=str, script_file=str),
+    output_locs=[
+        OutputLocation(
+            var="k",
```
kumare3, line 271: it is also assuming that k is always the output?

flytekit/extras/tasks/shell.py
myz540: Done, `d` is `typing.Dict`, can enforce the keys and values to be `str, str`, but technically, as long as the key-value can be cast to str, it should be ok.

flytekit/extras/tasks/shell.py
myz540: Yes, I had originally considered this but I had two reservations:

1. The desired behavior I wanted to add is so minimal that stamping out a new class entirely, only to copy 95% of the `execute()` methods felt overly verbose.
2. The main issue of setting arbitrary environment variables at execution time is still not easily addressed. 

I am open to your suggestion but could you provide a more concrete plan for implementation? I supposed I could re-work the class in such a way that the intended "raw" script be passed in via `script_file` but I still need the `script_args` and `env` variables to be passed in at execution time, and there is no easy way to add additional params to `execute()` since it goes through the `flyte_entity_call_handler` and all `inputs` are eventually exposed via the `kwargs` in `execute()`. So I could either accept `script_args` and `env` as "special" inputs, or accept no other inputs but these two and limit the use-case. The class would still need to be instantiated with a the generic template wrapper (likely provided by the class with ability to override). I'm happy to discuss this further, my solution here was intended to be as minimal since I am not familiar with the patterns and customs of this code base

flytekit/extras/tasks/shell.py
myz540: I wasn't sure how best to handle this, I think most standalone shell scripts either write files to some known location or accept an output file/dir as input parameter. I could have no outputs but in the spirit of returning some output to be passed to a downstream task, I just return the `ctx.working_directory`. I suppose the variable name could be anything here but I used `k` as in the examples

kumare3:
cc @palchicz any thoughts? cc @wild-endeavor / @eapolinario can you help review?

flytekit/extras/tasks/shell.py
palchicz: I'm strongly in favor of creating a new class instead of modifying the current one (i.e. favor extension over modification), especially since this would change the API of the class and it is already in use. 

It would also have a positive impact on the code base if the internals of this class were reworked such that new but slightly different classes are easier to stamp out. No real suggestions on implementation to get there though :-/

myz540:
Thanks for the review, I will refactor the implementation into a new class which extends `ShellTask`. I will close this PR when the new PR is ready.

flytekit/extras/tasks/shell.py
```diff
@@ -237,3 +237,145 @@ def execute(self, **kwargs) -> typing.Any:
 
     def post_execute(self, user_params: ExecutionParameters, rval: typing.Any) -> typing.Any:
         return self._config_task_instance.post_execute(user_params, rval)
+
+
+class _RawShellTask(ShellTask):
```
kumare3, line 242:
```suggestion
class RawShellTask(ShellTask):
```

kumare3: why the `_`

flytekit/extras/tasks/shell.py
```diff
@@ -237,3 +237,145 @@ def execute(self, **kwargs) -> typing.Any:
...
+            return final_outputs[0]
+        if len(final_outputs) > 1:
+            return tuple(final_outputs)
+        return None
+
+
+# The raw_shell_task is an instance of _RawShellTask and wraps a 'pure' shell script
+# This utility function allows for the specification of env variables, arguments, and the actual script within the
+# workflow definition rather than at `_RawShellTask` instantiation
+def get_raw_shell_task(name: str = None) -> _RawShellTask:
```
kumare3, line 356: can we make name required? Else if you use the same function twice, then it would clash right? or does it fail?

tests/flytekit/unit/extras/tasks/test_shell.py
```diff
@@ -249,3 +252,81 @@ def test_shell_script():
 
     assert t.script_file == script_sh
     t(f=test_csv, y=testdata, j=datetime.datetime(2021, 11, 10, 12, 15, 0))
+
+
+def test_raw_shell_task_with_args(capfd):
+    if script_sh_2 is None:
+        return
+    pst = get_raw_shell_task()
```
kumare3, line 260: can you write an example of the usage in a workflow, for better understanding and maybe, we should add example later to [Flytesnacks](https://github.com/flyteorg/flytesnacks/blob/master/cookbook/core/flyte_basics/shell_task.py)  to evnagelize this

kumare3:
@myz540 firstly this is outstanding. Thank you for the patience. Just a couple more comments. I will work on this with priority. Lets merge after you handle couple of these comments.

So i understand you have an utility function to reduce  boilerplate, but can we still use name. It is easy to make a mistake?

flytekit/extras/tasks/shell.py
myz540: I did not think there was too much value in exposing the class as without a template, it doesn't really differ too much from the regular `ShellTask`. I suppose I could add usage examples of supplying your own template into `flytesnacks`

flytekit/extras/tasks/shell.py
myz540: done

tests/flytekit/unit/extras/tasks/test_shell.py
myz540: Sure, where should I save the example? Should I open a PR into flytesnacks with the example?

flytekit/extras/tasks/shell.py
myz540: > @myz540 firstly this is outstanding. Thank you for the patience. Just a couple more comments. I will work on this with priority. Lets merge after you handle couple of these comments.
> 
> So i understand you have an utility function to reduce boilerplate, but can we still use name. It is easy to make a mistake? But, you do more, you have a better usecase

@kumare3 I have updated the PR with your proposed changes and also run black, isort, and flake8 on the code prior to the github actions. It should be in a good state. My plan is open a PR into flytesnacks with some examples and further documentation but that doesn't necessarily need to block.

myz540:
> 

@kumare3 I have updated the PR with your proposed changes and also run black, isort, and flake8 on the code prior to the github actions. It should be in a good state. My plan is open a PR into flytesnacks with some examples and further documentation but that doesn't necessarily need to block.

kumare3 (dismissed):

kumare3:
Will merge after all tests pass and also will add you as a contributor- welcome to the group 👏

kumare3 (approved):

PR #955, opened 2022-04-15
Add support deck in jupyter notebook 

pingsutw:
# TL;DR
Add support deck in jupyter notebook. Example: https://github.com/pingsutw/flytekit/blob/deck_jupyter/example_deck.ipynb

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

image

image

image

## Tracking Issue

## Follow-up issue
_NA_


flytekit/core/base_task.py
```diff
@@ -463,7 +463,7 @@ def dispatch_execute(
         new_user_params = self.pre_execute(ctx.user_space_params)
         from flytekit.deck.deck import _output_deck
 
-        new_user_params._decks = [ctx.user_space_params.default_deck]
```
wild-endeavor, line 466: does this change the user interface?  If we don't have one yet, can we get a cookbook example that shows the usage of a flyte deck.

pingsutw: no, it doesn't change the interface. we append a new deck to `new_user_params` in deck.py, so we don't need to it again here. https://github.com/flyteorg/flytekit/blob/29237baa7a5a057af16776911e9dea42617864a9/flytekit/deck/deck.py#L63

cookbook example: https://github.com/flyteorg/flytesnacks/pull/735

pingsutw:
Few updates, 
- Refine the code
- Fix the bug in the jupyter lab
- Will not output any files when running in jupyter. 
- Combine all the output HTML files (input.html, output.html, default.html) into the single one (deck.html)

pingsutw:
@kumare3 mind taking a look. The new UX in the jupyter notebook will be like
```python
@task()
def t1():
    flytekit.Deck("demo", TopFrameRenderer().to_html(df))
    return

with flytekit.new_context() as ctx:
    a = t1()
    ctx.display_deck()
```

kumare3:
I think that looks good 👍🏽

flytekit/core/context_manager.py
```diff
@@ -600,6 +602,11 @@ def current_context() -> Optional[FlyteContext]:
         """
         return FlyteContextManager.current_context()
 
+    def display_deck(self):
```
kumare3, line 605: lets just make it - get_deck, instead of display.
I think we can ask users to use 
```
with ... as ctx:
      t1()
      display(ctx.deck())
```

kumare3 (approved):

PR #960, opened 2022-04-18
pyflyte run command with first class workflow input parameters

katrogan:
# TL;DR
Updates the run command with first class workflow options and typed workflow input parameters.

This is accomplished by using a nested multi command so that workflows for a specific file can be listed in the help:

```
$ pyflyte run example.py --help
Usage: pyflyte run example.py [OPTIONS] COMMAND [ARGS]...

  Run a workflow in a file using script mode

Options:
  --help  Show this message and exit.

Commands:
  typechecker_dataclass   Run example.typechecker_dataclass in script mode
  typechecker_datetime    Run example.typechecker_datetime in script mode
  typechecker_json_types  Run example.typechecker_json_types in script mode
  typechecker_primitives  Run example.typechecker_primitives in script mode
  wf                      Run example.wf in script mode
```

And workflow inputs are subsequently listed:

```
$ pyflyte run example.py wf --help
Usage: pyflyte run example.py wf [OPTIONS]

  Run example.wf in script mode

Options:
  --remote                Whether to register and run the workflow on a Flyte
                          deployment
  -p, --project TEXT      Project to register and run this workflow in
  -d, --domain TEXT       Domain to register and run this workflow in
  --destination-dir TEXT  Directory inside the image where the tar file
                          containing the code will be copied to
  -i, --image TEXT        Image used to register and run.
  --service-account TEXT  Service account used when executing this workflow
  --wait-execution        Whether wait for the execution to finish
  --dump-snippet          Whether dump a code snippet instructing how to load
                          the workflow execution using flyteremote
  --n INTEGER             [default: 200]
  --mean FLOAT            [default: 0.0]
  --sigma FLOAT           [default: 1.0]
  --help                  Show this message and exit.
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added
 - [x] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2241

## Follow-up issue
_NA_

PR #962, opened 2022-04-20
pr into #960

wild-endeavor:
* Add support for workflows with `StructuredDataset` types in addition to pd.DataFrame.
  * These types will need to be specified on the command line as a string of a path to a directory of one file instead of to the file directly. The better aligns to the rest of the SD encoding behavior.
* Don't use `lstrip`... that removes all the characters in the string passed in, it doesn't just remove the prefix. `"aaaaab".lstrip("a") == "b"` not `"aaaab"`
* Moved the options like `remote` and `service-account` up to the `run` command level to avoid conflicts with user wf input names.
* Move construction of the `FlyteRemote` context to the workflow command and store in context since we know at that point if this is a remote run or not.
* Make the `PandasToParquetDataProxyEncodingHandler` capable of handling a folder with one file in it.
* With respect to handling of `pd.DataFrame`/`StructuredDataset` it's still a bit hacky unf.
  * For `pd.DataFrame`, when running
    * locally: just read the parquet file and make a dataframe.
    * remotely: construct a Python `StructuredDataset` object with uri pointing to the local file. Had to add a condition to the `StructuredDatasetTransformerEngine.to_literal` function to handle this since it's kind of a weird case... the `python_type` remains `pd.DataFrame`, but it'll be getting a `StructuredDataset` object instead.
  * For `StructuredDataset`, when running
    * locally: just create a StructuredDataset object with uri pointing to the local file. 
    * remotely: call the data-proxy encoder to force the upload, get the flyte SD literal, and then construct a Python `StructuredDataset` object pointing to that literal.

This was tested on https://github.com/flyteorg/flytesnacks/blob/main-condition-lp/cookbook/core/type_system/sd_sample.py#L31
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f7ba68361bc2f46bfa51?duration=all

Commands
```
remote: pyflyte run --service-account demo --remote --project flytesnacks core/type_system/sd_sample.py my_wf --remote /Users/ytong/df.parquet --image /Users/ytong/my_sd
local: pyflyte run --service-account demo --project flytesnacks core/type_system/sd_sample.py my_wf --remote /Users/ytong/df.parquet --image /Users/ytong/my_sd
```

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -69,6 +69,7 @@ class DataframeType(click.ParamType):
 
         # The value here should be a string containing a path to a parquet file. If not running locally, then we have
         # no need of reading the parquet file.
+        # This relies on the TypeEngine to trigger the remote encoder.
         return StructuredDataset(
             uri=value, metadata=literals.StructuredDatasetMetadata(structured_dataset_type=self._sdt)
         )
@@ -91,6 +92,8 @@ class StructuredDatasetParamType(click.ParamType):
         )
 
         # If we're running remotely, as part of the translation, we need to upload the file as well
+        # TODO: Figure out the best way to get the SD transformer engine to trigger the remote encoder similar to the
+        #   pd.DataFrame example.
         if ctx.obj[REMOTE_FLAG_KEY]:
             encoder = PandasToParquetDataProxyEncodingHandler(ctx.obj[DATA_PROXY_CALLBACK_KEY])
             f_ctx = FlyteContextManager.current_context()
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -3,6 +3,7 @@ import importlib
 import inspect
 import json
 import os
+import pathlib
 import typing
 from dataclasses import is_dataclass
 from datetime import datetime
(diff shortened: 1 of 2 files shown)


flytekit/clis/sdk_in_container/run.py
```diff
@@ -48,10 +60,51 @@ def convert(self, value, param, ctx) -> typing.Optional[typing.Union[typing.Dict
...
+        self._sdt = TypeEngine.to_literal_type(input_type)
+
+    def convert(self, value, param, ctx) -> typing.Union[pd.DataFrame, str]:
+        if not ctx.obj[REMOTE_FLAG_KEY]:
             return pd.read_parquet(value)
 
+        # The value here should be a string containing a path to a parquet file. If not running locally, then we have
+        # no need of reading the parquet file.
+        # This relies on the TypeEngine to trigger the remote encoder.
+        return StructuredDataset(
```
kumare3, line 73: should it be a structured dataset, dont we have to upload it yet? and that has to be done using the signed url right?

wild-endeavor: in this instance the type engine will take care of uploading it yes.

flytekit/clis/sdk_in_container/run.py
```diff
@@ -48,10 +60,51 @@ def convert(self, value, param, ctx) -> typing.Optional[typing.Union[typing.Dict
 class DataframeType(click.ParamType):
     name = "dataframe"
 
-    def convert(self, value, param, ctx) -> pd.DataFrame:
-        if not ctx.obj[REMOTE_KEY]:
+    def __init__(self, input_type: typing.Type[StructuredDataset]):
+        self._sdt = TypeEngine.to_literal_type(input_type)
+
+    def convert(self, value, param, ctx) -> typing.Union[pd.DataFrame, str]:
```
kumare3, line 66: will we have to do this for all dataFrame types? This seems a little odd?
like sparkDataFrame etc?

I somehow feel, that we should probably do this on the LiteralType - the click conversion? This universe is much smaller and controlled.

katrogan:
> Moved the options like remote and service-account up to the run command level to avoid conflicts with user wf input names.

were you able to test this? i wasn't able to get click to support the same flag in the command group and a leaf command

PR #965, opened 2022-04-20
Always try to refresh token, start flow if fail

wild-endeavor:
# TL;DR
Fixing a small issue where we have to run commands twice.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin


flytekit/clients/raw.py
```diff
@@ -195,9 +194,10 @@ def _refresh_credentials_standard(self):
             # metadata field yet. Therefore, if there's a mismatch, copy it over.
             self.set_access_token(client.credentials.access_token, authorization_header_key)
             return
-        elif client.can_refresh_token:
+
+        try:
             client.refresh_access_token()
-        else:
+        except ValueError:
```
EngHabu, line 200: Can this be more specific to only look for `401` errors? 
403, for example, means forbidden and we shouldn't attempt to re-authenticate...

EngHabu (approved):

PR #969, opened 2022-04-21
Replace FlyteContextManager._OBJS with contextvars

pingsutw:

# TL;DR
Replace `FlyteContextManager._OBJS` with `contextvars`. We use global stack `_OBJS` to maintain execution context within Flytekit, which means we can't run the tasks in multi-thread.
Using contextvars can help us maintain the stack by python internally. and it also works in a multi-thread program.

Here is a simple example to prove it. The output should always be 2, but it doesn't work without `contextvar`.
```
import threading
import pandas as pd

import flytekit
from flytekit import task
from flytekit.deck import TopFrameRenderer


@task()
def t1():
    flytekit.Deck("demo", TopFrameRenderer().to_html(df))
    print(len(flytekit.current_context().decks)) 
    return


if __name__ == "__main__":
    for i in range(20):
        threading.Thread(target=t1).start()
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/condition.py
+++ b/flytekit/core/condition.py
@@ -4,7 +4,7 @@ import datetime
 import typing
 from typing import Optional, Tuple, Union, cast
 
-from flytekit.core.context_manager import ExecutionState, FlyteContextManager
+from flytekit.core.context_manager import ExecutionState, FlyteContextManager, flyte_context_Var
 from flytekit.core.node import Node
 from flytekit.core.promise import (
     ComparisonExpression,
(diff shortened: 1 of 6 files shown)


pingsutw:
Fixed it #955

PR #970, opened 2022-04-21
hash compilation settings

wild-endeavor:
pr into other pr

First run with image "-latest"
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f6f9e9a80575441239a1?duration=all
Second run with image "-0.32.6"
https://demo.nuclyde.io/console/projects/flytesnacks/domains/development/executions/f5852df55ff574e6bae9?duration=all

Flyte versions are different, but the uploaded s3 key.

Tried hashing the `SerializationSettings` dataclass but it's not frozen so can't hash.

Should probably add `Options` as well but it's not even dataclass json and also not frozen.


cosmicBboy:
> Should probably add Options as well but it's not even dataclass json and also not frozen.

Can we make decorate `Options` with `@dataclass_json`?

kumare3 (approved):

kumare3 (approved):

PR #986, opened 2022-05-03
Set upstream node for branch node

pingsutw:

# TL;DR
We didn't set the upstream node for the branch node. Therefore, the DAG of following example is wrong,
```python
from flytekit import task, workflow, conditional


@task
def square(n: float) -> float:
    return n * n


@task
def double(n: float) -> float:
    return 2 * n


@task
def t1() -> float:
    return 3.0


@workflow
def wf(my_input: float) -> float:                                                                            
    val = t1()
    return (
        conditional("fractions")
            .if_((my_input >= 0.1) & (my_input <= 1.0))
            .then(double(n=val))
            .else_()
            .then(square(n=my_input))
    )

```
Before:
image

After:
image

For more detail, check https://flyte-org.slack.com/archives/C03CY9S9MFE/p1651262585261169 and https://flyte-org.slack.com/archives/C03CY9S9MFE/p1651307245028089

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/condition.py
+++ b/flytekit/core/condition.py
@@ -107,11 +107,16 @@ class ConditionalSection:
                 upstream_nodes=list(upstream_nodes),  # type: ignore
                 flyte_entity=node,
             )
+            for case in self.cases:
+                if case.output_promise is not None and case.output_promise.ref is not None:
+                    for node in case._output_promise.ref.node.upstream_nodes:
+                        node.runs_before(n)
+
             FlyteContextManager.current_context().compilation_state.add_node(n)  # type: ignore
             return self._compute_outputs(n)
         return self._condition
 
-    def if_(self, expr: bool) -> Case:
+    def if_(self, expr: typing.Any) -> Case:
         return self._condition._if(expr)
 
     def compute_output_vars(self) -> typing.Optional[typing.List[str]]:
@@ -283,9 +288,7 @@ class Case(object):
         return self._err
 
     # TODO this is complicated. We do not want this to run
-    def then(
-        self, p: Union[Promise, Tuple[Promise]]
-    ) -> Optional[Union[Condition, Promise, Tuple[Promise], VoidPromise]]:
+    def then(self, p: typing.Any) -> typing.Any:
         self._output_promise = p
         # We can always mark branch as completed
         return self._cs.end_branch()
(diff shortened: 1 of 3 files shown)


wild-endeavor:
using your example, are these equivalent?
```
@workflow
def wf(my_input: float) -> float:                                                                            
    val = t1()
    return (
        conditional("fractions")
            .if_((my_input >= 0.1) & (my_input <= 1.0))
            .then(double(n=val))
            .else_()
            .then(square(n=my_input))
    )
```

```
@workflow
def wf(my_input: float) -> float:                                                                            
    return (
        conditional("fractions")
            .if_((my_input >= 0.1) & (my_input <= 1.0))
            .then(double(n=t1()))
            .else_()
            .then(square(n=my_input))
    )
```
@kumare3 @pingsutw

pingsutw:
yeah, but if IIRC, the second one doesn't work, right?

wild-endeavor:
should they be equivalent?

pingsutw:
cc @wild-endeavor 
sorry, I think I was wrong. They are not quite equivalent
In the first one, we'll run `t1()` no matter the condition is true or false.
In the second one, we'll run `t1()` only when the condition is true. Therefore, we should make t1 become part of the branch node as well. We don't support it for now, right? 

which means we can't have a workflow like
 ```
                 if  ->  t1 -> t2 
start ->                               -> end
                  else  -> t3
```
For now, we can only have exactly one node in `if` block and `else` block.

wild-endeavor:
yes @pingsutw agreed.  @kumare3 @enghabu what should we do about this?

EngHabu:
I don't think the problem with the example above is the graph... it's in the execution path...
If you look at the generated DAG, it has an upstream dependency between `n1-n0` (the first branch) to `n0` which should cause that node to wait until data from n0 become available before proceeding.

image

Somewhere between [this](https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/nodes/branch/handler.go) and the Node Handler this breaks....

Please let me know if you need further help investigating this.

eapolinario:

PR #987, opened 2022-05-03
Update PR based on comments

reverson:


pingsutw (commented):
should we add `Architecture` to `_task_models.Container`?

https://github.com/flyteorg/flytekit/blob/59d0dc7c9dcc964423f3648fd13620cc01426e3e/flytekit/common/tasks/raw_container.py#L92-L100

PR #991, opened 2022-05-05
Swap out inspect file location

wild-endeavor:
# TL;DR
If the decorator and the usage of your decorator are in different places, it seems the new logic will inadvertently use the module where the decorator is defined instead of where the task is defined.

[Slack thread](https://flyte-org.slack.com/archives/CNMKCU6FR/p1651698360544049).

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Since the module is being correctly set by functools, we should just use that to get the file path.

### DIFF
--- a/tests/flytekit/unit/core/functools/test_decorator_location.py
+++ b/tests/flytekit/unit/core/functools/test_decorator_location.py
@@ -1,23 +1,11 @@
+import importlib
+
 from flytekit import task
-from .decorator_source import task_decorator_1, task_decorator_2, task_setup
 from flytekit.core.tracker import extract_task_module
 
-@task
-@task_decorator_1
-@task_decorator_2
-def my_task(x: int) -> int:
-    print("running my_task")
-    return x + 1
 
-
-@task
-@task_setup
-def get_data(x: int) -> int:
-    print("running get_data")
-    return x + 1
-
-
-def test_fjdskla():
-    res = extract_task_module(get_data)
-    print(res)
-    print(get_data.name)
+def test_dont_use_wrapper_location():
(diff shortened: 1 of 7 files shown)


tests/flytekit/unit/core/functools/test_decorator_location.py
```diff
@@ -0,0 +1,11 @@
+import importlib
+
+from flytekit import task
+from flytekit.core.tracker import extract_task_module
```
eapolinario, line 4: ditto.

tests/flytekit/unit/core/functools/test_decorator_location.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_dont_use_wrapper_location():
+    m = importlib.import_module("tests.flytekit.unit.core.functools.decorator_usage")
+    get_data_task = getattr(m, "get_data")
+    assert "decorator_source" not in get_data_task.name
+    assert "decorator_usage" in get_data_task.name
```
eapolinario, line 11: do you think it's worth having a more explicit unit test that exercises `extract_task_module`?

wild-endeavor: probably

wild-endeavor: wait no... this does do that already.

eapolinario: it does that indirectly, right? I was talking about explicit calls to that function.

eapolinario (approved):

PR #993, opened 2022-05-07
FlyteRemote data context does not get used

wild-endeavor:
# TL;DR
@cosmicBboy found this issue when working on the unionml library.  To summarize,

Even though the `FlyteRemote` object creates a `FlyteContext`, when you use that remote object to fetch an execution, there is nothing tying that `LiteralsResolver` instance back to that context.  So when you try to download/upload anything, you end up using the default `FileAccessProvider` created at the [module layer](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/data_persistence.py#L455).

To remediate, we can
* Add a context to the `LiteralsResolver` and cache the remote object's context when we get execution inputs/outputs.
  * Is this the only object that will need access to the remote object's context?  Is there anything else the user might interact with that would also require it?
* More golang style context management - allow `ctx` to be specified in the get function (but then the user has to manually use it at call-time).
* Push the context from the FlyteRemote onto the global stack in the constructor.  This way anything that uses the context will see it.  Some issues,
  * What if you create more than one remote object with different data configs?
  * What if another object, or the user manually, pushes additional FlyteContext objects onto the global stack? We do this already when we run workflows locally (inside the workflow/task `__call__` function).  Granted using a prior execution's input/outputs in a workflow declaration would be quite weird, but still something to consider.
* Do the above but lock the context somehow.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Doing option 1 - it's the least amount of code.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -294,8 +294,6 @@ class FileAccessProvider(object):
     durable store.
     """
 
-    INSTANCE = 0
-
     def __init__(
         self,
         local_sandbox_dir: Union[str, os.PathLike],
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -293,6 +293,14 @@ class FileAccessProvider(object):
     This is the class that is available through the FlyteContext and can be used for persisting data to the remote
     durable store.
     """
+    INSTANCE = 0
+
+    @classmethod
+    def update(cls) -> int:
+        old = cls.INSTANCE
+        cls.INSTANCE += 1
+        print(f"Creating FileAccessProvider instance {old} cls ID: {id(cls)}")
+        return old
 
     def __init__(
         self,
@@ -317,6 +325,8 @@ class FileAccessProvider(object):
         )
         self._raw_output_prefix = raw_output_prefix
         self._data_config = data_config if data_config else DataConfig.auto()
+
+        self._counter = FileAccessProvider.update()
 
     @property
     def data_config(self) -> DataConfig:
@@ -416,6 +426,7 @@ class FileAccessProvider(object):
         :param Text local_path:
         :param bool is_multipart:
         """
+        print(f"Calling get_data from instance {self._counter} config: {self.data_config}")
         try:
             with PerformanceTimer(f"Copying ({remote_path} -> {local_path})"):
                 pathlib.Path(local_path).parent.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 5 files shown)


flytekit/core/data_persistence.py
```diff
@@ -293,6 +293,14 @@ class FileAccessProvider(object):
     This is the class that is available through the FlyteContext and can be used for persisting data to the remote
     durable store.
     """
+    INSTANCE = 0
```
kumare3, line 296: why?

kumare3: is this only for debugging? if so lets delete this

flytekit/remote/remote.py
```diff
@@ -145,7 +145,7 @@ def __init__(
         )
 
         # Save the file access object locally, but also make it available for use from the context.
-        FlyteContextManager.with_context(
+        FlyteContextManager.push_context(
```
kumare3, line 148: :(

flytekit/core/data_persistence.py
wild-endeavor: will do

flytekit/remote/remote.py
```diff
@@ -1380,11 +1382,11 @@ def _assign_inputs_and_outputs(
         """Helper for assigning synced inputs and outputs to an execution object."""
         with self.remote_context():
```
eapolinario, line 1383: we don't really need to add this extra context manager now that we're specifying the context in the constructor of flyteremote, right? This applies to the other uses of `FlyteContext` in this file as well.

wild-endeavor: hmm.  i think this line is probably not needed, but other usages of the remote_context function are still required right?  For things like pulling down data and stuff, we need to use the remote instance's data config.

eapolinario: yes, for example in places like [1](https://github.com/flyteorg/flytekit/blob/master/flytekit/remote/remote.py#L429), [2](https://github.com/flyteorg/flytekit/blob/master/flytekit/remote/remote.py#L498), and [3](https://github.com/flyteorg/flytekit/blob/master/flytekit/remote/remote.py#L1002), right?

eapolinario (dismissed):

eapolinario (approved):

eapolinario (approved):

eapolinario (approved):

PR #1020, opened 2022-05-24
Added support for insecureSkipVerify on pyflyte

RobertoRRW:
# TL;DR
Enables the insecureSkipVerify config flag on pyflyte. Allows bypassing SSL certificate verification, allowing connections with self-signed certificates.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
When insecureSkipVerify is set to true, fetches the SSL cert from the endpoint and creates a grpc secure channel with the 'ssl_target_name_override' option set.

Similar to https://stackoverflow.com/a/52855472

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2531

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -115,7 +115,7 @@ class RawSynchronousFlyteClient(object):
         elif cfg.insecure_skip_verify:
             # Get port from endpoint or use 443
             try:
-                port = int(cfg.endpoint.rsplit(":", 1))[1]
+                port = int(cfg.endpoint.rsplit(":", 1)[1])
             except (IndexError, ValueError):
                 port = 443
             cert = ssl.get_server_certificate((cfg.endpoint, port))
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -4,8 +4,10 @@ import base64 as _base64
 import subprocess
 import time
 import typing
+import ssl
 from typing import Optional
 
+import OpenSSL
 import grpc
 import requests as _requests
 from flyteidl.admin.project_pb2 import ProjectListRequest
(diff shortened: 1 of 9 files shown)


flytekit/clients/raw.py
```diff
@@ -110,6 +112,21 @@ def __init__(self, cfg: PlatformConfig, **kwargs):
         self._cfg = cfg
         if cfg.insecure:
             self._channel = grpc.insecure_channel(cfg.endpoint, **kwargs)
+        elif cfg.insecure_skip_verify:
+            # Get the certificate from the endpoint at port 443
+            cert = ssl.get_server_certificate((cfg.endpoint, 443))
```
EngHabu, line 117: the cfg.endpoint should have the "host:port", if it doesn't have a port only then can we assume 443...

EngHabu (commented):
That's a nice trick! thank you for your contribution

flytekit/clients/raw.py
RobertoRRW: ok, changed

flytekit/clients/raw.py
```diff
@@ -110,6 +112,24 @@ def __init__(self, cfg: PlatformConfig, **kwargs):
         self._cfg = cfg
         if cfg.insecure:
             self._channel = grpc.insecure_channel(cfg.endpoint, **kwargs)
+        elif cfg.insecure_skip_verify:
+            # Get port from endpoint or use 443
+            try:
+                port = int(cfg.endpoint.rsplit(":", 1)[1])
+            except (IndexError, ValueError):
+                port = 443
+            cert = ssl.get_server_certificate((cfg.endpoint, port))
```
EngHabu, line 121: I don't think this is correct... if `cfg.endpoint = "example.com:8443"` the pair generated here will be `("example.com:8443", 8443)` whereas it should have been `("example.com", "8443")` ... notice two differences here:
1. The port should continue to be a string per [docs](https://pythontic.com/ssl/ssl-module/get_server_certificate)
2. The first element of the tuple (pair) should NOT contain the port anymore... 

The actual implementation of the API could be more lax but I would rather stick to the official docs in case the implementation catches on later...

EngHabu:
Looks like the changes in requirements are not fully working in some setups...
Mind taking a look at that @RobertoRRW?
@eapolinario do you have insights in what's going on?

```bash
Incompatible requirements found: grpcio==1.46.3 (from -r dev-requirements.txt (line 155)) and grpcio==1.46.1 (from -r requirements-spark2.txt (line 59))
```

RobertoRRW:
I had regenerated the requirements.txt and dev-requirements.txt but hadn't done so for the other requirement files.

eapolinario:
As per to https://support.snyk.io/hc/en-us/articles/360004358517-Unknown-PR-test-error this snyk error should be temporary. We can force a re-run by closing and opening the PR again.

eapolinario:
@RobertoRRW , you can run the [requirements](https://github.com/flyteorg/flytekit/blob/master/Makefile#L77) make target to ensure all requirements files are generated at once.

RobertoRRW:
The errored check doesn't seem to be related to my changes:

```
Original exception: Value error!  Received: 406. Request for data @ https://www.sqlitetutorial.net/wp-content/uploads/2018/03/chinook.zip failed. Expected status code 200
```

wild-endeavor (approved):

eapolinario (approved):

PR #1021, opened 2022-05-25
Fix importing code in script mode

eapolinario:
# TL;DR
Fix importing code in script code by leveraging the script-mode assumption as opposed to the current working directory

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The main change is in `load_naive_entity` where we no longer add the working directory to `sys.path` prior to loading the python file, instead, we add the project root.

I also added a few opportunistic fixes, like removing commented code and ensuring that we check the existence of `__init__.py` files before we copy them.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2467

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/tools/script_mode.py
+++ b/flytekit/tools/script_mode.py
@@ -14,7 +14,7 @@ from flytekit.core.tracker import extract_task_module
 from flytekit.core.workflow import WorkflowBase
 
 
-def compress_single_script(absolute_project_path: str, destination: str, full_module_name: str):
+def compress_single_script(source_path: str, destination: str, full_module_name: str):
     """
     Compresses the single script while maintaining the folder structure for that file.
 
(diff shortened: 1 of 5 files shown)


flytekit/tools/script_mode.py
```diff
@@ -142,7 +141,7 @@ def _find_project_root(source_path) -> Path:
     Traverse from current working directory until it can no longer find __init__.py files
     """
     # Start from the directory right above source_path
-    path = Path(source_path).parents[0]
+    path = Path(source_path).parents[0].resolve()
```
wild-endeavor, line 144: what does the resolve() do?

wild-endeavor (dismissed):
just a question though

flytekit/tools/script_mode.py
eapolinario: it turns a relative path into an absolute one. This particular one is useful to find the correct path of the script relative to the project root.

wild-endeavor (approved):

PR #1023, opened 2022-05-26
Add a type transformer for whylogs dataset profiles

naddeoa:
# TL;DR
This adds a type plugin that enables whylog's dataset profiles to participate in the flytekit task framework.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This is the first step to enabling whylogs dataset profile visualization in flytedeck and enabling constraints. Viz will be hooked up through flytedeck and it will allow users to visualize the data in that workflow. Constraints are a way to specify properties that should hold in the data and conditionally fail the workflow otherwise.

## Tracking Issue
-

## Follow-up issue
We're in the process of releasing our 1.0.0 right now and some features (constraints in particular) are missing from our release candidate. I need that to exist before I can add the right example. I'll also have to update the dependency in setup.py.


plugins/flytekit-spark/tests/test_wf.py
```diff
@@ -4,7 +4,7 @@
 
 import flytekit
 from flytekit import kwtypes, task, workflow
-from flytekit.types.schema import FlyteSchema
```
wild-endeavor, line 7: Could we revert this?  It's used below.

plugins/flytekit-whylogs/setup.py
```diff
@@ -0,0 +1,43 @@
...
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3.8",
+        "Topic :: Scientific/Engineering",
+        "Topic :: Scientific/Engineering :: Artificial Intelligence",
+        "Topic :: Software Development",
+        "Topic :: Software Development :: Libraries",
+        "Topic :: Software Development :: Libraries :: Python Modules",
+    ],
+    # TODO Something about this isn't working as it is.
```
wild-endeavor, line 41: it is now working right?

plugins/flytekit-whylogs/flytekitplugins/whylogs/transformer.py
```diff
@@ -0,0 +1,49 @@
...
+        return Literal(scalar=Scalar(blob=Blob(uri=remote_path, metadata=BlobMetadata(type=self._TYPE_INFO))))
+
+    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[DatasetProfileView]) -> T:
+        local_dir = ctx.file_access.get_random_local_path()
+        ctx.file_access.download(lv.scalar.blob.uri, local_dir)
+        return DatasetProfileView.read(local_dir)
+
+    # TODO where is this visible from?
+    # TODO how can I test this out locally? Do I need to build a wheel of flytekit and use it from a real project?
+    def to_html(self, ctx: FlyteContext, python_val: DatasetProfileView, expected_python_type: Type[DatasetProfileView]) -> str:
```
wild-endeavor, line 45: @pingsutw - could you help please? sorry i still don't know what the right way to extend flyte deck is.  is there documentation somewhere?

plugins/flytekit-whylogs/flytekitplugins/whylogs/transformer.py
```diff
@@ -0,0 +1,49 @@
...
+from flytekit.models.literals import Literal, Scalar, Blob, BlobMetadata
+from flytekit.models.types import LiteralType
+
+
+class WhylogsDatasetProfileTransformer(TypeTransformer[DatasetProfileView]):
+    """
+    Transforms whylogs Dataset Profiles to and from a Schema (typed/untyped)
+    """
+
+    # TODO Spark plugin is a SchemaType. How do I pick? Is blob right?
```
wild-endeavor, line 16: blob is right if you just want a file yeah.  The SchemaType is kinda deprecated.  It's still supported, but for columnar data we wrote the StructuredDataset type to handle dataframes.  but this is a protobuf message we're talking about right?  not a dataframe.

wild-endeavor (commented):
minor comments we already talked about.  let me know when the whylogs dependency is published and we're ready for final review.

plugins/flytekit-whylogs/flytekitplugins/whylogs/transformer.py
pingsutw: We just need to convert `python_val` to HTML string or pure string. https://github.com/flyteorg/flytekit/blob/cc2a4e7d6a3763b4905334b87add324159da44e5/flytekit/core/type_engine.py#L139-L143
`python_val.to_pandas().to_html()` is correct.

plugins/flytekit-whylogs/setup.py
naddeoa: It is, removed this.

plugins/flytekit-spark/tests/test_wf.py
naddeoa: I don't think this is required in the latest version. LMK if otherwise.

plugins/flytekit-whylogs/flytekitplugins/whylogs/transformer.py
naddeoa: Ok. That's what I'll leave it as for now then. Our next update will be to use our actual html viewer to visualize the profiles. How can I see the render? Is it going to be a full page or a tiny window that pops up in a hover tooltip?

eapolinario:
Superseded by https://github.com/flyteorg/flytekit/pull/1100.

PR #1032, opened 2022-06-01
TypeTransformers for PyTorch Tensor, Module, and Checkpoint

samhita-alla:

# TL;DR
This PR adds support for:

- `torch.Tensor` as a native type.
- saving and loading PyTorch modules (models & layers) using `state_dict`. `PyTorchStateDict` is the custom type defined to handle serialization and deserialization of `state_dict`. `torch.nn.Module` (base class for all neural network modules) isn't considered a native type here; this is per the [docs](https://pytorch.org/docs/stable/notes/serialization.html#saving-and-loading-torch-nn-modules): "Instead of saving a module directly, for compatibility reasons it is recommended to instead save only its state dict." Subclassing `torch.nn.Module` in a TypeTransfomer might mislead the users into thinking that a PyTorch model is being serialized and deserialized when instead the model's `state_dict` is being considered. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Module/Model serialization example:

```python
@task
def generate_module() -> torch.nn.Module:
    bn = torch.nn.BatchNorm1d(3, track_running_stats=True)
    return bn

@task
def t4(model: torch.nn.Module) -> torch.nn.Module:
    return model.l1
```

Tensor serialization example:
```python
@task
def generate_tensor_1d() -> torch.Tensor:
    return torch.zeros(5, dtype=torch.int32)

@task
def t1(tensor: torch.Tensor) -> torch.Tensor:
    assert tensor.dtype == torch.int32
    tensor[0] = 1
    return tensor
```

Checkpoint example:
```python
class Net(nn.Module):
    def __init__(self):
        super(Net, self).__init__()
        self.conv1 = nn.Conv2d(3, 6, 5)
        self.pool = nn.MaxPool2d(2, 2)
        ...

    def forward(self, x):
        ...
        return x


@task
def generate_model(hyperparameters: Hyperparameters) -> PyTorchCheckpoint:
    bn = Net()
    optimizer = optim.SGD(bn.parameters(), lr=0.001, momentum=0.9)
    return PyTorchCheckpoint(module=bn, hyperparameters=hyperparameters, optimizer=optimizer)

@task
def t1(checkpoint: PyTorchCheckpoint):
    new_bn = Net()
    new_bn.load_state_dict(checkpoint["module_state_dict"])
    optimizer = optim.SGD(new_bn.parameters(), lr=0.001, momentum=0.9)
    optimizer.load_state_dict(checkpoint["optimizer_state_dict"])
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2544

## Follow-up issue
_NA_
OR


cosmicBboy (commented):
I feel like this should be a flytekit plugin and not part of core flytekit package. It sets a precedence of adding more and more additional deps to the core package that I'm not sure we want to shoulder (pytorch, tensorflow, sklearn, etc.)

thoughts @eapolinario @kumare3 ?

__edit: I see how we're handling the case if `torch` isn't installed. I suppose this is okay, as long as we're okay establishing this pattern for other types__

samhita-alla:
@cosmicBboy, I don't think this can be a standalone plugin because having so will require users to install `torch` which might not have the version that the user has already installed or may want to install. If a user wants to use `torch.Tensor` or any other native type as such, the library might have already been installed to construct the models; so providing it with Flyte type doesn't look right.

> In the to_python_value function, I think the nn.Module.load_state_dict method should be automatically called so that the user doesn't have to worry about loading it in the task function body.

How should we serialize and deserialize a PyTorch model? How do we construct the model and apply `load_state_dict` to it?

samhita-alla:
> - we don't know ahead of time what hyperparameters to initialize the nn.Module subclass, which is an issue because often times one can't simply instantiate a model with MyModel().
> - we could do something like Annotated[MyModel, dict(...)] to capture hyperparameters, but it's not ideal since in most use cases the hyperparameters are part of the task arguments and so we don't know that ahead of time.

Can we not store hyperparameters when the user returns a module with `PyTorchStateDict` as the attached type? The module must have already been initialized with the hyperparameters in case the architecture isn't hard-coded. WDYT?

```python
@task
def generate_model() -> PyTorchStateDict:
    bn = MyModel(...)
    return PyTorchStateDict(module=bn)
```

> Does it make sense to support nn.Module via torch.save and torch.load (i.e. using pickle)? Back-pedaling on my previous comment, I wonder if it's okay to support this and referencing the pytorch docs to highlight potential issues with doing this (I just have a feeling that this would be a welcome convenience for the lazy pytorch dev 🙃).

Yeah! `torch.save` on `nn.Module` stores the path to the file containing the class. The disadvantage with this approach as per their doc is that the code can break in various ways when the module is loaded in other projects or after refactors. [Reference](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-entire-model).

But isn't this the approach we'll have to follow in case we want to support applying `load_state_dict` on the model within the plugin because that's how we could store the subclass? If it's a module or a single network layer, we'd have to store what it is, say, `BatchNorm1D`. Please let me know if you disagree.

> I think the "right way" using a PytorchStateDict dataclass is the way to go, although I'd propose extending this to PytorchCheckpoint such that it includes additional metadata, going off of [this guide](https://pytorch.org/tutorials/recipes/recipes/saving_and_loading_a_general_checkpoint.html).:

I love this! We can have `PyTorchCheckpoint` rather than `PyTorchStateDict` to handle additional metadata.

samhita-alla:
> do we want to support nn.Module as a type that uses torch.save/load, with documentation of the risks and encouraging use of PyTorchCheckpoint? I'm inclined to say yes because (i) pytorch supports it and (ii) in certain circumstances Flyte actually reduces the risk of incompatibilities since tasks are hardened via containerization.

Do you mean to say that we support:

- `nn.Module` as a native type where we use `torch.save` and `torch.load` on the entire module under the hood, and 
- `PyTorchStateDict` to serialize checkpoint and let users run `load_state_dict` on the deserialized checkpoint?

samhita-alla:
I've dynamically created transformers for `torch.nn.Module` and `torch.Tensor` because they use the same logic under the hood. To eliminate code duplication, I opted for dynamic class creation, and the two transformers are present in the `native.py` file.

On the whole, the following are the types enclosed in this PR:

- PyTorch Tensor (`torch.Tensor`) native type
- Pytorch Module (`torch.nn.Module`) native type
- `PyTorchCheckpoint`

All the examples are available in the PR description.

Note: `PyTorchCheckpoint` supports serializing hyperparamters of type `dict`, `NamedTuple`, and `dataclass`.

flytekit/types/numpy/__init__.py
```diff
@@ -1,13 +1 @@
-"""
```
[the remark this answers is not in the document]
samhita-alla: Yeah. I don't think we'd want to have Transformer in the API reference cause the methods within the TypeTransformer class remain the same.

flytekit/types/pytorch/native.py
```diff
@@ -0,0 +1,110 @@
+from __future__ import annotations
```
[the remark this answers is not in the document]
samhita-alla: I followed a procedural way of doing it, but yes, the OOP approach is more readable. There shouldn't be a problem with auto-completion cause users wouldn't be importing any module in their code as such. However, I modified the code to use inheritance now. :) Thanks for looking into this!

cosmicBboy (dismissed):
great work!

cosmicBboy (commented):
oh, one thing I forgot to mention here is: how do we want to handle cpu/cuda tensors?

For example, if I serialize a cuda tensor/module, the type engine will fail if we try to deserialize it in a cpu machine.

Should we leave that up to the user? Or should we do some automagic to handle that?

samhita-alla:
@cosmicBboy, I'm trying to generate a failure when training a model on GPU and loading the checkpoint on CPU in two different tasks. However, the execution's always a success. I ran the following code (I'll have to check if "cuda" is being set):

```python
import torch

from flytekit import task, workflow, Resources

import torch.nn as nn
import torch.nn.functional as F

from flytekit.types.pytorch import PyTorchCheckpoint

from sklearn.datasets import load_iris
from sklearn.model_selection import train_test_split
from sklearn.preprocessing import StandardScaler

import numpy as np
from typing import Tuple
from torch.autograd import Variable

class Model(nn.Module):
    def __init__(self, input_dim):
        super(Model, self).__init__()
        self.layer1 = nn.Linear(input_dim, 50)
        self.layer2 = nn.Linear(50, 50)
        self.layer3 = nn.Linear(50, 3)

    def forward(self, x):
        x = F.relu(self.layer1(x))
        x = F.relu(self.layer2(x))
        x = F.softmax(self.layer3(x), dim=1)
        return x


@task(requests=Resources(gpu="1"))
def generate_model() -> Tuple[
    PyTorchCheckpoint, torch.Tensor, torch.Tensor, torch.Tensor
]:
    iris = load_iris()
    X = iris["data"]
    y = iris["target"]
    names = iris["target_names"]
    feature_names = iris["feature_names"]

    # Scale data to have mean 0 and variance 1
    # which is importance for convergence of the neural network
    scaler = StandardScaler()
    X_scaled = scaler.fit_transform(X)

    # Split the data set into training and testing
    X_train, X_test, y_train, y_test = train_test_split(
        X_scaled, y, test_size=0.2, random_state=2
    )
    device = torch.device("cuda" if torch.cuda.is_available() else "cpu")

    model = Model(X_train.shape[1])
    model.to(device)
    optimizer = torch.optim.Adam(model.parameters(), lr=0.001)
    loss_fn = nn.CrossEntropyLoss()

    EPOCHS = 5
    X_train = Variable(torch.from_numpy(X_train)).float()
    y_train = Variable(torch.from_numpy(y_train)).long()
    X_test = Variable(torch.from_numpy(X_test)).float()
    y_test = Variable(torch.from_numpy(y_test)).long()

    X_train, X_test = X_train.to(device), X_test.to(device)
    y_train, y_test = y_train.to(device), y_test.to(device)

    loss_list = np.zeros((EPOCHS,))

    for epoch in range(EPOCHS):
        y_pred = model(X_train)
        loss = loss_fn(y_pred, y_train)
        loss_list[epoch] = loss.item()

        # Zero gradients
        optimizer.zero_grad()
        loss.backward()
        optimizer.step()

    return (
        PyTorchCheckpoint(
            module=model,
            # hyperparameters=Hyperparameters(epochs=3, loss=0.4),
            optimizer=optimizer,
        ),
        X_train,
        X_test,
        y_test,
    )


@task
def t1(
    checkpoint: PyTorchCheckpoint,
    X_train: torch.Tensor,
    X_test: torch.Tensor,
    y_test: torch.Tensor,
):
    new_bn = Model(X_train.shape[1])
    new_bn.load_state_dict(checkpoint["module_state_dict"])

    accuracy_list = np.zeros((5,))

    with torch.no_grad():
        y_pred = new_bn(X_test)
        correct = (torch.argmax(y_pred, dim=1) == y_test).type(torch.FloatTensor)
        accuracy_list = correct.mean()


@workflow
def wf():
    checkpoint, x_train, x_test, y_test = generate_model()
    t1(checkpoint=checkpoint, X_train=x_train, X_test=x_test, y_test=y_test)
```
Could you share a PyTorch code snippet if you've encountered the failure before? And, should the above code fail or succeed in your opinion?

samhita-alla:
@cosmicBboy, so `map_location` is required in all the three types: `torch.nn.Module`, `torch.Tensor`, and `PyTorchCheckpoint` (`state_dict` needs it, too!). Previously, I missed adding the pytorch-cuda image to the dockerfile, and hence, the code worked; however, now it fails during the GPU -> CPU conversion. 

Here's the code I've added to handle device conversion:

```python
if torch.cuda.is_available():
    map_location = "cuda:0"
else:
    map_location = torch.device("cpu")
```

cosmicBboy:
Amazing! Let's merge this @samhita-alla @eapolinario

samhita-alla:
+1, please. @cosmicBboy

tests/flytekit/integration/remote/mock_flyte_repo/workflows/requirements.txt
```diff
@@ -63,13 +59,16 @@ grpcio-status==1.46.3
 idna==3.3
     # via requests
 importlib-metadata==4.11.4
+<<<<<<< HEAD
+=======
```
pingsutw, line 63: nit: should remove it

tests/flytekit/integration/remote/mock_flyte_repo/workflows/requirements.txt
```diff
@@ -63,13 +59,16 @@ grpcio-status==1.46.3
...
     # via
     #   click
     #   keyring
 jeepney==0.8.0
+>>>>>>> origin/master
```
pingsutw, line 68: ditto

pingsutw (commented):
Couple of minor comments. LGTM otherwise

cosmicBboy (approved):

kumare3:
should we make this part of `pip install flytekitplugins-ml` plugin -> cc @cosmicBboy / @wild-endeavor

cosmicBboy:
just for the paper trail:

> [@Yee](https://unionai.slack.com/team/U01F63S25UN) and I chatted and we’re okay with the current approach on the PR. I think we figured it makes sense to have pytorch (and other “core” ML types) in flytekit.extras Our general heuristic is that: if it’s part of the “core” ML stack (i.e. it’s super popular, e.g. pytorch, tensorflow, keras, etc) then it can be in flytekit.extras but if not it goes into flytekitplugins … acknowledging that what is considered “core” will change over time.

flytekit/__init__.py
```diff
@@ -183,6 +183,7 @@
 from flytekit.core.workflow import WorkflowFailurePolicy, reference_workflow, workflow
 from flytekit.deck import Deck
 from flytekit.extras.persistence import GCSPersistence, HttpPersistence, S3Persistence
+from flytekit.extras.pytorch import PyTorchModuleTransformer, PyTorchTensorTransformer
```
cosmicBboy, line 186: I think this will break if torch isn't installed (?)

Maybe we can do something like:

```python
from flytekit.extras import pytorch
```

Which would automatically register the tensor and module type transformers (if `torch` is installed)

cosmicBboy (commented):
Module organization makes sense to me, see comment about `flytekit/__init__.py` import

flytekit/__init__.py
samhita-alla: My bad. Fixed the import now.

cosmicBboy (approved):

pingsutw (approved):

flytekit/extras/pytorch/checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol
+
+
+class IsDataclass(Protocol):
+    __dataclass_fields__: Dict
+    __dataclass_params__: Dict
+    __post_init__: Optional[Callable]
```
eapolinario, line 24: just for my own understanding, why does this work to detect if an object is a dataclass?

wild-endeavor (approved):

flytekit/extras/pytorch/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+.. autosummary::
+   :template: custom.rst
+   :toctree: generated/
+
+    PyTorchCheckpoint
+"""
+from flytekit.loggers import logger
+
+try:
+    from .checkpoint import PyTorchCheckpoint, PyTorchCheckpointTransformer
```
wild-endeavor, line 15: we should stick with the full import in the future, just for consistency.  merge as is, i'll update it in the future.

flytekit/extras/pytorch/checkpoint.py
samhita-alla: Every dataclass has `__dataclass_fields__` and `__dataclass_params__` defined. So this class is to define the type for reference. It doesn't really verify the type of the object.  That's handled by `__post_init__`.

flytekit/extras/pytorch/checkpoint.py
```diff
@@ -0,0 +1,141 @@
...
+from flytekit.core.context_manager import FlyteContext
+from flytekit.core.type_engine import TypeEngine, TypeTransformer, TypeTransformerFailedError
+from flytekit.models.core import types as _core_types
+from flytekit.models.literals import Blob, BlobMetadata, Literal, Scalar
+from flytekit.models.types import LiteralType
+
+try:
+    from typing import Protocol
+except ImportError:
+    from typing_extensions import Protocol
```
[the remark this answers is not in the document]
samhita-alla: Yep! I'll merge this now but will make sure to modify the import to use `typing_extensions` in a different PR.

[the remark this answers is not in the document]
samhita-alla: Modified the import — I had to resolve a merge conflict.

flytekit/extras/pytorch/checkpoint.py
eapolinario: Nice, thanks for explaining. Is this documented anywhere? I just want to make sure this check is not a one-off and it's actually the way people usually type-check dataclasses.

eapolinario (approved):

flytekit/extras/pytorch/checkpoint.py
samhita-alla: I don't think there's any official documentation. I came across this notation on [stackoverflow](https://stackoverflow.com/questions/54668000/type-hint-for-an-instance-of-a-non-specific-dataclass). If there's a better way to type-hint dataclass, let me know!

PR #1065, opened 2022-06-15
Fix read authorization_metadata_key from public_client_config in _refresh_credentials_from_command

sonjaer:
# TL;DR

`FLYTE_CREDENTIALS_AUTHORIZATION_METADATA_KEY` environment variable was removed in this PR https://github.com/flyteorg/flytekit/pull/858 in favor of reading the authorization metadata key from Admin. 

There was a miss in the PR where the `_refresh_credentials_from_command` did not switch to get the authorization key from self.public_client_config.authorization_metadata_key but instead used the default value `authorization`. 

Spotify uses `flyte-authorization` so we received an error about calls being UNAUTHENTICATED since the header name didn't correspond to the one expected by admin. 

This PR adds support for reading `authorization_metadata_key` from `public_client_config` in `_refresh_credentials_from_command`

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -275,7 +275,9 @@ class RawSynchronousFlyteClient(object):
             cli_logger.error("Failed to generate token from command {}".format(command))
             raise _user_exceptions.FlyteAuthenticationException("Problems refreshing token with command: " + str(e))
 
-        authorization_header_key = self.public_client_config.authorization_metadata_key or Credentials.COMMAND.read() or None
+        authorization_header_key = self.public_client_config.authorization_metadata_key or Credentials.COMMAND.read()
+        if not authorization_header_key:
+            self.set_access_token(output.stdout.strip())
         self.set_access_token(output.stdout.strip(), authorization_header_key)
 
     def _refresh_credentials_noop(self):
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -265,7 +265,7 @@ class RawSynchronousFlyteClient(object):
         :return:
         """
 
-        command = self._cfg.command or Credentials.COMMAND.read()
+        command = self._cfg.command
         if not command:
             raise FlyteAuthenticationException("No command specified in configuration for command authentication")
         cli_logger.debug("Starting external process to generate id token. Command {}".format(command))
@@ -283,7 +283,7 @@ class RawSynchronousFlyteClient(object):
         pass
 
     def refresh_credentials(self):
-        cfg_auth = self._cfg.auth_mode or Credentials.AUTH_MODE.read()
+        cfg_auth = self._cfg.auth_mode
         if type(cfg_auth) is str:
             try:
                 cfg_auth = AuthType[cfg_auth.upper()]
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -21,7 +21,6 @@ from google.protobuf.json_format import MessageToJson as _MessageToJson
 
 from flytekit.clis.auth import credentials as _credentials_access
 from flytekit.configuration import AuthType, PlatformConfig
-from flytekit.configuration.internal import Credentials
 from flytekit.exceptions import user as _user_exceptions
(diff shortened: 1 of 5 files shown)


flytekit/clis/flyte_cli/main.py
```diff
@@ -277,6 +277,7 @@ def _get_client(host: str, insecure: bool) -> _friendly_client.SynchronousFlyteC
     if parent_ctx.obj["cacert"]:
         kwargs["root_certificates"] = parent_ctx.obj["cacert"]
     cfg = parent_ctx.obj["config"]
+    print(cfg)
```
sonjaer, line 280:
```suggestion
```

flytekit/configuration/internal.py
```diff
@@ -65,6 +65,17 @@ class Credentials(object):
     This command is executed to return a token using an external process.
     """
 
+    AUTHORIZATION_METADATA_KEY = ConfigEntry(LegacyConfigEntry(SECTION, "authorization_metadata_key", list), YamlConfigEntry("admin.authorizationHeader", list))
```
sonjaer, line 68: I think we could remove this actually

flytekit/clients/raw.py
```diff
@@ -264,7 +265,7 @@ def _refresh_credentials_from_command(self):
         :return:
         """
 
-        command = self._cfg.command
+        command = self._cfg.command or Credentials.COMMAND.read()
```
eapolinario, line 268: We shouldn't need this after https://github.com/flyteorg/flytekit/pull/1063/. 

If you follow click's code, `self._cfg` in the case of `flyte-cli` commands is set [here](https://github.com/flyteorg/flytekit/blob/master/flytekit/clis/flyte_cli/main.py#L514). Notice the use of `PlatformConfig.auto()`, this should [take care](https://sourcegraph.com/github.com/flyteorg/flytekit@87d13903993d5339723ecd859bd362a64e780aec/-/blob/flytekit/configuration/__init__.py?L333-346) of loading the config values using env vars if those are set.

tests/flytekit/unit/clients/test_raw.py
```diff
@@ -195,6 +225,15 @@ def test_basic_strings(mocked_method):
...
-    cc = RawSynchronousFlyteClient(PlatformConfig(auth_mode=AuthType.EXTERNAL_PROCESS))
+    cc = RawSynchronousFlyteClient(PlatformConfig(auth_mode=AuthType.EXTERNALCOMMAND))
+    cc.refresh_credentials()
+    assert mocked_method.called
+
+
+@patch("flytekit.configuration.internal.Credentials.AUTH_MODE.read")
+@patch.object(RawSynchronousFlyteClient, "_refresh_credentials_from_command")
+def test_refresh_from_environment_variable(mocked_method, mock_read_auth_mode_from_env):
+    cc = RawSynchronousFlyteClient(PlatformConfig(auth_mode=None))
```
eapolinario, line 236: Can you use `PlatformConfig.auto` and mock the relevant env vars using https://docs.pytest.org/en/6.2.x/reference.html#pytest.MonkeyPatch.setenv ?

eapolinario (commented):
Thanks for this fix, the change to pass the authorization key in the case of external commands was missing.

flytekit/clients/raw.py
sonjaer: Ah ok sorry, missed that

sonjaer:
@eapolinario  what takes precedence from config, env and passing a flag?

eapolinario:
> @eapolinario what takes precedence from config, env and passing a flag?

`flyte-cli` rules are a bit more complicated, but in general, for configuration settings, environment variables take precedence over config data, with the exception of the `host` and `insecure` flags.

eapolinario (approved):

PR #1084, opened 2022-06-29
Housekeeping for pyflyte package and register commands

madhur-tandon:
# TL;DR

This PR shows a help message along with showing a nice error at the bottom, while also exiting with a non-zero exit code for commands such as `pyflyte package` and `pyflyte register` when they fail. It also fixes the issue of rebundling an existing `.tgz` file while using fast registration.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
1. A simple fix for `pyflyte register` when its argument `package_or_module` is not supplied, would have been to add `required=True` in its definition. This would have resulted in the following output after running `pyflyte register`:
```
Usage: pyflyte register [OPTIONS] PACKAGE_OR_MODULE...
Try 'pyflyte register --help' for help.

Error: Missing argument 'PACKAGE_OR_MODULE...'.
```

But, the documentation of `click` here: https://click.palletsprojects.com/en/8.1.x/arguments/#variadic-arguments in the section `Note on Non-Empty Variadic Arguments` doesn't recommend it.

Thus, an explicit check is done instead with the condition of `len(package_or_module) == 0`

2. Although the `help` message is displayed along with the error, it can be quite verbose. So we can also display only `usage` if required, for brevity. But for now, I have still chosen displaying of `help` for now as the original issue demands.

3. `Help` is only displayed for errors and not warnings aka things that lead to an exit with a non-zero status code, where error is displayed in Red. So it doesn't cover other cases where for eg: we display warning messages in Yellow and exit with a zero status code.

4. Initially, I wasn't able to reproduce the issue of doubling file size of the `.tgz` file during fast registration. The steps I followed are:
    - Make a directory called `core`, have a `sample.py` file in it with the contents from https://github.com/flyteorg/flytekit#a-simple-example
    - Use `pyflyte --pkgs core package --image core:v1 --fast` in the terminal to get a `flyte-package.tgz` in my current directory.
    - Use the same command again, but this time instead of getting a new `.tgz` file with double the size, I get the following:
    ```
    Usage: pyflyte package [OPTIONS]
    Try 'pyflyte package --help' for help.

    Error: Invalid value: Output file /Users/madhur/Desktop/unionml/flytekit/flyte-package.tgz already exists, specify -f to
    override.
    ```
    which is good I guess.
    - I then tried the same command again but with additional `--force` flag upon which I was able to reproduce the issue.

5. The idea of using temporary files for storing the `.tgz` files as suggested by the issue isn't the best one. According to the link here: https://stackoverflow.com/questions/18249314/are-tempfile-directories-eventually-cleared-by-system, it is suggested to handle the cleanup ourselves. On investigating a bit, it seems like the issue only occurs when the path of the `.tgz` file is present within the directory whose contents have to be packaged. A simple check for this along with an existence check for the `.tgz` file will resolve this, provided we clean-up the previous file when the above checks pass. I have also added logging and tests for this.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2638

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -1,10 +1,10 @@
 import os
 import pathlib
-import sys
 import typing
 
 import click
 
+from flytekit.clis.helpers import display_help_with_error
 from flytekit.clis.sdk_in_container import constants
 from flytekit.clis.sdk_in_container.helpers import get_and_save_remote_with_click_context
 from flytekit.configuration import FastSerializationSettings, ImageConfig, SerializationSettings
@@ -123,12 +123,10 @@ def register(
         raise ValueError("Unimplemented, just specify pkgs like folder/files as args at the end of the command")
 
     if len(package_or_module) == 0:
-        click.echo(f"{ctx.get_help()}\n")
-        click.secho(
+        display_help_with_error(
+            ctx,
             "Missing argument 'PACKAGE_OR_MODULE...', at least one PACKAGE_OR_MODULE is required but multiple can be passed",
-            fg="red",
         )
-        sys.exit(1)
 
     cli_logger.debug(
         f"Running pyflyte register from {os.getcwd()} "
@@ -170,9 +168,7 @@ def register(
         serialization_settings, detected_root, list(package_or_module), options
     )
     if len(registerable_entities) == 0:
-        click.echo(f"{ctx.get_help()}\n")
-        click.secho("No Flyte entities were detected. Aborting!", fg="red")
-        sys.exit(1)
+        display_help_with_error(ctx, "No Flyte entities were detected. Aborting!")
     cli_logger.info(f"Found and serialized {len(registerable_entities)} entities")
 
     if not version:
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -128,7 +128,7 @@ def register(
             "Missing argument 'PACKAGE_OR_MODULE...', at least one PACKAGE_OR_MODULE is required but multiple can be passed",
             fg="red",
         )
-        sys.exit(-1)
+        sys.exit(1)
 
     cli_logger.debug(
         f"Running pyflyte register from {os.getcwd()} "
@@ -170,6 +170,7 @@ def register(
         serialization_settings, detected_root, list(package_or_module), options
     )
     if len(registerable_entities) == 0:
+        click.echo(f"{ctx.get_help()}\n")
         click.secho("No Flyte entities were detected. Aborting!", fg="red")
         sys.exit(1)
     cli_logger.info(f"Found and serialized {len(registerable_entities)} entities")
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -122,6 +122,14 @@ def register(
     if pkgs:
         raise ValueError("Unimplemented, just specify pkgs like folder/files as args at the end of the command")
 
+    if len(package_or_module) == 0:
+        click.echo(f"{ctx.get_help()}\n")
+        click.secho(
+            "Missing argument 'PACKAGE_OR_MODULE...', at least one PACKAGE_OR_MODULE is required but multiple can be passed",
+            fg="red",
+        )
+        sys.exit(-1)
+
     cli_logger.debug(
         f"Running pyflyte register from {os.getcwd()} "
         f"with images {image_config} "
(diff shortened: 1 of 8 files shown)


wild-endeavor:
Hey just read your comment about the temp directories.  Multiple users including myself all the time have fallen into this trap.  Could we maybe try writing them into a certain folder in the tmp directory and maybe clearing that folder on runs?  or something like that?  I feel the UX improvement outweighs possible downsides from lingering files right?  What do you think?

flytekit/tools/repo.py
```diff
@@ -91,6 +91,10 @@ def package(
 
         # If Fast serialization is enabled, then an archive is also created and packaged
         if fast:
+            # If output exists and is a path within source, delete it so as to not re-bundle it again.
```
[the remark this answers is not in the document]
madhur-tandon: `output_tmpdir` is just a vessel for contents of `source` to be packaged into. Even if we save it somewhere else, it doesn’t matter since the size of `.tar.gz` will still double as the real underlying issue is `output` being a subpath of `source`. I will make it so that if `output` is not supplied, then save it to a temporary directory. Else, use it. With that said, I still think we need a check for `output` being within `source` — and maybe display an error there — so as to prevent it from saving altogether — This is for the case when the user supplies `output` such that it is still inside `source` , so it is worth handling I guess.

wild-endeavor (commented):
Okay hey sorry - you're right.   I think for the package command, leaving the output tgz file in the source directory is actually okay.  sorry about that.  Can you move that logic over to `register` instead?  Register still seems to be leaving the "fast....tgz" file in the source folder. I didn't think of it earlier for some reason but yeah the tgz file is actually needed in the subsequent command (`flytectl`) and having it in the current folder actually makes that easier to type.

The error messaging looks good though.

Thanks!

madhur-tandon:
> Can you move that logic over to `register` instead? Register still seems to be leaving the "fast....tgz" file in the source folder.

Done, I have:
- retained check for `output` within `source` to prevent doubling of `.tgz` for `package` command
- moved the storage of `.tgz` file inside a temporary directory for the `register` command instead of `package` when `output` is not provided
- also clean up the temporary directory created above once it has been sent to Flyte admin

flytekit/clis/sdk_in_container/register.py
```diff
@@ -142,6 +149,13 @@ def register(
     md5_bytes, native_url = remote._upload_file(pathlib.Path(zip_file))
     cli_logger.debug(f"Uploaded zip {zip_file} to {native_url}")
 
+    # Clean-up temporary directory that was used for fast registration
```
wild-endeavor, line 152: Can we delete this section?  i like the idea of keeping it around, just in case it needs to be inspected.

madhur-tandon: Yeah sure, not sure how the OS will clean it up then, is that fine?

flytekit/tools/repo.py
```diff
@@ -91,6 +91,10 @@ def package(
...
+            if os.path.abspath(output).startswith(os.path.abspath(source)) and os.path.exists(output):
```
wild-endeavor, line 95: does this ever get triggered?

madhur-tandon: If you do `pyflyte --pkgs core package --image core:v1 --fast --force` again and again, then yes

madhur-tandon: This is the original issue of `.tar.gz` doubling in size every time it is run

tests/flytekit/unit/cli/pyflyte/test_package.py
```diff
@@ -44,6 +63,29 @@ def test_get_registrable_entities():
         assert False, f"found unknown entity {type(e)}"
 
 
+def test_package_with_fast_registration():
```
wild-endeavor, line 66: 🙏

tests/flytekit/unit/cli/pyflyte/test_package.py
```diff
@@ -114,3 +156,11 @@ def test_package():
 def test_pkgs():
     pp = pyflyte.validate_package(None, None, ["a.b", "a.c,b.a", "cc.a"])
     assert pp == ["a.b", "a.c", "b.a", "cc.a"]
+
+
+def test_package_with_no_pkgs():
```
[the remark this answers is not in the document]
madhur-tandon: welcome 🙏🏼

flytekit/tools/repo.py
madhur-tandon: I am keeping this and resolving the comment

flytekit/clis/sdk_in_container/register.py
wild-endeavor: yeah it's the OS's problem

wild-endeavor (approved):

PR #1085, opened 2022-07-01
Increase laziness of SerializationSettings

madhur-tandon:
# TL;DR

This PR makes passing of `SerializationSettings` for APIs like `register_task` and `register_workflow` lazy aka a user may not pass them. This is helpful in cases like:

1. Using a task which doesn't explicitly require a container, eg: `SQLTask`
2. Using a fetched task in a workflow, which is already registered and thus doesn't require the use of `SerializationSettings`

In cases where it is required, the error is bubbled up to `translator.py` which throws a `ValueError` at some point.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
1. The approach is to make `SerializationSettings` optional. But since we cannot change the API i.e. function signatures -- a dummy `SerializationSettings` object is created when the user doesn't pass one (i.e. it is None). This dummy object doesn't have any image associated with it. Hence, the error of not having an image is bubbled up and is handled later in `translator.py`

2. The following cases are tested (without passing `SerializationSettings` of course):
- Registering a SQL Task -- should allow
- Registering a Python based task -- should throw an error
- Registering a Workflow that only has SQL Tasks (or tasks used to fetch something etc.) -- should allow
- Registering a Workflow that uses both SQL based Tasks and Python based Tasks -- should throw an error
- Registering a Workflow only based on fetched tasks (even if they were originally Python based) -- should allow
- Registering a Workflow based on fetched tasks as well as some local Python Tasks -- should throw an error

## Some other observations
1. `register_launch_plan` is inconsistent with `register_task` and `register_workflow` — the latter 2 have `SerializationSettings` in their arguments, but the former does not. In my opinion, these 3 functions should be consistent aka have similar function signatures. But this is out of scope for the given task. I have added a TODO comment.

2. Inside the function `register_task`, the call to `_serialize_and_register` can probably replaced by the following snippet:
```
from flytekit.tools.translator import  get_serializable_task
...
...
ident = self._resolve_identifier(ResourceType.TASK, entity.name, version, serialization_settings)
m = OrderedDict()
idl_task = get_serializable_task(m, serialization_settings, entity)
self.client.create_task(task_identifer=ident, task_spec=idl_task.spec)
```
Similar stuff can be done for `register_workflow`.
But I am not sure if this is even required. But if this is indeed done, we can possibly eliminate the use of `_serialize_and_register` altogether.
Regardless, this PR doesn't attempt to do this.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2647

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -367,8 +367,8 @@ class FlyteRemote(object):
     def _serialize_and_register(
         self,
         entity: FlyteLocalEntity,
-        settings: typing.Optional[SerializationSettings],
         version: str,
+        settings: typing.Optional[SerializationSettings] = None,
         options: typing.Optional[Options] = None,
     ) -> Identifier:
         """
@@ -376,19 +376,13 @@ class FlyteRemote(object):
         :return: Identifier of the registered entity
         """
         m = OrderedDict()
-        # Create dummy serialization settings for now.
-        # TODO: Clean this up by using lazy usage of serialization settings in translator.py
-        serialization_settings = (
-            settings
-            if settings
-            else SerializationSettings(
-                ImageConfig.auto_default_image(),
-                project=self.default_project,
-                domain=self.default_domain,
-                version=version,
+
+        if settings is None:
+            settings = SerializationSettings(
+                ImageConfig(), project=self.default_project, domain=self.default_domain, version=version
             )
-        )
-        _ = get_serializable(m, settings=serialization_settings, entity=entity, options=options)
+
+        _ = get_serializable(m, settings=settings, entity=entity, options=options)
 
         ident = None
         for entity, cp_entity in m.items():
@@ -432,7 +426,7 @@ class FlyteRemote(object):
                     default_lp = LaunchPlan.get_default_launch_plan(self.context, entity)
                     lp_entity = get_serializable_launch_plan(
                         OrderedDict(),
-                        settings or serialization_settings,
+                        settings,
                         default_lp,
                         recurse_downstream=False,
                         options=options,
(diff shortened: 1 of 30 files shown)


flytekit/remote/remote.py
```diff
@@ -367,28 +367,22 @@ def _resolve_identifier(self, t: int, name: str, version: str, ss: Serialization
...
-            settings
-            if settings
-            else SerializationSettings(
-                ImageConfig.auto_default_image(),
-                project=self.default_project,
-                domain=self.default_domain,
-                version=version,
+
+        if settings is None:
+            settings = SerializationSettings(
```
wild-endeavor, line 381: can we not have the dummy settings here?  it feels a little unclean - honestly the places where we already use dummy settings should be removed.  can we update the get_serializable so that it only looks for the SerializationSettings object when it's necessary?

madhur-tandon: hmm, I will have to change the function signature and the default value. I cannot let it be `None` since there are further conditions that check for `None` inside `_serialize_and_register` beneath. The dummy settings are now a part of the default value and the API has been refactored for every affected file. Hope this helps!

flytekit/tools/translator.py
```diff
@@ -542,8 +542,8 @@ def get_reference_spec(
 
 def get_serializable(
     entity_mapping: OrderedDict,
-    settings: SerializationSettings,
     entity: FlyteLocalEntity,
+    settings: Optional[SerializationSettings] = SerializationSettings(ImageConfig()),
```
wild-endeavor, line 546: Can we make this `None` instead of a dummy `SerializationSettings`?

madhur-tandon: good catch, thanks! will update

PR #1088, opened 2022-07-02
Fix PyPI package source for `v1.1.0`

sugatoray:
## Use jinja2 autoescaping for security purposes

- [x] add `select_autoescape(enabled_extensions=('html',))` to line-114.
  https://github.com/flyteorg/flytekit/blob/a968a5aa8338adbec10a38f79c5fdddb72c7c161/flytekit/deck/deck.py#L112-L115

  - See: https://github.com/conda-forge/flytekit-feedstock/pull/14#issuecomment-1172955010

  ```sh
  # 🔥 include autoescaping for security purposes
  # sources:
  # - https://jinja.palletsprojects.com/en/3.0.x/api/#autoescaping
  # - https://stackoverflow.com/a/38642558/8474894 (see in comments)
  # - https://stackoverflow.com/a/68826578/8474894
  ```

- [x] Fix non-inclusion of `flytekit/deck/html/template.html` in `v1.1.0`: Add `MANIFEST.in` to fix this bug.

  - The `setup.py` file tries to package `template.html` from the root of the repository. But, it has been moved to `flytekit/deck/html/template.html`.
    https://github.com/flyteorg/flytekit/blob/a968a5aa8338adbec10a38f79c5fdddb72c7c161/setup.py#L25

  - There is no `template.html` at the root of the package.
  
    image

  - The PyPI package (`v1.1.0`) does not have any `template.html`.

  image

  - The PyPI package (`v1.1.0`) also does not include any `template.html` inside `flytekit/deck/html` directory.
  
   image

## Type
 - [x] Bug Fix: :orange_circle: 
 - [x] Security vulnerability

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,0 +1,49 @@
+## Include files and folders
+
+recursive-include flytekit *
+recursive-include flytekit_scripts *
+recursive-include plugins *
+
+include README.md
+include flytekit/py.typed  # marker file for PEP 561
+include flytekit/deck/html/template.html
+
+include CHANGELOG.md
+include LICENSE
+
+include MANIFEST.in
+include pyproject.toml
+include setup.py
+include setup.cfg
+
+
+## Exclude files and folders
+
+recursive-exclude tests *
+recursive-exclude docs *
+recursive-exclude boilerplate *
+recursive-exclude .github *
+
+# exclude .dist folder: 
+#  - contains the generated *.tar.gz and .whl files.
+recursive-exclude .dist *
+
+# exclude requirements files
+exclude requirements.*
+exclude requirements-*.*
+exclude doc-requirements.*
+exclude dev-requirements.*
+
+# exclude config files
+exclude .gitignore
+exclude .readthedocs.yaml
+exclude .pre-commit-config.yaml
+exclude codecov.yml
+
+# exclude other repository management files
+exclude Dockerfile.py*
+exclude Makefile
+exclude NOTICE
+exclude pull_request_template.md
+exclude CODEOWNERS
+exclude CODE_OF_CONDUCT.md
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -27,9 +27,9 @@ recursive-exclude docs *
 recursive-exclude boilerplate *
 recursive-exclude .github *
 
-# exclude .dist folder: 
+# exclude dist folder: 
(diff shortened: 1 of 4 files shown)


pingsutw:
@sugatoray It seems like the dataframe doesn't be rendered correctly if using autoescaping, check the example below
```python=
import pandas as pd

@task
def t1() -> pd.DataFrame:
    return pd.DataFrame({"Name": ["Tom", "Joseph"], "Age": [20, 22]})

@workflow
def wf() -> pd.DataFrame:
    return t1()

wf()
``` 

image

The correct output should be like
image

sugatoray:
Interesting! Let me look into it.

pingsutw:
@sugatoray You're able to reproduce it locally as well. check this [example](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/deck.html#sphx-glr-auto-core-flyte-basics-deck-py)

sugatoray:
## Some Notes

The following is taken from Flask. It provides more clarity on what `autoescaping` means in Jinja2 and how to allow expected html tags safely.

I would like to try them in the following order:

1. Use the `{% autoescape %}` block while rendering the template.

   ```j2
   {% autoescape false %}
       <p>autoescaping is disabled here
       <p>{{ will_not_be_escaped }}
   {% endautoescape %}
   ```

2. Use `{{ variable | safe }}` inside the template (`template.html`) to specify each variable as safe.

> The other alternative will be to pass the generated html template into `IPython.display.HTML()` function. (This is just a hunch; am not sure if this will work or not)

source: https://flask.palletsprojects.com/en/2.1.x/templating/#controlling-autoescaping

image

wild-endeavor:
did this work?  tests are passing but lint is now complaining.  would it be possible to codify the original issue pingsutw raised into a unit test btw?

wild-endeavor:
@sugatoray were you able to test this?  does the `safe` render things correctly again?
Thank you.

wild-endeavor:
@sugatoray mind giving me write access to your fork so i can clean up the lint error?

pingsutw (approved):
I just tested it, and render stuff looks good to me. Thank you @sugatoray

wild-endeavor (approved):

PR #1093, opened 2022-07-05
Ray Task Support

pingsutw:
# TL;DR
- [Design Doc](https://docs.google.com/document/d/1-695lxz8a-GFz4cFamGkF1NhspKMkDGIDdkY9EDLMz8/edit#)
- [Slide](https://docs.google.com/presentation/d/1d9qyswQppPUlEpSVoXdXPeAV2oEWlaSKLpvPtCv8QMw/edit#slide=id.g10d1fdf7a50_0_14)

Flyte ray task contains Ray Cluster Config that will be used by propeller to create a Ray cluster.

There are three ways to run Ray job
1. submit to the existing cluster
```python
import typing
import ray
from flytekit import task, workflow, Resources
from flytekitplugins.ray import RayJobConfig, WorkerNodeConfig, HeadNodeConfig

@ray.remote
def f(x):
    return x * x

@task(task_config=RayJobConfig(
    address=<RAY_CLUSTER_ADDRESS>
    runtime_env={"pip": ["numpy", "pandas"]})
)
def ray_task() -> typing.List[int]:
    futures = [f.remote(i) for i in range(5)]
    return ray.get(futures)
```
2. Running the below task locally, and flyte will create a local Ray cluster
3. Running the below task in Flyte Cluster, Flyte will create a [RayJob](https://github.com/ray-project/kuberay/blob/76f96bf5a05651611f9a8734c34c2a7f23f54b54/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L35-L53) CR in Kubernetes.
```python
@task(task_config=RayJobConfig(
    worker_node_config=[WorkerNodeConfig(group_name="test-group", replicas=10)],
)
def ray_task() -> typing.List[int]:
    futures = [f.remote(i) for i in range(5)]
    return ray.get(futures)
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2641

## Follow-up issue
_NA_


flytekit/models/core/resource.py
```diff
@@ -0,0 +1,289 @@
...
+
+class WorkerGroupSpec(_common.FlyteIdlEntity):
+    def __init__(
+        self,
+        group_name: str,
+        replicas: int,
+        min_replicas: typing.Optional[int] = 0,
+        max_replicas: typing.Optional[int] = None,
+        compute_template: typing.Optional[str] = None,
+        image: typing.Optional[str] = "rayproject/ray:1.8.0",
```
kumare3, line 16: can we not use the `container_image` directly? or this for the worker image. 
Even if so, can we not re-use the worker image?

plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -0,0 +1,71 @@
...
+            task_config=task_config,
+            task_type=self._RAY_TASK_TYPE,
+            task_function=task_function,
+            resource=task_config.resource,
+            **kwargs
+        )
+        self._task_config = task_config
+
+    def pre_execute(self, user_params: ExecutionParameters) -> ExecutionParameters:
+        ray.init(self._task_config.address, **self._task_config.extra_args)
```
kumare3, line 62: if we are running locally, how can we do one of 2 things
1. Run in local mode
2. Connect to a remote Ray cluster if `RAY_REMOTE_CLUSTER` ip is set?

plugins/flytekit-ray/tests/test_ray.py
```diff
@@ -0,0 +1,74 @@
...
+config = RayConfig(
+    address=None,
+    ray_cluster=RayCluster(
+        name="test_ray",
+        cluster_spec=ClusterSpec(
+            head_group_spec=HeadGroupSpec(image="ray.io/ray:1.10.2"),
+            worker_group_spec=[WorkerGroupSpec(group_name="test_group", replicas=3)],
+        ),
+    ),
+    runtime_env={"pip": ["numpy"]},
```
kumare3, line 18: is it possible to run our own image in ray?

kumare3:
looks pretty good

.github/workflows/pythonbuild.yml
```diff
@@ -112,6 +114,7 @@ jobs:
           pip install -r requirements.txt
           if [ -f dev-requirements.txt ]; then pip install -r dev-requirements.txt; fi
           pip install --no-deps -U https://github.com/flyteorg/flytekit/archive/${{ github.sha }}.zip#egg=flytekit
+          pip install --no-deps -U git+https://github.com/flyteorg/flyteidl@ray
```
pingsutw, line 117: Reminder: Remove it after merging idl pr

flytekit/models/core/resource.py
pingsutw: Updated it, Now we use [task image](https://github.com/flyteorg/flyteidl/blob/7b1fd01ff429d28755765204f00dea57598d8851/protos/flyteidl/core/tasks.proto#L154) by default for the head node and worker nodes, so we have to use `rayproject/ray:1.8.0` as our base image

plugins/flytekit-ray/flytekitplugins/ray/task.py
pingsutw: if we run the ray task locally, ray tasks will connect to the remote cluster if `self._task_config.address ` is set. 
On the other hand, Ray will create a local standalone cluster if the address is None.

plugins/flytekit-ray/tests/test_ray.py
daikeshi: > yup,but we have to use `rayproject/ray:1.8.0` as our base image

hmm, is there any restriction on which Ray version users can use with Flyte?

daikeshi: We will also need to be able to configure Ray head/worker images

wild-endeavor:
trying to be better about pr comments - since this is a new feature/integration, can you write a couple examples and put them in the pr description?  we'll need the examples for the documentation in any case so it's not wasted work.

wild-endeavor (approved):

PR #1102, opened 2022-07-13
Don't override target name when using self-signed certs

wild-endeavor:
# TL;DR
We were getting
```
Error getpublicclient config <_InactiveRpcError of RPC that terminated with:
        status = StatusCode.UNKNOWN
        details = "Stream removed"
        debug_error_string = "{"created":"@1657750494.962014000","description":"Error received from peer ipv4:54.80.18.147:443","file":"src/core/lib/surface/call.cc","file_line":967,"grpc_message":"Stream removed","grpc_status":2}"
>
```
when using the admin client with a self-signed cert.

The reason is because the `cn` from the x509 subject had a * in it.  `*.elb.us-east-1.amazonaws.com` which was causing the ssl check to fail in nginx ingress.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2420

### DIFF
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -120,7 +120,6 @@ class RawSynchronousFlyteClient(object):
             else:
                 server_address = (cfg.endpoint, "443")
             cert = ssl.get_server_certificate(server_address)
-            x509 = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, cert)
             credentials = grpc.ssl_channel_credentials(str.encode(cert))
             options = kwargs.get("options", [])
             self._channel = grpc.secure_channel(
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -119,13 +119,11 @@ class RawSynchronousFlyteClient(object):
                 server_address = tuple(endpoint_parts)
             else:
                 server_address = (cfg.endpoint, "443")
-
             cert = ssl.get_server_certificate(server_address)
             x509 = OpenSSL.crypto.load_certificate(OpenSSL.crypto.FILETYPE_PEM, cert)
             cn = x509.get_subject().CN
             credentials = grpc.ssl_channel_credentials(str.encode(cert))
             options = kwargs.get("options", [])
-            options.append(("grpc.ssl_target_name_override", cn))
             self._channel = grpc.secure_channel(
                 target=cfg.endpoint,
                 credentials=credentials,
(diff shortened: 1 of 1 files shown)


brucearctor:
Great if that fixes this self-signed-cert/opta-setup issue….  But will this break others’ flows ( is that that override needed for others, ex: @RobertoRRW  who wrote the [override](https://github.com/flyteorg/flytekit/commit/065b8995c6246c55f8d1ba6fd2b74ee7c26314e9#diff-54d6ca794a724ff2e2b60e6dee730cb73865428d81f9e23c68e46e8fedf3ded6R128) in the first place )?

brucearctor (approved):
LGTM

flytekit/clients/raw.py
```diff
@@ -119,13 +119,9 @@ def __init__(self, cfg: PlatformConfig, **kwargs):
                 server_address = tuple(endpoint_parts)
             else:
                 server_address = (cfg.endpoint, "443")
-
```
brucearctor, line 122: what's style guide say about this?

wild-endeavor: I just `make fmt` and what it doesn't change is not regulated I guess.

eapolinario (approved):

flytekit/clients/raw.py
brucearctor: Is there a published flyte styleguide though?  Yes, if `fmt` addresses that's great -- also good if I can write code that the formatter doesn't need to change :-)

eapolinario (approved):

flytekit/clients/raw.py
wild-endeavor: No there is not unf.  Should we consider adding one?  I don't think I'd personally mind as long as it's a guide and not a strict set of rules.

brucearctor: Suggesting that the guide would be the things that the formatter/linter would test for [ and reject if not adhered to ].  Anything more than that is getting too nitpicky.  I don't recall whether the project is using things like spotless, black/yapf, etc.  But, effectively a writeup of the things that these things enforce [ so people don't need to know/look-at the configurations of such tools ].

wild-endeavor: oh yeah - `make fmt` just calls black.  and isort i think

PR #1126, opened 2022-08-05
Use joblib hashing to generate cache key to ensure repeatability

wild-endeavor:
# TL;DR
Thanks to @bethebunny for finding/fixing this.

When the type of a literal is a `dict` or `Dict` (i.e. no type specified for the value), the type engine converts it to a protobuf struct.  If the protobuf struct contains more than one field, then the cache key generated under the old code which ran base64 encoded a cloudpickle dump would sometimes change between calls, which led to local caching not working.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


wild-endeavor:
nm ignore this for now.  need to add more to this.

wild-endeavor:
@eapolinario can you take a look at this when you get a chance please?  I don't understand the unit tests that are failing and can't find any references to explain what's happening.  Why do we not want ints to cache?

eapolinario (approved):

PR #1131, opened 2022-08-12
pyflyte run imperative workflows

pingsutw:
# TL;DR
support running imperative workflow by using pyflyte run.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2706

## Follow-up issue
_NA_


wild-endeavor:
will take a look tomorrow

flytekit/tools/script_mode.py
```diff
@@ -97,16 +97,14 @@ def tar_strip_file_attributes(tar_info: tarfile.TarInfo) -> tarfile.TarInfo:
 
 
 def fast_register_single_script(
-    wf_entity: WorkflowBase, create_upload_location_fn: typing.Callable
+    source_path: str, module_name: str, create_upload_location_fn: typing.Callable
 ) -> (_data_proxy_pb2.CreateUploadLocationResponse, bytes):
-    _, mod_name, _, script_full_path = extract_task_module(wf_entity)
```
pingsutw, line 102: We can't get the module name of the imperative workflow by using [inspect.getmodule](https://github.com/flyteorg/flytekit/blob/4f44056bc76e372ac23ce8d388c5be4144ff4b97/flyte) because `inspect.getmodule(<imperative_wf>).__name__` is always equal to `flytekit.core.workflow`

To address this issue, we can use the module name passed in the pyflyte run instead. 
`pyflyte run --remote imperative_wf.py (model_name) wf

wild-endeavor (approved):
yeah sounds good, but can we make it like pyflyte register and pyflyte package in the future?  can we make another issue for that?  That will get us around this problem and maybe we can revert this change in the future.

you tested this in a subfolder right?  like if you have `a/b/c/wf.py` running `pyflyte run` in the c/ folder, still gets you a.b.c.wf as the name.

pingsutw:
yes, It's working in the subfolder as well. I've tested it.

> can we make it like pyflyte register and pyflyte package in the future?

Not sure what you mean. The reason why it only doesn't work in the pyflyte run is that `mod_name` in the [compress_single_script](https://github.com/flyteorg/flytekit/blob/795fd779886a07a38e16fb4b2692f818eaefd230/flytekit/tools/script_mode.py#L109) is generated from `extract_task_module`, but `extract_task_module` doesn't work for the imperative workflow. Therefore, we have to use the module name passed from the pyflyte run command.

On the other hand, the pyflyte register uses [fast_package](https://github.com/flyteorg/flytekit/blob/3137609c6d838a92077b7be27fc92cef78144066/flytekit/clis/sdk_in_container/register.py#L145), which doesn't require module name. That's why the pyflyte register works now.

wild-endeavor (approved):

PR #1134, opened 2022-08-16
Allow None protocol to mean all data persistence supported storage options in Structured Dataset

wild-endeavor:
# TL;DR
Contributors of dataframe handling plugins use the `StructuredDatasetTransformerEngine.register` method to make the system aware of specific encoders and decoders.  Up until now, the user was asked to specify the "protocol" associated with each handler.

This is awkward for two reasons:
* If your encoder/decoder just relies on the base data persistence layer flytekit provides, there is no reason to ask the contributor to register once per protocol.  If we add a new protocol to the persistence layer, then someone would have to go through and update all these.
* Registering multiple protocols while having the default_for_type option in the signature makes it awkward.  The impression is that you have to somehow decide a priori which protocol should be the default.  This doesn't make sense - the protocol to pick should default to the storage layer determined by the raw output prefix instead.

We are:
* making the `protocol` field optional now.
* when `None`, all protocols that the data persistence layer knows about will be registered with that handler.
* default_for_type will now default to False.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2755

### DIFF
--- a/flytekit/types/structured/structured_dataset.py
+++ b/flytekit/types/structured/structured_dataset.py
@@ -9,7 +9,7 @@ import typing
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
 from typing import Dict, Generator, Optional, Type, Union
-
+from flytekit.core.data_persistence import DataPersistencePlugins
 import _datetime
 import numpy as _np
 import pandas
@@ -168,7 +168,7 @@ def extract_cols_and_format(
 
 
 class StructuredDatasetEncoder(ABC):
-    def __init__(self, python_type: Type[T], protocol: str, supported_format: Optional[str] = None):
+    def __init__(self, python_type: Type[T], protocol: Optional[str] = None, supported_format: Optional[str] = None):
         """
         Extend this abstract class, implement the encode function, and register your concrete class with the
         StructuredDatasetTransformerEngine class in order for the core flytekit type engine to handle
@@ -184,7 +184,7 @@ class StructuredDatasetEncoder(ABC):
           for does not exist, the transformer enginer will look for the "" endcoder instead and write a warning.
         """
         self._python_type = python_type
-        self._protocol = protocol.replace("://", "")
+        self._protocol = protocol.replace("://", "") if protocol else None
         self._supported_format = supported_format or ""
 
     @property
@@ -192,7 +192,7 @@ class StructuredDatasetEncoder(ABC):
         return self._python_type
 
     @property
-    def protocol(self) -> str:
+    def protocol(self) -> Optional[str]:
         return self._protocol
 
     @property
@@ -228,7 +228,7 @@ class StructuredDatasetEncoder(ABC):
 
 
 class StructuredDatasetDecoder(ABC):
-    def __init__(self, python_type: Type[DF], protocol: str, supported_format: Optional[str] = None):
+    def __init__(self, python_type: Type[DF], protocol: Optional[str] = None, supported_format: Optional[str] = None):
         """
         Extend this abstract class, implement the decode function, and register your concrete class with the
         StructuredDatasetTransformerEngine class in order for the core flytekit type engine to handle
@@ -243,7 +243,7 @@ class StructuredDatasetDecoder(ABC):
           for does not exist, the transformer enginer will look for the "" decoder instead and write a warning.
         """
         self._python_type = python_type
-        self._protocol = protocol.replace("://", "")
+        self._protocol = protocol.replace("://", "") if protocol else None
         self._supported_format = supported_format or ""
 
     @property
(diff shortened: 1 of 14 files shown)


tests/flytekit/unit/core/test_data_persistence.py
```diff
@@ -14,3 +14,8 @@ def test_is_remote():
     assert fp.is_remote("/tmp/foo/bar") is False
     assert fp.is_remote("file://foo/bar") is False
     assert fp.is_remote("s3://my-bucket/foo/bar") is True
+
+
+def test_lister():
+    x = DataPersistencePlugins.supported_protocols()
+    assert [y.replace("://", "") for y in x] == ["file", "/", "gs", "http", "https", "s3"]
```
wild-endeavor, line 21: @pingsutw can you help me take a look at this test?  It's failing when I have the fsspec plugin installed because the fsspec plugin registers itself with all these protocols:
```
['file',
 '/',
 'gs',
 'http',
 'https',
 's3',
 'memory',
 'dropbox',
 'zip',
 'tar',
 'gcs',
 'gdrive',
 'sftp',
 'ssh',
 'ftp',
 'hdfs',
 'arrow_hdfs',
 'webhdfs',
 's3a',
 'wandb',
 'oci',
 'adl',
 'abfs',
 'az',
 'cached',
 'blockcache',
 'filecache',
 'simplecache',
 'dask',
 'dbfs',
 'github',
 'git',
 'smb',
 'jupyter',
 'jlab',
 'libarchive',
 'reference',
 'generic',
 'oss',
 'webdav']
```
Does it really handle all that?!  There's no way right?

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/__init__.py
```diff
@@ -25,13 +25,15 @@
...
 
 from flytekit import StructuredDatasetTransformerEngine, logger
-from flytekit.configuration import internal
-from flytekit.types.structured.structured_dataset import ABFS, GCS, S3
 
 from .arrow import ArrowToParquetEncodingHandler, ParquetToArrowDecodingHandler
 from .pandas import PandasToParquetEncodingHandler, ParquetToPandasDecodingHandler
 from .persist import FSSpecPersistence
 
+S3 = "s3"
```
pingsutw, line 33: We can import this from https://github.com/flyteorg/flytekit/blob/31277351c831b195951f1bc1152a4d08338e5633/flytekit/types/structured/structured_dataset.py#L41-L43

tests/flytekit/unit/core/test_data_persistence.py
pingsutw: I think Fsspec can handle all of [them](https://filesystem-spec.readthedocs.io/en/latest/api.html#built-in-implementations). 
Could we change it to
`assert {"file", "/", "gs", "http", "https", "s3"}.issubset(set([y.replace("://", "") for y in x]))`

plugins/flytekit-data-fsspec/flytekitplugins/fsspec/__init__.py
wild-endeavor: oh crap, no i'm deleting these.

wild-endeavor: structured dataset should not know about protocols at all really.  deleted.

flytekit/types/structured/structured_dataset.py
```diff
@@ -388,22 +382,67 @@ def __init__(self):
...
+            for persistence_protocol in DataPersistencePlugins.supported_protocols():
+                # TODO: Clean this up when we get to replacing the persistence layer.
+                # The behavior of the protocols given in the supported_protocols and is_supported_protocol
+                # is not actually the same as the one returned in get_protocol.
+                stripped = DataPersistencePlugins.get_protocol(persistence_protocol)
+                logger.debug(f"Automatically registering {persistence_protocol} as {stripped} with {h}")
+                try:
+                    cls.register_for_protocol(h, stripped, False, override)
+                except DuplicateHandlerError:
+                    ...
```
pingsutw, line 415: LGTM. nit: should we add debug message here?

pingsutw (dismissed):

flytekit/types/structured/structured_dataset.py
wild-endeavor: yeah.  i will.  will need another +1 later, will get from @eapolinario .  need to add more tests.

wild-endeavor:
@esadler-hbo this is the pr i'm talking about btw... hoping to get it merged by eod.

pingsutw (dismissed):

eapolinario (approved):

PR #1138, opened 2022-08-19
Require docker<7.0.0 in setup.py

rahul-theorem:
# TL;DR
Bumps the max version of `docker` in flytekit to `7.0.0`.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Upgrades the max version of `docker` in `flytekit` to `7.0.0`.

## Tracking Issue
N/A

## Follow-up issue
N/A


wild-endeavor:
Thanks!  Would you mind also `make requirements` with python3.7?  (until we have more time to split things up, we're running requirements with the minimum supported python version).  Did you bump this cuz it was causing a conflict somewhere else?

eapolinario:
@wild-endeavor , original context: https://flyte-org.slack.com/archives/CREL4QVAQ/p1660922404113919. We should make this less

wild-endeavor:
yea that sounds good.  we're not using it for much really doesn't need to be that strict.

wild-endeavor:
@rahul-theorem would you mind updating the pr?

rahul-theorem:
Hey @wild-endeavor updating this right now, just needed to switch my python env to 3.7

wild-endeavor:
yeah could you try that?  that is weird.  and you're running with python 3.7 right?  i think we still rely on that.  if a clean venv is still causing downgrades let us know.  thanks!

wild-endeavor (approved):

wild-endeavor:
actually holding off... will try to run this later on a new venv

wild-endeavor:
oh it looks like all the downgrades are happening in the doc requirements file.  i think that's okay.  it looks like the previous one was generated with 3.9 and now we're using 3.7.  i think that's fine. we can bump again separately if needed.  3.9 is probably okay for docs though. cc @eapolinario 

merging.

PR #1144, opened 2022-08-29
Add literal type to union literal

pingsutw:

# TL;DR
https://github.com/flyteorg/flyte/issues/2817

```python=
class Color(Enum):
    RED = "red"
    GREEN = "green"
    BLUE = "blue"


@task
def coalesce(a:  Union[List[str], str, Color]) -> str:
    if type(a) == Color:
        print(a.name)
        return a.name
    print(a)
    return a


@workflow
def wf(a:  Union[List[str], str, Color]) -> str:
    coalesced = coalesce(a=a)
    return coalesced


if __name__ == "__main__":
    wf(a=Color("red"))
```
Take the above code as an example. If we run `pyflyte run --remote example_union.py wf --a "flyte"`, we'll get the below error.

```bash
Request failed with status code 400 invalid a input wrong type, expected union_type:<variants:<collection_type:
<simple:STRING > structure:<tag:"Typed List" > > variants:<simple:STRING structure:<tag:"str" > > variants:<enum_type:
<values:"red" values:"green" values:"blue" > structure:<tag:"DefaultEnumTransformer" > > > , got simple:STRING instead
```
That's because literal(string) can be converted to str or enum, which is [ambiguous](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/validators/typing.go#L214-L222). To address this, we should add [literal type](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/core/literals.proto#L58) in Union type, and it can force [validator](https://github.com/flyteorg/flytepropeller/blob/0301f2a216001bdb988eef4ae58293f525076ad3/pkg/compiler/validators/typing.go#L33-L35) to cast "flyte" to string 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -264,7 +264,6 @@ class FlyteLiteralConverter(object):
                 # and then use flyte converter to convert it to literal.
                 python_val = converter._click_type.convert(value, param, ctx)
                 literal = converter.convert_to_literal(ctx, param, python_val)
-                self._python_type = python_type
                 return Literal(scalar=Scalar(union=Union(literal, variant)))
             except (Exception or AttributeError) as e:
                 logging.debug(f"Failed to convert python type {python_type} to literal type {variant}", e)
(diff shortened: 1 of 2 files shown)


eapolinario:
Can you add a unit test?

eapolinario (dismissed):

wild-endeavor:
this is dependent on the propeller pr going out first right?

wild-endeavor (approved):

wild-endeavor (approved):

PR #1135, opened 2022-08-16
Fix the type of optional[int] in dataclass

pingsutw:
Signed-off-by: Kevin Su <pingsutw@apache.org>

# TL;DR
https://flyte-org.slack.com/archives/C037MMU74FN/p1660670745909239

```
@dataclass
class MyDataclass(DataClassJsonMixin):
    value: Optional[int] = None

my_dataclass = MyDataclass(value=1)
```

`my_dataclass.value` appears to be converted to float and becomes 1.0 instead of remaining int and 1 as I expected.

Explicitly convert the `value` to int as well.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2775

## Follow-up issue
_NA_

PR #1148, opened 2022-09-06
Fix the type of optional[int] in nested dataclass

pingsutw:

# TL;DR
The value of `optional[int]` in nested dataclass is converted to float. https://github.com/flyteorg/flytekit/pull/1135 doesn't handle `optional[int]` in the nested dataclasses

In this pr, forced cast the value of `optional[int]` in the dataclass or nested databases to an integer.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
The code to reproduce
```python
from typing import Optional


@dataclass_json
@dataclass
class Args(object):
    max_iter: int = 0
    init_eval: int = 0
    max_eval: int = 0


@dataclass_json
@dataclass
class Schema(object):
    kwargs: Optional[Args] = None

@task
def t1(schema: Schema):
    print(schema)


@workflow
def wf(schema: Schema = Schema(Args(1, 2, 3))):
    t1(schema=schema)


if __name__ == "__main__":
    wf(schema=Schema(Args(1, 2, 3)))
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2839

## Follow-up issue
_NA_


flytekit/core/type_engine.py
```diff
@@ -463,6 +463,10 @@ def _fix_val_int(self, t: typing.Type, val: typing.Any) -> typing.Any:
             # Handle nested Dict. e.g. {1: {2: 3}, 4: {5: 6}})
             return {self._fix_val_int(ktype, k): self._fix_val_int(vtype, v) for k, v in val.items()}
 
+        if get_origin(t) is typing.Union and type(None) in get_args(t):
+            # Handle optional type. e.g. Optional[int], Optional[dataclass]
+            return self._fix_val_int(get_args(t)[0], val)
```
wild-endeavor, line 468: Does this work if the user doesn't do Optional, but instead does `typing.Union[int, None]`.

Also unrelated, can we add a unit test to document what happens in the case of `typing.Union[int, float]` (where the value is like 5 or something, something that's an int?

pingsutw: Marshmellow doesn't support union type. https://github.com/marshmallow-code/marshmallow/issues/1191
Therefore, we can't use `typing.Union[int, float]` in dataclass. btw, `Union[None, int]` is also an optional type, but Marshmallow does not support it.

will also update the doc of the dataclass example in flytesnack.

wild-endeavor (approved):

PR #1149, opened 2022-09-06
Add assert_type in dataclass transformer

pingsutw:

# TL;DR
[slack thread](https://flyte-org.slack.com/archives/CREL4QVAQ/p1661266231263149)

Add `assert_type()`in dataclass transformer, so we don't need to specify `type_hints` when using flyte remote.

We got an error when using dataclass as input in the flyte remote without passing `type_hints`, because `guess_python_type` is not equal to the user-defined dataclass type

```
@dataclass_json
@dataclass
class Foo(object):
    a: int = 0

@task
def t1(a: Foo):
    ...
```
In this example, we use guess_python_type (FooSchema) as our `expected_type` by default if using flyte remote to execute the above task.
However, FooSchema is created by flytekit and it's not equal to the user-defined dataclass (Foo), so [assert_type](https://github.com/flyteorg/flytekit/blob/aff19cb8bd59c5376cd638cef4ee39f05786a3a8/flytekit/core/type_engine.py#L95) raises an error.

We can override `assert_type` in the dataclass transformer, and check that the type of each attribute is the same as the `guess_type` by iterating all attributes in dataclass.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

The code to reproduce

```python
from typing import Optional


@dataclass_json
@dataclass
class Args(object):
    max_iter: int = 0
    init_eval: float = 0
    max_eval: str = 0


@dataclass_json
@dataclass
class Schema(object):
    kwargs: Optional[Args] = None


@task
def t11(schema: Schema):
    print(schema)


@workflow
def wf3(schema: Schema = Schema(Args(1, 2.0, "hello"))):
    t11(schema=schema)


if __name__ == "__main__":
    wf3(schema=Schema(Args(1, 2.0, "hello")))
```

```python
from example_dataclass import Args, Schema

remote = FlyteRemote(
    config=Config.auto("/Users/kevin/.flyte/config-remote.yaml"),
    default_domain="development",
)

flyte_workflow = remote.fetch_workflow(
    name="example_dataclass.wf3", project="flytesnacks", domain="development"
)

execution = remote.execute(
    flyte_workflow, inputs={"schema": Schema(Args(1, 2, "hello"))}, wait=False,
)
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2844

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -282,9 +282,9 @@ class DataclassTransformer(TypeTransformer[object]):
         # def t1(a: Foo):
         #     ...
         #
-        # we use guess_python_type (FooSchema) as our expected_type by default if using flyte remote to execute the above workflow.
-        # FooSchema is created by flytekit and it's not equal to the user-defined dataclass (Foo). Therefore, we can't
-        # use `assert type(v) == expected_type` here.
+        # we use guess_python_type (FooSchema) as our expected_type by default if using flyte remote to execute the above task.
+        # FooSchema is created by flytekit and it's not equal to the user-defined dataclass (Foo).
+        # Therefore, we can't use `assert type(v) == expected_type` here.
 
         expected_fields_dict = {}
         for f in dataclasses.fields(expected_type):
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -271,7 +271,7 @@ class DataclassTransformer(TypeTransformer[object]):
         super().__init__("Object-Dataclass-Transformer", object)
 
     def assert_type(self, expected_type: Type[DataClassJsonMixin], v: T):
-        # Skip iterating all attributes in the dataclasses if the type of v already matches the expected_type
+        # Skip iterating all attributes in the dataclass if the type of v already matches the expected_type
         if type(v) == expected_type:
             return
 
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
```diff
@@ -270,6 +270,46 @@ class Test():
...
+            if UnionTransformer.is_optional_type(original_type):
+                original_type = UnionTransformer.get_sub_type_in_optional(original_type)
+            if UnionTransformer.is_optional_type(expected_type):
+                expected_type = UnionTransformer.get_sub_type_in_optional(expected_type)
+
+            val = v.__getattribute__(f.name)
+            if dataclasses.is_dataclass(val):
+                self.assert_type(expected_type, val)
+            elif original_type != expected_type:
+                raise TypeTransformerFailedError(f"Type of Val '{original_type}' is not an instance of {expected_type}")
```
wild-endeavor, line 311: is it easy to add a test for this line?

wild-endeavor (approved):
is there really nothing in dataclass json or json schema that will do this? we have to write our own code?

flytekit/core/type_engine.py
pingsutw: Updated it. 
I've tried to find a kind of function that can compare the attribute in dataclass, but I didn't find that.

wild-endeavor (approved):

PR #1151, opened 2022-09-08
Added symlink dereferencing in fast packaging and tests

vchowdhary:
# TL;DR
Adds a switch to enable dereferencing symlinks during fast packaging in `pyflyte serialize fast workflows`.
cc @wild-endeavor 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added
 - [x] Code documentation added --- added documentation for click switch

## Complete description
Added a click switch to `pyflyte serialize fast workflows` that enables/disables dereferencing any symlinks encountered during packaging. This only applies to `fast_package`. 

## Tracking Issue
_NA_

No issue associated with this -- this is part of a request from Stripe. Our use case for this is as follows:
Our workflows often import code from utils files that are in sibling directories -- this makes the current `pyflyte package` not quite work for us. Our workaround is to use our bazel build target, which generates a `runfiles` directory that contains all the source code and dependencies, and then serialize from that folder. Some symlinks are generated during the build, so we needed to add `dereference=True` when opening the tar file so that the actual target contents are loaded and downloaded during execution. 

This change should not effect any of the current DataPersistence plugins that are used since they are all remote and would require dereferencing symlinks anyways if they wanted to use the contents. 

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/clis/sdk_in_container/package.py
+++ b/flytekit/clis/sdk_in_container/package.py
@@ -77,8 +77,16 @@ from flytekit.tools.repo import NoSerializableEntitiesError, serialize_and_packa
     default="/root",
     help="Filesystem path to where the code is copied into within the Dockerfile. look for `COPY . /root` like command.",
 )
+@click.option(
+    "--deref-symlinks",
+    default=False,
+    is_flag=True,
+    help="Enables symlink dereferencing when packaging files in fast registration",
+)
 @click.pass_context
-def package(ctx, image_config, source, output, force, fast, in_container_source_path, python_interpreter):
+def package(
+    ctx, image_config, source, output, force, fast, in_container_source_path, python_interpreter, deref_symlinks
+):
     """
     This command produces a Flyte backend registrable package of all entities in Flyte.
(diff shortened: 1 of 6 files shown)


eapolinario (commented):
This change looks good. Can you also make sure that new flag is added to the other places where `fast_package` is used?

vchowdhary:
> This change looks good. Can you also make sure that new flag is added to the other places where `fast_package` is used?

Thanks for catching that! There was one place that I missed.

eapolinario:
What about https://github.com/flyteorg/flytekit/blob/3137609c6d838a92077b7be27fc92cef78144066/flytekit/tools/repo.py#L98 ?

vchowdhary:
Ah my bad, missed that.

eapolinario (approved):

vchowdhary:
@eapolinario  
Any idea on when this would be released? We see that there are [monthly releases usually](https://pypi.org/project/flytekit/#history) and one hasn't happened yet for september.

PR #1157, opened 2022-09-12
Fixing structured dataset for empty bq tables

dylanwilder:
# TL;DR
Structured dataset bq read fails on empty table because `pandas.concat()` fails if the passed array is empty.  Fix it to create an empty dataframe with the required schema

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/types/structured/bigquery.py
+++ b/flytekit/types/structured/bigquery.py
@@ -51,7 +51,14 @@ def _read_from_bq(
     frames = []
     for message in reader.rows().pages:
         frames.append(message.to_dataframe())
-    return pd.concat(frames)
+    if len(frames) > 0:
+        df = pd.concat(frames)
+    else:
+        schema = pa.ipc.read_schema(
+            pa.py_buffer(read_session.arrow_schema.serialized_schema)
+        )
+        df = schema.empty_table().to_pandas()
+    return df
 
 
 class PandasToBQEncodingHandlers(StructuredDatasetEncoder):


eapolinario (commented):
Can you add a unit-test for the empty case [here](https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/types/structured_dataset/test_bigquery.py)?

kumare3 (approved):

flytekit/types/structured/bigquery.py
```diff
@@ -51,7 +51,14 @@ def _read_from_bq(
     frames = []
     for message in reader.rows().pages:
         frames.append(message.to_dataframe())
-    return pd.concat(frames)
+    if len(frames) > 0:
+        df = pd.concat(frames)
```
pingsutw, line 55: If the table is empty, the length of `read_session.streams` is 0. so I think we should update the code to below.
Let me know if it doesn't work for you
```suggestion
    frames = []
    if len(read_session.streams) > 0:
        stream = read_session.streams[0]
        reader = client.read_rows(stream.name)
        for message in reader.rows().pages:
            frames.append(message.to_dataframe())

    if len(frames) > 0:
        df = pd.concat(frames)
```

wild-endeavor: @dylanwilder could you take a look at this snippet please? Does this work for you?

gitgraghu: @wild-endeavor I tested this snippet. Seems like there will be a stream even for empty tables. So this does not work.

PR #1159, opened 2022-09-13
Set default format of structured dataset to empty

pingsutw:
# Compatibility Note
This PR partly relies on https://github.com/flyteorg/flytepropeller/commit/69158eab688dd9c44423a2b1ff47f65d6ac44efe which was merged into propeller v1.1.36 and Flyte v1.2.  If you are running an older version of propeller, you will experience cache-misses with this PR for cached tasks.  Please see the tl;dr for more info, but essentially since the format in the signatures of tasks with dataframes in their inputs or outputs will now be `""`, if you don't upgrade, propeller will think that the signature has changed and result in a data catalog cache-miss.

# TL;DR
https://github.com/flyteorg/flyte/issues/2864

The current default format type is `Parquet,` and it leads to some issues

1. SD transformer will always convert dataframe to parquet instead of [cls.DEFAULT_FORMATS[df_type]](https://github.com/flyteorg/flytekit/blob/28ffc6346fd87c19f15d922e654280146ab3a40c/flytekit/types/structured/structured_dataset.py#L332)
```python
def t1() -> StructuredDataset: # The default format of structured dataset is Parquet here
```

2. Failed to run BQ task if the cache is enabled because [type validation](https://github.com/flyteorg/flytepropeller/blob/9cf5368793e5d60113cb819a6e496a7b6accd619/pkg/compiler/validators/typing.go#L157) is failing.
```python
def t1() -> StructuredDataset: # The default format of structured dataset is Parquet here
    df = pd.DataFrame({"len": [len(sd.open(pd.DataFrame).all())]})
    return StructuredDataset(df, uri=bq_uri) # The format of structured dataset is "" 
```

In this PR, we set the default format to ""
```python
def t1() -> StructuredDataset: # The default format of structured dataset is "" here
```

1. By default, we use `handlers[df_type][protocol][""]` to encode a dataframe. if `handlers[df_type][protocol][""]` doesn't exist, then use `handlers[df_type][protocol][cls.DEFAULT_FORMATS[df_type]]`.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2864


wild-endeavor:
We are going to hold off on this change for a while, need to get the propeller change out and people migrated first.  https://github.com/flyteorg/flytepropeller/pull/483

flytekit/types/structured/structured_dataset.py
```diff
@@ -334,16 +334,17 @@ class StructuredDatasetTransformerEngine(TypeTransformer[StructuredDataset]):
...
-    @staticmethod
-    def _finder(handler_map, df_type: Type, protocol: str, format: str):
+    @classmethod
+    def _finder(cls, handler_map, df_type: Type, protocol: str, format: str):
         try:
             return handler_map[df_type][protocol][format]
         except KeyError:
             try:
-                hh = handler_map[df_type][protocol][""]
+                default_format = cls.DEFAULT_FORMATS.get(df_type, PARQUET)
```
wild-endeavor, line 343: I don't fully understand this.  why search for parquet?  What happens if we leave it as it is now?

The thinking for all this was - if you request format X, and there's a handler for X then that's what is used.  If format X is not available, then check to see if there's a generic handler.  But I didn't have a good way of specifying what's "generic", so I just picked an empty string.

I think there's a difference between a generic handler, and a default format.  If a user specified a format of Y, but a handler for Y cannot be found, even if the default format is X, I don't think it should be used.  We should look to see if there's a generic (""), and if not then fail.

flytekit/types/structured/structured_dataset.py
```diff
@@ -762,17 +759,9 @@ def _get_dataset_type(self, t: typing.Union[Type[StructuredDataset], typing.Any]
         # Get the column information
         converted_cols = self._convert_ordered_dict_of_columns_to_list(column_map)
 
-        # Get the format
-        default_format = (
-            original_python_type.DEFAULT_FILE_FORMAT
-            if issubclass(original_python_type, StructuredDataset)
-            else self.DEFAULT_FORMATS.get(original_python_type, PARQUET)
```
wild-endeavor, line 769: shouldn't we still keep this?  The extract_cols_and_format call above, which returns the `storage_format` variable, will only look at Python `Annotate` annotations.  If the function signature specifies only a naked dataframe like (eg `pd.DataFrame`) wouldn't this now return `""`, even if there is a default format specified?

wild-endeavor (commented):
I think unf. there's some negative interaction with this [PR](https://github.com/flyteorg/flytekit/pull/1107) which removed the setting of the default for a lot of the types.  I think it's time to address the flaw introduced in that PR - there we took out the `default_for_type` switch because it was setting the default for both the format and the storage backend.  I think we should add two new args, `set_default_format` and `set_default_storage` or something like that.

wild-endeavor (approved):

PR #1161, opened 2022-09-13
Auto Display of Deck in Jupyter notebook

kumare3:

# TL;DR
Jupyter notebook can now display FlyteDecks natively

FlyteDecks are html objects. In IPython, its own HTML objects are natively display-able. This changes detects that the environment is a notebook and makes the object match the expectation of IPython

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added


flytekit/core/context_manager.py
```diff
@@ -617,7 +617,26 @@ def current_context() -> Optional[FlyteContext]:
         """
         return FlyteContextManager.current_context()
 
-    def get_deck(self) -> str:
+    def get_deck(self) -> typing.Union[str, "IPython.core.display.HTML"]:
```
eapolinario, line 620: this is not going to work in mypy, right? In other words, this is purely for documentation purposes.

flytekit/deck/deck.py
```diff
@@ -89,12 +90,17 @@ def _ipython_check() -> bool:
     return is_ipython
 
 
-def _get_deck(new_user_params: ExecutionParameters) -> str:
+def _get_deck(new_user_params: ExecutionParameters) -> typing.Union[str, "IPython.core.display.HTML"]:
```
eapolinario, line 93: ditto.

flytekit/deck/deck.py
```diff
@@ -89,12 +90,17 @@ def _ipython_check() -> bool:
...
-    return template.render(metadata=deck_map)
+    raw_html = template.render(metadata=deck_map)
+    if _ipython_check():
+        from IPython.core.display import HTML
+
+        return HTML(raw_html)
+    return raw_html
 
 
 def _output_deck(task_name: str, new_user_params: ExecutionParameters):
```
eapolinario, line 106: Should we special-case ipython in this case as well (as in "don't need to write the deck file in the jupyter notebook case")?

flytekit/core/context_manager.py
kumare3: yes

kumare3: should i delete it?

flytekit/deck/deck.py
kumare3: wdym?

flytekit/core/context_manager.py
eapolinario: At the very least we should leave a comment explaining that this is not mypy-compliant. 

How about we type-ignore this for now?
```suggestion
def _get_deck(
    new_user_params: ExecutionParameters,
) -> typing.Union[str, "IPython.core.display.HTML"]:  #type:ignore

```

eapolinario (approved):

PR #1221, opened 2022-10-05
Handle special characters in snowflake query

pingsutw:

# TL;DR
Related to https://flyte-org.slack.com/archives/CP2HDHKE1/p1664792800827789?thread_ts=1663843915.889099&cid=CP2HDHKE1.

[snowflake docs](https://docs.snowflake.com/en/sql-reference/data-types-text.html#dollar-quoted-string-constants)

Replace "\n" to "\\n" to fix the 400 error.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/base_sql_task.py
+++ b/flytekit/core/base_sql_task.py
@@ -41,7 +41,7 @@ class SQLTask(PythonTask[T]):
             task_config=task_config,
             **kwargs,
         )
-        self._query_template = query_template
+        self._query_template = query_template.replace("\n", "\\n").replace("\t", "\\t")
 
     @property
     def query_template(self) -> str:
--- a/plugins/flytekit-snowflake/tests/test_snowflake.py
+++ b/plugins/flytekit-snowflake/tests/test_snowflake.py
@@ -10,7 +10,7 @@ from flytekit.types.schema import FlyteSchema
 
 query_template = """
             insert overwrite directory '{{ .rawOutputDataPrefix }}' stored as parquet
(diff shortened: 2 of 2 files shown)


flytekit/core/base_sql_task.py
```diff
@@ -41,7 +41,7 @@ def __init__(
             task_config=task_config,
             **kwargs,
         )
-        self._query_template = query_template
+        self._query_template = query_template.replace("\n", "\\n").replace("\t", "\\t")
```
eapolinario, line 44: are there any other characters or escape sequences that are specific to snowflake? I tried to read https://docs.snowflake.com/en/sql-reference/data-types-text.html, but I'm still not sure if we have to handle some of those.

pingsutw: I feel like some of those can handle by users themself. if we triple quote like
```
 query = """select 1
 from table"""
```
python will convert it to `select 1\nfrom table`, so that's why we help them handle this special character in flytekit.

However, if they want to run something like `select $1, $2 from ('Backslash','C:\user');`, they are able to explicit convert it to `select $1, $2 from ('Backslash','C:\\user');` by themself to make it work in snowflake.

eapolinario (approved):

PR #1227, opened 2022-10-06
Enable use of image stanza in `pyflyte register` and other small changes

eapolinario:
# TL;DR
`pyflyte register` is now able to load images present in the config file. That sub-command also requires a version in non-fast case.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Changes to `pyflyte register`:
1. The output of the command now is more explicit about which serialization settings are used, which include the images allowed to be used.
2. Force version in non fast. The idea is to make it clear that this option requires the user to be deliberate about versions when registering entities using the non-fast option.
3. Allow for images present in a config file to be used during registration

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -137,6 +137,9 @@ def register(
     if pkgs:
         raise ValueError("Unimplemented, just specify pkgs like folder/files as args at the end of the command")
 
+    if non_fast and not version:
+        raise ValueError("Version is a required parameter in case --non-fast is specified.")
+
     if len(package_or_module) == 0:
         display_help_with_error(
             ctx,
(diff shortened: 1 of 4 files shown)


flytekit/clis/sdk_in_container/register.py
```diff
@@ -194,5 +202,9 @@ def register(
         version = remote._version_from_hash(md5_bytes, serialization_settings, service_account, raw_data_prefix)  # noqa
         cli_logger.info(f"Computed version is {version}")
 
+    cli_logger.warn(
```
wild-endeavor, line 205: can we use info or lower?  if this is something we want to print then click.echo?

wild-endeavor (approved):

flytekit/clis/sdk_in_container/register.py
eapolinario: this should be printed. I'm going to change it to `click.echo`.

wild-endeavor (approved):

PR #1236, opened 2022-10-13
Remove use TypeVar to denote the extension of a Flytefile #2870

jerempy:
# TL;DR
Do not use TypeVar to denote the extension of a Flytefile. Pylint was breaking in a couple of places. This fixes it.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Ran pylint and got the stackstrace. Fixed it and then ran again without the errors. details below.

`typing.TypeVar` requires that the name of the var it is assigned to is a match. meaning `T = TypeVar("T")` is ok but `C = TypeVar("T")` is not. So when we have a file extension like "tar.gz" there is no way to assign that as a `tar.gz` variable. So instead we are just switching it to str type.

I left the TypeVar on file extensions in https://github.com/flyteorg/flytekit/blob/master/flytekit/types/file/__init__.py#L28-L82 since they don't seem to break anything.

I searched the rest of the repo looking for any other regex matches for something like with a period in the file extesnion and didn't find any others.

I also corrected a return type for `split_traintest_dataset()` since my linter was warning me about it 

## Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/2870

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/types/directory/types.py
+++ b/flytekit/types/directory/types.py
@@ -15,6 +15,7 @@ from flytekit.models import types as _type_models
 from flytekit.models.core import types as _core_types
 from flytekit.models.literals import Blob, BlobMetadata, Literal, Scalar
 from flytekit.models.types import LiteralType
+from flytekit.types.file import FileExt
 
 T = typing.TypeVar("T")
 
@@ -144,7 +145,9 @@ class FlyteDirectory(os.PathLike, typing.Generic[T]):
     def __class_getitem__(cls, item: typing.Union[typing.Type, str]) -> typing.Type[FlyteDirectory]:
         if item is None:
             return cls
-        item_string = str(item)
+
+        item_string = FileExt.check_and_convert_to_str(item)
+
         item_string = item_string.strip().lstrip("~").lstrip(".")
         if item_string == "":
             return cls
--- a/flytekit/types/file/__init__.py
+++ b/flytekit/types/file/__init__.py
@@ -25,58 +25,77 @@ import typing
 
 from .file import FlyteFile
 
+
+class FileExt:
+    """Used for annotating types for file extensions for FlyteFile
+    This is usefule for extensions that have periods in them, i.e. "tar.gz"
+
+    Uses:
+    TAR_GZ = typing.Annotated[str, FileExt("tar.gz")]
+    """
+
+    def __init__(self, ext: str):
+        self._ext = ext
+
+    def __str__(self):
+        return self._ext
+
+    def __repr__(self):
+        return self._ext
+
+
 # The following section provides some predefined aliases for commonly used FlyteFile formats.
 # This makes their usage extremely simple for the users. Please keep the list sorted.
 
-hdf5 = typing.TypeVar("hdf5")
+hdf5 = typing.Annotated[str, FileExt("hdf5")]
 #: This can be used to denote that the returned file is of type hdf5 and can be received by other tasks that
 #: accept an hdf5 format. This is usually useful for serializing Tensorflow models
 HDF5EncodedFile = FlyteFile[hdf5]
 
-html = typing.TypeVar("html")
+html = typing.Annotated[str, FileExt("html")]
 #: Can be used to receive or return an HTMLPage. The underlying type is a FlyteFile type. This is just a
 #: decoration and useful for attaching content type information with the file and automatically documenting code.
 HTMLPage = FlyteFile[html]
 
-joblib = typing.TypeVar("joblib")
+joblib = typing.Annotated[str, FileExt("joblib")]
 #: This File represents a file that was serialized using `joblib.dump` method can be loaded back using `joblib.load`.
 JoblibSerializedFile = FlyteFile[joblib]
 
-jpeg = typing.TypeVar("jpeg")
+jpeg = typing.Annotated[str, FileExt("jpeg")]
 #: Can be used to receive or return an JPEGImage. The underlying type is a FlyteFile type. This is just a
 #: decoration and useful for attaching content type information with the file and automatically documenting code.
 JPEGImageFile = FlyteFile[jpeg]
 
-pdf = typing.TypeVar("pdf")
+pdf = typing.Annotated[str, FileExt("pdf")]
 #: Can be used to receive or return an PDFFile. The underlying type is a FlyteFile type. This is just a
 #: decoration and useful for attaching content type information with the file and automatically documenting code.
 PDFFile = FlyteFile[pdf]
 
-png = typing.TypeVar("png")
+png = typing.Annotated[str, FileExt("png")]
 #: Can be used to receive or return an PNGImage. The underlying type is a FlyteFile type. This is just a
 #: decoration and useful for attaching content type information with the file and automatically documenting code.
 PNGImageFile = FlyteFile[png]
 
-python_pickle = typing.TypeVar("python_pickle")
+python_pickle = typing.Annotated[str, FileExt("python_pickle")]
 #: This type can be used when a serialized Python pickled object is returned and shared between tasks. This only
 #: adds metadata to the file in Flyte, but does not really carry any object information.
 PythonPickledFile = FlyteFile[python_pickle]
 
-ipynb = typing.TypeVar("ipynb")
+ipynb = typing.Annotated[str, FileExt("ipynb")]
 #: This type is used to identify a Python notebook file.
 PythonNotebook = FlyteFile[ipynb]
 
-svg = typing.TypeVar("svg")
+svg = typing.Annotated[str, FileExt("svg")]
 #: Can be used to receive or return an SVGImage. The underlying type is a FlyteFile type. This is just a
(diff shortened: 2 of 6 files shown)


samhita-alla:
@jerempy, thanks for creating the PR! We'll also need to fix the FlyteFile extensions cause they'll fail when you run `pylint` on a file that contains say, `FlyteFile[typing.TypeVar("...")]`. You can test [this example](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/files.html) by including `TypeVar` for your reference.

@eapolinario suggested using `typing.Annotated` rather than `TypeVar` for `FlyteFile` objects. Would you mind working on it? You may also need to take a look at https://github.com/flyteorg/flytekit/blob/master/flytekit/types/file/file.py file to support the `Annotated` case. Here's an example where we're using `Annotated` to accept the user metadata: https://github.com/flyteorg/flytekit/blob/ab9aa65c420afed1ec3b80e0d650fcee3d63852a/flytekit/types/numpy/ndarray.py#L16-L24

jerempy:
@samhita-alla thanks for the comment! I tried running pylint on files with `FlyteFile[typing.TypeVar("...")]` and it wasn't causing breakages. 

the original breakage was caused by assigning types to variables with different names - which does cause breakages with pylint. You cannot do `VAR = TypeVar("OTHERVAR")` pylint would break with that, but this is okay: `VAR = TypeVar("VAR")`. TypeVar is used 55 times in this repo, and I looked at all of them and they all look good to go.

The specific ones named here in the bug that broke pylint were assigned different names because they had a .(period) in their extenions name, i.e. `tar.gz` so they are unworkable.

I took a look at typing.Annotated, but per my understanding of it it doesnt seem like it would fir here since it is used to add metadata to types, rather than defining them.

At any rate, I fixed a typehint that broke the tests in CI - should pass CI now 🤞

pingsutw (approved):

samhita-alla:
@eapolinario, would you mind taking a look at this PR?

jerempy:
sorry @pingsutw if you could review again 🙏 :I noticed that I accidentally changed the return type in https://github.com/flyteorg/flytekit/blob/98acf1bd7a76f0b6f2d04cbeb87ac17b84b3836c/tests/flytekit/unit/core/test_realworld_examples.py#L87

so i reverted it back to featured_columns and class_columns

jerempy:
@pingsutw think we could merge this one? October is almost over 😅

pingsutw (dismissed):
@eapolinario I think we can merge first, and then update FlyteFile after this. btw, I didn't see any new mypy error after changing from `type` to `string`

eapolinario:
@pingsutw , how did you run mypy?

For example, given this file:
```
import typing

def f_no_extension() -> FlyteFile:
    ...

reveal_type(f_no_extension)

def f_string_extension() -> FlyteFile["ext1"]:
    ...

reveal_type(f_string_extension)

def f_typevar() -> FlyteFile[typing.TypeVar("ext2")]:
    ...

reveal_type(f_typevar)
```

mypy returns:
```
❯ mypy example.py
example.py:2: error: Cannot find implementation or library stub for module named "flytekit.types.file"
example.py:2: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
example.py:7: note: Revealed type is "def () -> Any"
example.py:9: error: Name "ext1" is not defined
example.py:12: note: Revealed type is "def () -> Any"
example.py:17: note: Revealed type is "def () -> Any"
Found 2 errors in 1 file (checked 1 source file)
```

Note how in the case of `f_string_extension` mypy return an error.

pingsutw:
sorry, my bad. I used `make lint`, and it didn't check those two files.

samhita-alla:
@jerempy, let's not rush things. I can assign a "hacktoberfest-accepted" label in case the PR doesn't get merged before Oct. 31. :)

Can you take a look at the above comments and do the necessary? Here's Eduardo's comment:

>I don't believe we should go ahead with this change.
>Simply replacing mentions to typevar with raw strings is going to start showing up as errors in mypy.

hajapy:
Given the discussion about how using a plain `str` isn't valid and causes issues with mypy, I think this solution is not truly an improvement over the status quo, where mypy is fine but pylint is unhappy. I think the idea should be to work out how `typing.Annotated` can be used instead of `typing.TypeVar`. Also, I am not sure whether there needs to be some deprecation schedule for the old way. If it were to simply stop working with `TypeVar`, that could be problematic. I'd suggest supporting `Annotated` as an extra case, and emitting deprecation warnings for uses of `TypeVar` or `str`, before eventually dropping support for those.

jerempy:
@samhita-alla thanks for that!
@eapolinario and @pingsutw  you guys able to take  a new look? I re-did the whole thing for `typing.Annotated`
It looks good to both mypy and pylint

eapolinario (dismissed):
This looks amazing! Thanks for persisting with this, @jerempy!

flytekit/types/file/file.py
```diff
@@ -151,6 +152,13 @@ def extension(cls) -> str:
     def __class_getitem__(cls, item: typing.Union[str, typing.Type]) -> typing.Type[FlyteFile]:
         if item is None:
             return cls
+
+        if typing.get_origin(item) is Annotated:
+            if typing.get_args(item)[0] == str:
+                item = typing.get_args(item)[1]
```
eapolinario, line 158: you should also import from `typing_extensions`.

flytekit/types/file/__init__.py
```diff
@@ -21,62 +21,81 @@
...
 
 from .file import FlyteFile
 
+
+class FileExt:
+    """Used for annotating types for file extensions for FlyteFile
+    This is usefule for extensions that have periods in them, i.e. "tar.gz"
+
+    Uses:
+    TAR_GZ = Annotated[str, FileExt("tar.gz")]
```
samhita-alla, line 34:
```suggestion
    """
    Used for annotating file extension types of FlyteFile.
    This is useful for extensions that have periods in them, e.g., "tar.gz".

    Example:
    TAR_GZ = Annotated[str, FileExt("tar.gz")]
```

samhita-alla:
@jerempy, two more changes to go! :)

hajapy:
Should the project dependencies change  to declare dep on typing_extensions?

Should we only use this for python <3.9 and use import logic like trying to import Annotated from typing and if that fails fall back to typing_extensions?

How about docs and or release notes for these changes? It's a user visible change in expectation.

flytekit/types/file/__init__.py
jerempy: cool! changed

flytekit/types/directory/types.py
```diff
@@ -144,7 +145,9 @@ def extension(cls) -> str:
     def __class_getitem__(cls, item: typing.Union[typing.Type, str]) -> typing.Type[FlyteDirectory]:
         if item is None:
             return cls
-        item_string = str(item)
+
+        item_string = FileExt.check_and_convert_to_str(item)
```
jerempy, line 149: @eapolinario this is what I'm refering to, since the FlyteDirectory takes in `Annotated[str, FileExt()]`

eapolinario:
> ~Should the project dependencies change to declare dep on typing_extensions?~
> 
> ~Should we only use this for python <3.9 and use import logic like trying to import Annotated from typing and if that fails fall back to typing_extensions?~
> 
> Edit: Ah, flytekit already requires typing_extensions regardless of python version.
> 
> How about docs and or release notes for these changes? It's a user visible change in expectation.

That's a good call and it's exactly how `typing_extensions` is supposed to work, i.e. if you're in in python >=3.9 importing `Annotated` from `typing` or `typing_extensions` is ok, whereas if you're using python <3.9 you're going to get an error if you try to use `typing.Annotated`. In other words, `typing_extensions` offers a type alias for the same underlying type if it exists.

That said, we're going to make it clear that `FlyteFile` and `FlyteDirectory` now can use this new mode of setting extensions in the release notes. Also worth noting that this change is backwards compatible (as long as the extension doesn't change), for example:
```
@task
def t_return_old_style_flytefile() -> FlyteFile[TypeVar("ext")]:
   ...

@task 
def t_takes_new_flytefile(f: Annotated[FlyteFile, FileExt("ext")]):
   ...

@workflow
def wf():
    f =  t_return_old_style_flytefile()
    t_takes_new_flytefile(f=f)
```

eapolinario:
@jerempy , can you fix the sagemaker test? Same error (`Annotated` being imported from `typing`)

jerempy:
@eapolinario woops! missed that one. got it now. should be good to go 😎

jerempy:
happy halloween! Looks like missing 7% diff coverage, based on lines of code - but when I look at the PR I;d say everything is adequately covered - if it fine to merge anyway?

eapolinario (approved):
Thanks a lot for this contribution, @jerempy !

PR #1237, opened 2022-10-14
Improve pyflyte register

kumare3:
# TL;DR
this PR improves the code behind pyflyte register accomplishing, 
- faster performance
- better visualization
- simplified logic with a clean flow
- all programmatic APIs, to invoke programmatically

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
pyflyte register used legacy methods and some techniques that are not required if you are within the python VM. For example, since we have SerializationSettings, it is not necessary to re-hydrate the objects simply to replace the identifiers. Moreover, the logic is further simplified and it is possible to perform repo level registration using a simple api and/or through flyteremote.

TODO: UnionML can now use this lower level api


wild-endeavor (approved):
I think this is fine.  Just looked at the PR though.  Maybe give this another day or two so I can try to pull this and test more?  Or we cut a beta release right away and test from there.  what do you think @eapolinario ?

wild-endeavor (approved):

PR #1238, opened 2022-10-17
Set reasonable defaults for pandas dataframe default renderer

eapolinario:
# TL;DR
pandas dataframe default renderer should be cheap to render

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The [default deck renderer for pandas dataframes](https://github.com/flyteorg/flytekit/blob/master/flytekit/deck/renderer.py#L27) relies on pandas dataframe [to_html](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.to_html.html), but big dataframes take a long time to return, so in this PR we set reasonable defaults for both the max number of rows and columns.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/deck/renderer.py
+++ b/flytekit/deck/renderer.py
@@ -14,17 +14,22 @@ class Renderable(Protocol):
         raise NotImplementedError
 
 
+DEFAULT_MAX_ROWS = 10
+DEFAULT_MAX_COLS = 100
+
+
 class TopFrameRenderer:
     """
     Render a DataFrame as an HTML table.
     """
 
-    def __init__(self, max_rows: Optional[int] = None):
+    def __init__(self, max_rows: Optional[int] = DEFAULT_MAX_ROWS, max_cols: Optional[int] = DEFAULT_MAX_COLS):
         self._max_rows = max_rows
+        self._max_cols = max_cols
 
     def to_html(self, df: pandas.DataFrame) -> str:
         assert isinstance(df, pandas.DataFrame)
-        return df.to_html(max_rows=self._max_rows)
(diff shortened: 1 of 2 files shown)


pingsutw (dismissed):

flytekit/deck/renderer.py
```diff
@@ -14,17 +14,22 @@ def to_html(self, python_value: Any) -> str:
...
+DEFAULT_MAX_COLS = 100
+
+
 class TopFrameRenderer:
     """
     Render a DataFrame as an HTML table.
     """
 
-    def __init__(self, max_rows: Optional[int] = None):
+    def __init__(self, max_rows: Optional[int] = DEFAULT_MAX_ROWS, max_cols: Optional[int] = DEFAULT_MAX_COLS):
```
wild-endeavor, line 26: not optional if you give a default.
```suggestion
    def __init__(self, max_rows: int = DEFAULT_MAX_ROWS, max_cols: int = DEFAULT_MAX_COLS):
```

wild-endeavor: right?  or not?  i never figured out how this works?  i think it's from the perspective of the body of the function right?  not the caller

wild-endeavor (dismissed):

flytekit/deck/renderer.py
eapolinario: that's true, we only need `Optional` if we want to explicitly support passing `None`, which is not the case here.

wild-endeavor (approved):

PR #1241, opened 2022-10-18
feat: add TypeTransformer for Tensorflow Model

techytushar:
# TL;DR
This PR adds support for:
1. TypeTransformer for Tensorflow Model type

**Note:** I was not able to run the `pip-compile` commands to generate the new `requirements` files, it failed on `protobuf` version error

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
[https://github.com/flyteorg/flyte/issues/2570](https://github.com/flyteorg/flyte/issues/2570)

## Follow-up issue
_NA_


pingsutw:
> I was not able to run the pip-compile commands to generate the new requirements files, it failed on protobuf version error

Adding `grpcio-status<1.49.0` to the requirement.in can probably fix it.

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,28 @@
+"""
```
pingsutw, line 1: Could we move these files to `flytekit/extras/tensorflow/model`? we'll have two transformers (tensor and model) under the `tensorFlow` directory. https://github.com/flyteorg/flytekit/pull/1243

pingsutw (commented):
Could we move these files to `flytekit/extras/tensorflow/model`?

pingsutw (commented):
Two tests are failing. `ModuleNotFoundError: No module named 'tensorflow'`
You have to run `make dev-requirements.txt` to update the requirements in `dev-requirements.txt`, and then GA will be able to install TensorFlow.

flytekit/extras/tensorflow/model/native.py
```diff
@@ -0,0 +1,78 @@
...
+            literal_type.blob is not None
+            and literal_type.blob.dimensionality == _core_types.BlobType.BlobDimensionality.SINGLE
+            and literal_type.blob.format == self.TENSORFLOW_FORMAT
+        ):
+            return T
+
+        raise ValueError(f"Transformer {self} cannot reverse {literal_type}")
+
+
+class TensorflowModelTransformer(TensorflowTypeTransformer[tf.keras.Model]):
```
samhita-alla, line 71: @techytushar, `tf.Module` is the parent class for `tf.keras.Model` and `tf.keras.layers.Layer`. Can you use that here? Also, can you merge this class into the parent class cause I don't think we need to use Generic in the first place if there's only one override?

samhita-alla:
@pingsutw, what do you think about having a `model.py` rather than a model folder, or better `module.py` since the type transformer will be for `tf.Module`? It seems to make more sense to me. Since we also have a PR for tensor, we can have a `tensor.py` file.

tests/flytekit/unit/extras/tensorflow/test_native.py
```diff
@@ -0,0 +1,29 @@
...
+def generate_model() -> tf.keras.Model:
+    inputs = tf.keras.Input(shape=(32,))
+    outputs = tf.keras.layers.Dense(1)(inputs)
+    model = tf.keras.Model(inputs, outputs)
+    return model
+
+
+@task
+def get_model_layers(model: tf.keras.Model) -> List[tf.keras.layers.Layer]:
+    return model.layers
```
samhita-alla, line 18: This test shouldn't use your type transformer; it should just pickle. So it doesn't make sense to have unless you add a type transformer for `tf.Module`.

samhita-alla: Also, I'm unsure if List[...] will work. `tf.keras.layers.Layer` should, however, work after you add `tf.Module`.

tests/flytekit/unit/extras/tensorflow/test_native.py
```diff
@@ -0,0 +1,29 @@
...
+
+
```
samhita-alla, line 20: @techytushar, can you also add a test for `tf.keras.Sequential` and check if that's working?

pingsutw:
>  do you think about having a model.py rather than a model folder
yes. make sense. I prefer `module.py`

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+.. autosummary::
+   :template: custom.rst
+   :toctree: generated/
+"""
+from flytekit.loggers import logger
+
+# TODO: abstract this out so that there's an established pattern for registering plugins
+# that have soft dependencies
+try:
+    # isolate the exception to the torch import
```
samhita-alla, line 15:
```suggestion
    # isolate the exception to the tensorflow import
```

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    import tensorflow
+
+    _tensorflow_installed = True
+except (ImportError, OSError):
+    _tensorflow_installed = False
+
+
+if _tensorflow_installed:
+    from .model import TensorflowModelTransformer
```
samhita-alla, line 24:
```suggestion
    from .module import TensorFlowModuleTransformer
```

Can you change the file name to `module.py` and the transformer name to `TensorFlowModuleTransformer`?

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,26 @@
+"""
+Flytekit Tensorflow
+=========================================
+.. currentmodule:: flytekit.extras.tensorflow
+
+.. autosummary::
+   :template: custom.rst
+   :toctree: generated/
```
samhita-alla, line 8: I don't think we need any documentation for this. Can you remove this docstring, delete `extras.tensorflow.rst` and remove the tensorflow entry in `types.extend.rst`?

flytekit/extras/tensorflow/model.py
```diff
@@ -0,0 +1,78 @@
...
+import tensorflow as tf
+
+from flytekit.core.context_manager import FlyteContext
+from flytekit.core.type_engine import TypeEngine, TypeTransformer, TypeTransformerFailedError
+from flytekit.models.core import types as _core_types
+from flytekit.models.literals import Blob, BlobMetadata, Literal, Scalar
+from flytekit.models.types import LiteralType
+
+
+class TensorflowModelTransformer(TypeTransformer[tf.keras.Model]):
```
samhita-alla, line 13:
```suggestion
class TensorflowModelTransformer(TypeTransformer[tf.Module]):
```

Please check the comments we've added.

tests/flytekit/unit/extras/tensorflow/test_native.py
techytushar: added

flytekit/extras/tensorflow/model.py
techytushar: I tried using the `tf.Module` but the `tf.keras.models.save_model` method fails when saving the `Layers` (in the unit tests).
I don't think we can save the whole layer in tensorflow, we can only save its weights

How should I go about it ?
1. keep the type as `tf.keras.Model`
2. or use `tf.Module` and add separate check for `Layers` and save them separately

samhita-alla: @techytushar, you're right. How about then you create two files, one each for model and layers? You can name as `model.py` and `layer.py`, respectively. The model transformer also needs to subclass `tf.keras.Sequential` along with `tf.keras.Model`. Please fine the layer serialization [here](https://www.tensorflow.org/api_docs/python/tf/keras/layers/serialize).

samhita-alla: For layer one, set the literal to `generic`, e.g., https://github.com/flyteorg/flytekit/blob/6530601c2538a5d804127a97f63291730b1ba1d8/plugins/flytekit-dolt/flytekitplugins/dolt/schema.py#L76.

techytushar: The `Sequential` [class](https://www.tensorflow.org/api_docs/python/tf/keras/Sequential) already inherits from the `Model` class, so is the subclassing needed ? If it is a `Sequential` type then it automatically use the `ModelTransformer`

samhita-alla: Oh, then not needed.

samhita-alla:
@techytushar, let me know once the changes are in.

techytushar:
@samhita-alla I have pushed the changes.

**Note:** I could not use the generic Struct type for storing the layer config because in protobuf format the `int` values gets converted to `float` due to which when we convert it back to the `tf layer` it was not able to parse it. So I have stored them as simple `json string`

More info here: https://stackoverflow.com/questions/51818125/how-to-use-ints-in-a-protobuf-struct

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+    # isolate the exception to the tensorflow import
+    import tensorflow
+
+    _tensorflow_installed = True
+except (ImportError, OSError):
+    _tensorflow_installed = False
+
+
+if _tensorflow_installed:
+    from .layer import TensorflowLayerTransformer
```
samhita-alla, line 21: Can you please change the class names to `TensorFlow...` with `F` capitalized?

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,26 @@
+"""
+Flytekit Tensorflow
```
samhita-alla, line 2:
```suggestion
Flytekit TensorFlow
```

flytekit/extras/tensorflow/__init__.py
```diff
@@ -0,0 +1,26 @@
...
+except (ImportError, OSError):
+    _tensorflow_installed = False
+
+
+if _tensorflow_installed:
+    from .layer import TensorflowLayerTransformer
+    from .model import TensorflowModelTransformer
+else:
+    logger.info(
+        "Unable to register TensorflowModelTransformer, TensorflowLayerTransformer because tensorflow is not installed."
```
samhita-alla, line 25:
```suggestion
        "Unable to register TensorFlowModelTransformer and TensorFlowLayerTransformer because tensorflow is not installed."
```

flytekit/extras/tensorflow/layer.py
```diff
@@ -0,0 +1,48 @@
...
+from flytekit.core.context_manager import FlyteContext
+from flytekit.core.type_engine import TypeEngine, TypeTransformer, TypeTransformerFailedError
+from flytekit.models.literals import Literal, Primitive, Scalar
+from flytekit.models.types import LiteralType, SimpleType
+
+
+class TensorflowLayerTransformer(TypeTransformer[tf.keras.layers.Layer]):
+    def __init__(
+        self,
+    ):
```
samhita-alla, line 15:
```suggestion
    def __init__(self):
```

flytekit/extras/tensorflow/layer.py
```diff
@@ -0,0 +1,48 @@
+import json as _json
```
samhita-alla, line 1: I don't think an alias is required here. Let's just import json: `import json`.

flytekit/extras/tensorflow/layer.py
```diff
@@ -0,0 +1,48 @@
...
+        self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[tf.keras.layers.Layer]
+    ) -> tf.keras.layers.Layer:
+        if not (lv and lv.scalar and lv.scalar.primitive):
+            raise TypeTransformerFailedError(f"Cannot convert from {lv} to {expected_python_type}")
+
+        layer_config = _json.loads(lv.scalar.primitive.string_value)
+        return tf.keras.layers.deserialize(layer_config)
+
+    def guess_python_type(self, literal_type: LiteralType) -> Type[tf.keras.layers.Layer]:
+        if literal_type.simple == literal_type.simple.STRUCT:
```
samhita-alla, line 40:
```suggestion
        if literal_type is not None and (literal_type.simple == SimpleType.STRING):
```

flytekit/extras/tensorflow/model.py
```diff
@@ -0,0 +1,78 @@
...
+        expected: LiteralType,
+    ) -> Literal:
+        meta = BlobMetadata(
+            type=_core_types.BlobType(
+                format=self.TENSORFLOW_FORMAT,
+                dimensionality=_core_types.BlobType.BlobDimensionality.SINGLE,
+            )
+        )
+
+        local_path = ctx.file_access.get_random_local_path() + ".h5"
```
samhita-alla, line 41:
```suggestion
        local_path = ctx.file_access.get_random_local_path()
```

flytekit/extras/tensorflow/model.py
```diff
@@ -0,0 +1,78 @@
...
+        pathlib.Path(local_path).parent.mkdir(parents=True, exist_ok=True)
+
+        # save tensorflow model in h5 format
```
samhita-alla, line 44:
```suggestion
        # save tensorflow model in tf format
```

tests/flytekit/unit/extras/tensorflow/test_layer.py
```diff
@@ -0,0 +1,48 @@
...
+        loss=tf.keras.losses.BinaryCrossentropy(),
+        metrics=[
+            tf.keras.metrics.BinaryAccuracy(),
+        ],
+    )
+    return model.layers
+
+
+@task
+def get_layers_weights(layers: List[tf.keras.layers.Layer]) -> List[Any]:
```
samhita-alla, line 34: Shouldn't be `List[Any]`. If they are floats, use `List[float]`.

tests/flytekit/unit/extras/tensorflow/test_layer.py
```diff
@@ -0,0 +1,48 @@
...
+
+@workflow
+def wf():
+    dense_layer = get_layer()
+    layers = generate_sequential_model()
+    get_layers_weights(layers=[dense_layer])
+    get_layers_weights(layers=layers)
+
+
+@workflow
```
samhita-alla, line 46: Please remove this.

flytekit/extras/tensorflow/model.py
techytushar: `tf.keras.models.save_model` supports [2 formats](https://www.tensorflow.org/guide/keras/save_and_serialize#whole-model_saving_loading), by default it is `SavedModel` format but it stores the model as a folder and not as a single file. Due to which this line fails
```
ctx.file_access.put_data(local_path, remote_path, is_multipart=False)
```
So I used the `h5` format since it stores it as a single file.

samhita-alla: You should be able to copy a directory as well by setting `is_multipart` to True.

techytushar:
@samhita-alla since its the last day of Hacktoberfest, please can you merge this PR or add label so that it can count towards my contribution

samhita-alla:
@techytushar, I'll merge the PR later as I need to push some changes.

cosmicBboy:
fixes https://github.com/flyteorg/flyte/issues/2570.

We're going to need to move these changes into the tensorflow plugin @eapolinario

eapolinario:
> fixes [flyteorg/flyte#2570](https://github.com/flyteorg/flyte/issues/2570).
> 
> We're going to need to move these changes into the tensorflow plugin @eapolinario

Great, I'll make sure to incorporate those there.

PR #1245, opened 2022-10-19
MyPy Cleanup Incompatible Types Errors

ryankarlos:
Signed-off-by: Ryan Nazareth <ryankarlos@gmail.com>

# TL;DR
This PR fixes all the mypy Incompatible Types errors

## Complete description

This PR fixes the following mypy errors thrown when running  `make lint | grep Incompatible`

Screenshot 2022-10-19 at 21 01 38

Additonally, mypy complains that  plugins/flytekit-kf-mpi` invalid package name -  the `__init__.py` has been removed 
in this PR as this should be a directory anyway  based on other plugins structures.

## Tracking Issue

Addresses part of issue https://github.com/flyteorg/flyte/issues/1030

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -80,14 +80,14 @@ class ExecutionParameters(object):
     @dataclass(init=False)
     class Builder(object):
         stats: taggable.TaggableStats
-        execution_date: datetime
-        logging: _logging.Logger
+        execution_date: typing.Optional[datetime]
+        logging: Optional[_logging.Logger]
         execution_id: typing.Optional[_identifier.WorkflowExecutionIdentifier]
         attrs: typing.Dict[str, typing.Any]
-        working_dir: typing.Union[os.PathLike, utils.AutoDeletingTempDir]
+        working_dir: typing.Optional[utils.AutoDeletingTempDir]
         checkpoint: typing.Optional[Checkpoint]
         decks: List[Deck]
-        raw_output_prefix: str
+        raw_output_prefix: Optional[str]
         task_id: typing.Optional[_identifier.Identifier]
 
         def __init__(self, current: typing.Optional[ExecutionParameters] = None):
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -80,15 +80,15 @@ class ExecutionParameters(object):
     @dataclass(init=False)
     class Builder(object):
         stats: taggable.TaggableStats
-        execution_date: typing.Optional[datetime]
-        logging: Optional[_logging.Logger]
-        execution_id: typing.Optional[_identifier.WorkflowExecutionIdentifier]
         attrs: typing.Dict[str, typing.Any]
-        working_dir: typing.Optional[utils.AutoDeletingTempDir]
-        checkpoint: typing.Optional[Checkpoint]
         decks: List[Deck]
-        raw_output_prefix: Optional[str]
-        task_id: typing.Optional[_identifier.Identifier]
+        raw_output_prefix: Optional[str] = None
+        execution_id: typing.Optional[_identifier.WorkflowExecutionIdentifier] = None
+        working_dir: typing.Optional[utils.AutoDeletingTempDir] = None
+        checkpoint: typing.Optional[Checkpoint] = None
+        execution_date: typing.Optional[datetime] = None
+        logging: Optional[_logging.Logger] = None
+        task_id: typing.Optional[_identifier.Identifier] = None
 
         def __init__(self, current: typing.Optional[ExecutionParameters] = None):
             self.stats = current.stats if current else None
--- a/flytekit/core/python_customized_container_task.py
+++ b/flytekit/core/python_customized_container_task.py
@@ -105,7 +105,7 @@ class PythonCustomizedContainerTask(ExecutableTemplateShimTask, PythonTask[TC]):
         self._resources = ResourceSpec(
             requests=requests if requests else Resources(), limits=limits if limits else Resources()
         )
-        self._environment = environment
+        self._environment = environment  # type: ignore
         self._container_image = container_image
         self._task_resolver = task_resolver or default_task_template_resolver
 
--- a/flytekit/core/python_customized_container_task.py
+++ b/flytekit/core/python_customized_container_task.py
@@ -105,7 +105,7 @@ class PythonCustomizedContainerTask(ExecutableTemplateShimTask, PythonTask[TC]):
         self._resources = ResourceSpec(
             requests=requests if requests else Resources(), limits=limits if limits else Resources()
         )
-        self._environment = environment  # type: ignore
+        self._environment = environment or {}
         self._container_image = container_image
         self._task_resolver = task_resolver or default_task_template_resolver
 
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -229,9 +229,9 @@ def extract_task_module(f: Union[Callable, TrackedInstance]) -> Tuple[str, str, 
         mod_name = mod.__name__
         name = f.lhs
         # We cannot get the sourcefile for an instance, so we replace it with the module
-        f = mod
+        f = mod  # type: ignore
     else:
-        mod = inspect.getmodule(f)
+        mod = inspect.getmodule(f)  # type: ignore
         if mod is None:
             raise AssertionError(f"Unable to determine module of {f}")
         mod_name = mod.__name__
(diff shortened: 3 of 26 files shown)


flytekit/types/directory/types.py
```diff
@@ -17,6 +17,7 @@
 from flytekit.models.types import LiteralType
 
 T = typing.TypeVar("T")
+PathType = typing.Union[str, os.PathLike]
```
ryankarlos, line 20: aliased the type here to keep line length within black constraints to avoid wrapped lines - as not sure `# type :ignore` comment works  for multi line expressions

flytekit/core/python_customized_container_task.py
```diff
@@ -105,7 +105,7 @@ def __init__(
         self._resources = ResourceSpec(
             requests=requests if requests else Resources(), limits=limits if limits else Resources()
         )
-        self._environment = environment
+        self._environment = environment  # type: ignore
```
pingsutw, line 108: It's better to make these two variables become the same type instead of using ignore here.
```suggestion
        self._environment = environment or {}
```

pingsutw (commented):
I'd feel like some places we shouldn't use ignore. we can use `cast` or update the type hint in the class or function. However, It's fine to use ignore after dataclass fields.

flytekit/core/tracker.py
```diff
@@ -229,16 +229,18 @@ def extract_task_module(f: Union[Callable, TrackedInstance]) -> Tuple[str, str,
         mod_name = mod.__name__
         name = f.lhs
         # We cannot get the sourcefile for an instance, so we replace it with the module
-        f = mod
+        g = mod
+        inspect_file = inspect.getfile(g)
     else:
-        mod = inspect.getmodule(f)
+        mod = inspect.getmodule(f)  # type: ignore
```
ryankarlos, line 235: couldn't find an easy way to fix this 
mypy error is `expression has type "Optional[Module]", variable has type Module`

I tried  `mod = inspect.getmodule(f) or None` (and also to ensure that variable is None if  `inspect.getmodule(f)` is None due to check afterwards) ....... however that didn't seem to work. Any ideas ?

Maybe im being silly but I actually don't understand why mypy is expecting mod to be  `Module`  - im guessing this is based on previous assignment to `mod = importlib.import_module(f.instantiated_in)` but should it matter if we are reassigning it anyway ?

flytekit/core/context_manager.py
```diff
@@ -80,14 +80,14 @@ class ExecutionParameters(object):
     @dataclass(init=False)
     class Builder(object):
         stats: taggable.TaggableStats
-        execution_date: datetime
-        logging: _logging.Logger
+        execution_date: typing.Optional[datetime]
```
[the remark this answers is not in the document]
ryankarlos: Should i set them for all the attributes bound to dataclass which have optional type ? Also not sure where these are being used elsewhere (or redundant if they don't have defaults anyway) - also they are being set in `__init__`
```
    @dataclass(init=False)
    class Builder(object):
        stats: taggable.TaggableStats
        execution_date: typing.Optional[datetime]
        logging: Optional[_logging.Logger]
        execution_id: typing.Optional[_identifier.WorkflowExecutionIdentifier]
        attrs: typing.Dict[str, typing.Any]
        working_dir: typing.Optional[utils.AutoDeletingTempDir]
        checkpoint: typing.Optional[Checkpoint]
        decks: List[Deck]
        raw_output_prefix: Optional[str]
        task_id: typing.Optional[_identifier.Identifier]

        def __init__(self, current: typing.Optional[ExecutionParameters] = None):
            self.stats = current.stats if current else None
            self.execution_date = current.execution_date if current else None
            self.working_dir = current.working_directory if current else None
            self.execution_id = current.execution_id if current else None
            self.logging = current.logging if current else None
            self.checkpoint = current._checkpoint if current else None
            self.decks = current._decks if current else []
            self.attrs = current._attrs if current else {}
            self.raw_output_prefix = current.raw_output_prefix if current else None
            self.task_id = current.task_id if current else None

```

ryankarlos:
@pingsutw  i've  fixed the type hints in most places where previously had `# type:ignore`.  Also, see my comments above.

flytekit/core/tracker.py
pingsutw: I see, using ignore is fine here

flytekit/core/context_manager.py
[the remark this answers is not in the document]
pingsutw: good catch, they are duplicates, I'm not sure if we have a specific reason to do so. Maybe keep it for now. 

> Should i set them for all the attributes bound to dataclass which have optional type

yup, Let's set the default value of all the optional variables to None in the dataclass.

flytekit/types/structured/structured_dataset.py
```diff
@@ -326,12 +326,13 @@ class StructuredDatasetTransformerEngine(TypeTransformer[StructuredDataset]):
         str: type_models.LiteralType(simple=type_models.SimpleType.STRING),
     }
 
-    ENCODERS: Dict[Type, Dict[str, Dict[str, StructuredDatasetEncoder]]] = {}
-    DECODERS: Dict[Type, Dict[str, Dict[str, StructuredDatasetDecoder]]] = {}
+    ENCODERS: Dict[Type, Dict[str, Dict[str, Handlers]]] = {}
```
pingsutw, line 329: I feel like using `Handlers` is not correct here because the type of the class in the `ENCODERS` must be `StructuredDatasetEncoder`

flytekit/types/structured/structured_dataset.py
```diff
@@ -326,12 +326,13 @@ class StructuredDatasetTransformerEngine(TypeTransformer[StructuredDataset]):
...
+    DECODERS: Dict[Type, Dict[str, Dict[str, Handlers]]] = {}
     DEFAULT_PROTOCOLS: Dict[Type, str] = {}
     DEFAULT_FORMATS: Dict[Type, str] = {}
 
-    Handlers = Union[StructuredDatasetEncoder, StructuredDatasetDecoder]
+    # Handlers = Union[StructuredDatasetEncoder, StructuredDatasetDecoder]
```
pingsutw, line 334: Remove?

flytekit/types/structured/structured_dataset.py
```diff
@@ -359,7 +360,7 @@ def get_decoder(cls, df_type: Type, protocol: str, format: str):
         return cls._finder(StructuredDatasetTransformerEngine.DECODERS, df_type, protocol, format)
 
     @classmethod
-    def _handler_finder(cls, h: Handlers, protocol: str) -> Dict[str, Handlers]:
+    def _handler_finder(cls, h: Handlers, protocol: str) -> Dict[str, Any]:
```
pingsutw, line 363:
```suggestion
    def _handler_finder(cls, h: Handlers, protocol: str) -> Dict[str, Handlers]:
```

flytekit/types/structured/structured_dataset.py
```diff
@@ -569,7 +570,19 @@ def encode(
         sd_model.metadata._structured_dataset_type.format = handler.supported_format
         return Literal(scalar=Scalar(structured_dataset=sd_model))
 
+    @overload
     def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> T:
+        ...
+
+    @overload
```
pingsutw, line 577: Why do we need to add `to_python_value`? we have three `to_python_value`...

flytekit/types/structured/structured_dataset.py
ryankarlos: if I do this , then mypy complains:
```
flytekit/types/structured/structured_dataset.py:373: error: Incompatible return value type (got "Dict[str, StructuredDatasetEncoder]", expected "Dict[str, TypeVar]")
```

flytekit/types/structured/structured_dataset.py
ryankarlos: i was doing this function type overloading https://mypy.readthedocs.io/en/stable/more_types.html#function-overloading as mypy was complaining when I had union on both param and return value so it couldn't map the correct ones.

However, i removed it now in latest changes and it seemed to be fine running locally - so ill wait for ci to see if it passes.

flytekit/types/structured/structured_dataset.py
ryankarlos: done

flytekit/types/structured/structured_dataset.py
ryankarlos: mypy making things painful 😄

pingsutw:
@ryankarlos do you mind if I push some commits to your branch?

ryankarlos:
@pingsutw  yup go for it

pingsutw (approved):

samhita-alla:
@pingsutw, can we merge this PR?

PR #1255, opened 2022-10-25
Fixing Fast Registration for non root dirs

kumare3:

# TL;DR
In the case when using a container with a non root user, the inflation dir for the code has to be specified. It is common practice to skip a trailing slash "/mypath`/`". so when the distribution is downloaded, it will instead write this to this filename
And this fails the tar logic, which loooks for a similarly named file under the destination path.

The solution simply enforces that the given destination path is actually a directory

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3022

### DIFF
--- a/flytekit/tools/fast_registration.py
+++ b/flytekit/tools/fast_registration.py
@@ -103,10 +103,11 @@ def download_distribution(additional_distribution: str, destination: str):
     :param Text additional_distribution:
     :param os.PathLike destination:
     """
-    if not destination:
+    if not os.path.isdir(destination):
         raise ValueError("Destination path is required to download distribution and it should be a directory")
-    # NOTE the os.path.join(destitation, ''). This is to ensure that the given path is infact a directory and all
-    # downloaded data should be copied into this directory
+    # NOTE the os.path.join(destination, ''). This is to ensure that the given path is infact a directory and all
+    # downloaded data should be copied into this directory. We do this to account for a difference in behavior in 
+    # fsspec, which requires a trailing slash in case of pre-existing directory.
     FlyteContextManager.current_context().file_access.get_data(additional_distribution, os.path.join(destination, ""))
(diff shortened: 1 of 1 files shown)


wild-endeavor (dismissed):

pingsutw (approved):

flytekit/tools/fast_registration.py
```diff
@@ -105,10 +103,14 @@ def download_distribution(additional_distribution: str, destination: str):
     :param Text additional_distribution:
     :param os.PathLike destination:
     """
-    file_access.get_data(additional_distribution, destination)
+    if not destination:
+        raise ValueError("Destination path is required to download distribution and it should be a directory")
```
eapolinario, line 107: should we use https://docs.python.org/3/library/os.path.html#os.path.isdir and raise if it returns false?

flytekit/tools/fast_registration.py
```diff
@@ -105,10 +103,14 @@ def download_distribution(additional_distribution: str, destination: str):
...
+    # NOTE the os.path.join(destitation, ''). This is to ensure that the given path is infact a directory and all
```
eapolinario, line 108: typo: destitation.

flytekit/tools/fast_registration.py
```diff
@@ -105,10 +103,14 @@ def download_distribution(additional_distribution: str, destination: str):
...
+    # downloaded data should be copied into this directory
+    FlyteContextManager.current_context().file_access.get_data(additional_distribution, os.path.join(destination, ""))
```
eapolinario, line 110: Can you also leave a comment saying that we're doing this because of `fsspec` ?

eapolinario (approved):

PR #1259, opened 2022-10-26
[WIP] Install python packages at runtime

pingsutw:
# TL;DR
Install python package before running the task. It has some benefits,
1. Make development or iteration easier and faster because If we just want to update the Python package in the Docker image, we do not need to rebuild the Dockerfile.
2. Install python package only in a specific tasks (pods). 
3. Install two different version of tensorflow respectively in the tasks for testing. t1 can install `TF1` and t2 can install `TF2`
4. It can make size of base image smaller and reduce the times of pulling docker image. Due to the fact that we don't rebuild the image, nodes don't need to pull the new image. 

Example
```python
from flytekit import task, workflow, RuntimeEnv


@task(runtime_env=RuntimeEnv(pip_packages=["tensorflow", "pandas==1.5.1"]))
def t1(a: int) -> int:
    import tensorflow as tf
    print(tf.__version__)
    return a

def t2(a: int) -> int:
    return a


@workflow
def my_wf():
    t1(a=2)
    t2(a=3)


if __name__ == "__main__":
    my_wf()
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2573

## Follow-up issue
- apt install or yum install
- Add support Conda

### DIFF
--- a/flytekit/core/runtime_env.py
+++ b/flytekit/core/runtime_env.py
@@ -35,8 +35,8 @@ class RuntimeEnv:
         for package in self._pip_packages:
             pip_install_cmd.append(" " + package)
 
-        print(pip_install_cmd)
+        logger.info(f"Install python packages with command: {pip_install_cmd}")
         try:
             subprocess.run(pip_install_cmd, capture_output=True, text=True, check=True)
         except subprocess.CalledProcessError as e:
-            logger.error("Failed to install python packages with error: {}".format(e))
+            logger.error(f"Failed to install python packages with error: {e}")
--- a/flytekit/core/runtime_env.py
+++ b/flytekit/core/runtime_env.py
@@ -1,0 +1,35 @@
+from typing import List, Optional
+
+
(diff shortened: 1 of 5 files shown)


eapolinario:

flytekit/core/runtime_env.py
```diff
@@ -0,0 +1,42 @@
...
+            "python",
+            "-m",
+            "pip",
+            "install",
+            "--disable-pip-version-check",
+            "--no-cache-dir",
+        ]
+
+        for package in self._pip_packages:
+            pip_install_cmd.append(" " + package)
```
nelsonjr, line 36: pip_install_cmd is an array, no need for the space.

PR #1307, opened 2022-11-04
add missing secret_requests to ContainerTask

flixr:
Signed-off-by: Felix Ruess <felix.ruess@roboception.de>

# TL;DR
add missing secret_requests to ContainerTask

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Do the same as in other container tasks...

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -7,6 +7,7 @@ from flytekit.core.interface import Interface
 from flytekit.core.resources import Resources, ResourceSpec
 from flytekit.core.utils import _get_container_definition
 from flytekit.models import task as _task_model
+from flytekit.models.security import Secret, SecurityContext
 
 
 class ContainerTask(PythonTask):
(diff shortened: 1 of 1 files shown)


flixr:
So while with this I can inject secrets into my container "the flyte way", it doesn't quite do what I expected/need.
What I really need is a way to [define env vars using secret data](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-container-environment-variables-using-secret-data).

wild-endeavor (approved):

wild-endeavor:
@flixr - the flytekit `Secret` does have a mount type.  Does that do what you need?  If not, you can use a straight up pod task too.

flixr:
@wild-endeavor yeah, I used `Secret(group="minio-write-creds", key="AWS_ACCESS_KEY_ID", mount_requirement=Secret.MountType.ENV_VAR)` and it did inject it as an env var.
The point is that the name of the injected env var needs to be `AWS_ACCESS_KEY_ID`, but flyte injects it as `_FSEC_MINIO_WRITE_CREDS_AWS_ACCESS_KEY_ID`.
And there seems to be no way to influence this, as it is hardcoded in flytepropeller.
With normal k8s secrets I can specify the name of the env var, so there I can do it:
https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#define-a-container-environment-variable-with-data-from-a-single-secret
But how to achieve that for my flyte ContainerTasks?

wild-endeavor:
Merging this for now... thanks!  Can you make a new issue in the flyte repo and just copy in your last comment?  I'm going to ask around and see what can be done.  It does feel a bit weird that we're prepending seemingly random characters to the env var.

Just curious though, wrt this env var specifically, you're going with secrets instead of linking up iam roles to the eks service accounts?

flixr:
I'm not running on EKS, but on my on-prem k3s cluster and how would I even change the service account used for my tasks?
And depending on the task, I need to access different S3 buckets (actually separate minio cluster), also with different credentials, some with write and some with read-only access.

PR #1309, opened 2022-11-05
Bump requirements and restrict grpcio and grpcio-status

eapolinario:
# TL;DR
We have to restrict grpcio in setup.py otherwise we won't be able to bump requirements in plugins

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Two main changes:
1. regenerate requirements
2. move restriction to `grpcio` and `grpcio-status` to `setup.py` so that we can regenerate requirements in plugins (otherwise plugins try to pull in latest `grpcio`, which has a restriction on `protobuf>=4.21.6` (and it's incompatible with [flyteidl](https://github.com/flyteorg/flyteidl/blob/master/setup.py#L18))

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -48,8 +48,10 @@ setup(
         "deprecated>=1.0,<2.0",
         "docker>=5.0.3,<7.0.0",
         "python-dateutil>=2.1",
-        "grpcio>=1.43.0,!=1.45.0,<2.0",
-        "grpcio-status>=1.43,!=1.45.0",
+        # Restrict grpcio and grpcio-status.  The 1.49.1 version pulls in a version of protobuf that is not compatible.
+        # More details in https://github.com/flyteorg/flyte/issues/3006
+        "grpcio>=1.43.0,!=1.45.0,<1.49.1,<2.0",
+        "grpcio-status>=1.43,!=1.45.0,<1.49.1",
         "importlib-metadata",
         "pyopenssl",
         "joblib",
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -122,7 +122,7 @@ exceptiongroup==1.0.1
     # via pytest
 filelock==3.8.0
     # via virtualenv
-flyteidl==1.2.3
+flyteidl==1.2.4
     # via
     #   -c requirements.txt
     #   flytekit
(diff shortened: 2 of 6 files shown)


wild-endeavor (dismissed):
this goes away with buf right?

pingsutw (dismissed):

eapolinario:
> this goes away with buf right?

Correct. Once we move to buf we'll be able to remove https://github.com/flyteorg/flyteidl/blob/master/setup.py#L18 (or at least relax it to the accommodate the next major version of `protobuf`).

pingsutw (approved):

setup.py
```diff
@@ -48,8 +48,10 @@
         "deprecated>=1.0,<2.0",
         "docker>=5.0.3,<7.0.0",
         "python-dateutil>=2.1",
-        "grpcio>=1.43.0,!=1.45.0,<2.0",
-        "grpcio-status>=1.43,!=1.45.0",
+        # Restrict grpcio and grpcio-status.  The 1.49.1 version pulls in a version of protobuf that is not compatible.
+        # More details in https://github.com/flyteorg/flyte/issues/3006
+        "grpcio>=1.43.0,!=1.45.0,<1.49.1,<2.0",
```
honnix, line 53: Is it intended to have `<2.0`? After I install flytekit-1.2.3, I got
```
grpcio                   1.50.0
grpcio-status            1.48.2
```
which doesn't seem to be good.

eapolinario: 1.2.3 does not contain this section. I'm going to create a patch release shortly.

honnix: Ah I see. But is `<2.0` intended?

honnix: Hmm, it is there: https://github.com/flyteorg/flytekit/blob/v1.2.3/setup.py#L51-L52

honnix: BTW I worked on https://github.com/flyteorg/flytekit/pull/1329

PR #1311, opened 2022-11-10
Compile the workflow only at compile time

pingsutw:
# TL;DR
Compile is done on module load, which we should delay to execution or registration time, else module load causes failures.
For example, we use the module loader to load the task function when running a task on flyte cluster. In the meantime, it also compiles the workflow, which is unnecessary. 

Make compiling the workflow only happen while serializing the PythonFunctionWorkflow.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
I've tested it by running dynamic workflow on flyte cluster. we compile the workflow only when running `count_characters`
```python
import typing
from time import sleep

from flytekit import dynamic, task, workflow


@task
def return_index(character: str) -> int:
    """
    Computes the character index (which needs to fit into the 26 characters list)"""
    sleep(1)
    if character.islower():
        return ord(character) - ord("a")
    else:
        return ord(character) - ord("A")


@task
def update_list(freq_list: typing.List[int], list_index: int) -> typing.List[int]:
    """
    Notes the frequency of characters"""
    freq_list[list_index] += 1
    return freq_list


@task
def derive_count(freq1: typing.List[int], freq2: typing.List[int]) -> int:
    """
    Derives the number of common characters"""
    count = 0
    for i in range(26):
        count += min(freq1[i], freq2[i])
    return count


@workflow
def wf1():
    return_index(character="h")


@dynamic
def count_characters(s1: str, s2: str) -> int:
    freq1 = [0] * 26
    freq2 = [0] * 26

    # looping through the string s1
    for i in range(len(s1)):

        # index and freq1 are not accessible as they are promises
        index = return_index(character=s1[i])
        freq1 = update_list(freq_list=freq1, list_index=index)

    # looping through the string s2
    for i in range(len(s2)):

        # index and freq2 are not accessible as they are promises
        index = return_index(character=s2[i])
        freq2 = update_list(freq_list=freq2, list_index=index)

    wf1()

    return derive_count(freq1=freq1, freq2=freq2)


@workflow
def wf(s1: str = "Pear", s2: str = "Earth") -> int:
    return count_characters(s1=s1, s2=s2)


if __name__ == "__main__":
    print(wf(s1="Pear", s2="Earth"))
```

## Tracking Issue

## Follow-up issue
_NA_


flytekit/core/workflow.py
```diff
@@ -731,7 +731,10 @@ def wrapper(fn):
             default_metadata=workflow_metadata_defaults,
             docstring=Docstring(callable_=fn),
         )
-        workflow_instance.compile()
+
+        ctx = FlyteContextManager.current_context()
+        if ctx.execution_state.mode != ctx.execution_state.Mode.TASK_EXECUTION:
```
pingsutw, line 736: cc @kumare3 could we just skip it in `TASK_EXECUTION` mode (running the workflow on flyte cluster)?

wild-endeavor:
will this lead to a lot of user errors?

pingsutw:
No, it won't. I run both locally and remotely, and don't see any errors.

kumare3:
@pingsutw what about registration?

pingsutw:
Registration works as well

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #1343, opened 2022-11-20
Dynamic local execution

wild-endeavor:
# TL;DR
Local execution for dynamic workflows never correctly mimicked a production run.  Because of the execution context that was set, tasks would return native Python values instead of promises, leading to issues like the `>>` operator not working.

This limitation was also introducing runtime bugs in user code because users would use the output of tasks in ways that required them to be native Python values, though this issue was partially mitigated in https://github.com/flyteorg/flytekit/pull/1121 by forcing a compilation when run locally.

This bug is also holding up implementation of the human-in-the-loop work.  And also makes it so that you can't [set node level overrides](https://discuss.flyte.org/t/5143267/hey-all-does-anyone-have-experience-using-a-dynamic-with-ove) if you want to run locally.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Dynamic tasks are supposed to mimic workflows, so we are taking some code from the workflow local_execute pattern.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3093


flytekit/core/python_function_task.py
```diff
@@ -264,14 +267,47 @@ def dynamic_execute(self, task_function: Callable, **kwargs) -> Any:
...
-            updated_exec_state = ctx.execution_state.with_params(mode=ExecutionState.Mode.TASK_EXECUTION)
-            with FlyteContextManager.with_context(
-                ctx.with_execution_state(updated_exec_state).with_serialization_settings(_LOCAL_ONLY_SS)
-            ) as ctx:
-                logger.debug(f"Running compilation for {self} as part of local run as check")
-                self.compile_into_workflow(ctx, task_function, **kwargs)
-                logger.info("Executing Dynamic workflow, using raw inputs")
-                return exception_scopes.user_entry_point(task_function)(**kwargs)
+            if self._wf is None:
+                with FlyteContextManager.with_context(ctx.with_serialization_settings(_LOCAL_ONLY_SS)) as ctx:
```
kumare3, line 271: why should we run this?

wild-endeavor: why not?  i think it's cleaner to compile it once no?  won't we do this for regular workflows too?

eapolinario (approved):

eapolinario (approved):

PR #1350, opened 2022-11-23
Add support to choose docker build target architecture

andresgomezfrr:
# TL;DR
Add new `TARGET_PLATFORM_BUILD` ENV var to set the target platform architecture to allow to build docker image on a machine with a different architecture than the docker image will be executed. 

## Type
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
_NA_

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3097

## Follow-up issue
_NA_

### DIFF
--- a/flytekit_scripts/flytekit_build_image.sh
+++ b/flytekit_scripts/flytekit_build_image.sh
@@ -73,7 +73,7 @@ fi
 echo "Building: $FLYTE_INTERNAL_IMAGE using $TARGET_PLATFORM_BUILD architecture"
 
 # This build command is the raison d'etre of this script, it ensures that the version is injected into the image itself
-eval "$(docker build ${DOCKER_PLATFORM_OPT} . --build-arg tag=${FLYTE_INTERNAL_IMAGE} -t ${FLYTE_INTERNAL_IMAGE} -f ${DOCKERFILE_PATH})"
+docker build ${DOCKER_PLATFORM_OPT} . --build-arg tag=${FLYTE_INTERNAL_IMAGE} -t ${FLYTE_INTERNAL_IMAGE} -f ${DOCKERFILE_PATH}
 
 echo "$IMAGE_NAME built locally."
 
--- a/flytekit_scripts/flytekit_build_image.sh
+++ b/flytekit_scripts/flytekit_build_image.sh
@@ -61,10 +61,17 @@ if [ -n "$REGISTRY" ]; then
 else
   FLYTE_INTERNAL_IMAGE=${IMAGE_NAME}:${PREFIX}${TAG}
 fi
-echo "Building: $FLYTE_INTERNAL_IMAGE"
+
+# Check if the user set the target build architecture, if not use the default instead.
+if [ -z "$TARGET_PLATFORM_BUILD" ]; then
+  TARGET_PLATFORM_BUILD=$(docker system info --format '{{.OSType}}/{{.Architecture}}')
+fi
+
+echo "Building: $FLYTE_INTERNAL_IMAGE using $TARGET_PLATFORM_BUILD as target architecture"
 
 # This build command is the raison d'etre of this script, it ensures that the version is injected into the image itself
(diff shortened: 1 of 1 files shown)


flytekit_scripts/flytekit_build_image.sh
```diff
@@ -61,10 +61,17 @@ if [ -n "$REGISTRY" ]; then
 else
   FLYTE_INTERNAL_IMAGE=${IMAGE_NAME}:${PREFIX}${TAG}
 fi
-echo "Building: $FLYTE_INTERNAL_IMAGE"
+
+# Check if the user set the target build architecture, if not use the default instead.
+if [ -z "$TARGET_PLATFORM_BUILD" ]; then
```
honnix, line 66: Maybe we can simplify this by skipping `--platform` so Docker will do things accordingly:
* If there is a base image matching the current architecture, it will build an image matching the architecture
* If there is no base image matching the current architecture, it will do a build matching the base image architecture

I think if we force `--platform` and there is base image matching that, it will fail. If we think it is better to fail loudly, then it is fine to pass `--platform` explicitly, but this will be a breaking change that we need to note.

andresgomezfrr: But I think that this doesn't solve the main problem. The problem is that you want to build the new image to run in a different architecture than where you are running this image.

>If there is a base image matching the current architecture, it will build an image matching the architecture

If we go into this case, it is possible that you want to run the image in a different architecture. Is your solution to this only having the image base available in the running architecture?

andresgomezfrr: ok I think that I understood your proposal, I'm going to change it :)

andresgomezfrr: 👍🏼 done!

honnix (approved):
I assume you tested the change.

honnix:
Please fix DCO by following https://github.com/flyteorg/flytekit/pull/1350/checks?check_run_id=9764766999

honnix (dismissed):

pingsutw (commented):
nit: There is a lint error, could you fix it?

flytekit_scripts/flytekit_build_image.sh
```diff
@@ -61,10 +62,19 @@ if [ -n "$REGISTRY" ]; then
...
+  DOCKER_PLATFORM_OPT="--platform $TARGET_PLATFORM_BUILD"
+else
+  TARGET_PLATFORM_BUILD="default"
+fi
+
+echo "Building: $FLYTE_INTERNAL_IMAGE using $TARGET_PLATFORM_BUILD architecture"
 
 # This build command is the raison d'etre of this script, it ensures that the version is injected into the image itself
-docker build . --build-arg tag="$FLYTE_INTERNAL_IMAGE" -t "$FLYTE_INTERNAL_IMAGE" -f "${DOCKERFILE_PATH}"
+eval "$(docker build ${DOCKER_PLATFORM_OPT} . --build-arg tag=${FLYTE_INTERNAL_IMAGE} -t ${FLYTE_INTERNAL_IMAGE} -f ${DOCKERFILE_PATH})"
```
honnix, line 76: I don't understand this change. Would the following not work?
```
docker build ${DOCKER_PLATFORM_OPT} . --build-arg tag=${FLYTE_INTERNAL_IMAGE} -t ${FLYTE_INTERNAL_IMAGE} -f ${DOCKERFILE_PATH}
```

honnix, line 76:
```suggestion
docker build ${DOCKER_PLATFORM_OPT} . --build-arg tag=${FLYTE_INTERNAL_IMAGE} -t ${FLYTE_INTERNAL_IMAGE} -f ${DOCKERFILE_PATH}
```

honnix:
Hopefully https://github.com/flyteorg/flytekit/pull/1350/commits/bd96c45993f53178bce4da278b02e9d6ba093aa0 could fix the shellcheck issue.

honnix (approved):

wild-endeavor (approved):

PR #1354, opened 2022-11-28
TypeTransformer for TensorFlow Tensor

samhita-alla:
# TL;DR
Primary PR: https://github.com/flyteorg/flytekit/pull/1269

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/extras/tensorflow/tensor.py
```diff
@@ -0,0 +1,83 @@
...
+    def __init__(self):
+        super().__init__(name="TensorFlow Tensor", t=tf.Tensor)
+
+    def get_literal_type(self, t: Type[tf.Tensor]) -> LiteralType:
+        return LiteralType(
+            blob=_core_types.BlobType(
+                format=self.TENSORFLOW_FORMAT,
+                dimensionality=_core_types.BlobType.BlobDimensionality.SINGLE,
+            )
+        )
```
samhita-alla, line 29: @eapolinario / @pingsutw, is it okay if the `format` specified in `get_literal_type` isn't the same as the one in `to_literal`? `format` contains the tensor type in `to_literal` and the same cannot be fetched in `get_literal_type`.

pingsutw: I feel like it causes some issues when using `Tensor` as input of the workflow. Propeller will expect the format of the literal to be `TensorFlowTensor`, but seems like the format of literal should be `int32` or `float64`.

Should we extend the structured dataset? Tensor is a kind of dataframe, so I think we can register a decoder/encoder for Tensor. btw, we could register a renderer for Tensor, so that we can see the Tensor schema or top 10 rows of Tensor on the flyte deck.

samhita-alla: I'm unsure if we can extend the structured dataset cause tensors can go beyond 2D. I don't think it fits in the typical dataframe style format cause it doesn't come with any notion of columns. Let me know what you think.

> we could register a renderer for Tensor, so that we can see the Tensor schema or top 10 rows of Tensor on the flyte deck.

`tf.Tensor` can have multiple dimensions. So I'll need to find the shape and then print the tensor. Let me look into it.

kumare3:
Eventually we should use the streaming constructs for all of these types

eapolinario:

PR #1381, opened 2022-12-21
fix(pyflyte register): change -d to -D for --destination-dir as -d is already used for --domain

mcloney-ddm:
# TL;DR

`pyflyte register` currently lists `-d` as the short option for both `--domain` and `--destination-dir`, which can result in unexpected behavior (e.g., causing executions of workflows to crash because it's pointing at the `development` directory, which does not exist). 

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Changed `-d` flag for `--destination-dir` to `-D`, left `-d` flag as valid short flag for `--domain`.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3178

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -67,7 +67,7 @@ Note: This command only works on regular Python packages, not namespace packages
     help="Directory to write the output zip file containing the protobuf definitions",
 )
 @click.option(
-    "-d",
+    "-D",
     "--destination-dir",
     required=False,
     type=str,


wild-endeavor:
@eapolinario what do you think? can we merge this?  we definitely should.  @mcloney-ddm do you know how the conflict resolves today?  Wondering if there are people relying on this behavior hard-coded in CI commands.

And thanks!

eapolinario:
> @mcloney-ddm do you know how the conflict resolves today? Wondering if there are people relying on this behavior hard-coded in CI commands.

The [previously](https://github.com/flyteorg/flytekit/blob/e63d017c6d8732ebe616b2d49d4dbc89989a0ca9/flytekit/clis/sdk_in_container/register.py#L39) defined `-d` is used (it actually is a side-effect of how decorators are applied, in other words, the definition in line 39 clobbers the one in line 70). It is very unlikely that we have some user relying on this behavior in the wild as it would require an extraordinary series of coincidences for this to work by accident.

eapolinario (approved):

eapolinario:
It's unfortunate that `click` doesn't provide any facilities to detect this case.

PR #1380, opened 2022-12-21
PodSpec should not require primary_container name

kumare3:
if primary_container name is not specified it should default.

Signed-off-by: Ketan Umare <ketan.umare@gmail.com>

# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
 _How did you fix the bug, make the feature etc. Link to any design docs etc_

## Tracking Issue
https://github.com/flyteorg/flyte/issues/<number>

## Follow-up issue
_NA_
OR
_https://github.com/flyteorg/flyte/issues/<number>_

PR #1382, opened 2022-12-22
Added more pod tests and an example pod task

kumare3:

# TL;DR
Related to https://github.com/flyteorg/flytekit/pull/1380
Adds more tests and an example

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/plugins/flytekit-k8s-pod/tests/test_pod.py
+++ b/plugins/flytekit-k8s-pod/tests/test_pod.py
@@ -12,6 +12,7 @@ from flytekit import Resources, TaskMetadata, dynamic, map_task, task
 from flytekit.configuration import FastSerializationSettings, Image, ImageConfig, SerializationSettings
 from flytekit.core import context_manager
 from flytekit.core.type_engine import TypeEngine
+from flytekit.exceptions import user
 from flytekit.extend import ExecutionState
 from flytekit.tools.translator import get_serializable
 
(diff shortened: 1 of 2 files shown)


plugins/flytekit-k8s-pod/tests/test_pod.py
```diff
@@ -473,3 +474,32 @@ def dynamic_task_with_pod_subtask(dummy_input: str) -> str:
...
+
+
+def test_pod_config():
+    with pytest.raises(user.FlyteValidationException):
+        Pod(pod_spec=None)
+
+    with pytest.raises(user.FlyteValidationException):
+        Pod(pod_spec=V1PodSpec(containers=[]), primary_container_name=None)
+
+    selector = ({"node_group": "memory"},)
```
eapolinario, line 486: just for the sake of correctness (e.g. in case someone copies this example), can you make this a dictionary?

eapolinario (commented):
the [default name](https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-k8s-pod/flytekitplugins/pod/task.py#L15) for the primary container is invalid because it doesn't pass [this test](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/task_compiler.go#L88) (underscores cannot show up in dns names).

plugins/flytekit-k8s-pod/tests/test_pod.py
kumare3: good catch. also great catch with the container name. I just changed that as well.

pingsutw (approved):

eapolinario (approved):

PR #1384, opened 2022-12-23
Update default config to work out-of-the-box with flytectl demo

cosmicBboy:
Fixes https://github.com/flyteorg/flyte/issues/3182

This PR makes the getting started UX smoother when users try to invoke `pyflyte run` against the flytectl demo cluster. This change means that they no longer have to `export FLYTECTL_CONFIG=...` to get it working.

Signed-off-by: Niels Bantilan <niels.bantilan@gmail.com>

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/configuration/file.py
```diff
@@ -54,7 +59,8 @@ def read_from_file(
             return None
         try:
             v = cfg.get(self)
-            return transform(v) if transform else v
+            if _exists(v):
```
wild-endeavor, line 62: was this causing a problem?

flytekit/configuration/file.py
```diff
@@ -80,10 +86,11 @@ def read_from_file(
             return None
         try:
             v = cfg.get(self)
-            if v:
+            if _exists(v):
```
wild-endeavor, line 89: but `isinstance(False, bool)` is true, so if `v` was `False` now this will be true.  Or is `v` never a bool?

cosmicBboy: Yep, it wasn't picking up `insecure: false` in the yaml config because `if v` would evaluate to false, so the config would just always use the default value.

cosmicBboy: I guess the `transform` here doesn't actually do what it's intended to do before yaml will automatically parse values to the expected types

wild-endeavor (approved):

eapolinario (approved):

PR #1400, opened 2022-12-28
wip - Delete model files

eapolinario:

# TL;DR
Remove all mentions to model files 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
 Beware:
1. Not all models are removed, so far I had to bring back `common.py`, `array_job.py`, `filters.py`, and `schedule.py`, either because there is no equivalent protobuf messages (like the objects built in `filters.py`) or because there's adhoc logic in the model files that was not straightforward to bring back (for example, in some cases we were inheriting from the model files, but now that we're moving to protobuf messages, inheriting from protobuf messages is not allowed)
2. for `oneof` fields, use `WhichOneof("field")` to check which field is set (read https://developers.google.com/protocol-buffers/docs/reference/python-generated#oneof for more details).
3. to check if a field is set, use `HasField("field")`
4. use `ClearField` to set a [singular field](https://developers.google.com/protocol-buffers/docs/reference/python-generated) to null.
5. remote is a mess. We also do not have great test coverage there, so proceed with caution
6. No positional arguments allowed in protobuf stub constructors.
7. Enums are a little weird. For example, look at the [generated code](https://github.com/flyteorg/flyteidl/blob/master/gen/pb_python/flyteidl/core/types_pb2.pyi#L22-L28 ) for the [BlobDimensionality enum](https://github.com/flyteorg/flyteidl/blob/master/protos/flyteidl/core/types.proto#L74-L77). We can refer to each value using `BlobType.MULTIPART` or `BlobType.BlobDimensionality.MULTIPART`, where in the latter case `pyright` (the static type-checker I'm using in emacs) was complaining, even though it works in both cases. You'll notice that I chose the former case in most search-and-replace, but it might be confusing to use that form in the long run (since it won't be clear which enum is being used as the source for the value).

## Tracking Issue

## Follow-up issue
_NA_
OR


eapolinario:

PR #1411, opened 2023-01-13
Pluck retry from flytekit and into sagemaker

eapolinario:

# TL;DR
`retry` is only used in the sagemaker plugin - it should no longer be a flytekit dependence

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
 `retry` is only used in the sagemaker flytekit plugin, which means that we only need `retry` there. 

We're going to release a minor version of flytekit and in a follow up PR we can then update the dependence in the sagemaker plugin's requirements.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3052

## Follow-up issue
_NA_


pingsutw (dismissed):

kumare3 (dismissed):

plugins/flytekit-aws-sagemaker/requirements.txt
```diff
@@ -190,6 +192,8 @@ responses==0.22.0
     # via flytekit
 retry==0.9.2
     # via flytekit
+retry2==0.9.5
+    # via flytekitplugins-awssagemaker
```
hajapy, line 196: I don't think you'd want both of these to get installed, might need to regenerate with the updated flytekit?

pingsutw: good catch, thanks. we'll remove it

pingsutw (approved):

PR #1419, opened 2023-01-23
DuckDB plugin

samhita-alla:

# TL;DR
This PR adds a `DuckDBQuery` task plugin that runs queries using DuckDB as the DBMS. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

Capturing the crucial assumptions I made while building the task plugin:

- The `DuckDBQuery` task parameter that a user needs to send argument to includes `query` and can contemplate adding includes `inputs`. 
- `query` can include a set of queries that'll be run sequentially. The last query needs to be a SELECT query. 
- `inputs` can include structured dataset or a list of parameters to be sent to the queries.
- The `output` is a pyarrow table. Can be converted to any structured dataset compatible type.
- The connection mode is set to `:memory`, i.e., the data is always stored in an in-memory, non-persistent database. It can be set to a file, but it's difficult to make the file accessible to different `DuckDBQuery` pods, which otherwise wouldn't make sense because file is persistent, and it needs to be leveraged.

Example:

```python
duckdb_query = DuckDBQuery(
    name="read_parquet",
    query=[
        "INSTALL httpfs",
        "LOAD httpfs",
        """SELECT hour(lpep_pickup_datetime) AS hour, count(*) AS count FROM READ_PARQUET(?) GROUP BY hour""",
    ],
    inputs=kwtypes(params=list[str]),
)

@workflow
def parquet_wf(parquet_file: str) -> pd.DataFrame:
    return duckdb_query(params=[parquet_file])

assert isinstance(
    parquet_wf(parquet_file="https://d37ci6vzurychx.cloudfront.net/trip-data/green_tripdata_2022-02.parquet"),
    pd.DataFrame,
)
```

## Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/3246

## Follow-up issue
_NA_
OR


kumare3:
@samhita-alla can we support accepting StructuredDataset as an input?

samhita-alla:
@kumare3, `DuckDBQuery` task accepts structured dataset already. Here's an example:

```python
duckdb_task = DuckDBQuery(
    name="duckdb_sd_df",
    query="SELECT * FROM pandas_df WHERE i = 2",
    inputs=kwtypes(pandas_df=StructuredDataset),
)

@task
def get_pandas_df() -> StructuredDataset:
    return StructuredDataset(
        dataframe=pd.DataFrame.from_dict({"i": [1, 2, 3, 4], "j": ["one", "two", "three", "four"]})
    )

@workflow
def pandas_wf(pandas_df: StructuredDataset) -> pd.DataFrame:
    return duckdb_task(pandas_df=pandas_df)

assert isinstance(pandas_wf(pandas_df=get_pandas_df()), pd.DataFrame)
```

Let me know if you're looking for something different.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -0,0 +1,102 @@
...
+class DuckDBQuery(PythonInstanceTask):
+    _TASK_TYPE = "duckdb"
+
+    def __init__(
+        self,
+        name: str,
+        query: Union[str, List[str]],
+        inputs: Optional[Dict[str, Union[StructuredDataset, list]]] = None,
+        **kwargs,
+    ):
```
cosmicBboy, line 26: can we add a docstring with `:param: name` annotations to this method?

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -0,0 +1,102 @@
...
+                    yield QueryOutput(output=self._con.execute(query, params), counter=counter)
+                else:
+                    raise ValueError("Parameter not specified.")
+        else:
+            yield QueryOutput(output=self._con.execute(query), counter=counter)
+
+    def execute(self, **kwargs) -> StructuredDataset:
+        """
+        TODO: Enable iterative download after adding the functionality to structured dataset code.
+        """
```
cosmicBboy, line 63: can we add a docstring with `:param: name` annotations to this method?

also move the TODO out of the docstring and place it into an inline comment `# TODO`

plugins/flytekit-duckdb/setup.py
```diff
@@ -0,0 +1,36 @@
...
+    name=microlib_name,
+    version=__version__,
+    author="flyteorg",
+    author_email="admin@flyte.org",
+    description="DuckDB Plugin for Flytekit",
+    namespace_packages=["flytekitplugins"],
+    packages=[f"flytekitplugins.{PLUGIN_NAME}"],
+    install_requires=plugin_requires,
+    license="apache2",
+    python_requires=">=3.7",
```
cosmicBboy, line 21: let's restrict this to <=3.10 as well, [like this](https://github.com/flyteorg/flytekit/blob/master/setup.py#L84)

cosmicBboy (commented):
looking great! added some comments for docstrings

samhita-alla:
@cosmicBboy, thanks for reviewing the PR! Can you look through it again?

cosmicBboy:
image

DuckDB api reference is blank, I think we need to update the https://github.com/flyteorg/flytekit/blob/master/doc-requirements.in file with `duckdb` as a dependency, or else it would be rendered.

cosmicBboy:
@samhita-alla we'll also need to invest a bit in enable warnings as errors in the sphinx build process `-W` like we did with flytesnacks. It'll let us catch these kinds of docs build errors

samhita-alla:
@cosmicBboy, fixed the docs and added a GitHub action to show warnings as errors.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -0,0 +1,117 @@
...
+
+class DuckDBQuery(PythonInstanceTask):
+    _TASK_TYPE = "duckdb"
+
+    def __init__(
+        self,
+        name: str,
+        query: Union[str, List[str]],
+        inputs: Optional[Dict[str, Union[StructuredDataset, list]]] = None,
+        **kwargs,
```
pingsutw, line 25: Could we add an Output Schema type here like [snowflake](https://github.com/flyteorg/flytekit/blob/707fc03b88ec08fb10d60ac642f3c80e275bd90c/plugins/flytekit-snowflake/flytekitplugins/snowflake/task.py#L45)? if output_schema_type is none, we won't generate output dataset. 
And the type should change to StructuredDataset, because we already deprecated FlyteSchema

samhita-alla: @pingsutw, this task isn't helpful if there's no output dataset. The `DuckDBQuery` task runs some queries and returns the output of a `SELECT` statement. Hence, it must return the query output, and in this case, it's `StructuredDataset`. Also, I can definitely add `output_schema_type`. But the output has to always be a `StructuredDataset`. So is it necessary? I'm already hard-coding the output type in the initialization. Let me know what you think.

pingsutw: I see. do we support `insert` or some other operations? If not, I think we don't need schema type for now

pingsutw (approved):

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
samhita-alla: Yeah, we do. So you can give a bunch of queries to a single `DuckDBQuery` task. But the last one needs to be a SELECT query because after say, you insert the data, you need to retrieve the data, right? Else, it's of no use. I'm using the non-persistent offering by DuckDB. So all the data will be available only within the query. Does that make sense?

pingsutw: make sense. Thanks for the explanation.

samhita-alla:
Thanks, Kevin! Will merge this PR after @cosmicBboy approves as well.

samhita-alla:
@kumare3, let me know if this PR looks good to you.

cosmicBboy (approved):

PR #1425, opened 2023-01-26
Add `pod_template` and `pod_template_name` arguments for `PythonAutoContainerTask`, its downstream tasks, and `@task`.

ByronHsu:
# TL;DR
Add `pod_template` and `pod_template_name` arguments for `PythonAutoContainerTask`, its downstream tasks, and `@task`. (Following [this guide](https://github.com/flyteorg/flytekit/compare/sample-pod?expand=1))

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

1. Users can pass in `pod_template` in `PythonAutoContainerTask`, its downstream tasks, and `@task` to customize `pod_template` for each task instance. It also allows developers to set the default `pod_template` for plugins in flytekit.
2. Users can pass in `pod_template_name` in `PythonAutoContainerTask`, its downstream tasks, and `@task` to select which `PodTemplate` resources to use on the server side.

### Example

- instance
```
task_with_minimum_pod_template = DummyAutoContainerTask(
    name="x",
    task_config=None,
    task_type="t",
    container_image="repo/image:0.0.0",
    pod_template=PodTemplate(
        primary_container_name="primary",
        labels={"lKeyA": "lValA"},
        annotations={"aKeyA": "aValA"},
        pod_spec=V1PodSpec(
            containers=[
                V1Container(
                    name="primary",
                ),
            ]
        ),
    ),
    pod_template_name="A",
)
```

- decorator
```
    @task(
        container_image="repo/image:0.0.0",
        pod_template=PodTemplate(
            primary_container_name="primary",
            labels={"lKeyA": "lValA"},
            annotations={"aKeyA": "aValA"},
            pod_spec=V1PodSpec(
                containers=[
                    V1Container(
                        name="primary",
                    ),
                ]
            ),
        ),
        pod_template_name="A",
    )
    def func_with_pod_template(i: str):
        print(i + 3)
```

## Tracking Issue

https://github.com/flyteorg/flyte/issues/3123

## TODO
- [ ] Update `requirement.txt` to include `kubernetes`
- [ ] Improve comments
- [ ] `pod_template` and `pod_template_name` in container task?
- [ ] Test in pyspark task? (Already did in its parent `PythonFunctionTask` though)


flytekit/core/container_task.py
```diff
@@ -10,6 +10,7 @@
 from flytekit.models.security import Secret, SecurityContext
 
 
+# TODO: do we need pod_template here? Seems that it is a raw container not running in pods
```
hamersaw, line 13: Yes, the goal is to add this support to `ContainerTask` as well. So these are defined by containers, but Flyte executes them as a `Pod`. So if a user provides a `PodTemplate` argument, we should populate the oneof in idl as a `K8sPod` rather than a `Contianer`.

hamersaw:
Also, support in [ShellTask](https://github.com/flyteorg/flytekit/blob/7b48fff5cfe2fcf19e28fde8e35c20acdb9059e2/flytekit/extras/tasks/shell.py#L100) would be great too! Happy to meet up and discuss any of this.

flytekit/core/container_task.py
ByronHsu: 1. I didn't see any reference of `ContainerTask` in flytekit source code. Was it deprecated in the core and only used by developers in the plugin?
2. Why the `get_command` part only print instead of return `list[str]`? And why the printing message is`docker run...`?

ByronHsu:
@hamersaw Shelltask should be working because it inherits `PythonAutoContainerTask`. @wild-endeavor Do we need to add a pod_template test in all  `PythonAutoContainerTask` downstream tasks?

ByronHsu:
How to safely update `requirement.txt`?
image

flytekit/core/pod_template.py
```diff
@@ -0,0 +1,22 @@
...
+
+from kubernetes.client.models import V1PodSpec
+
+from flytekit.exceptions import user as _user_exceptions
+
+PRIMARY_CONTAINER_DEFAULT_NAME = "primary"
+
+
+@dataclass
+class PodTemplate(object):
```
byjiang1996, line 12: Trying to understand some design decisions.
Why not use V1PodTemplate directly but use customized PodTemplate object here: https://k8s-python.readthedocs.io/en/stable/kubernetes.client.models.html#kubernetes.client.models.v1_pod_template.V1PodTemplate

ByronHsu: probably because we don’t want users to expose to that many things. We just want them to be able to customize pod_spec, label, annotations, etc, so they won’t get confused

ByronHsu:
I added `"kubernetes>=12.0.1"` in `setup.py` and ran `make requirements.txt`, but many deps got added. Is this the expected result?

eapolinario:
> I added `"kubernetes>=12.0.1"` in `setup.py` and ran `make requirements.txt`, but many deps got added. Is this the expected result?

Yes, you can see in the rendered `requirements.txt` file the source for these new deps. 

That said, we have an issue right now where the `dev-requirements.txt` requirements doesn't match the new `requirements.txt`. I'm going to push a commit to your branch to fix this.

eapolinario:
https://github.com/flyteorg/flyteidl/pull/358 needs to be merged and incorporated here (i.e. by bumping version of flyteidl in [setup.py](https://github.com/flyteorg/flytekit/blob/master/setup.py#L43))

ByronHsu:
@esadler-hbo tysm!

flytekit/core/pod_template.py
```diff
@@ -0,0 +1,22 @@
...
+    pod_spec: V1PodSpec
+    primary_container_name: str = PRIMARY_CONTAINER_DEFAULT_NAME
+    labels: Optional[Dict[str, str]] = None
+    annotations: Optional[Dict[str, str]] = None
+
+    def __post_init__(self):
+        if not self.pod_spec:
+            raise _user_exceptions.FlyteValidationException("A pod spec cannot be undefined")
```
hamersaw, line 20: Does this need to be required? If a user doesn't provide it, we can just generate an empty one right? It would make the API more ergonomic IMO.

flytekit/core/pod_template.py
```diff
@@ -0,0 +1,22 @@
...
+
+from flytekit.exceptions import user as _user_exceptions
+
+PRIMARY_CONTAINER_DEFAULT_NAME = "primary"
+
+
+@dataclass
+class PodTemplate(object):
+    pod_spec: V1PodSpec
+    primary_container_name: str = PRIMARY_CONTAINER_DEFAULT_NAME
```
hamersaw, line 14: This is going to be a point of contention. We should probably rename this to something like `task_container_name` and have it default to something other than `primary`.

This is because the `primary_container_name` in the `Pod` `task_config` is very specific, and not only for applying configuration flytekit-side. On the backend, if the `primary_container_name` is specified through `TaskTemplate.Config` we declare the task completed when the primary container completes. This allows The `Pod` `task_config` to inject sidecars that run indefinitely. However, this is the opposite of what is necessary for `ContainerTasks` to complete. Where we inject a container(s) (copilot) to download inputs and then output the task outputs. For a task to complete, all of the containers need to completed (rather than just the primary container in the case of `Pod` `task_config`). So if we want to support this `PodTemplate` from `ContainerTasks` (which we certainly do), we need to say that all containers need to complete for the task to complete. I think using the same naming (ie. `primary_container_name`) as the `Pod` `task_config` would make this confusing as the functionality is different.

hamersaw: TL;DR IMO we should require all containers specified in the `PodTemplate` to complete for the task to complete. This is necessary if we want to support this functionality from the `ContainerTask`. The only thing we lose is the ability to apply this with the `Pod` `task_config`. But since the `Pod` task already defines a `Pod` it already has the same support that this provides, so there is no need.

hamersaw:
Also, I noticed that right now we can not specify a `PodSpec` with an empty `Containers` list. This means we need verbosity if a user wants to so something simple like add a toleration to the `Pod`. It looks like we could use `local_vars_configuration` internally to [stop client-side validation](https://github.com/kubernetes-client/python/blob/13988e46123377a7b59afa9fccb0fb2615bc82fa/kubernetes/client/models/v1_pod_spec.py#L331) of the `PodTemplate. I think this is worth spending a little time looking into - because then we can do something like:
```
@task(
    pod_template=PodTemplate(
        spec=V1PodSpec{
            tolerations=[
                V1Toleration{
                    key= "num-gpus"
                    operator= "Equal"
                    value= 1
                    effect= "NoSchedule"
                },
            ]
        )
    )
)
def foo(bar: int):
    # omitted
```

Without having to add the bloat of definition a list with an empty container.
```
@task(
    pod_template=PodTemplate(
        spec=V1PodSpec{
            containers=[
                V1Container(
                    name="primary",
                ),
            ],
            tolerations=[
                V1Toleration{
                    key= "num-gpus"
                    operator= "Equal"
                    value= 1
                    effect= "NoSchedule"
                },
            ]
        )
    )
)
def foo(bar: int):
    # omitted
```

flytekit/core/python_auto_container.py
```diff
@@ -40,6 +49,8 @@ def __init__(
         environment: Optional[Dict[str, str]] = None,
         task_resolver: Optional[TaskResolverMixin] = None,
         secret_requests: Optional[List[Secret]] = None,
+        pod_template: Optional[PodTemplate] = None,
+        pod_template_name: Optional[str] = None,
```
hamersaw, line 53: Should we briefly explore name changes here? `pod_template` and `pod_template_name` are not very descriptive and certainly don't capture a client-side `PodTemplate` application vs server-side k8s resource `PodTemplate` configuration. Thoughts?

flytekit/core/pod_template.py
ByronHsu: I copied this from k8s_pod task, but I think your statement makes sense. The empty pod spec will be filled with default values in `_serialize_pod_spec`

flytekit/core/pod_template.py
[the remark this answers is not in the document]
ByronHsu: I don't have strong feelings, but currently we only have `loggers.py` in the root.
image

ByronHsu:
> 

Why we can not specify a PodSpec with an empty containers list? Which part will raise error?

flytekit/core/pod_template.py
ByronHsu: I don't really understand this. Seems to lack some context. Will contact you offline.

flytekit/core/python_auto_container.py
ByronHsu: yeah it is very confusing now, but honestly I don't have a better idea on my mind. I also feel they should be `one of`. If users can set both, it will be more confusing.

flytekit/core/pod_template.py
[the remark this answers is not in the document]
eapolinario: We should do it. Simply import it in [this file](https://github.com/flyteorg/flytekit/blob/master/flytekit/__init__.py#L165-L202).

flytekit/core/pod_template.py
ByronHsu: drop it

flytekit/core/python_auto_container.py
```diff
@@ -179,6 +207,56 @@ def get_container(self, settings: SerializationSettings) -> _task_model.Containe
...
+    def _serialize_pod_spec(self, settings: SerializationSettings) -> Dict[str, Any]:
+        containers = self.pod_template.pod_spec.containers
+        primary_exists = False
+        for container in containers:
+            if container.name == self.pod_template.primary_container_name:
+                primary_exists = True
+                break
+        if not primary_exists:
+            # insert a placeholder primary container if it is not defined in the pod spec.
+            containers.append(_task_model.Container(name=self.pod_template.primary_container_name))
```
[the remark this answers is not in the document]
ByronHsu: image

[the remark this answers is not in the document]
eapolinario: TaskTemplate comes from [here](https://github.com/flyteorg/flytekit/blob/ea39054223a9432692c03314b90c45240f2d8419/flytekit/tools/translator.py#L201-L213).

flytekit/core/pod_template.py
ByronHsu: addressed in the new commits

flytekit/core/pod_template.py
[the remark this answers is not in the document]
ByronHsu: addressed in the new commits

flytekit/core/python_auto_container.py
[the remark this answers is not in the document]
ByronHsu: addressed in the new commits

flytekit/core/python_auto_container.py
ByronHsu: Still needs discussion!

ByronHsu:
I did the following changes:
1. Import `PodTemplate` in the root level
2. Return the primary container name in `get_config` in `pythonAutoContainerTask` and its descendants (`awsbatch` are `papermili` the two using `get_config` currently). 
3.  Allow empty `V1PodSpec`
4. Bypass required container assert

**Bullet 4** is a bit tricky. I overwrote `V1PodSpec` `containers` setter to not do the validation. I felt this overwrite things should be moved to some entrypoint in flytekit. Need python experts' help.

eapolinario:
> Also, I noticed that right now we can not specify a `PodSpec` with an empty `containers` list. This means we need unecessary verbosity if a user wants to so something simple like add a toleration to the `Pod`. It looks like we could use `local_vars_configuration` internally to [stop client-side validation](https://github.com/kubernetes-client/python/blob/13988e46123377a7b59afa9fccb0fb2615bc82fa/kubernetes/client/models/v1_pod_spec.py#L331) of the `PodTemplate`. I think this is worth spending a little time looking into - because then we can do something like:
> 
> ```
> @task(
>     pod_template=PodTemplate(
>         spec=V1PodSpec{
>             tolerations=[
>                 V1Toleration{
>                     key= "num-gpus"
>                     operator= "Equal"
>                     value= 1
>                     effect= "NoSchedule"
>                 },
>             ]
>         )
>     )
> )
> def foo(bar: int):
>     # omitted
> ```
> 
> Without having to add the bloat of definition a list with an empty container.
> 
> ```
> @task(
>     pod_template=PodTemplate(
>         spec=V1PodSpec{
>             containers=[
>                 V1Container(
>                     name="primary",
>                 ),
>             ],
>             tolerations=[
>                 V1Toleration{
>                     key= "num-gpus"
>                     operator= "Equal"
>                     value= 1
>                     effect= "NoSchedule"
>                 },
>             ]
>         )
>     )
> )
> def foo(bar: int):
>     # omitted
> ```

Unfortunately, there's no way to turn off validation through this method you mentioned, Dan. Take a look at this code:
```
In [35]: from kubernetes.client import Configuration

In [36]: conf = Configuration()

In [37]: conf.client_side_validation = True  # the default

In [38]: V1PodSpec(local_vars_configuration=conf)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[38], line 1
----> 1 V1PodSpec(local_vars_configuration=conf)

File /tmp/tmp.m5bYaiOZ4k/venv/lib/python3.10/site-packages/kubernetes/client/models/v1_pod_spec.py:166, in V1PodSpec.__init__(self, active_deadline_seconds, affinity, automount_service_account_token, containers, dns_config, dns_policy, enable_service_links, ephemeral_containers, host_aliases, host_ipc, host_network, host_pid, host_users, hostname, image_pull_secrets, init_containers, node_name, node_selector, os, overhead, preemption_policy, priority, priority_class_name, readiness_gates, restart_policy, runtime_class_name, scheduler_name, security_context, service_account, service_account_name, set_hostname_as_fqdn, share_process_namespace, subdomain, termination_grace_period_seconds, tolerations, topology_spread_constraints, volumes, local_vars_configuration)
    164 if automount_service_account_token is not None:
    165     self.automount_service_account_token = automount_service_account_token
--> 166 self.containers = containers
    167 if dns_config is not None:
    168     self.dns_config = dns_config

File /tmp/tmp.m5bYaiOZ4k/venv/lib/python3.10/site-packages/kubernetes/client/models/v1_pod_spec.py:322, in V1PodSpec.containers(self, containers)
    314 """Sets the containers of this V1PodSpec.
    315
    316 List of containers belonging to the pod. Containers cannot currently be added or removed. There must be at least one container in a Pod. Cannot be updated.  # noqa: E501
   (...)
    319 :type: list[V1Container]
    320 """
    321 if self.local_vars_configuration.client_side_validation and containers is None:  # noqa: E501
--> 322     raise ValueError("Invalid value for `containers`, must not be `None`")  # noqa: E501
    324 self._containers = containers

ValueError: Invalid value for `containers`, must not be `None`

In [39]: conf.client_side_validation = False

In [40]: V1PodSpec(local_vars_configuration=conf)
Out[40]:
... # this works
```

That said, we don't necessarily have to define a container, the only restriction is that the list of containers is not `None`, i.e. this works:
```
In [41]: V1PodSpec(containers=[])
Out[41]:
{'active_deadline_seconds': None,
 'affinity': None,
 'automount_service_account_token': None,
 'containers': [],
 'dns_config': None,
 'dns_policy': None,
...
```

I'm conflicted. On one hand, we could wrap k8s `V1PodSpec` with our own `V1PodSpec` (so that we could disable client-side validation), OTOH, leaning on this behavior of the current k8s object where we define an empty list is not terrible (and we have a short feedback loop in case we break it). 

One thing that feels off for me with this idea of replacing the validation of the k8s client validation for the containers specifically is that if there's code that relies on that validation (for whatever reason) in a task, that code will no longer work.

wdyt?

hamersaw:
> wdyt?

I think this sounds great. Lets leave it as is, users have to define an empty containers list is the pod spec is defined. Going this route also means if it's a problem we can address it in the future - much easier then going the opposite direction.

flytekit/core/container_task.py
eapolinario, line 13: can you remove this comment?

tests/flytekit/unit/core/test_python_function_task.py
```diff
@@ -122,3 +125,83 @@ def foo_missing_cache_version(i: str):
...
+        "flytekit.core.python_auto_container.default_task_resolver",
+        "--",
+        "task-module",
+        "tests.flytekit.unit.core.test_python_function_task",
+        "task-name",
+        "func_with_pod_template",
+    ]
+
+    #################
+    # Test pod_teamplte_name
```
eapolinario, line 196: nit: typo: teamplte -> template

tests/flytekit/unit/core/test_python_function_task.py
```diff
@@ -122,3 +125,83 @@ def foo_missing_cache_version(i: str):
         @task(cache_serialize=True)
         def foo_missing_cache(i: str):
             print(f"{i}")
+
+
+def test_pod_template():
```
eapolinario, line 130: Can you also add a test where no pod spec is defined?

eapolinario (commented):
Just a few minor comments and we should be good to go.

tests/flytekit/unit/core/test_python_function_task.py
ByronHsu: I changed `task_with_minimum_pod_template` to test the case without podspec. The original intention of `task_with_minimum_pod_template` to test pod spec's volume is already covered in `test_pod_template`.

eapolinario (approved):
This is awesome!

PR #1428, opened 2023-02-01
explicitly handle integers for untyped dict type

cosmicBboy:
fixes #3158

Signed-off-by: Niels Bantilan <niels.bantilan@gmail.com>

# TL;DR

Protobuf <> json will auto-cast integers to floats when serializing dict values when the type is annotated with a bare `dict` or `typing.Dict` annotation. This will break code that relies on precise types, such as sklearn, which explicitly checks the types of hyperparameter values in Estimator constructor.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

Because the interchange between protobuf and json casts ints to floats, this PR introduces a workaround to handle converting floats to integers by parsing the protobug `struct_pb2.Value` with regex and inferring integers based on its `__repr__`, which actually preserves the integer representation, e.g. `number_value: 100`, but accessing the actual `number_value` attribute will still return a float `assert value.number_value == 100.0`.

See [here](https://github.com/protocolbuffers/protobuf/blob/main/python/google/protobuf/json_format.py#L704-L705) in the protobuf codebase where ints and floats are treated the same.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3158

## Follow-up issue
_NA_


cosmicBboy:
yeah, the regex approach doesn't actually work... will need to create some kind of special encoding of this in the flyte literal

fg91 (commented):
@cosmicBboy thank you for tackling this! Both at my previous team and my current one, several of the ML engineers ran into this, e.g. for the batch size parameter, causing quite some confusion each time.

I wonder what effect the different serialization format has on how workflow default args of dictionary type are displayed in the flyte console when launching a workflow. Did you check this?

PR #1523, opened 2023-02-25
prefer FLYTE_ prefixed AWS creds env vars

flixr:
# TL;DR
use FLYTE_AWS_ACCESS_KEY_ID and FLYTE_AWS_SECRET_ACCESS_KEY via config if they exist instead of AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY

so that user code can inject and use these default AWS creds env vars

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
flytekit will used the `AWS_ACCESS_KEY_ID` env var instead of the also existing `FLYTE_AWS_ACCESS_KEY_ID` env if the former exists in the container.... this made my task which needs to retrieve data from a different bucket with different creds (injected via secret) fail upon downloading/uploading flyte inputs/output.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/plugins/flytekit-data-fsspec/tests/test_persist.py
+++ b/plugins/flytekit-data-fsspec/tests/test_persist.py
@@ -1,7 +1,7 @@
 import os
 import pathlib
 import tempfile
-from unittest import mock
+import mock
 
 from flytekitplugins.fsspec.persist import FSSpecPersistence, s3_setup_args
 from fsspec.implementations.local import LocalFileSystem
--- a/plugins/flytekit-data-fsspec/tests/test_persist.py
+++ b/plugins/flytekit-data-fsspec/tests/test_persist.py
@@ -1,6 +1,7 @@
 import os
 import pathlib
 import tempfile
+from unittest import mock
 
 from flytekitplugins.fsspec.persist import FSSpecPersistence, s3_setup_args
 from fsspec.implementations.local import LocalFileSystem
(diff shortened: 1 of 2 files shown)


flixr:
@kumare3 would be great if you can take a look at this, since it looks like you refactored that part last...

I haven't tested this yet, but it should do the trick...

flixr:
Friendly ping @kumare3 @wild-endeavor @eapolinario 

This fixes the issue for me and I added a test...

plugins/flytekit-data-fsspec/tests/test_persist.py
```diff
@@ -18,6 +18,17 @@ def test_s3_setup_args():
     kwargs = s3_setup_args(S3Config(access_key_id="access"))
     assert kwargs == {"key": "access"}
 
+    os.environ["AWS_ACCESS_KEY_ID"] = "ignore-user"
```
wild-endeavor, line 21: Could we use this approach for handling the env vars https://stackoverflow.com/questions/31582750/python-mock-patch-os-environ-and-return-value

and also add a few more tests to cover at least the three interesting possibilities?
1. Std env vars present, `FLYTE_` env vars not present
2. Std env vars present, `FLYTE_` env vars also present
3. Std env vars not present, `FLYTE_` env vars present

wild-endeavor: Could you also run a local test using pyflyte run for sanity check please?

1. Build an image with this new flytekit code and push the image to the sandbox registry.
    a. Basically on line [22 in the dockerfile](https://github.com/flyteorg/flytekit/blob/be24c52f3fb8f7949db172f490cca95fa0f0e413/Dockerfile#L22) add a pip command to install this git sha of flytekit.
        It'll look something like
            ```RUN pip install --no-deps -U "https://github.com/flyteorg/flytekit/archive/abcabc.zip#egg=flytekit"
            ```
    b. `docker build . --tag localhost:30000/test_stuff:v1 --build-arg=PYTHON_VERSION=3.10 --build-arg=VERSION=1.4.1`
    c. `docker push localhost:30000/test_stuff:v1`
1. Then `pyflyte run --remote --image localhost:30000/test_stuff:v1 some/workflow.py`

Somewhat annoying I know, but a simple test like this would've prevented the 1.4.0 regression we had earlier today.  just trying to be more careful and we'll work on making this easier/automated.

flixr: updated the tests.

wild-endeavor:
```
RUN pip install --no-deps -U "https://github.com/flyteorg/flytekit/archive/291f9aeb520e8fd1ca6837a6ea78f6722a3d3d16.zip#subdirectory=plugins/flytekit-data-fsspec&egg=flytekitplugins-data-fsspec"
```

plugins/flytekit-data-fsspec/tests/test_persist.py
```diff
@@ -1,6 +1,7 @@
 import os
 import pathlib
 import tempfile
+from unittest import mock
```
wild-endeavor, line 4:
```suggestion
import mock
```

plugins/flytekit-data-fsspec/tests/test_persist.py
flixr: ok, I can confirm that this works as expected in the sandbox

wild-endeavor:
also i don't think the test does anything.  for example this test passes:

```python
@mockenv(
    AWS_ACCESS_KEY_ID="ignore-user",
    AWS_SECRET_ACCESS_KEY="ignore-secret",
    FLYTE_AWS_ACCESS_KEY_ID="flyte",
    FLYTE_AWS_SECRET_ACCESS_KEY="flyte-secret",
)
def test_s3_setup_args_env_both():
    print("hello world")
    raise Exception("hello all")
```

wild-endeavor:
yeah let's drop `mockenv` and just call `mock.patch.dict` directly on each test, that seems to work, just tested it.

flixr:
strange... thanks for catching that.
updated

wild-endeavor (approved):

wild-endeavor:
btw, we're actually deleting all this code and revamping the data layer completely.  I'll port this over to that pr after this is merged.

PR #1535, opened 2023-03-06
[Core feature] Convert List[Any] to a single pickle file

Yicheng-Lu-llll:
# TL;DR
Currently, flyte create N (size of the list) pickle files if the output type is List[Any]. See https://github.com/flyteorg/flyte/issues/3207 for more details.

This PR:
- converts the entire list into a single pickle file by default.
- takes batches of list items and uploads each batch as a separate pickle file if the user annotates the batch size(see below).
```python
@task
def t1() -> Annotated[List[Any],  100]

```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

Originally, in `ListTransformer.to_literal`, Items are converted one by one.
https://github.com/flyteorg/flytekit/blob/8ccb5dde044f9dce3e216a8cea58dd464fb3ab6b/flytekit/core/type_engine.py#L977

This PR checks if python_type contains the type flytePickle. If so, batch the list of items and call `FlytePickleTransformer.to_literal` directly.

   ```python
     if flytePickleTransformer is not None and self.conatinFlytePickle(python_type, flytePickle):
            lit_list = [flytePickleTransformer.to_literal(ctx, python_val[i : i + batchSize], t, expected.collection_type) for i in range(0, len(python_val), batchSize)]  # type: ignore
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3207

## Follow-up issue
_NA_

### CODE REVIEW

File: flytekit/core/type_engine.py [hunk 1]
```diff
@@ -969,22 +969,43 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
         except Exception as e:
             raise ValueError(f"Type of Generic List type is not supported, {e}")
 
+    @staticmethod
+    def isBatchable(t: Type, FlytePickle: Type[T]):
```
pingsutw, line 973:
```suggestion
    def isBatchable(t: Type):
```

flytekit/core/type_engine.py
```diff
@@ -969,22 +969,43 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+            return False
+
     def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
+        from flytekit.types.pickle import FlytePickle
         if type(python_val) != list:
             raise TypeTransformerFailedError("Expected a list")
 
         t = self.get_sub_type(python_type)
-        lit_list = [TypeEngine.to_literal(ctx, x, t, expected.collection_type) for x in python_val]  # type: ignore
+        if self.isBatchable(python_type, FlytePickle):
```
pingsutw, line 988: should use snake case
```suggestion
        if self.is_batchable(python_type, FlytePickle):
```

flytekit/core/type_engine.py
```diff
@@ -969,22 +969,43 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+            batchSize = len(python_val)  # default batch size
```
pingsutw, line 989: ditto

flytekit/core/type_engine.py
```diff
@@ -969,22 +969,43 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+        from flytekit.types.pickle import FlytePickle
         try:
             lits = lv.collection.literals
         except AttributeError:
             raise TypeTransformerFailedError()
-
         st = self.get_sub_type(expected_python_type)
-        return [TypeEngine.to_python_value(ctx, x, st) for x in lits]
+        if self.isBatchable(expected_python_type, FlytePickle):
+            batchList = [TypeEngine.to_python_value(ctx, batch, FlytePickle) for batch in lits]
```
pingsutw, line 1005:
```suggestion
            batch_list = [TypeEngine.to_python_value(ctx, batch, FlytePickle) for batch in lits]
```

flytekit/core/type_engine.py
```diff
@@ -969,22 +969,43 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
         except Exception as e:
             raise ValueError(f"Type of Generic List type is not supported, {e}")
 
+    @staticmethod
+    def isBatchable(t: Type, FlytePickle: Type[T]):
+        print("hi!!!!!!!!!!!!!!",)
```
pingsutw, line 974:
```suggestion
```

flytekit/core/type_engine.py
```diff
@@ -969,22 +969,46 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+
     def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
+        from flytekit.types.pickle import FlytePickle
+
         if type(python_val) != list:
             raise TypeTransformerFailedError("Expected a list")
 
-        t = self.get_sub_type(python_type)
-        lit_list = [TypeEngine.to_literal(ctx, x, t, expected.collection_type) for x in python_val]  # type: ignore
+        if self.is_batchable(python_type):
```
pingsutw, line 988: nit
```suggestion
        if ListTransformer.is_batchable(python_type):
```

pingsutw (commented):
LGTM

flytekit/core/type_engine.py
```diff
@@ -969,22 +969,46 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+            batchSize = len(python_val)  # default batch size
+            # parse annotated to get the number of items saved in a pickle file.
+            if get_origin(python_type) is Annotated:
+                batchSize = get_args(python_type)[1]
```
pingsutw, line 992: we should check `get_args(python_type)[1]` is int here because we may have an output type that looks like this annotated[list, [HashMethod](https://docs.flyte.org/projects/flytekit/en/latest/generated/flytekit.HashMethod.html#flytekit.HashMethod)(...)]

pingsutw: https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/task_cache.html#caching-of-non-flyte-offloaded-objects

pingsutw:
> Prior to these changes, , converting a list with type List[Dict[str, Any]] with 1000 items needs 271.448063737 seconds.
After this PR, convertinga list with type List[Dict[str, Any]] with 1000 items only needs 2.990736610998283 seconds.

Amazing, thanks for testing it

flytekit/core/type_engine.py
Yicheng-Lu-llll: Thank you!

pingsutw (approved):

pingsutw (commented):
Thanks, Have you tried to run it with cache enabled in the sandbox?

flytekit/core/type_engine.py
```diff
@@ -968,22 +968,55 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
         except Exception as e:
             raise ValueError(f"Type of Generic List type is not supported, {e}")
 
+    @staticmethod
+    def is_batchable(t: Type):
```
wild-endeavor, line 972: can we add a bunch of unit tests just for this function

flytekit/core/type_engine.py
```diff
@@ -968,22 +968,55 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+        if get_origin(t) is Annotated:
+            return ListTransformer.is_batchable(get_args(t)[0])
+        if get_origin(t) is list:
+            subtype = get_args(t)[0]
+            if subtype == FlytePickle or (hasattr(subtype, "__origin__") and subtype.__origin__ == FlytePickle):
+                return True
+        return False
+
     def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
+        from flytekit.types.pickle import FlytePickle
```
wild-endeavor, line 988: can we move this down inside the is batchable condition?

flytekit/core/type_engine.py
```diff
@@ -968,22 +968,55 @@ def get_literal_type(self, t: Type[T]) -> Optional[LiteralType]:
...
+                        batchSize = annotation
+                        break
+            lit_list = [TypeEngine.to_literal(ctx, python_val[i : i + batchSize], FlytePickle, expected.collection_type) for i in range(0, len(python_val), batchSize)]  # type: ignore
+        else:
+            t = self.get_sub_type(python_type)
+            lit_list = [TypeEngine.to_literal(ctx, x, t, expected.collection_type) for x in python_val]  # type: ignore
         return Literal(collection=LiteralCollection(literals=lit_list))
 
     def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> typing.List[typing.Any]:  # type: ignore
+        from flytekit.types.pickle import FlytePickle
```
wild-endeavor, line 1008: and also here please

Yicheng-Lu-llll:
> thanks! can we also do a quick end to end test please on sandbox if you haven't already

Certainly! May I know which test you are referring to?  I have tested the following two cases(shown in the PR description): 
- old task's cached output -> new task
- new task(image include this PR) -> old task(image without this PR)

Do you mean to add cases that demonstrate successful execution?

pingsutw (approved):

elibixby:
I think with the `Annotation` functionality I added in #1615 you can accomplish this much more clearly and cleanly without bringing the functionality into TypeEngine, where it adds a lot of complexity, and doesn't even really provide this feature in many cases (for example what if I want to batch types which are already registered in the TypeEngine).


```
BatchedList = Annotated[List[Any], FlytePickleTransformer()]

@task
def my_task(...) -> BatchedList:
   ...
```

Or even using a custom transformer with a fixed batch size, and explicit type;

```

class BatchedList:

    def __class_getitem__(self, type: Type, batch_size: int):
         return  Annotated[List[type],  MyBatchingTransformer(t=type, batch_size=int)]


@task
def my_batching_task(...) -> BatchedList[int, 5E10]:
   ...
```

@Yicheng-Lu-llll Does this meet your needs? 

If so perhaps we can revert this as it adds significant complexity. @wild-endeavor 
I'm sorry I didn't catch this before I went in, the Annotation override feature is newer and not well documented.

Yicheng-Lu-llll:
Hi @elibixby, Thank you for commenting! We can further discuss about this.

> I think with the Annotation functionality I added in https://github.com/flyteorg/flytekit/pull/1615 you can accomplish this much more clearly and cleanly without bringing the functionality into TypeEngine, where it adds a lot of complexity, and doesn't even really provide this feature in many cases (for example what if I want to batch types which are already registered in the TypeEngine).


1. The primary goal of this PR is to address the performance issues. Some users find it is extremely slow if they have large list with type `Any`.  This issue is not apparent with other types. Hence, this PR focuses on batching the `List[Any] `type to enhance performance, while leaving all other types unaffected.

3. In my initial implementation,I also change the behavior of `get_transformer`,  However, as you pointed out, , this approach will lose the ability to change the batch size. From a user's perspective, creating a custom transformer to support batch size can be overly complex. Also, adding batch size to `FlytePickleTransformer` would introduce equivalent complexity.

5. But if we do not care about the batch size, I do strongly agree your approach is much much better. But if we choose to disregard batch size and proceed with your suggestions , more modifications will be required beyond #1615. For example, We need to make sure [translate_inputs_to_literals](https://github.com/flyteorg/flytekit/blob/f671fb68092979b0328b97453d0ba0d396a6398e/flytekit/core/promise.py#L35) and TypeEngine are consistent.

elibixby:
> 2. In my initial implementation,I also change the behavior of `get_transformer`,  However, as you pointed out, , this approach will lose the ability to change the batch size. From a user's perspective, creating a custom transformer to support batch size can be overly complex. Also, adding batch size to `FlytePickleTransformer` would introduce equivalent complexity.

I agree if we were to use annotations, we should provide a convenience wrapper for batch size to users, so they don't need to implement their own type transformer. I disagree that it would introduce equivalent complexity to this PR. It would be a completely isolated subclass of `FlytePickleTransformer`.  (as a side note I am working on a PR right now which alters the way `FlytePickleTransformer` is used to be a default in `get_transformer` rather than a special case only for function signatures as implmented in #1565 )

> 3. If we do not care about the batch size, I do strongly agree your approach is much better. But even we choose to disregard batch size and proceed with your suggestions, more modifications will be required beyond https://github.com/flyteorg/flytekit/pull/1615. For example, We need to make sure [translate_inputs_to_literals](https://github.com/flyteorg/flytekit/blob/f671fb68092979b0328b97453d0ba0d396a6398e/flytekit/core/promise.py#L35) and TypeEngine are consistent.

This is something I've discussed pretty extensively with @wild-endeavor . I think the preferred solution here is to make it so that adding a `TypeTransformer` by itself (either through registration or annotation), is sufficient to fully the support the type in Flyte. 

The main thrust of this is in supporting `pyflyte run` (as you point out), but it could fairly easily be supported by adding a `from_json` to the `TypeTransformer` base class which `pyflyte` uses to unmarshal yaml data. 

Would you be open to a solution that addresses both of these issues through an Annotated type as I've just described?

Yicheng-Lu-llll:
>  I disagree that it would introduce equivalent complexity to this PR. It would be a completely isolated subclass of FlytePickleTransformer

If no batch size is specified, when a list comes in, the `FlytePickleTransformer` can simply transform the entire list into a single pickle file. However, with a batch size, things may be different. For example, we can split the list into three, and then convert them into three separate pickle files. But then, how to store and handle it as a literal seems to be a problem. Because it appears we have to change the literal type (from `blob` to `LiteralCollection`), otherwise we can not simply deal with the three separate pickle files. This is why I said "may introduce equivalent complexity to the current `FlytePickleTransformer`". It would make things easier to handle the batching process within the `ListTransformer`, which naturally supports the `LiteralCollection`.  


> Would you be open to a solution that addresses both of these issues through an Annotated type as I've just described?

Overall make sense to me. But I am not sure how to deal the above concern .

PR #1544, opened 2023-03-09
Make `FlyteFile` compatible with `Annotated[..., HashMethod]`

AdrianoKF:
See issue #3424

# TL;DR
This PR makes `FlyteFile`s usable with offloaded hash calculation via `Annotated[FlyteFile, HashMethod(...)]`, so that they can be cached based on file contents, e.g.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
In `FlyteFilePathTransformer.to_literal()`, `Annotated` type aliases are now
handled by unwrapping the first type argument to the alias, before checking if
they are a subclass of `FlyteFile`.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3424

## Follow-up issue
_NA_


flytekit/types/file/file.py
```diff
@@ -272,6 +272,10 @@ def to_literal(
         if python_val is None:
             raise TypeTransformerFailedError("None value cannot be converted to a file.")
 
+        # Correctly handle `Annotated[FlyteFile, ...]` by extracting the origin type
+        if typing.get_origin(python_type) is typing.Annotated:
```
eapolinario, line 276: import `Annotated` from `typing_extensions` so that we can use it in python 3.8.

AdrianoKF:
> import `Annotated` from `typing_extensions` so that we can use it in python 3.8. 

Fixed here and in test code in aa18baa6

flytekit/types/file/file.py
AdrianoKF: Fixed here and in test code in aa18baa6

flytekit/types/file/file.py
```diff
@@ -272,6 +273,10 @@ def to_literal(
         if python_val is None:
             raise TypeTransformerFailedError("None value cannot be converted to a file.")
 
+        # Correctly handle `Annotated[FlyteFile, ...]` by extracting the origin type
+        if typing.get_origin(python_type) is Annotated:
```
pingsutw, line 277: `get_origin` should also import from typing_extensions

AdrianoKF: Pesky 3.8 compatibility, sorry for the additional turnaround - have been using 3.9+ exclusively for some time, so my mental radar doesn't catch these.

Fixed for `get_origin` and `get_args` in bffa9b8a.

tests/flytekit/unit/core/test_flyte_file.py
```diff
@@ -433,3 +435,19 @@ def wf(path: str) -> os.PathLike:
...
+
+    @task
+    def t1(path: str) -> Annotated[FlyteFile, HashMethod(calc_hash)]:
+        return FlyteFile(path)
+
+    @workflow
+    def wf(path: str) -> None:
+        t1(path=path)
+
+    with tempfile.NamedTemporaryFile() as f:
```
eapolinario, line 452: `NamedTemporaryFile` objects are weird on [windows](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile) (essentially they can't be opened twice). Specifying `delete=False` should fix the issue.

AdrianoKF: Sorry for the slow response, I was OOO last week.

I noticed the `local_dummy_file` fixture in the test module and switched to using that instead in c74dca15.

eapolinario (approved):

PR #1565, opened 2023-03-27
Sketch of non-JSON dataclass serialization

elibixby:
# TL;DR
See  https://github.com/flyteorg/flyte/issues/3359#issuecomment-1484905200 

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -7,6 +7,7 @@ import datetime as _datetime
 import enum
 import inspect
 import json as _json
+import logging
 import mimetypes
 import textwrap
 import typing
@@ -217,6 +218,49 @@ class RestrictedTypeTransformer(TypeTransformer[T], ABC):
 
 
 class DataclassTransformer(TypeTransformer[object]):
+
+    def __init__(self, registry: Type[TypeEngine]):
+        super().__init__(name='dataclass', t=object, enable_type_assertions=True)
+        self._registry = registry
+
+    @property
+    def registry(self) -> Type[TypeEngine]:
+        return self._registry
+
+    def assert_type(self, t: Type[T], v: T):
+        if not dataclasses.is_dataclass(v):
+            raise TypeTransformerFailedError(f'Object {v} must be a dataclass')
+
+    def get_literal_type(self, t: Type[T]) -> LiteralType:
+        return _type_models.LiteralType(simple=_type_models.SimpleType.STRUCT)
+
+    def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
+        fields = {}
+        for field in dataclasses.fields(python_type):
+            sub_val = getattr(python_val, field.name)
+            transformer = self.registry.get_transformer(field.type)
+            fields[field.name] = transformer.to_literal(sub_val).to_flyte_idl()
+
+        struct = _struct.Struct()
+        struct.update(fields)
+        return Literal(scalar=Scalar(generic=struct))
+
+    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> Optional[T]:
+        fields = {}
+        struct: _struct.Struct = lv.scalar.generic
+
+        for field in dataclasses.fields(expected_python_type):
+            if struct.HasField(field.name):
+                transformer = self.registry.get_transformer(field.type)
+                fields[field.name] = transformer.to_python_value(
+                    ctx=ctx,
+                    lv=Literal.from_flyte_idl(struct[field.name]),
+                    expected_python_type=field.type)
+
+        return expected_python_type(**fields)
+
+
+class _JsonDataclassTransformer(TypeTransformer[object]):
     """
     The Dataclass Transformer, provides a type transformer for arbitrary Python dataclasses, that have
     @dataclass and @dataclass_json decorators.
@@ -668,7 +712,7 @@ class TypeEngine(typing.Generic[T]):
 
     _REGISTRY: typing.Dict[type, TypeTransformer[T]] = {}
     _RESTRICTED_TYPES: typing.List[type] = []
-    _DATACLASS_TRANSFORMER: TypeTransformer = DataclassTransformer()  # type: ignore
+    _JSON_DATACLASS_TRANSFORMER: TypeTransformer = _JsonDataclassTransformer()  # type: ignore
 
     @classmethod
     def register(
@@ -764,7 +808,11 @@ class TypeEngine(typing.Generic[T]):
 
         # Step 4
         if dataclasses.is_dataclass(python_type):
-            return cls._DATACLASS_TRANSFORMER
+            if isinstance(python_type, DataClassJsonMixin):
+                logging.warning('dataclasses no longer require `dataclasses_json.dataclasses` annotation')
+                return cls._JSON_DATACLASS_TRANSFORMER
+            else:
+                return DataclassTransformer(registry=cls)
 
         raise ValueError(f"Type {python_type} not supported currently in Flytekit. Please register a new transformer")
 


elibixby:
There doesn't appear to be a valid `LiteralType` for a `LiteralMap` with differently typed values. This doesn't really matter though, because we never need the hint of the `LiteralType` at least in unit tests. I'm not sure what the functionality of the second order typing is... (`get_literal_type` and `guess_python_type`. It doesn't seem to be used anywhere in the codebase outside of the corresponding methods in `TypeEngine`).

tests/flytekit/unit/core/test_dataclass.py
```diff
@@ -27,3 +28,79 @@ def wf() -> AppParams:
 
     res = wf()
     assert res.region == "us-west-3"
+
+
+def test_dataclass_no_json():
+    @dataclass
+    class AppParams(object):
```
pingsutw, line 35: Could we add more type here? like structuredDataset and FlyteDirectory.

pingsutw (commented):
@elibixby Thanks, I like this idea. we should also update flyteConsole, right? otherwise, it won't be able to show the inputs / outputs.

tests/flytekit/unit/core/test_dataclass.py
pingsutw: and Any or Pickle type

elibixby:
> @elibixby Thanks, I like this idea. we should also update flyteConsole, right? otherwise, it won't be able to show the inputs / outputs.

Yeah I think we will need to update both the console (for displaying and building input forms), as well as `flytectl` for generating `execution.yaml` files. I imagine they should see similar code reductions from not needing to deal with jsonSchema, and being able to fall back on pre-existing parsing logic. 

 Would love someone more familiar with these side of things to take this on, or at least provide pointers to what needs to be changed.

tests/flytekit/unit/core/test_dataclass.py
elibixby: I added `StructuredDataset` and `FlyteDirectory` types to the complex type test. Can you clarify what you mean by Any/Pickle types?

pingsutw:
yup. I'll help. let's get it merge first.

tests/flytekit/unit/core/test_dataclass.py
pingsutw: I mean `typing.List[Any]`. https://docs.flyte.org/projects/cookbook/en/latest/auto/core/type_system/flyte_pickle.html

elibixby: I get `ValueError: Type of Generic List type is not supported, Type typing.Any not supported currently in Flytekit. Please register a new transformer`

Is this supposed to work? I don't see registration of a `TypeTransformer` for `typing.Any`. I see the `PickleTransformer` but it doesn't seem to register for `typing.Any`.

elibixby: Ok I've dug in a little more. Some thoughts:

I copied how `test_type_engine.py` tests the pickle transformer, and it works fine. But it is worth noting that for some reason resolution of something as a `FlytePickle` type, happens in `interface.py` and not in `TypeEngine`, which I think is leading to some weird behavior unexpected behavior, due to the interface only following along the generic type tree, rather than recursing via `TypeEngine.get_transformer`. 
Some examples:
* `List[Foo]` works in a function signature
* `List[Foo]` doesn't work in a dataclass
* `Foo` works both in a function signature and a dataclass
* `A dataclass containing a dataclass field `Foo` works in neither a function signature nor a dataclass.
* A custom `MyGeneric[Foo]` where the user  creates a `TypeTransformer` registered for `MyGeneric` will work in neither.

Rather than replacing the type, I would just "fall back" to `PickleTransformer` inside of `TypeEngine.get_transformer` (with a warning ofc). 
* Gets rid of the `FlytePickle` type completely.
* `get_transformer` could no longer raise a `TypeTransformerFailedError` killing a ton of code (67 usages in flytekit).
Instead e.g. `List[Any]` would fall back to pickle for the list type.

I'm not sure if this level of permissiveness is desirable. That said, I don't think it's any more permissive than the current state of things, it's just less arbitrary. (i.e. I couldn't accurately tell you right now in what situations things will "just work" using pickle, and in what situations they won't, so in that sense, having it "just work" in every situation seems better). 

WDYT. I'm happy to bundle that up in this PR, or I can split it out into a separate PR on top of this.

I have a hunch there's a few more inconsistencies between `interface.py` and the way this PR is doing things, but I haven't explored further

ALTERNATIVE APPROACH

My personal feeling is that pickle should be entirely explicit (i.e. just `TypeEngine.register(PickleTransformer(), additional_types=[Foo])` isn't that hard), but if we wanted to leave the current inconsistent approach in place, we could at least add an `Annotation` based escape hatch that doesn't interact with the type registry at all.

```
from flytekit import PickleTransformer

class Foo:
    ...

PickledFoo = Annotated[Foo, PickleTransformer()]

@task
def my_task(foo: PickledFoo) -> PickledFoo:
    ...
```

then in `TypeEngine.get_transformer` when you test for annotation, check for an explicit transformer, and short circuit the search in the registry.

This would get me out a lot of sticky situations with the registry in fact. 

For example when you're trying to define a second more custom transformer for a type that's already defined, and you only want to use some of the time, you could provide a type alias that explicitly uses that transformer.

pingsutw: > My personal feeling is that pickle should be entirely explicit

we can't do that now. it will break people, but can definitely add a hook like you said to override default transformer in flytekit

one of the reason we have pickleType is that we save [original_python_type](https://github.com/flyteorg/flytekit/blob/87c665becca35567063472bcf8477e0d3876f088/flytekit/types/pickle/pickle.py#L36) in the pickle type, and it allows us to check if the type matches between two tasks.

For example, below wf should fail to be registered.
```python
@task
def t1() -> Foo:
  ...

@task
def t2(b: Bar):
  ...

@workflow
def wf():
  t2(b=t1())
```

pingsutw: yes, I think we can remove `interface._change_unrecognized_type_to_pickle ` and `FlytePickle`, but we should keep `pickle format` and pickleTransformer.
If get_transformer() fail to find the transformer for the type, flytekit falls back to `PickleTransformer`

pingsutw: we can add origin_python_type to literalType [here](https://github.com/flyteorg/flytekit/blob/87c665becca35567063472bcf8477e0d3876f088/flytekit/types/pickle/pickle.py#L92) instead

elibixby:
Another note: I needed to rework `DictTransformer` a little. The only reason the tests were passing using non-`str` type keys was because dataclass was using a different code-path for dict serialization. In reality there's no way to serialize an int-key map in protobuf (structs always use string keys). There were even a couple tests that codified this behavior (creating a dict with an int key, and expecting it to be a string key), but this felt really quite confusing to users, so I removed it. Technically this is a breaking change, but I kinda doubt anyone is relying on this functionality given how weird it is.

tests/flytekit/unit/core/test_dataclass.py
elibixby: Sounds good. I took a shot at this if you want to take a look. 

I think the Annotated override is probably worth doing, but seems isolated enough it can probably go in a separate PR

wild-endeavor (commented):
hey @elibixby - thinking about it a bit more, is it a big deal from here to preserve the current behavior for dataclass_json?  that is, if it's a dataclass_json, still use the existing code.  If it's a dataclass dataclass, then use the new logic.

I'm concerned about breaking existing users.  Currently a dataclass_json gets encoded to a literal Scalar Generic.  In the new world it would be a Literal with a map, and also represents a scalar object in Python as a map object in Flyte.  Let me think about this a bit more, it feels a bit awkward right?

cc @pingsutw - what do you think?

elibixby:
@wild-endeavor I agree that leaving old logic in place for backwards compatibility is worth considering especially if there are concerns with flytectl or console version mismatch.

However I don't agree that this is an awkward use of the python <-> proto mapping. A dataclass is much closer to a map than it is to a scalar. A dataclass is defined by having a map `__dataclass_fields__` as a class member. Not only that but my personal experience using flyte has been that the duplicate nesting logic of dataclasses is the source of lots of frustrating unexpected behavior, (why do unions work in functions but not in dataclasses, why are enums serialized differently, why don't custom types work in dataclasses, etc), and by removing json serialization you can see from the test cases that several other weird edge cases are removed as well (forward references breaking dataclasses, etc).

wild-endeavor:
Okay that sounds good.  We can make this new transformer operate as a map/dictionary type.  But yeah if we could preserve the existing dataclass_json behavior, I think that will be good.  So no change for dataclass_json, this new logic for pure dataclass.

is that possible?

And then we can do a generic json transformer, and think about how to pull out the logic from run.py as a second/third step.  what do you think?

pingsutw:
@elibixby let's not remove the code in current dataclass transformer for now. 
Could we rename `dataclassTransformer` to `dataclassJsonTransformer`? and create a new transformer called `dataclassTransformer` that can serialize the python value to literal instead of json schema.  wdyt?

elibixby:
> Any update on this? I think this would help resolve an issue I'm seeing with the existing transformer! Happy to pair/help on any remaining outstanding issues!

Hey, I just haven't had time to restore the original json dataclass transformer to work along side. I've been focused on a couple more pressing issues with the type system that came out of this CL (namely writing an escape hatch for the type system in #1615 and changing pickling to be a universal default rather than only for top level types)

I think everything that is needed for merged is here, just some old code/tests need to be restored alongside the new stuff. If you'd be interested in helping with that I'd appreciate it!

pingsutw (commented):
We need to rename current `DataclassTransformer` to `DataclassJsonTransformer`, and move the code that @elibixby added to `DataclassTransformer`. last, we need to update this [line](https://github.com/flyteorg/flytekit/blob/06fffc786edfe9d4cbb891d6c0e14181a5823998/flytekit/core/type_engine.py#L779). if python_type has @dataclass_json decorator, it should return `DataclassJsonTransformer`; otherwise, it should return `DataclassTransformer`

@ggydush would you like to help updating this PR

PR #1576, opened 2023-04-03
Feat: Add torchrun plugin

fg91:
# TL;DR
Work in progress

This plugin allows running torch elastic (torchrun) distributed training with Flyte.

```python

import torch
from flytekit import dynamic, task, workflow
from flytekitplugins.kfpytorch import PyTorch

from .torch_elastic_task import Elastic


@dataclass_json
@dataclass
class Config:
    lr: float = 1e-5
    bs: int = 64
    name: str = "foo"


@task
def init_model() -> torch.nn.Module:
    model = torch.nn.Linear(11, 22)

    return model


"""
This doesn't start a kubelfow pytorch job yet but a single python task Pod which then
runs a local worker group in sub-processes.
The changes in the flyteidl protobuf definitions, the flytekit python api, and the
flytepropeller (operator) which we need to actually make this distributed on multiple nodes
are easy (see RFC document linked in PR description).
"""
@task(
    task_config=Elastic(
        min_replicas=1,
        max_replicas=1,
        start_method="spawn",
    )
)
def train(config: Config, model: torch.nn.Module) -> tuple[str, Config, torch.nn.Module]:
    import os

    import torch

    local_rank = os.environ["LOCAL_RANK"]

    out_model = torch.nn.Linear(1000, int(local_rank) * 2000 + 1)
    print(f"Training with config {config}")
    config.name = "modified"
    return f"result from local rank {local_rank}", config, out_model


@workflow
def wf(config: Config=Config()) -> tuple[str, Config, torch.nn.Module]:
    model = init_model()
    return train(config=config, model=model)


if __name__ == "__main__":
    print(wf(config=Config()))
```

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/plugins/flytekit-torch-elastic/flytekitplugins/torch_elastic/task.py
+++ b/plugins/flytekit-torch-elastic/flytekitplugins/torch_elastic/task.py
@@ -1,0 +1,117 @@
+from dataclasses import dataclass
+from typing import Any, Callable
+
+from flytekit import FlyteContextManager, PythonFunctionTask
+from flytekit.core.context_manager import ExecutionParameters
+from flytekit.extend import TaskPlugins
+from torch.distributed.launcher.api import LaunchConfig, elastic_launch
+from torch.distributed.run import parse_min_max_nnodes
+
+
+@dataclass
+class Elastic(object):
+    min_replicas: int
+    max_replicas: int
+    start_method: str = "spawn"
+
+
+def mp_helper(fn, kwargs):
+    print("Using start method spawn")
+    import dill
+
+    fn = dill.loads(fn)
+
+    return_val = fn(**kwargs)
+
+    return return_val
+
+
+class PytorchElasticFunctionTask(PythonFunctionTask[Elastic]):
+    """
+    Actual Plugin that transforms the local python code for execution within a spark context
+    """
+
+    _ELASTIC_TASK_TYPE = "torch-elastic"
+
+    def __init__(self, task_config: Elastic, task_function: Callable, **kwargs):
+        super(PytorchElasticFunctionTask, self).__init__(
+            task_config=task_config,
+            task_type=self._ELASTIC_TASK_TYPE,
+            task_function=task_function,
+            **kwargs,
+        )
+
+    def execute(self, **kwargs) -> Any:
+        """
+        This method will be invoked to execute the task. If you do decide to override this method you must also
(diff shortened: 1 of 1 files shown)


plugins/flytekit-torch-elastic/flytekitplugins/torch_elastic/task.py
```diff
@@ -0,0 +1,117 @@
...
+            min_nodes=min_nodes,
+            max_nodes=max_nodes,
+            nproc_per_node=4,
+            rdzv_backend="c10d",
+            max_restarts=1,
+            # rdzv_endpoint = "foo"
+            start_method=self.task_config.start_method,
+        )
+
+        if self.task_config.start_method == "spawn":
```
fg91, line 65: @kumare3 what do you think of this workaround?

Unfortunately we can't just pass `self._task_function` to `elastic_launch` since it is not pickleable.

fg91:
@kumare3 I also opened draft PRs in [flyteidl](https://github.com/flyteorg/flyteidl/pull/389) and [flyteplugins](https://github.com/flyteorg/flyteplugins/pull/342). Please note that these are super early prototypes because I wanted to first hack everything together to derisk any potential deal breakers such as the limitations due to pickle.

I think this could be part of the existing kfpytorch plugin as an optional install `pip install flytekitplugins-kfpytorch[elastic]` (since it will have torch as a dependency).

In flytekit, this could look like this:

```python
@task(
    task_config=PyTorch,
        num_workers=2,
        elastic_policy=ElasticPolicy(  # <- optional
            n_proc_per_node = ...,
            ....
        ),
    ),
)
def train():
```

In flyteplugins most of the existing logic could be reused. We would only need a check whether the user configured an elastic policy and if so, add this to the PytorchJob object. I [quickly hard-coded this in this draft PR](https://github.com/flyteorg/flyteplugins/pull/342/files#diff-eef9f32845c5409f19826b00bc2b9e5aad792dd23a89abccdf096947b6082830).

ByronHsu:
Do you mind creating an issue in [flyte mp](https://github.com/flyteorg/flyte/issues/3123) and reference the issues in all related prs. By that way, we can have a central place to track the feature

PR #1579, opened 2023-04-04
Eager workflows to support async workflows

cosmicBboy:
Fixes https://github.com/flyteorg/flyte/issues/3672

# TL;DR

Implement an experimental API for "eager workflows": workflows that enable the use of Python async code.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Flytesnacks Example Link
...

## Complete description

This PR adds support for `@eager` workflows. These are similar to `@dynamic` workflows in that they are flyte tasks that dynamically execute a graph. However, dynamic workflows can only use the materialized inputs to inform the shape of the graph -- at runtime, the flyte backend still gets a statically compiled workflow, but it's one that's compiled at runtime. `@eager` workflows take this to the other extreme of the static <-> dynamic spectrum.

Under the hood, an `@eager` "workflow" is actually a task that has access to a `FlyteRemote` object. This object needs to be configured correctly so that, at run-time, the remote object can authenticate to the configured flyte cluster. The eager workflow then uses this remote object to invoke the tasks that inside the user-defined eager workflow function body to invoke individual task executions.

## Testing

Use the following Dockerfile and script to test this feature:

```Dockerfile
# Dockerfile.eager
FROM python:3.8-slim-buster

WORKDIR /root
ENV VENV /opt/venv
ENV LANG C.UTF-8
ENV LC_ALL C.UTF-8
ENV PYTHONPATH /root

RUN apt-get update && apt-get install -y build-essential git

# Install the AWS cli separately to prevent issues with boto being written over
RUN pip3 install awscli

ENV VENV /opt/venv
# Virtual environment
RUN python3 -m venv ${VENV}
ENV PATH="${VENV}/bin:$PATH"

# Install Python dependencies
COPY . /root/
WORKDIR /root

RUN pip install -e .
RUN pip install -e ./plugins/flytekit-deck-standard
RUN pip install scikit-learn pandas

ENV FLYTE_SDK_LOGGING_LEVEL 20
```

Python script:
```python
# flytekit_eager.py
"""Async workflows prototype."""

import asyncio
import time
from typing import NamedTuple

import pandas as pd
from sklearn.datasets import load_wine
from sklearn.linear_model import LogisticRegression
from sklearn.model_selection import train_test_split
from sklearn.metrics import accuracy_score

from flytekit import workflow, task
from flytekit.configuration import Config, PlatformConfig


class CustomException(Exception): ...

BestModel = NamedTuple("BestModel", model=LogisticRegression, metric=float)


@task
def get_data() -> pd.DataFrame:
    """Get the wine dataset."""
    return load_wine(as_frame=True).frame


@task
def process_data(data: pd.DataFrame) -> pd.DataFrame:
    """Simplify the task from a 3-class to a binary classification problem."""
    return data.assign(target=lambda x: x["target"].where(x["target"] == 0, 1))


@task
def train_model(data: pd.DataFrame, hyperparameters: dict) -> LogisticRegression:
    """Train a model on the wine dataset."""
    features = data.drop("target", axis="columns")
    target = data["target"]
    return LogisticRegression(max_iter=3000, **hyperparameters).fit(features, target)


@task
def evaluate_model(data: pd.DataFrame, model: LogisticRegression) -> float:
    """Train a model on the wine dataset."""
    features = data.drop("target", axis="columns")
    target = data["target"]
    return float(accuracy_score(target, model.predict(features)))


remote = FlyteRemote(
    config=Config.for_sandbox(),
    default_domain="development",
)

@eager(remote=remote)
async def main() -> BestModel:
    logging.info("STARTING MAIN")
    data = await get_data()
    processed_data = await process_data(data=data)

    # split the data
    try:
        train, test = train_test_split(processed_data, test_size=0.2)
    except Exception as exc:
        raise CustomException(str(exc)) from exc

    models = await asyncio.gather(*[
        train_model(data=train, hyperparameters={"C": x})
        # for x in [0.1, 0.01, 0.001, 0.0001, 0.00001]
        for x in [0.1]
    ])
    results = await asyncio.gather(*[
        evaluate_model(data=test, model=model) for model in models
    ])

    best_model, best_result = None, float("-inf")
    for model, result in zip(models, results):
        if result > best_result:
            best_model, best_result = model, result

    assert best_model is not None, "model cannot be None!"
    return best_model, best_result


@workflow
def wf() -> BestModel:
    return main()


if __name__ == "__main__":
    print("training model")
    model = asyncio.run(main())
    print(f"trained model: {model}")

```

## Tracking Issue
https://github.com/flyteorg/flyte/discussions/3396

## Follow-up issue
_NA_


flytekit/experimental/eager_function.py
```diff
@@ -0,0 +1,569 @@
...
+
+        @task
+        def add_one(x: int) -> int:
+            return x + 1
+
+        @task
+        def double(x: int) -> int:
+            return x * 2
+
+        @eager
```
pingsutw, line 373: qq: are we able to pass image or image spec here?

flytekit/core/base_task.py
```diff
@@ -498,9 +506,72 @@ def compile(self, ctx: FlyteContext, *args, **kwargs):
...
+                output_deck.append(TypeEngine.to_html(ctx, v, self.get_type_for_output_var(k, v)))
+
+            if ctx.execution_state and ctx.execution_state.is_local_execution():
+                # When we run the workflow remotely, flytekit outputs decks at the end of _dispatch_execute
+                _output_deck(self.name.split(".")[-1], new_user_params)
+
+    async def _async_execute(self, native_inputs, native_outputs, ctx, exec_ctx, new_user_params):
+        out = await native_outputs
+        native_outputs = self.post_execute(new_user_params, native_outputs)
+        literals_map, native_outputs_as_map = self._output_to_literal_map(out, exec_ctx)
```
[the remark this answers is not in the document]
cosmicBboy: I think this needs to be `out`, since `native_outputs` is a python awaitable no?

flytekit/experimental/eager_function.py
```diff
@@ -0,0 +1,569 @@
...
+        async def eager_workflow(x: int) -> int:
+            try:
+                out = await add_one(x)
+            except EagerException:
+                # The ValueError error is caught
+                # and raised as an EagerException
+                raise
+            return await double(one)
```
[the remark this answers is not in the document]
cosmicBboy: yes, need to update the docstring

flytekit/experimental/eager_function.py
```diff
@@ -0,0 +1,569 @@
...
+        if getattr(self.entity, "execution_mode", None) == PythonFunctionTask.ExecutionBehavior.DYNAMIC:
+            raise EagerException(
+                "Eager workflows currently do not work with dynamic workflows. "
+                "If you need to use a subworkflow, use a static @workflow or nested @eager workflow."
+            )
+
+        if self.execution_state in {
+            ExecutionState.Mode.LOCAL_TASK_EXECUTION,
+            ExecutionState.Mode.LOCAL_WORKFLOW_EXECUTION,
+        }:
```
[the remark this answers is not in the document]
cosmicBboy: 👍

flytekit/experimental/eager_function.py
```diff
@@ -0,0 +1,569 @@
...
+
+        url = self.remote.generate_console_url(execution)
+        node = AsyncNode(self, entity_name, execution, url)
+        self.async_stack.set_node(node)
+        logger.debug(url)
+
+        for _ in range(self._n_polls):
+            execution = self.remote.sync(execution)
+            if execution.closure.phase in {WorkflowExecutionPhase.FAILED}:
+                raise EagerException(f"Error executing {self.entity.name} with error: {execution.closure.error}")
```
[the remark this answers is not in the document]
cosmicBboy: this line: https://github.com/flyteorg/flytekit/pull/1579/files#diff-a9c76bab852a74a971a267146887f4e3dc32b350e127300e65120cc92d22e2b3R478-R480

should be able to run the cleanup function in case any error happens in the eager workflow

flytekit/experimental/eager_function.py
cosmicBboy: yep! you can forward any kwarg that `@task` accepts

cosmicBboy:
@pingsutw please feel free to make the necessary changes.

pingsutw (approved):

flytekit/core/tracker.py
```diff
@@ -92,6 +137,29 @@ def find_lhs(self) -> str:
                 # continue looping through m.
                 logger.warning("Caught ValueError {} while attempting to auto-assign name".format(err))
 
+        # try to find object in module when the tracked instance is defined in the __main__ module
+        module = import_module_from_file(self._instantiated_in, self._module_file)
+
+        def _candidate_name_matches(candidate) -> bool:
+            if not hasattr(candidate, "name") or not hasattr(self, "name"):
```
wild-endeavor, line 144: can we add a comment as to when is this the case?

flytekit/core/tracker.py
```diff
@@ -77,7 +122,7 @@ def find_lhs(self) -> str:
             raise _system_exceptions.FlyteSystemException(f"Object {self} does not have an _instantiated in")
 
         logger.debug(f"Looking for LHS for {self} from {self._instantiated_in}")
-        m = _importlib.import_module(self._instantiated_in)
+        m = importlib.import_module(self._instantiated_in)
         for k in dir(m):
```
wild-endeavor, line 126: previously if this loop didn't raise anything we would just raise an exception on line 96.  Can you elaborate on the circumstances in which we want to keep going?  is it just to support the case when the task is in __main__?

flytekit/core/tracker.py
```diff
@@ -92,6 +137,29 @@ def find_lhs(self) -> str:
...
 
+        # try to find object in module when the tracked instance is defined in the __main__ module
+        module = import_module_from_file(self._instantiated_in, self._module_file)
+
+        def _candidate_name_matches(candidate) -> bool:
+            if not hasattr(candidate, "name") or not hasattr(self, "name"):
+                return False
+            return candidate.name == self.name
+
+        for k in dir(module):
```
wild-endeavor, line 148: are we looking for an exact object match here?  why can't we just do object id equality if that's the case?

flytekit/core/tracker.py
```diff
@@ -216,6 +284,13 @@ def get_absolute_module_name(self, path: str, package_root: typing.Optional[str]
 _mod_sanitizer = _ModuleSanitizer()
 
 
+def _task_module_from_callable(f: Callable):
```
wild-endeavor, line 287: can we type hint this and add a small comment for what the three things are?   also why is this only for Callable?  because all tasks are callable?

flytekit/core/promise.py
```diff
@@ -1026,7 +1027,7 @@ def flyte_entity_call_handler(
                     return create_task_output(vals, cast(SupportsNodeCreation, entity).python_interface)
                 else:
                     return None
-            return cast(LocallyExecutable, entity).local_execute(child_ctx, **kwargs)
+            return cast(LocallyExecutable, entity).local_execute(ctx, **kwargs)
```
eapolinario, line 1030: why do we need this change?

flytekit/core/tracker.py
cosmicBboy: correct, this happens because eager workflows are just using a FlyteRemote object to execute tasks that are in the namespace at execution time. All the logic below is to try to load the module from the `_module_file`, hence we cannot find an exact match because the task functions are no longer equal to the equivalent functions in `main`

flytekit/core/tracker.py
cosmicBboy: not all the attributes in the module will have a `name` attr

flytekit/core/tracker.py
cosmicBboy: because `module` here is now a different module from `main`, task/dynamic/workflow functions can no longer be identified via equality checks.

PR #1581, opened 2023-04-05
Measuring the time taken for each component when executing a task

Yicheng-Lu-llll:
# TL;DR
The concept involves measuring the time taken by each component when executing a task, including both Flytekit and user-defined parts. Users have the option to view a timeline graph by setting the `disable_deck parameter=False`. For example:
```python
from flytekit import task, Resources, workflow
from flytekit.core.utils import timeit


@task(
    disable_deck=False,
    limits=Resources(mem="4Gi", cpu="1"),
)
def t1():
    import time

    # Test timeit used as a decorator
    @timeit("Download data from s3")
    def download_data():
        time.sleep(1)

    download_data()

    # Test timeit used as a context manager
    # Simulate a user using a very long name for a time measurement
    with timeit("Test long string "*20):
        time.sleep(1)

    # Simulate multiple time measurements
    for i in range(10):
        with timeit(f"Run small tasks {i}"):
            # The 'print' function will execute quickly compared to 'time.sleep(1)'.
            # This is used to test if the timeline graph can accurately represent
            # measurements with significantly varying execution times.
            print("hello world")     


@workflow
def wf():
    t1()

if __name__ == "__main__":
    wf()
```

Tests(also tested in sandbox):
image

image

image

Note:
- Setting the disable_deck parameter to True will only affect whether the user can see the timeline graph(whether HTML generated). The data will still be collected and stored in the background. However, for input and output decks, if the disable_deck parameter is set to True, these decks will not be created at all. 
## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

### DIFF
--- a/plugins/flytekit-mlflow/tests/test_mlflow_tracking.py
+++ b/plugins/flytekit-mlflow/tests/test_mlflow_tracking.py
@@ -29,4 +29,4 @@ def train_model(epochs: int):
 
 def test_local_exec():
     train_model(epochs=1)
-    assert len(flytekit.current_context().decks) == 5  # mlflow metrics, params, input, and output
+    assert len(flytekit.current_context().decks) == 5  # mlflow metrics, params, timeline, input, and output
--- a/plugins/flytekit-mlflow/tests/test_mlflow_tracking.py
+++ b/plugins/flytekit-mlflow/tests/test_mlflow_tracking.py
@@ -29,4 +29,4 @@ def train_model(epochs: int):
 
 def test_local_exec():
     train_model(epochs=1)
-    assert len(flytekit.current_context().decks) == 4  # mlflow metrics, params, input, and output
+    assert len(flytekit.current_context().decks) == 5  # mlflow metrics, params, input, and output
--- a/tests/flytekit/unit/deck/test_deck.py
+++ b/tests/flytekit/unit/deck/test_deck.py
@@ -1,3 +1,5 @@
+import datetime
+
 import pandas as pd
 import pytest
 from mock import mock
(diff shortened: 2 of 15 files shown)


tests/flytekit/unit/deck/test_deck.py
```diff
@@ -23,12 +25,30 @@ def test_deck():
...
+        ProcessTime=1.0,
+    )
+    ctx = FlyteContextManager.current_context()
+    ctx.user_space_params._decks = []
+    time_line_deck = ctx.user_space_params.time_line_deck
+    time_line_deck.append_time_info(time_info)
+    assert time_line_deck.name == "time line"
+    assert len(time_line_deck.time_info) == 1
+    assert time_line_deck.time_info[0] == time_info
+    assert len(ctx.user_space_params.decks) == 1
```
Yicheng-Lu-llll, line 43: Unlike other decks, time_line_deck will the conversion of data to HTML is delayed until the html property is accessed since using partial data to render timeline graph is meaningless.


So, the time_line_deck.html file makes use of both GanttChartRenderer and TableRenderer for visualization. To enable these renderers, the 'flytekit-deck-standard' pluginneed be installed. However, as this plugin is not included in the ci testing environment, `time_line_deck.html` is not incorporated in this test.

plugins/flytekit-mlflow/tests/test_mlflow_tracking.py
```diff
@@ -29,4 +29,4 @@ def train_model(epochs: int):
 
 def test_local_exec():
     train_model(epochs=1)
-    assert len(flytekit.current_context().decks) == 4  # mlflow metrics, params, input, and output
+    assert len(flytekit.current_context().decks) == 5  # mlflow metrics, params, input, and output
```
Yicheng-Lu-llll, line 32: [TODO] Add comment: `# mlflow metrics, params, timeline, input, and output`

pingsutw (approved):

PR #1582, opened 2023-04-05
Feature/3506: Improve Type Conversion errors, use rich to prettify error messages

cosmicBboy:
# TL;DR

This PR makes input/output type conversion error messages a lot more useful by providing more context to the user.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added
 - ~~[ ] Code documentation added~~
 - ~~[ ] Any pending items have an associated Issue~~

## Complete description

This PR makes input/output type conversion error messages a lot more actionable by:
- providing context on what workflow and task was executing when the error was raised
- providing context on the input name / output name and positional index that raised the error
- modifying the string representation of flyte literal types to be more explicit
- pretty-printing errors using `rich`

## Appearance Changes

Given the following script:

```python
import random

@task
def add_rand(n: int) -> float:
    return float(n + random.randint(-1000, 1000))

@task
def bad_types(a: int) -> float:
    return str(a)

@workflow
def wf(a: int):
    bad_types(a=add_rand(n=a))

@workflow
def wf_bad_output(a: int) -> int:
    return good_types(a=add_rand(n=a))
```

### Running `wf(a=1)` (incorrect type in intermediate task)

Before:
image

After:
image

### Running `wf(a=1.0)` (incorrect input type to workflow)

Before:
image

After:
image

### Running `wf_bad_output(a=1)` (incorrect workflow output)

Before:
image

After:
image

### Running `bad_types(a=1)` (incorrect task output)

Before:
image

After:
image

### Running `bad_types(a=1)` (incorrect task output)

Before:
image

After:
image

### Running `bad_types(a=str(1))` (incorrect task input)

Before:
image

After:
image

### Logging

Json format:

image

Text format:

image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3506

## Follow-up issue
_NA_


wild-endeavor (commented):
need to play around with this, but +1 for the most part.  I'm okay adding rich to it.  in general though we have way too many dependencies.  It's becoming a problem and i think at least partly why the load times have been so slow.  But one more won't make the problem meaningfully worse.

do we set the env var to be non-json in pyflyte run/register?

cosmicBboy:
> do we set the env var to be non-json in pyflyte run/register?

Planning on doing that in another PR

flytekit/loggers.py
```diff
@@ -63,10 +76,14 @@
             child_logger.setLevel(logging.WARNING)
 
 # create formatter
-formatter = jsonlogger.JsonFormatter(fmt="%(asctime)s %(name)s %(levelname)s %(message)s")
+logging_fmt = os.environ.get(LOGGING_FMT_ENV_VAR, "json")
```
wild-endeavor, line 79: so no change in the default case right?

cosmicBboy: correct, will update the logging format in a different PR

flytekit/loggers.py
```diff
@@ -33,8 +36,18 @@
 user_space_logger = child_loggers["user_space"]
 
 # create console handler
-ch = logging.StreamHandler()
```
honnix, line 36: This is unfortunately a breaking change but I guess a minor version bump (1.6.0) is fine? We might want to highlight this in release notes.

kumare3: Cc @wild-endeavor / @eapolinario

PR #1593, opened 2023-04-16
Backfill should fill in the right input vars

kumare3:
# TL;DR
- Backfill was using incorrect arguments
- backfill should use the argument that user provides or none at all

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3599

### DIFF
--- a/flytekit/clis/sdk_in_container/backfill.py
+++ b/flytekit/clis/sdk_in_container/backfill.py
@@ -168,7 +168,7 @@ def backfill(
             execute=execute,
             parallel=parallel,
         )
-        if entity:
+        if not dry_run:
             console_url = remote.generate_console_url(entity)
             if execute:
                 click.secho(f"\n Execution launched {console_url} to see execution in the console.", fg="green")
--- a/flytekit/clis/sdk_in_container/backfill.py
+++ b/flytekit/clis/sdk_in_container/backfill.py
@@ -168,11 +168,12 @@ def backfill(
             execute=execute,
             parallel=parallel,
         )
-        if not dry_run:
-            console_url = remote.generate_console_url(entity)
-            if execute:
-                click.secho(f"\n Execution launched {console_url} to see execution in the console.", fg="green")
-                return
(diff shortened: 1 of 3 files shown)


MrMathias (commented):
Thank you for being so quick with a fix! :) 

I have two thoughts:

If we accept one input argument only (kick_off_time), I think this code section should reflect that, by expecting, naming and type checking this one "kick_off_time" input.

If we backfill "previous 3 days", I don't think I understand what that would mean without a kick_of_time input (or without a scheduled launch plan)? - I image no lp schedule would mean no launches at all and no "kick_of_time" input would mean n launches of identical workflow contexts, resolving into n identical task executions.

My brain might be molded a bit round Airflow's approach, but I hope it makes sense.

samhita-alla:
@kumare3, the checks are failing. 
https://github.com/flyteorg/flytekit/actions/runs/4711901760/jobs/8356566316?pr=1593

kumare3:
@MrMathias I do think you are thinking like Airflow :). Remember in Flyte, we know when a launchplan is scheduled. `backfill` works only for launchplans which are scheduled and specifically scheduled using the CronSchedule. currently fixed_rate scheduling is not supported.
Thus the backfill system will auto-generate a plan which is a workflow that will contain invocations of the `scheduled` launchplan, with kick_off_time (whatever argument) fixed to the time instance in the schedule. 
This uses 2 properties of Flyte
1. Scheduled launchplans can always be launched manually, simply specifying a time input
2. A workflow can implicitly launch - launchplans - by statically verifying the interface.

A final property is, that the kick_off_time value is locked at workflow generation time. So every invocation of the workflow will result in the same execution.

samhita-alla (approved):

wild-endeavor (approved):

PR #1603, opened 2023-04-23
Enable torch elastic training (torchrun)

fg91:
# TL;DR
This PR brings [torch elastic training (`torchrun`)](https://pytorch.org/docs/stable/elastic/run.html) to the pytorch plugin:

```python
from flytekitplugins.kfpytorch import Elastic

@task(
    task_config=Elastic(
        replicas=4,
        nproc_per_node=4,
        ...
    ),
    ...
)
def train(...):
    ...
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
See https://github.com/flyteorg/flyte/issues/3614 for motivation and a more detailed description.

## Tracking Issue
[https://github.com/flyteorg/flyte/issues/3614](https://github.com/flyteorg/flyte/issues/3614)

## Follow-up issue
-


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -46,9 +78,174 @@ def __init__(self, task_config: PyTorch, task_function: Callable, **kwargs):
...
+    def execute(self, **kwargs) -> Any:
+        """
+        This method will be invoked to execute the task.
+
+        Handles the exception scope for the `_execute` method.
+        """
+        from flytekit.exceptions import scopes as exception_scopes
+
+        if self.execution_mode == self.ExecutionBehavior.DEFAULT:
+            return exception_scopes.user_entry_point(self._execute)(**kwargs)
```
fg91, line 224: @kumare3 do we need the else case/dynamic here? Coped from python func task.

kumare3: i do not think you need to worry about that

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -29,6 +31,36 @@ class PyTorch(object):
...
+        min_replicas int: Lower limit for the number of replicas to which the training job can scale down
+        max_replicas int: Upper limit for the number of replicas to which the training job can scale up.
+            Cannot be smaller than min_replicas.
+        nproc_per_node (Union[int, str]): Number of workers per node. Supported values are [auto, cpu, gpu, int].
+        start_method (str): Multiprocessing start method to use when creating workers.
+        monitor_interval (int): Interval, in seconds, to monitor the state of workers.
+        max_restarts (int): Maximum number of worker group restarts before failing.
+    """
+
+    replicas: int = 1
```
kumare3, line 55: can we just get rid of replicas?
and just make num_workers = max_replicas?
is that not right?
is replicas supposed to be desired_replicas?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -213,19 +212,6 @@ def fn_partial():
...
-    def execute(self, **kwargs) -> Any:
-        """
-        This method will be invoked to execute the task.
-
-        Handles the exception scope for the `_execute` method.
-        """
-        from flytekit.exceptions import scopes as exception_scopes
-
-        if self.execution_mode == self.ExecutionBehavior.DEFAULT:
-            return exception_scopes.user_entry_point(self._execute)(**kwargs)
```
fg91: @kumare3 I think we do need to wrap the execution with `exception_scopes.user_entry_point()`. I first didn't have this and every exception in user code appeared to lead to a retry even if `retries=0`. If I understand correctly, because without this any user-code exception is counted against the system failures retry budget?

kumare3: I agree, I was going to refactor the base class to make execute the leaf method so that we won't have to do this again. 
But we'll let it be, put the if block and do it again.
Sorry my bad

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
kumare3: @fg91 I seem to agree. but I was thinking instead of nnodes keep min_replica and max_replica? Why not? though I will go with whatever you decide.
I actually am not really sure if HPA works with gloo correctly. Its not really easy to change the membership and work correctly. lets reserve this for later. Lets get the basic version out. If you make the change I can +1 tonight and merge. I have been testing on single node and it seems to work great.

fg91: I have a slight preference for `nnodes=1 / "1:2"` over `min_replicas`, `max_replicas` for the following reasons:

* Not introduce a 3rd way in addition to `torchrun`'s `nnodes` and Kubeflow's `replicas`, `minReplicas`, `maxReplicas` that neither user knows
* I think most users will want to have a fixed number of workers but use `torchrun` simply because either a project like Alpacca or a library like ignite assumes it or because they want to do distributed training on a single node without operator.  It should be most simple for those users who don't think about min/max and I feel that `nnodes=2` is simpler for them than `min_replicas=2`, `max_replicas=2`.

If it's ok for you I will change back to `nnodes` and ping you again for review :)

fg91: https://github.com/flyteorg/flytekit/pull/1603/commits/5552413312f6a662cec13661c75befebb4d2497c

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: https://github.com/flyteorg/flytekit/pull/1603/commits/4b2ce0c1d95db731b3e691f203dc47c788ee2d62

This should work as well right?
I tested by running an elastic task from within a dynamic subworkflow.

The if/else is only because both `@task` and `@dynamic` inherit from this class?

ByronHsu:
any updates on this pr? We would love to try this out.

kumare3:
This is working, we will be merging soon

kumare3 (approved):

PR #1620, opened 2023-05-05
Feat: Add pydantic plugin with `BaseModel` type transformer

fg91:
# TL;DR
The plugin adds type support for pydantic `BaseModel`.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed: The transformer does not yet support all Flyte types in a pydantic model (e.g. data frames)
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

The plugin allows passing pydantic models to and between tasks:

## Type Example
```python
import flytekitplugins.pydantic
class Config(BaseModel):
    lr: float = 1e-3
    batch_size: int = 32
@task
def train(cfg: Config):
    ...
```

## Tracking Issue
[https://github.com/flyteorg/flyte/issues/2686](https://github.com/flyteorg/flyte/issues/2686)

## Follow-up issue
_NA_


ArthurBook:
Hi

super useful. thanks for sharing.
Trying this on my end but couldnt get it to play with the flytefiles/directories because they require arbitrary_types_allowed which disables the .schema() method. Here is an example :

class Rax(pydantic.BaseModel, arbitrary_types_allowed=True):
    hax: directory.FlyteDirectory

    @classmethod
    def from_json(cls, json: str) -> "Rax":
        self = cls.parse_raw(json)
        return self
    
Any thoughts on how to best solve it? I am looking at the built in DataclassTransformer

fg91:
Closing in favor of https://github.com/flyteorg/flytekit/pull/1660 where @ArthurBook improved upon this implementation of the pydantic base model type transformer.

PR #1629, opened 2023-05-10
Fallback to python task if worker is zero for pytorch

ByronHsu:
# TL;DR

Fallback to python task if worker is zero for pytorch

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyteplugins/pull/348

### DIFF
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -65,12 +65,16 @@ class PyTorchFunctionTask(PythonFunctionTask[PyTorch]):
     """
 
     _PYTORCH_TASK_TYPE = "pytorch"
+    _PYTORCH_TASK_TYPE_STANDALONE = "python-task"
 
     def __init__(self, task_config: PyTorch, task_function: Callable, **kwargs):
+
+        task_type = self._PYTORCH_TASK_TYPE_STANDALONEE if task_config.num_workers == 0 else self._PYTORCH_TASK_TYPE
+
         super().__init__(
             task_config,
             task_function,
-            task_type=self._PYTORCH_TASK_TYPE,
+            task_type=task_type,
             **kwargs,
         )
 
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -180,7 +180,7 @@ jobs:
           tags: ${{ steps.external-plugin-service-names.outputs.tags }}
           build-args: |
             VERSION=${{ needs.deploy.outputs.version }}
-          file: ./Dockerfile.external-plugin-service
+          file: ./Dockerfile
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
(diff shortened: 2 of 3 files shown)


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -65,12 +65,16 @@ class PyTorchFunctionTask(PythonFunctionTask[PyTorch]):
...
+
+        task_type = self._PYTORCH_TASK_TYPE_STANDALONEE if task_config.num_workers == 0 else self._PYTORCH_TASK_TYPE
+
         super().__init__(
             task_config,
             task_function,
-            task_type=self._PYTORCH_TASK_TYPE,
+            task_type=task_type,
             **kwargs,
         )
```
fg91, line 80: Two questions:

* I think we might need to handle `workers=0` separately in `get_custom` since in this case we don't want to create a `PytorchJob`. (See how this is done for elastic task below)
* Currently, `task_config=Pytorch(workers=0)` is equivalent to no `task_config` at all. However, `torch.distributed.init_process_group()` will not work without the env vars set by the operator. We could solve this by overwriting the `execute` method and simply setting the env vars `WORLD_SIZE=1`, `RANK=0`, and potentially the master address (would have to try whether it is required).

wild-endeavor (commented):
maybe @peridotml can chime in on this pr?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
peridotml: What do you y'all think about throwing an error if workers=0 and telling people to use a standard python config if they want to run it on a single machine?

If people really want to set workers to 0 then I understand having a smooth fallback, but otherwise it could confuse people if they make a mistake.

fg91: @ByronHsu @wild-endeavor the new pytorch elastic task can run locally and in a single k8s pod but also with multiple workers using kubeflow training operator. I'd say its functionality is a superset of the already existing `PyTorch` task config. What do you think about using this one in order to debug dist training with a single worker @ByronHsu ?

I think falling back to a normal pod (without kubeflow operator) when doing `task_config=PyTorch(num_workers=0)` doesn't make much sense because the env vars like `MASTER_ADDR`, `RANK`, ... required by `torch.distributed.init_process_group()`, ... will not be set, neither by the kubeflow operator, nor by the pytorch task logic and distributed training, thus, cannot be tested.

I would propose to either allow `num_workers=0` in `PyTorch` task but use kubeflow training operator also in this case (when users don't want to use the training operator, they can use `Elastic`) or 2) not allow `num_workers=0` as is the case now.

kumare3:
should this be closed?

PR #1637, opened 2023-05-14
feat: add py.typed

tekumara:
# TL;DR

Add `py.typed` file so that when type checkers are run in codebases that use flytekit, its types will be checked.

Resolves https://github.com/flyteorg/flyte/issues/3681

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

Add `py.typed` file

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3681


kumare3:
It's more than adding the file, we saw that some types are not correct yet

tekumara:
Yeh, I've noticed them in pyright, which will use the types whether there a `py.typed` or not, see for example #3682

tekumara:
Yeh, I've noticed that with pyright, which will use the types whether there is a `py.typed` or not, see for example [#3682](https://github.com/flyteorg/flyte/issues/3682) ... 

But adding this, it will flush out these issues for users of mypy.

tekumara:
Oh I also see the types will also need to be exported to avoid:

```
  test.py:4:22 - error: "task" is not exported from module "flytekit"
    Import from "flytekit.core.task" instead (reportPrivateImportUsage)
  test.py:4:28 - error: "workflow" is not exported from module "flytekit"
```  

as per either:
- [PEP 008](https://www.python.org/dev/peps/pep-0008/#id50) states modules should explicitly declare the names in their public API using the `__all__` attribute .
- [PEP 484](https://www.python.org/dev/peps/pep-0484/#stub-files) also mentions that imported modules are not considered exported unless they use the `import ... as ...` syntax or equivalent.

tekumara:
Oh I also see the types will need to be exported to avoid:

```
  test.py:4:22 - error: "task" is not exported from module "flytekit"
    Import from "flytekit.core.task" instead (reportPrivateImportUsage)
  test.py:4:28 - error: "workflow" is not exported from module "flytekit"
```  

as per either:
- [PEP 008](https://www.python.org/dev/peps/pep-0008/#id50) states modules should explicitly declare the names in their public API using the `__all__` attribute .
- [PEP 484](https://www.python.org/dev/peps/pep-0484/#stub-files) also mentions that imported modules are not considered exported unless they use the `import ... as ...` syntax or equivalent.

kumare3:
@eapolinario how can we get this added into flytekit? what needs to be done?

PR #1641, opened 2023-05-16
Pass verify flag to all authenticators

ByronHsu:
# TL;DR

Pass verify flag to all authenticators

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

### DIFF
--- a/flytekit/clients/auth/authenticator.py
+++ b/flytekit/clients/auth/authenticator.py
@@ -48,7 +48,14 @@ class Authenticator(object):
     Base authenticator for all authentication flows
     """
 
-    def __init__(self, endpoint: str, header_key: str, credentials: Credentials = None, http_proxy_url: typing.Optional[str] = None, verify: typing.Optional[typing.Union[bool, str]] = None):
+    def __init__(
+        self,
+        endpoint: str,
+        header_key: str,
+        credentials: Credentials = None,
+        http_proxy_url: typing.Optional[str] = None,
+        verify: typing.Optional[typing.Union[bool, str]] = None,
+    ):
         self._endpoint = endpoint
         self._creds = credentials
         self._header_key = header_key if header_key else "authorization"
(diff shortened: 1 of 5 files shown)


flytekit/clients/auth/authenticator.py
```diff
@@ -54,11 +54,13 @@ def __init__(
         header_key: str,
         credentials: Credentials = None,
         http_proxy_url: typing.Optional[str] = None,
+        verify: typing.Optional[typing.Union[bool, str]] = None,
```
wild-endeavor, line 57: why is this type a union of a bool and a string?

flytekit/clients/auth_helper.py
```diff
@@ -58,12 +58,13 @@ def get_authenticator(cfg: PlatformConfig, cfg_store: ClientConfigStore) -> Auth
             logging.warning(f"Authentication type {cfg_auth} does not exist, defaulting to standard")
             cfg_auth = AuthType.STANDARD
 
+    verify = None
+    if cfg.insecure_skip_verify:
+        verify = False
+    elif cfg.ca_cert_file_path:
+        verify = cfg.ca_cert_file_path
```
wild-endeavor, line 65: oh i guess this is why, this is pretty ugly though, can we split this out and pass down both?

flytekit/clients/auth/authenticator.py
ByronHsu: https://requests.readthedocs.io/en/latest/api/#requests.request
http requests is designed like this 
> verify – (optional) Either a boolean, in which case it controls whether we verify the server’s TLS certificate, or a string, in which case it must be a path to a CA bundle to use. Defaults to True.

wild-endeavor: oh i see

wild-endeavor (approved):

PR #1650, opened 2023-05-19
Map over notebook task

pingsutw:
# TL;DR
This PR allows map over notebook task

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
import os
import pathlib
import typing

from flytekit import kwtypes, task, workflow, ImageSpec, map_task
from flytekitplugins.papermill import NotebookTask

new_flytekit = "git+https://github.com/flyteorg/flytekit@022ade936d2fe1eb6a2d705c319cce7d2c66ade6"
new_papermill = "git+https://github.com/flyteorg/flytekit.git@022ade936d2fe1eb6a2d705c319cce7d2c66ade6#subdirectory=plugins/flytekit-papermill"
image = ImageSpec(registry="pingsutw", packages=[new_papermill, new_flytekit], apt_packages=["git"])

nb = NotebookTask(
    name="simple-nb",
    notebook_path=os.path.join(
        pathlib.Path(__file__).parent.absolute(), "nb-simple.ipynb"
    ),
    render_deck=True,
    inputs=kwtypes(a=float),
    outputs=kwtypes(square=float),
    container_image=image,
)


@workflow
def wf(a: float) -> typing.List[float]:
    return map_task(nb)(a=[a])


if __name__ == "__main__":
    print(wf(a=3.14))
```

image

## Tracking Issue
https://flyte-org.slack.com/archives/CP2HDHKE1/p1684471512753359

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -79,9 +79,9 @@ class MapPythonTask(PythonTask):
 
         collection_interface = transform_interface_to_list_interface(actual_task.python_interface, self._bound_inputs)
         self._run_task: PythonFunctionTask = actual_task
-        if issubclass(type(actual_task), PythonInstanceTask):
+        if isinstance(actual_task, PythonInstanceTask):
             mod = actual_task.task_type
-            f = actual_task.name
+            f = actual_task.lhs
         else:
             _, mod, f, _ = tracker.extract_task_module(actual_task.task_function)
         h = hashlib.md5(collection_interface.__str__().encode("utf-8")).hexdigest()
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -64,7 +64,7 @@ class MapPythonTask(PythonTask):
         else:
             actual_task = python_function_task
 
-        if not isinstance(actual_task, PythonTask) or not issubclass(type(actual_task), PythonInstanceTask):
+        if not issubclass(type(actual_task), PythonTask):
             raise ValueError("Map tasks can only compose of Python Functon Tasks currently")
 
         if len(actual_task.python_interface.outputs.keys()) > 1:
@@ -76,8 +76,8 @@ class MapPythonTask(PythonTask):
 
         collection_interface = transform_interface_to_list_interface(actual_task.python_interface, self._bound_inputs)
         self._run_task: PythonFunctionTask = actual_task
-        if hasattr(actual_task, "_IMPLICIT_OP_NOTEBOOK_TYPE"):
-            mod = "papermill"
+        if issubclass(type(actual_task), PythonInstanceTask):
+            mod = actual_task.task_type
             f = actual_task.name
         else:
             _, mod, f, _ = tracker.extract_task_module(actual_task.task_function)
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -34,7 +34,7 @@ class MapPythonTask(PythonTask):
 
     def __init__(
         self,
-        python_function_task: typing.Union[PythonFunctionTask, functools.partial],
+        python_function_task: typing.Union[PythonFunctionTask, PythonInstanceTask, functools.partial],
         concurrency: Optional[int] = None,
         min_success_ratio: Optional[float] = None,
         bound_inputs: Optional[Set[str]] = None,
@@ -65,7 +65,7 @@ class MapPythonTask(PythonTask):
             actual_task = python_function_task
 
         if not isinstance(actual_task, PythonFunctionTask):
-            if issubclass(type(actual_task), PythonInstanceTask):
+            if isinstance(actual_task, PythonInstanceTask):
                 pass
             else:
                 raise ValueError("Map tasks can only compose of PythonFuncton and PythonInstanceTasks currently")
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -64,8 +64,11 @@ class MapPythonTask(PythonTask):
         else:
             actual_task = python_function_task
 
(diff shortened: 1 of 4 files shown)


flytekit/core/map_task.py
```diff
@@ -64,7 +64,7 @@ def __init__(
         else:
             actual_task = python_function_task
 
-        if not isinstance(actual_task, PythonFunctionTask):
+        if not isinstance(actual_task, PythonTask) or not issubclass(type(actual_task), PythonInstanceTask):
```
kumare3, line 67: You don't need the second check instance is subclass of pythontask

flytekit/core/map_task.py
```diff
@@ -76,7 +76,11 @@ def __init__(
 
         collection_interface = transform_interface_to_list_interface(actual_task.python_interface, self._bound_inputs)
         self._run_task: PythonFunctionTask = actual_task
-        _, mod, f, _ = tracker.extract_task_module(actual_task.task_function)
+        if hasattr(actual_task, "_IMPLICIT_OP_NOTEBOOK_TYPE"):
```
kumare3, line 79: This seems to special for notebook- we should make it for instance tasks in general

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
```diff
@@ -277,16 +270,14 @@ def execute(self, **kwargs) -> Any:
         output_list = []
 
         for k, type_v in self.python_interface.outputs.items():
-            if k == self._IMPLICIT_OP_NOTEBOOK:
-                output_list.append(self.output_notebook_path)
```
kumare3, line 281: Are you just removing these?

pingsutw: cc @rahul-theorem I know you're using papermill plugin, do you use these paths in the output?

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
```diff
@@ -165,13 +166,14 @@ def __init__(
         if not os.path.exists(self._notebook_path):
             raise ValueError(f"Illegal notebook path passed in {self._notebook_path}")
 
-        if outputs:
+        if outputs and output_notebooks:
```
wild-endeavor, line 169:
```suggestion
        if output_notebooks:
```

wild-endeavor: what if you just want a notebook?  don't want to have to make a fake output.

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
```diff
@@ -287,6 +289,8 @@ def execute(self, **kwargs) -> Any:
             else:
                 raise TypeError(f"Expected output {k} of type {type_v} not found in the notebook outputs")
 
+        if len(output_list) == 1:
```
wild-endeavor, line 292: what happens if we don't do this?

flytekit/core/map_task.py
```diff
@@ -76,7 +76,11 @@ def __init__(
 
         collection_interface = transform_interface_to_list_interface(actual_task.python_interface, self._bound_inputs)
         self._run_task: PythonFunctionTask = actual_task
-        _, mod, f, _ = tracker.extract_task_module(actual_task.task_function)
+        if issubclass(type(actual_task), PythonInstanceTask):
```
wild-endeavor, line 82: is there a difference between this and isinstance(actual_task, PythonInstanceTask)

flytekit/core/map_task.py
```diff
@@ -76,7 +76,11 @@ def __init__(
...
+            mod = actual_task.task_type
```
wild-endeavor, line 83: should this be `actual_task.instantiated_in`?  what happens if there are two of the same notebook tasks, named the same, with the same interface, but in two different .py files?  will there be confusion?

wild-endeavor (commented):
the only thing is that this will change the interface for the user right?  we will need to add a disclaimer to the release notes.

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
pingsutw: There will be a mismatch between the output type and the downstream task's input type.

flytekit/core/map_task.py
pingsutw: updated it to `isinstance(actual_task, PythonInstanceTask)`

flytekit/core/map_task.py
pingsutw: used lhs instead

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
pingsutw: updated it

plugins/flytekit-papermill/flytekitplugins/papermill/task.py
```diff
@@ -165,13 +166,16 @@ def __init__(
         if not os.path.exists(self._notebook_path):
             raise ValueError(f"Illegal notebook path passed in {self._notebook_path}")
 
-        if outputs:
+        if output_notebooks:
+            if outputs is None:
```
wild-endeavor, line 170: can do `output = outputs or {}`

wild-endeavor (approved):

PR #1660, opened 2023-05-26
Feat: Add type support for pydantic BaseModels

ArthurBook:
# TL;DR
Lets you use pydantic basemodels as inputs and outputs in tasks.
Also extends the basemodels to have any flyte supported file as field.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Added serializer/deserializers for flytepath and flytedirectory s.t. pydantic can jsonify these objects too.
## Type Example
```python

class Config(BaseModel):
    lr: float = 1e-3
    batch_size: int = 32
    files: List[FlyteFile]
    directories: List[FlyteDirectory]

@task
def train(cfg: Config):
    ...
```

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2686

## Follow-up issue
_NA_


plugins/flytekit-pydantic/README.md
```diff
@@ -0,0 +1,30 @@
...
+```
+
+
+## Type Example
+```python
+from pydantic import BaseModel
+import flytekitplugins.pydantic
+
+
+class Config(BaseModel, **flytekitplugins.pydantic.pydantic_flyteobject_config):
```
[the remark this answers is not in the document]
ArthurBook: Sure! I made it a bit easier and added an example in the latest commit.

pingsutw (commented):
Thanks @ArthurBook for working on this. 
To serialize `pydantic.baseModel`, we have to serialize the object to json. Do I understand correctly?
If so, I have some thoughts 
1. Add `to_json` to [typeTransformer](https://github.com/flyteorg/flytekit/blob/06fffc786edfe9d4cbb891d6c0e14181a5823998/flytekit/core/type_engine.py#L62)?
2. Make [lv](https://github.com/flyteorg/flytekit/blob/06fffc786edfe9d4cbb891d6c0e14181a5823998/flytekit/core/type_engine.py#L122) become Union type (union[json, Literal]).
3. Create a generic FlyteDirSerializerDeserializer, and use `transformer.to_json` to serialize all the python value.
4. I think it's worth to move pydantic transformer to `flytekit/core/type`

wdyt?

plugins/flytekit-pydantic/flytekitplugins/pydantic/schema.py
```diff
@@ -0,0 +1,59 @@
...
+        python_val: pydantic.BaseModel,
+        python_type: Type[pydantic.BaseModel],
+        expected: types.LiteralType,
+    ) -> literals.Literal:
+        """This method is used to convert from given python type object pydantic ``pydantic.BaseModel`` to the Literal representation."""
+
+        s = struct_pb2.Struct()
+        schema = python_val.schema_json()
+        data = python_val.json()
+        s.update({"schema": schema, "data": data})
```
pingsutw, line 37: are we able to inject `flytekitplugins.pydantic.pydantic_flyteobject_config` to `python_val` somehow here, so we don't need that config when we declare the pydantic model.
```
class TrainConfig(BaseModel, **flytekitplugins.pydantic.pydantic_flyteobject_config):
   ...
```

pingsutw (commented):
@ArthurBook sorry for the late reply, we had some discussion internally. let's keep pydantic transformer in the plugin first. overall LGTM, leave a comment below.

plugins/flytekit-pydantic/flytekitplugins/pydantic/schema.py
ArthurBook: I like that. Applied the change in the latest commit!

plugins/flytekit-pydantic/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+```bash
+pip install flytekitplugins-pydantic
+```
+
+
+## Type Example
+```python
+from pydantic import BaseModel
+import flytekitplugins.pydantic # This import will enable you to add FlyteFiles and FlyteDirectories to you BaseModels
```
pingsutw, line 17: we don't need it. flytekit will auto load the transformer if the plugin is installed.

plugins/flytekit-pydantic/flytekitplugins/pydantic/basemodel_extensions.py
```diff
@@ -0,0 +1,152 @@
...
+    PydanticSerializerDeserializerBase[flyte_directory_types.FlyteDirectory, FlyteDirJsonEncoded]
+):
+    def __init__(
+        self,
+        t: Type[flyte_directory_types.FlyteDirectory] = flyte_directory_types.FlyteDirectory,
+    ):
+        super().__init__(t)
+
+    def serialize(self, obj: flyte_directory_types.FlyteDirectory) -> FlyteDirJsonEncoded:
+        return {"path": obj.remote_source if obj.remote_source else obj.path}
```
pingsutw, line 81: should we upload the local file to s3? like https://github.com/flyteorg/flytekit/blob/master/flytekit/types/file/file.py#L393-L397

plugins/flytekit-pydantic/flytekitplugins/pydantic/basemodel_extensions.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+FLYTEFILE_DESERIALIZABLE_TYPES = Union[flyte_directory_types.FlyteFile, str, FlyteFileJsonEncoded]
+
+
+class FlyteFileSerializerDeserializer(PydanticSerializerDeserializerBase[flyte_file.FlyteFile, FlyteFileJsonEncoded]):
+    def __init__(self, t: Type[flyte_file.FlyteFile] = flyte_file.FlyteFile):
+        super().__init__(t)
+
+    def serialize(self, obj: flyte_file.FlyteFile) -> FlyteFileJsonEncoded:
```
pingsutw, line 124: ditto

plugins/flytekit-pydantic/flytekitplugins/pydantic/basemodel_extensions.py
ArthurBook: Latest commit should cover that - how does this look?

plugins/flytekit-pydantic/flytekitplugins/pydantic/flytepath_creation.py
```diff
@@ -14,6 +14,16 @@
 FlytePath = TypeVar("FlytePath", flyte_file.FlyteFile, flyte_directory_types.FlyteDirectory)
 
 
+def upload_to_s3(flytepath: FlytePath, ctx: Optional[context_manager.FlyteContext] = None) -> None:
```
pingsutw, line 17: nit: could we remove `ctx: Optional[context_manager.FlyteContext] = None`? and always create `ctx` in this function.

plugins/flytekit-pydantic/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+class TrainConfig(BaseModel):
+    lr: float = 1e-3
+    batch_size: int = 32
+    files: List[FlyteFile]
+    directories: List[FlyteDirectory]
+
+@task
+def train(cfg: TrainConfig):
+    ...
```
pingsutw, line 27: @fg91 mind also taking a look? I think the interface looks better now!

pingsutw (dismissed):
LGTM, could you also update the pr description

plugins/flytekit-pydantic/README.md
fg91: Will do a review on the weekend :+1:

plugins/flytekit-pydantic/tests/test_type_transformer.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+class ConfigWithFlyteFiles(BaseModel):
+    """Config BaseModel for testing purposes with flytekit.files.FlyteFile type hint."""
+
+    flytefiles: List[file.FlyteFile]
+
+
+class ConfigWithFlyteDirs(BaseModel):
+    """Config BaseModel for testing purposes with flytekit.files.FlyteFile type hint."""
```
fg91, line 45:
```suggestion
    """Config BaseModel for testing purposes with flytekit.directory.FlyteDirectory type hint."""
```

plugins/flytekit-pydantic/tests/test_type_transformer.py
```diff
@@ -0,0 +1,167 @@
...
+    flytedirs: List[directory.FlyteDirectory]
+
+
+class ChildConfig(Config):
+    """Child class config BaseModel for testing purposes."""
+
+    d: List[int] = [1, 2, 3]
+
+
+@pytest.mark.parametrize(
```
fg91, line 56: I think it would be nice to add an example of `ConfigWithFlyteFiles` and `ConfigWithFlyteDirs` to this roundtrip test.

plugins/flytekit-pydantic/tests/test_type_transformer.py
```diff
@@ -0,0 +1,167 @@
...
+    def read(cfg: ConfigWithFlyteFiles) -> str:
+        with open(cfg.flytefiles[0], "r") as f:
+            return f.read()
+
+    @workflow
+    def wf(cfg: ConfigWithFlyteFiles) -> str:
+        return read(cfg=cfg)
+
+    string = wf(cfg=cfg)
+    assert string == "love sosa"
```
fg91, line 129: My personal preference would be to not include any pop culture references but stick with e.g. "foo bar" which developers from any background will recognize as placeholders. Your call though.

setup.py
```diff
@@ -40,8 +40,8 @@
         "python-dateutil>=2.1",
         # Restrict grpcio and grpcio-status.  Version 1.50.0 pulls in a version of protobuf that is not compatible
         # with the old protobuf library (as described in https://developers.google.com/protocol-buffers/docs/news/2022-05-06)
-        "grpcio>=1.50.0,<2.0",
-        "grpcio-status>=1.50.0,<2.0",
+        "grpcio>=1.50.0,!=1.55.0,<2.0",
```
fg91, line 43: What is the reason for excluding these versions and removing `singledispatchmethod`?

plugins/flytekit-pydantic/flytekitplugins/pydantic/flytepath_creation.py
```diff
@@ -0,0 +1,76 @@
...
+from flytekit.models import literals
+from flytekit.models.core import types as core_types
+from flytekit.types.directory import types as flyte_directory_types
+from flytekit.types.file import file as flyte_file
+
+FlytePath = TypeVar("FlytePath", flyte_file.FlyteFile, flyte_directory_types.FlyteDirectory)
+
+
+def upload_to_s3(flytepath: FlytePath) -> None:
+    """Upload a FlytePath to S3"""
```
fg91, line 18: Is this really specific to s3 or blob storage in general?

fg91 (changes requested):
Thanks for improving upon [my initial attempt](https://github.com/flyteorg/flytekit/pull/1620) to implement a type transformer for pydantic base models. Being able to have flyte files and directories in the models is a very nice feature!! I closed my PR in favor of this one here.

The PR contains a large number of changes in e.g. `flytekit/clis`, `core`, `remote`, `plugins/flytekit-papermill`, ... I suspect that these changes are not related to this plugin and would go away when updating the branch/rebasing? If this is correct, could you please fix the diff to only contain your changes? This would make a thorough review much easier. Thanks!

Some questions/thoughts:

1. I like that we inject flyte specific json (de-) serializers so that users don't have to modify their models. I also like that all subclasses of `PydanticSerializerDeserializerBase` are automatically registered.
2. Is my understanding correct that with the current state flyte files and directories are supported in base models but e.g. pandas dataframe or `torch.nn.Module`s, which the flyte type system supports, are not?
    * In order to support those as well, would we have to subclass `PydanticSerializerDeserializerBase` for each additonal type, leading to a situation where for every type flyte supports we would maintain  pydantic json (de-) serializer in addition to the type transformer? While an implementation that supports flyte files and directories is 100% better than my attempt which doesn't support them at all, this would make me long for a more general solution:
        * Since the flyte type transformers convert to and from protobuf, do you think there could be a way to maintain only a single json (de-) serializer which is called by pydantic for all types that the flyte type system supports? The json (de-)serializer would then convert the python value to protobuf using the flyte type transformer and then maybe to json? Or maybe even leave as protobuf? The reason I'm asking whether you think something like this would be possible is because this would alleviate us from the burden to maintain a json (de-) serializer for every type transformer in the type system.

fg91:
Thanks a lot for cleaning up the diff @ArthurBook, much easier to review! :pray:

fg91:
I've been thinking about this since [reviewing](https://github.com/flyteorg/flytekit/pull/1660#pullrequestreview-1473744858) and I'm interested in what you think about such an approach  @ArthurBook @pingsutw:


* I pick up on your idea to inject `json_serializers` for Flyte types into the base model @ArthurBook 
* I inject only a single serializer that used for all Flyte types
* This serializer uses the Flyte type engine to convert the respective python value/Flyte types in the base model to a Flyte literal
* The flyte literal is *not* saved in the base model itself (not json serializable). Instead, we save it in something I called `store` (would need a better name) and in the base model we only injejct a json serializable placeholder with reference to the "offloaded" Flyte literal
* In the Flyte literal which is returned by the `to_literal` method of the type transformer, we transport the json serialized base model and the "offloaded" Flyte types separately.


```python
import pandas as pd
import pydantic
import torch.nn as nn
import uuid
from functools import partial
from google.protobuf import struct_pb2
from typing import TypeVar, Type, Dict


from flytekit.types.file import file as flyte_file
from flytekit.models import literals, types
from flytekit.models.literals import LiteralMap
from flytekit.core.type_engine import TypeEngine
from flytekit.core.context_manager import FlyteContextManager


def serialize_using_flyte_type_engine(value, store: Dict):
    """
    Replace Flyte types in a pydantic BaseModel with json serializable place holders and
    serialize the real values into the `store`.
    """
    
    ctx = FlyteContextManager().current_context()
    python_type = type(value)

    flyte_literal = TypeEngine.to_literal(ctx, value, python_type, TypeEngine.to_literal_type(python_type))
    
    identifier = uuid.uuid4().hex
    
    # TODO: Don't store str in store but e.g. an instance of some placeholder class
    # Also use a human readble identifier instead of random uuid
    store[identifier] = flyte_literal
    
    return identifier


# TODO: the following needs to happen within the type transformer's `to_literal` method (see below)
# once the json_encoder is injected again instead of being hard-coded in the `class Config` of the base model
# For me the injection of the json encoders in `to_literal` currently doesn't work yet
store = {}
serialize = partial(serialize_using_flyte_type_engine, store=store)


# Now we define for which Flyte types we want to defer json serialization to our custom json serializer above
PYDANTIC_SUPPORTED_FLYTE_TYPES = [
    nn.Module, pd.DataFrame, flyte_file.FlyteFile,  # And so on, ...
]


class TestModel(pydantic.BaseModel):
    
    # When using `TypeVar('torch.nn.Linear')` instead of `nn.Linear`, we don't have to set `allow_arbitrary_types`
    torch_module: TypeVar('torch.nn.Linear') = nn.Linear(1,1)
    df: TypeVar('pandas.core.frame.DataFrame') = pd.DataFrame([{'key 1': 'value 1', 'key 2': 'value 2', 'key 3': 'value 3'}])
        
        
    # TODO: inject json_encoders in the `to_literal` method again once fixed
    class Config:
        json_encoders = {t: serialize for t in PYDANTIC_SUPPORTED_FLYTE_TYPES}


def to_literal(
    python_val: pydantic.BaseModel,
    python_type: Type[pydantic.BaseModel],
    expected: types.LiteralType,
) -> literals.Literal:
    """This method is used to convert from given python type object pydantic ``pydantic.BaseModel`` to the Literal representation."""
    s = struct_pb2.Struct()
    
    # TODO: injejct json_encoders again instead of setting them explicitly in `class Config`.
    # python_val.__config__.json_encoders.update({t: serialize for t in PYDANTIC_SUPPORTED_FLYTE_TYPES})
    
    # TODO: Should we try to use this again? Currently not working despite the json serializer
    # schema = python_val.schema_json()
    
    base_model_json = python_val.json()
    """
    When calling `.json()`, our custom flyte engine json serializer was invoked for each instance of a Flyte type.
    The `base_model_json` now contains references/placeholders for the torch.nn.Module/pd.DataFrame/FlyteDirectory, ... 
    and the actual literal values are stored in `store`.
    """
    
    s.update({"data": base_model_json})  # Evaluate whether schema is needed and can be fixed
    
    # Now we save the json serialized base model and the "offloaded" instances of the Flyte types separately into a literal
    # TODO: This needs a better structure, I just pushed everything into the literal to demo the idea somehow.
    literal_data = literals.Literal(scalar=literals.Scalar(generic=s))  # This is the base model
    store["base_model"] = literal_data
    literal_store = LiteralMap(literals=store)

    return literal_store

"""
def to_python_value(...):
    # Convert the json serialized base model into a dictionary.
    # Put back all the de-serialized instances of Flyte types at the respective place holder locations
    # Convert back to `BaseModel` using `parse_raw`
"""

t = TestModel()


print(to_literal(t, pydantic.BaseModel, None))
```

What I like about this idea is that the flyte type engine is used to do the serialization and regardless of how many Flyte types are added in the future, all of them will be supported without implementing any pydantic json serializers.

The json serializer injection is not working for me, do you maybe see what I'm doing wrong @ArthurBook ?

What do you think about this idea?

ArthurBook:
Think I can't reply to the comment above so writing here.

Thanks for spotting that - think its a bug. Those encoders seem to be a classvariable that can't be updated fully after initialization. Below is a workaround that could be used instead. What do you think?

```python
# Changed this to a tuple allow using it in isinstance
PYDANTIC_SUPPORTED_FLYTE_TYPES = (
    nn.Module,
    pd.DataFrame,
    flyte_file.FlyteFile,  # And so on, ...
)

def encode_with_flytetype_support(basemodel: pydantic.BaseModel) -> str:
    """
    Serialize a pydantic BaseModel to a json string with support for Flyte types.
    """

    def encoder(obj: Any) -> str:
        if isinstance(obj, PYDANTIC_SUPPORTED_FLYTE_TYPES):
            return serialize(obj)
        return basemodel.__json_encoder__(obj)

    return basemodel.json(encoder=encoder)

class TestModel(pydantic.BaseModel):
    # When using `TypeVar('torch.nn.Linear')` instead of `nn.Linear`, we don't have to set `allow_arbitrary_types`
    # torch_module: TypeVar("torch.nn.Linear") = nn.Linear(1, 1)
    df: TypeVar("pandas.core.frame.DataFrame") = pd.DataFrame(
        [{"key 1": "value 1", "key 2": "value 2", "key 3": "value 3"}]
    )


def to_literal(
    python_val: pydantic.BaseModel,
    python_type: Type[pydantic.BaseModel],
    expected: types.LiteralType,
) -> literals.Literal:
    """This method is used to convert from given python type object pydantic ``pydantic.BaseModel`` to the Literal representation."""
    ...
    base_model_json = encode_with_flytetype_support(python_val)
    ...

```

pingsutw:
This is a great idea. Instead of using UUID, I'll use the property names of pydantic model, but I'm sure if it's possible. UUID is fine though. 

And I think we should add `python_type` arg to the `serialize_using_flyte_type_engine`
```python
def serialize_using_flyte_type_engine(value, python_type, store: Dict):
    ...

class Config:
      json_encoders = {t: partial(serialize_using_flyte_type_engine, python_type=t, store=store) for t in PYDANTIC_SUPPORTED_FLYTE_TYPES}
```

ArthurBook:
Got inspired and rewrote the code to a V2 working off Fabio's idea.
This is now only leveraging the transformers from the core types and uses a FlyteObjectStore namespace for holding the literals while serializing/deserializing.

P happy with the outcome.. Ended up much lighter in terms of logic and also generalizes to any supported flyte types.

I marked some TODO's in the code in the points where I'd be especially curious for your opinion. Let me know what you think... 
Since you both are more familiar with all the use cases I'd ask if you can help testing it a bit more to make sure I covered all edge cases?

many thanks

flytekit/clis/sdk_in_container/run.py
```diff
@@ -399,7 +399,10 @@ def convert_to_struct(
         Convert the loaded json object to a Flyte Literal struct type.
         """
         if type(value) != self._python_type:
-            o = cast(DataClassJsonMixin, self._python_type).from_json(json.dumps(value))
+            if hasattr(self._python_type, "parse_raw"):  # e.g pydantic basemodel
```
fg91, line 402: Question: This is for passing base models as json string to `pyflyte run`? (Just to make sure I understand.)

fg91, line 402: Suggestion (non blocking): could we simply check if the python type is a child class of `pydantic.BaseModel`?

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,77 @@
...
+PYDANTIC_VALIDATOR_METHOD_NAME = pydantic.BaseModel.__get_validators__.__name__
+PythonType = TypeVar("PythonType")  # target type of the deserialization
+Serializable = TypeVar("Serializable")  # flyte object type
+
+
+def set_validators_on_supported_flyte_types() -> None:
+    """
+    Sets the validator on the pydantic model for the
+    type that is being serialized/deserialized
+    """
```
fg91, line 19:
```suggestion
    """Set the validator on the pydantic model for the type that is being (de-)serialized.
    """
```

Nitpick (non-blocking): [PEP257](https://peps.python.org/pep-0257/) says 

> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

flytekit/clis/sdk_in_container/run.py
ArthurBook: Correct!

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+import pandas as pd
+import torch.nn as nn
+from typing_extensions import Annotated, NewType
+
+from flytekit.core import context_manager, type_engine
+from flytekit.models import literals
+from flytekit.types import directory
+from flytekit.types.file import file
+
+PYDANTIC_SUPPORTED_FLYTE_TYPES = (
```
fg91, line 13: Suggestion: What do you think about simply using [this](https://github.com/flyteorg/flytekit/blob/20427ee9cdbe068f17a3e1f67910fcd231a48c3d/flytekit/core/type_engine.py#L952) to retrieve the supported types from the type engine? Benefit would be that types that users create type transformers for would automatically be picked up.

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+    file.FlyteFile,
+    directory.FlyteDirectory,
+    # TODO - add all supported types
+)
+
+LiteralObjID = Annotated[str, "Key for unique object in literal map."]
+PythonType = TypeVar("PythonType")  # target type of the deserialization
+
+
+class FlyteObjectStore:
```
fg91, line 25: Suggestion: From the name I would expect that this class had a wider scope in Flyte than supporting serialization in a type transformer plugin. Might confuse users who look through the code. What do you think about calling this e.g. `PydanticTransformerLiteralStore`?

flytekit/clis/sdk_in_container/run.py
ArthurBook: Yes, I like that - had missed that pydantic was a dependency in doc-requirements

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+    This class is an intermediate store for python objects that are being serialized/deserialized.
+
+    On serialization of a basemodel, flyte objects are serialized and stored in this object store.
+    On deserialization of a basemodel, flyte objects are deserialized from this object store.
+    """
+
+    _literal_store: Dict[LiteralObjID, literals.Literal] = {}
+
+    def __setattr__(self, name: str, value: Any) -> None:
+        raise Exception("Attributes should not be set on the FlyteObjectStore.")
```
fg91, line 36: Question: Is this required to avoid a problem? Typically one wouldn't implement this for a class in python.

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def __init__(self) -> None:
+        raise Exception("This should not be instantiated, it is a singleton object store.")
```
fg91, line 39: Question: what happens when in a single process multiple base models get (de-)serialized? Are we at risk that data serialized for one model is available when deserializing another model? I wonder whether not making this a singleton would be safer. In this case it could have a `to_literal` and `from_literal` method and would be instantiated both in the `to_literal` and the `to_python_value` method of the type transformer.

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+        return item in self.get_literal_store()
+
+    @classmethod
+    def get_literal_store(cls):
+        """Accessor to the class variable"""
+        return cls._literal_store
+
+    @classmethod
+    def register_python_object(cls, python_object: object) -> LiteralObjID:
+        """serializes to literal and returns a unique identifier"""
```
fg91, line 51:
```suggestion
        """Serialize to literal and return a unique identifier."""
```

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+        literal = cls.get_literal_store()[identifier]
+        python_object = deserialize_flyte_literal(literal, expected_type)
+        return python_object
+
+    @classmethod
+    def as_literalmap(cls) -> literals.LiteralMap:
+        """
+        Converts the object store to a literal map
+        """
+        return literals.LiteralMap(literals=cls.get_literal_store())
```
fg91, line 69: Praise: very nice

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+        Reads a literal map and populates the object store from it
+        """
+        literal_store = cls.get_literal_store()
+        literal_store.update(literal_map.literals)
+
+
+def deserialize_flyte_literal(flyteobj_literal: literals.Literal, python_type: Type[PythonType]) -> PythonType:
+    """
+    Deserializes a Flyte Literal into the python object instance.
+    """
```
fg91, line 83:
```suggestion
    """Deserialize a Flyte Literal into the python object instance."""
```

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
ArthurBook: Not necessary. 

The goal was to emphasize that this class's function is essentially to serve as a namespace with only a few auxiliary methods for storing/retrieving objects. Given this is somewhat less generic than how most classes are used, I considered it beneficial to insert this sort of defensive mechanism, analogous to the init block I implemented.

It's a good sign that you're questioning this. It suggests that the class's purpose is indeed coming across clearly. Let me know whichever way you prefer it and ill make it so

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
ArthurBook: Good question. This was something that took some design on my side. 

I went for the singleton pattern because the validators need to be set on supported flyte objects **before** they are set as a fields in a basemodel subclass. This is because the pydantic.BaseModel gathers all its validators by calling __get_validators__ on each field **at the moment that it is subclassed**, without awareness of which pydantic.BaseModel it is that is calling it.
The singleton pattern where all serialized flyteobjects get saved to the same store enusres that we have all the serialized flyte types in the literalmap that gets sent. While I agree that a single BaseModel independent store is not as slick as having one per item, I went for this because its simplicity. Because we are using UUIDs + class name as the keys in the object store, the risk of two different objects sharing the same key is pretty low. 

Nevertheless, I appreciate your point and I think it could be accommodated without too much changes. We can use a context manager in the to_literal, which creates a namespace that is unique for the current basemodel, locks the access to the object store s.t. all calls to register register_python_object will write into that namespace, and on __exit__ returns the literalmap + releases the lock. Ill commit a proposal for it in a bit.

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
```diff
@@ -0,0 +1,107 @@
...
+    ctx = context_manager.FlyteContext.current_context()
+    transformer = type_engine.TypeEngine.get_transformer(python_type)
+    python_obj = transformer.to_python_value(ctx, flyteobj_literal, python_type)
+    return cast(PythonType, python_obj)
```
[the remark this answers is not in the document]
ArthurBook: Think the get_transformer didnt use the TypeVar so changed that

plugins/flytekit-pydantic/flytekitplugins/pydantic/object_store.py
ArthurBook: Thanks for this thought. Solved now - think this improved readability of the code easier too!

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,134 @@
...
+import pydantic
+from flytekit.core import context_manager, type_engine
+from flytekit.models import literals
+
+from flytekit.types import directory, file
+
+from flytekitplugins.pydantic import commons, serialization
+
+# this field is used by pydantic to get the validator method
+PYDANTIC_VALIDATOR_METHOD_NAME = pydantic.BaseModel.__get_validators__.__name__
```
fg91, line 13: Change request (blocking): Could you please account for [this](https://github.com/pydantic/pydantic/blob/f742ca4313fc936a4701f8005b829c78834434f2/docs/migration.md?plain=1#L527) point in the migration guide and make sure that what we build here will work with V2?

fg91 (changes requested):
Pretty happy with how this is developing!

* Everything I marked with "non-blocking" in the comments below is my personal taste but feel free to ignore and resolve the conversations if you don't agree. Many points are about style of docstrings, naming of functions, ...
* Let's talk about/iterate on everything I marked as blocking, especially compatibility with pydantic version V2, in which format (str or struct) the base model is transported, how we deal with the schema, ...

fg91:
I've been thinking about this after my review last night.

I'm pretty happy with the serialization part. I like that the custom encoder simply gets passed to `base_model.json(encoder=)`. This means the user doesn't notice anything of the serialization part.

About the deserialization part I like that we got it to work with a single validator that leverages the flyte type engine, this is great. The singleton and more importantly having to inject/monkeypatch the validator before instantiation of the base model are two compromises in my opinion, the latter being the more severe because the user might notice this.

I would like to propose an alternative approach - not necessarily saying we need to do it this way - but to discuss the pros and cons of both approaches:

```python
from typing import TypeVar

import torch.nn as nn

class Baz(BaseModel):
    e: TypeVar("nn.Module")

class Foo(BaseModel):
    a: int
    b: TypeVar("nn.Module")
    c: list[Baz]

class Bar(BaseModel):
    d: Foo


# Let's say this is what we extract from the Flyte LiteralMap in `to_python_value`:

deserialized_obj= {"d": {"a": 1, "b": "placeholder_123", "c": [{"e": "placeholder_234"}, {"e": "placeholder_345"}]}}
deserialized_store = {
    "placeholder_123": nn.Linear(1, 1),
    "placeholder_234": nn.Linear(2, 2),
    "placeholder_345": nn.Linear(3, 3),
}


def replace_placeholders(obj: dict, store: dict):
    """Recursively walk a serialized base model and replace place holders with real values."""
    if isinstance(obj, str):    # Not all strings but place holders
        obj = store[obj]    # Here we need to also convert from flyte literal to python value
    elif isinstance(obj, dict):
        for k, _ in obj.items():
            obj[k] = replace_placeholders(obj[k], store)
    elif isinstance(obj, list):
        for i in range(len(obj)):
            obj[i] = replace_placeholders(obj[i], store)
    return obj


deserialized = replace_placeholders(deserialized_obj, deserialized_store)

b = Bar.parse_obj(deserialized)

print(b)
```

```
Bar(d=Foo(a=1, b=Linear(in_features=1, out_features=1, bias=True), c=[Baz(e=Linear(in_features=2, out_features=2, bias=True)), Baz(e=Linear(in_features=3, out_features=3, bias=True))]))
```

For the deserialization part we would need to know the python type and I don't yet see what the best place would be to retrieve this info from.

This approach would mean that we are responsible for walking the base model and finding the place holders, a bit more responsibility on our side, but it would allow us to trade that for reduced complexity and magic. Interested in your thoughts on this tradeoff @ArthurBook!

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,134 @@
...
+def make_validators_for_type(
+    flyte_obj_type: Type[Serializable],
+) -> Callable[[Any], Iterator[Callable[[Any], Serializable]]]:
+    """
+    Returns a validator that can be used by pydantic to deserialize the object
+    """
+
+    previous_validators = getattr(flyte_obj_type, PYDANTIC_VALIDATOR_METHOD_NAME, lambda *_: [])()
+
+    def validator(object_uid_maybe: Union[commons.LiteralObjID, Any]) -> Union[Serializable, Any]:
```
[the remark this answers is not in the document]
ArthurBook: These validators will be run on any object that is set on that field so we need to accommodate cases like a dict that gets validated to a pd.DataFrame or a string that gets validated into a file.FlyteFile.

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,134 @@
...
+        """partial of deserialize_flyte_literal with the object_type fixed"""
+        if not isinstance(object_uid_maybe, str):
```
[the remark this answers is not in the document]
ArthurBook: If I've understood you correctly, you're asking for an example when the object_uid_maybe is a string but not an ID?
One example of that is when a BaseModel with a file.FlyteFile field is first initialized with a path passed as a string. 

```python
class Config(pydantic.BaseModel):
   file: file.FlyteFile

Config(file='s3://...') # here a non-placeholder string is given
```

Think its good to have a double-safe here to make sure that the flytestore is only accessed in the very specific case that it truly is an id.

I committed an extra check to also make sure that the serialized object ID exists in the store too...

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,134 @@
...
+            return object_uid_maybe  # this validator should only trigger for the placholders
+        if not PydanticDeserializationLiteralStore.is_attached():
+            return object_uid_maybe
```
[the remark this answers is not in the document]
ArthurBook: I might be misunderstanding the q. The purpose of the checks here are to pin down the exact, only case where the store should be used (during de-serialization, when a object-uuid is passed and it actually exists in the store) else, let the other validators do their job in whichever way the user has defined them.

I committed some comments to help clarity of these checks.

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,134 @@
...
+
+    return validator_generator
+
+
+def validate_flytefile(flytefile: Union[str, file.FlyteFile]) -> file.FlyteFile:
+    """validator for flytefile (i.e. deserializer)"""
+    if isinstance(flytefile, file.FlyteFile):
+        return flytefile
+    if isinstance(flytefile, str):  # when e.g. initializing from config
+        return file.FlyteFile(flytefile)
```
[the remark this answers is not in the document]
ArthurBook: Yeah, apols, we tend to read it cli args in from a YAML files we refer to as configs.. 
This aligns the pydantic basemodel with the syntax one would use for a dataclass json when running the workflow from command line.

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,134 @@
...
+        return cls.literal_store is not None
+
+    @classmethod
+    def get_python_object(
+        cls, identifier: commons.LiteralObjID, expected_type: Type[PythonType]
+    ) -> Optional[PythonType]:
+        """Deserialize a literal and return the python object"""
+        if not cls.is_attached():
+            raise Exception("Must attach to a literal map before deserializing")
+        assert cls.literal_store is not None
```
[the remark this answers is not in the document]
ArthurBook: I left it here in case someone would call get_python_object without the attach context manager.

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
```diff
@@ -0,0 +1,128 @@
...
+from flytekit.models import literals
+from flytekit.core import context_manager, type_engine
+
+from . import commons
+
+
+BASEMODEL_JSON_KEY = "BaseModel JSON"
+FLYTETYPE_OBJSTORE__KEY = "Flyte Object Store"
+
+SerializedBaseModel = Annotated[str, "A pydantic BaseModel that has been serialized with placeholders for Flyte types."]
```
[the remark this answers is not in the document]
ArthurBook: Yeah let's discuss that. I think it could work. Need to have a look at how the dict() method on the BaseModel works in the context of this integration.

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
```diff
@@ -0,0 +1,128 @@
...
+    basemodel_literal = make_literal_from_json(basemodel_json)
+    return basemodel_literal
+
+
+def make_literal_from_json(json: str) -> literals.Literal:
+    """
+    Converts the json representation of a pydantic BaseModel to a Flyte Literal.
+    """
+    # serialize as a string literal
+    ctx = context_manager.FlyteContext.current_context()
```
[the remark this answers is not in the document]
ArthurBook: Thanks, this is the part I am least familiar with so thought I'd rely on the existing code as much as possible..

The solution with struct seems pretty verbose. How does this look to you?

```python
def make_literal_from_json(json: str) -> literals.Literal:
    """
    Converts the json representation of a pydantic BaseModel to a Flyte Literal.
    """
    # serialize as a string literal
    base_model_literal = struct_pb2.Struct()
    base_model_literal.update({BASEMODEL_JSON_KEY: json})
    return literals.Literal(scalar=literals.Scalar(primitive=literals.Primitive(string_value=json)))  # type: ignore
```

plugins/flytekit-pydantic/tests/test_type_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+        return cfg
+
+    @flytekit.workflow
+    def wf(cfg: config_type) -> config_type:
+        return train(cfg=cfg)
+
+    returned_cfg = wf(cfg=cfg)
+
+    # assert returned_cfg == cfg
+    # TODO these assertions are not valid for all types
```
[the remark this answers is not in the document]
ArthurBook: e.g. a pandas dataframe.

>> ValueError: The truth value of a DataFrame is ambiguous. Use a.empty, a.bool(), a.item(), a.any() or a.all().

plugins/flytekit-pydantic/tests/test_type_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+
+    @flytekit.workflow
+    def wf(cfg: ConfigWithFlyteDirs) -> List[str]:
+        return listdir(cfg=cfg)  # type: ignore
+
+    dirs = wf(cfg=cfg)
+    assert len(dirs) == 2  # type: ignore
+
+
+def test_double_config_in_wf():
```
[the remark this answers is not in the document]
ArthurBook: I used this to make sure that the flyte object store isnt for whatever reason mixed up when deserializing two otherwise similar objects. Think it could catch something in the future with a test that aims to ensure accuracy when multiple basermodels are passed in the wf kwargs.

plugins/flytekit-pydantic/tests/test_type_transformer.py
```diff
@@ -0,0 +1,203 @@
...
+    def wf(cfg1: TrainConfig, cfg2: TrainConfig) -> bool:
+        return are_different(cfg1=cfg1, cfg2=cfg2)  # type: ignore
+
+    assert wf(cfg1=cfg1, cfg2=cfg2), wf(cfg1=cfg1, cfg2=cfg2)  # type: ignore
+
+
+# TODO: //Arthur to Fabio this was differente before but now im unsure what the test is doing
+# previously a pattern match error was checked that its raised, but isnt it OK that the ChildConfig
+# is passed since its a subclass of Config?
+# I modified the test to work the other way around, but im not sure if this is what you intended
```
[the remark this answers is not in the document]
ArthurBook: I see. 

To support the `schema()` method, we would need to define a schema for each of the flytetypes. 
Think it would require some design to make that generic across python versions and that leaving this one to the end user is likely to be OK.

What do you think?

ArthurBook:
Thanks for the review! 
Most comments are addressed in my latest commit and left replies on the remainder.

Agree that the necessity to add to the `__get_validators__` on supported flyte types can be considered ugly. On the other hand, it is a [documented](https://docs.pydantic.dev/latest/usage/types/#classes-with-__get_validators__) solution that I would read as the currently preferred way for custom validation? From a user-perspective, I would argue that the risks related adding to the generator in our solution are mitigated because the [checks](https://github.com/flyteorg/flytekit/pull/1660/files#diff-5cfa7734bdca34e06edc47528750533a0acad516d11a36ac9f0fa098b73dd612R98-R105:~:text=%22%22%22Partial%20of%20deserialize_flyte_literal,%2C%20flyte_obj_type) ensure that the flyteliteral deserializer __only__ triggers on very explicit conditions met within the context manager block during deserialization.

Serializing to a dict instead of JSON could nevertheless be neat because of the mentioned advantages - especially because it lets us encapsulate the entire deserialization logic execution to stay within the plugin.

Think the changes required for that would be the following:
1) serialize the same way as now to JSON, and finally call `json.loads()` on the serialized basemodel to convert it to a dictionary with (list/dicts) of strings that gets passed on from the task as a LiteralMap.
2) when deserializing, use the `replace_placeholders` from your idea to put back the non-pydantic-native stuff such as flytefiles etc. before calling `BaseModel.parse_obj()` on this de-serialized dict representation.

The advantage is that we avoid singleton object store, although since this one is used solely as a temporary namespace with a strict context manager lock, I think of it more as a stylistic improvement?

My concern with this approach is that we would take on more code to maintain. In the current version we let pydantic decide when to tap into the store by giving a validator that is called by pydantic's deserializer whenever needed by pydantic's logic. In the dict solution, the flyte code would bear the responsibility to insert the objects exactly where needed by pydantic's deserializer. The latter is probably OK since JSON only has two data structures: objects (=python dict) and arrays (=python list). I am just not sure if there is any edge case that we could miss due to that we would be possibly be modifying the string representation within how each [pydantic supported datatype](https://docs.pydantic.dev/latest/usage/types/) gets deserialized...

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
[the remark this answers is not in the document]
ArthurBook: Gave my long comment at the bottom so marking this note as resolved

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
```diff
@@ -0,0 +1,119 @@
...
+    """
+
+    def __init__(self) -> None:
+        self.literal_store: LiteralStore = dict()
+
+    def register_python_object(self, python_object: object) -> LiteralObjID:
+        """Serialize to literal and return a unique identifier."""
+        serialized_item = serialize_to_flyte_literal(python_object)
+        identifier = make_identifier_for_serializeable(python_object)
+        self.literal_store[identifier] = serialized_item
```
fg91, line 48: Could we please check here that `identifier` isn't already contained in `literal_store`.

plugins/flytekit-pydantic/tests/test_type_transformer.py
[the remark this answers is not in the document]
fg91: Notes from huddle:

We cannot use `.schema()` unless with use e.g. `TypeVar("pd.Dataframe")` as typehint  instead of `pd.Dataframe` but if we do that the validators are not invoked.

This test can be removed then.

pingsutw (commented):
Overall LGTM. Could you run a task with pydantic inputs, and enable cache? We have to make sure the [type checker](https://github.com/flyteorg/flytepropeller/blob/1541ae5e7792c545c64b7ea356957e5efe155c9a/pkg/compiler/validators/typing.go#L19) in the propeller can work as well.
When the cache is enabled, Propeller will check the literal store in the cache (datacatalog) matches the input literal type. If it matches, then the propeller will create a pod to run the task.

plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py
```diff
@@ -0,0 +1,148 @@
...
+    def validator_generator(*args, **kwags) -> Iterator[Callable[[Any], type_engine.T]]:
+        """Generator that returns validators."""
+        yield validator
+        yield from previous_validators
+        yield from ADDITIONAL_FLYTETYPE_VALIDATORS.get(flyte_obj_type, [])
+
+    return validator_generator
+
+
+def validate_flytefile(flytefile: Union[str, file.FlyteFile]) -> file.FlyteFile:
```
wild-endeavor, line 115: don't really understand these validators.  is this really required?  is this a pydantic thing?  i feel like we should let the base flytekit type engine handle the transformations

wild-endeavor (commented):
Left comments on https://github.com/flyteorg/flytekit/pull/1718/files - ended up being easier just to add comments through an editor rather than the web.  My main point of concern is the scalar/literalmap bit.  (the formatting changes are a result of `make fmt` btw)

wild-endeavor:
@ArthurBook @fg91 will you guys be at the meeting today?  maybe we can talk about this pr

ArthurBook:
hi! yeah I'll join.

fg91:
> 

Yes, will be there :+1:

plugins/flytekit-pydantic/flytekitplugins/pydantic/basemodel_transformer.py
```diff
@@ -0,0 +1,65 @@
...
+
+class BaseModelTransformer(type_engine.TypeTransformer[pydantic.BaseModel]):
+    _TYPE_INFO = types.LiteralType(simple=types.SimpleType.STRUCT)
+
+    def __init__(self):
+        """Construct pydantic.BaseModelTransformer."""
+        super().__init__(name="basemodel-transform", t=pydantic.BaseModel)
+
+    def get_literal_type(self, t: Type[pydantic.BaseModel]) -> types.LiteralType:
+        return types.LiteralType(simple=types.SimpleType.STRUCT)
```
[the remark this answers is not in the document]
fg91: I have used `types.LiteralType(simple=types.SimpleType.STRUCT)` before in an internal plugin and never saw any suspicious behavior. Doesn't mean there isn't anything I have missed though.

> i guess my main question is why the need for the separate store? is there a binary/byte type in pydantic that can be somehow flagged that we can just serialize the pb message into?

The pydantic base model is serialized into a json string by pydantic. The offloaded objects are stored in a `dict[str, Literal]` by the logic of this type transformer.

Parsing the json string into a protobuf struct (see [here](https://github.com/ArthurBook/flytekit/blob/0e9f21559ee688f907e651777e70cb070efce91a/plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py#L110)) and then saving both the "json struct" and the "store" into a `LiteralMap` was my first idea to put these two into a single literal but I'm open to any other way as well.

ArthurBook:
Hi all.

Addressed some comments:
- Changed [BaseModelTransformer](https://github.com/ArthurBook/flytekit/blob/4c463dee4088e5e44dd1086cff6d7d2903df724d/plugins/flytekit-pydantic/flytekitplugins/pydantic/basemodel_transformer.py#L25C22-L25C22) to return Literal instead of LiteralMap 
- Added `type_engine.TypeEngine.lazy_import_transformers()` call to grab all TypeEngine supported types [here](https://github.com/ArthurBook/flytekit/blob/4c463dee4088e5e44dd1086cff6d7d2903df724d/plugins/flytekit-pydantic/flytekitplugins/pydantic/commons.py#L24C9-L24C9). Noticed that some of those, namely np.ndarray and pyarrow.lib.table prevent the __set_attr__, which hinders us to add the deserialization validator [here](https://github.com/ArthurBook/flytekit/blob/4c463dee4088e5e44dd1086cff6d7d2903df724d/plugins/flytekit-pydantic/flytekitplugins/pydantic/deserialization.py#L82C16-L82C16). The current workaround is to exclude those types.
- Formatting updates

Started testing pydantic 2 and at a first glance it seems trickier than expected.
Several of the methods used for serialization/deserialization in <2 are deprecated. I'm wondering at this point whether @fg91 your idea to internalize the serialization logic works better than to add the validators to the flyte supported types (it seems hackier to do this in pydantic 2). Also seems that the encoder argument to model.json() is removed, so the deserialization will also require some change.

plugins/flytekit-pydantic/setup.py
```diff
@@ -4,7 +4,7 @@
 
 microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
 
-plugin_requires = ["flytekit>=1.1.0b0,<2.0.0", "pydantic"]
+plugin_requires = ["flytekit>=1.7.0b0,<2.0.0", "pydantic"]
```
wild-endeavor, line 7: given the hesitation around pydantic 2, should we pin this to <2?

wild-endeavor (commented):
thoughts on making it a literal scalar instead of a literal map?  otherwise looks good.  thank you

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def register_python_object(self, python_object: object) -> LiteralObjID:
+        """Serialize to literal and return a unique identifier."""
+        serialized_item = serialize_to_flyte_literal(python_object)
+        identifier = make_identifier_for_serializeable(python_object)
+        assert identifier not in self.literal_store
+        self.literal_store[identifier] = serialized_item
+        return identifier
+
+    def as_literalmap(self) -> literals.LiteralMap:
```
pingsutw, line 52: nit
```suggestion
    def as_literal_map(self) -> literals.LiteralMap:
```

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
```diff
@@ -0,0 +1,119 @@
...
+def serialize_basemodel(basemodel: pydantic.BaseModel) -> literals.Literal:
+    """
+    Serializes a given pydantic BaseModel instance into a LiteralMap.
+    The BaseModel is first serialized into a JSON format, where all Flyte types are replaced with unique placeholder strings.
+    The Flyte Types are serialized into separate Flyte literals
+    """
+    store = BaseModelFlyteObjectStore()
+    basemodel_literal = serialize_basemodel_to_literal(basemodel, store)
+    store_literal = store.as_literalmap()
+    basemodel_literalmap = literals.LiteralMap(
```
pingsutw, line 68:
```suggestion
    basemodel_literal_map = literals.LiteralMap(
```

plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py
```diff
@@ -0,0 +1,119 @@
...
+        {
+            BASEMODEL_JSON_KEY: basemodel_literal,  # json with flyte types replaced with placeholders
+            FLYTETYPE_OBJSTORE_KEY: store_literal,  # placeholders mapped to flyte types
+        }
+    )
+    literal = literals.Literal(map=basemodel_literalmap)  # type: ignore
```
pingsutw, line 74:
```suggestion
    literal = literals.Literal(map=basemodel_literal_map)  # type: ignore
```

fg91:
> given the hesitation around pydantic 2, should we pin this to <2? (comment from @pingsutw)

In my opinion we shouldn't release a Pydantic 1.x Flyte plugin that is expected to not work the same way with version 2, given that v2 is now already released. If we can't find a good way that will work for both versions, I'd vote for building something for v2 only.

> Started testing pydantic 2 and at a first glance it seems trickier than expected. Several of the methods used for serialization/deserialization in <2 are deprecated. I'm wondering at this point whether @fg91 your idea to internalize the serialization logic works better than to add the validators to the flyte supported types (it seems hackier to do this in pydantic 2). Also seems that the encoder argument to model.json() is removed, so the deserialization will also require some change. (comment from @ArthurBook)

How would the UX look like in pydantic v2 (disregarding Flyte completely) to add e.g. a pandas Dataframe (as an example of a Flyte supported type) to a `BaseModel` without allowing arbitrary types?

Would you say that adding e.g. a dataframe to a base model is something that Pydantic (especially v2) expects users to do? Or is does this require hacks already in Pydantic, again disregarding any Flyte type system integration?

I'm asking these questions because, taking a step back, I'm wondering whether we would still come to the conclusion that "the pydantic base model type transformer should support all flyte types", a goal we formulated during the [brainstorming meeting on March 9th](https://github.com/flyteorg/flyte/issues/2686#issuecomment-1462674447). Does Pydantic really support those types? Or are we encountering these issues because we are trying to make the type transformer do something that even Pydantic isn't supposed to do?

Not saying this is the case but now I'd like to be sure it isn't. Interested in your thoughts about this @ArthurBook @pingsutw @wild-endeavor @cosmicBboy  @ggydush.

wild-endeavor:
I'm not sure about pydantic, but I would like to re-start the dataclass (non-dataclass_json dataclass) effort since that is also a container type.  That transformer should definitely support generalized types including Flyte types.  I think it's important for us to have better support for at least three of the popular containers soon (attrs, dc, pydantic), but I'm okay if pydantic doesn't make the cut.

fg91:
> I'm not sure about pydantic [...]  support generalized types including Flyte types [...] but I'm okay if pydantic doesn't make the cut.

For me the main question is whether it is within pydantic's typical scope to support general types - if yes, so should the type transformer in my opinion.

Not saying we should take this as an authority argument, but I asked GPT4 about this:

> I want to include a pandas dataframe and a pytorch nn.Module to be trained on the pandas dataframe into a pydantic BaseModel. How do I do this? [...] Would you say including something like a dataframe or nn.Module in a pydantic base model is something uncommon?

> You can include a pandas DataFrame and a PyTorch nn.Module in a Pydantic BaseModel by using the arbitrary_types_allowed configuration option. This allows you to use arbitrary types as fields in your Pydantic model. Here’s an example [...] It is not very common to include a pandas DataFrame or a PyTorch nn.Module directly in a Pydantic BaseModel. Pydantic is typically used for data validation and parsing, and its primary use case is to validate and parse data coming from external sources, such as APIs or user input. As such, it is more common to use Pydantic to validate and parse data into a format that can be used by other libraries, such as pandas or PyTorch. That being said, it is certainly possible to include a DataFrame or an nn.Module in a Pydantic model, as I showed in my previous response. This can be useful in certain situations where you want to include these objects as part of your data model. However, it is important to note that this is not the primary use case for Pydantic, and there may be other, more suitable tools for working with these types of objects.

(I went on to ask about a way to do this without allowing arbitrary types. It suggested to create a class inheriting from pandas DataFrame and giving this class a validator, i.e. similar to what we did by injecting the validator.)


@ArthurBook said:

> Several of the methods used for serialization/deserialization in <2 are deprecated. I'm wondering at this point whether @fg91 your idea to internalize the serialization logic works better than to add the validators to the flyte supported types (it seems hackier to do this in pydantic 2)

He is referring to [this](https://github.com/flyteorg/flytekit/pull/1660#issuecomment-1598296796) suggestion, @wild-endeavor.
Internalizing the serialization/deserialization logic this way (instead of monkeypatching the validators) at least wouldn't require any hacks in my opinion.
If the current way of injecting the validators requires more hacks in v2, I would suggest that we decide between 1) using arbitrary types = true + internalized (de-) serialization logic for flyte types or 2) not supporting arbitrary types.

Really interested what you think about especially the pydantic use case question @ArthurBook.

ArthurBook:
Hello! Apols for a delayed response.

Looking at [this](https://docs.pydantic.dev/latest/usage/serialization/#custom-serializers), I think we have Pydantic 2's recommended approach for our scenario. However, the overhaul in Pydantic 2's serialization/deserialization makes it challenging to create a single solution compatible with both Pydantic 1 and 2. We could probably work around this by crafting our own BaseModel serializer/deserializer, but I'd need to familiarize myself into Pydantic 2 before looking into how that would look.

About BaseModels and arbitrary types, I think it's a matter of taste. I have a strong personal preference for Pydantic as a JSON-centric framework, probably because of my primary usage of the package for reading/writing data from disk and receiving/sending fastapi payloads (v neat integration BTW).

As for the plugin, I think inclusion of flyte files and - directories is a must for the plugin's usability. While I'd keep other types out of my BaseModels, I think we've already done the groundwork for handling custom data structures, so it's a small step to apply to add that feature if we are anyways supporting flytefiles and -dirs. Just my two cents.

wild-endeavor:
@ArthurBook @fg91 should we just add the pydantic version gate and then merge this in?  tested on a `flytectl demo` environment right?  also lint is breaking, could you take a look please?

wild-endeavor:
@fg91 sorry i think i lost the slack thread. did you find issues in testing?

fg91:
> @fg91 sorry i think i lost the slack thread. did you find issues in testing?

Yes, I did, I boiled it down to this minimal working example:

```python

class Config(BaseModel):
    path: str


@task
def train(cfg: Config):
    print(cfg)


@dynamic(cache=True, cache_version="0.3")
def sub_wf(cfg: Config):
    train(cfg=cfg)


@workflow
def wf():
    sub_wf(cfg=Config(path="bar"))


if __name__ == "__main__":
    wf()
```

Upon local execution with `python test.py`, the error is:

```
│ /home/fabiogratz/miniconda3/envs/dev/lib/python3.9/site-packages/flytekit/core/workflow.py:271 in __call__                  │
│                                                                                                                             │
│ ❱ 271 │   │   │   raise exc                                                                                                 │
│                                                                                                                             │
│ /home/fabiogratz/miniconda3/envs/dev/lib/python3.9/site-packages/flytekit/core/workflow.py:268 in __call__                  │
│                                                                                                                             │
│ ❱ 268 │   │   │   return flyte_entity_call_handler(self, *args, **input_kwargs)                                             │
│                                                                                                                             │
│ /home/fabiogratz/miniconda3/envs/dev/lib/python3.9/site-packages/flytekit/core/promise.py:1022 in flyte_entity_call_handler │
│                                                                                                                             │
│ ❱ 1022 │   │   │   result = cast(LocallyExecutable, entity).local_execute(child_ctx, **kwargs)                              │
│                                                                                                                             │
│ /home/fabiogratz/miniconda3/envs/dev/lib/python3.9/site-packages/flytekit/core/workflow.py:290 in local_execute             │
│                                                                                                                             │
│ ❱ 290 │   │   function_outputs = self.execute(**kwargs_literals)                                                            │
│                                                                                                                             │
│ /home/fabiogratz/miniconda3/envs/dev/lib/python3.9/site-packages/flytekit/core/workflow.py:729 in execute                   │
│                                                                                                                             │
│ ❱ 729 │   │   return exception_scopes.user_entry_point(self._workflow_function)(**kwargs)                                   │
│                                                                                                                             │
│ /home/fabiogratz/miniconda3/envs/dev/lib/python3.9/site-packages/flytekit/exceptions/scopes.py:203 in user_entry_point      │
│                                                                                                                             │
│ ❱ 203 │   │   │   │   raise type(exc)(f"Error encountered while executing '{fn_name}':\n  {exc                              │
╰─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: Encountered error while executing workflow 'wf':
  Error encountered while executing 'wf':
  Parameter to MergeFrom() must be instance of same class: expected <class 'flyteidl.core.literals_pb2.Literal'> got <class 
'flyteidl.core.literals_pb2.LiteralMap'>.
```

When setting `cache=False` it works. But having forgotten to bump the cache version appears not to be the problem. When activating the cache and setting a cache version that 100% was never used before, it fails with this error on the first try. (edited)

ArthurBook:
Hi all, haven't forgotten about this - just busy with a product launch + travel.
I'll aim to get to this ASAP but need to prioritize work stuff first.

wild-endeavor:
is this a common occurrence though?
```
literals_pb2.LiteralMap(literals={'boo': literals_pb2.LiteralMap(literals={})})
```
wouldn't it be
```
literals_pb2.LiteralMap(literals={'boo': literals_pb2.Literal(map=literals_pb2.LiteralMap(literals={})}))
```
not sure if this is relevant at all.

ArthurBook:
That would work too! Let me know if you have a preference to serializing to that rather than a flat literalmap [like here](https://github.com/flyteorg/flytekit/blob/cf753266dafc2d927e7fa57a37f82caf115e9a80/plugins/flytekit-pydantic/flytekitplugins/pydantic/serialization.py#L61)?

wild-endeavor (commented):
can't commit to your branch but added a unit test that at least helps me understand what's going on. - https://github.com/flyteorg/flytekit/commit/01c065e86a891b06869f794a4a58311c648009f8 and updated some formatting.  can you pull this in and then we should merge.

Thanks!

wild-endeavor (dismissed):

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #1661, opened 2023-05-28
Pickle as a default Transformer rather than a special case of function signatures

elibixby:
# TL;DR
Right now only "top level" missing types (those defined directly in a function) are pickled if they don't have a registered `TypeTransformer`. This PR alters that behavior to anywhere `TypeTransformer`s are used.

## Type
 - [ ] Bug Fix
 - [X ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
The special casing for pickling as a serialization/deserialization method leads to some strange behavior where, e.g

```

@task
def my_task(foo: Foo):
   ...
```

Is valid, because that type is manually converted, but e.g. or `Foo` inside of a dataclass etc are not. Because the conversion to the placeholder `FlytePickle` type happens in a function which is only called once on top level type, and only traverses the `origin` stack, rather than being used anywhere `TypeTransformers` are used. This PR fixes that problem.

## Tracking Issue

## Follow-up issue
_NA_
OR


kumare3:
this seems to be abandoned

PR #1665, opened 2023-05-30
Updates to pyflyte init templating

zeryx:
# TL;DR
This PR removes the cookiecutter templating system for `pyflyte init`, and instead uses gitpython, without templating.
This is part of a wider effort to replace our Dockerfile / docker_build.sh systems with the much simpler `ImageSpec`.

This PR depends on https://github.com/flyteorg/flytekit-python-template/pull/40

**This is a breaking change**
All templates are now imagespec based, and cookiecutter is no longer used to generate custom readme titles or directory names.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
As discussed as part of the larger Flyte example improvement with newly introduced imagespec.

### DIFF
--- a/flytekit/clis/sdk_in_container/helpers.py
+++ b/flytekit/clis/sdk_in_container/helpers.py
@@ -1,6 +1,7 @@
+import os
+import shutil
 from dataclasses import replace
 from typing import Optional
-import shutil, os
 
 import rich_click as click
 from git import Repo
@@ -78,13 +79,13 @@ def clone_and_copy_repo_dir(git_url, branch, src_dir, dest_dir):
     - dest_dir: The local directory to copy into.
     """
     # Clone the repo
-    repo = Repo.clone_from(git_url, 'temp_repo')
+    repo = Repo.clone_from(git_url, "temp_repo")
 
     # Checkout to the branch
     repo.git.checkout(branch)
 
     # Define the source directory path
-    src_path = os.path.join('temp_repo', src_dir)
+    src_path = os.path.join("temp_repo", src_dir)
 
     # Check if source directory exists
     if not os.path.exists(src_path):
@@ -95,4 +96,4 @@ def clone_and_copy_repo_dir(git_url, branch, src_dir, dest_dir):
     shutil.copytree(src_path, dest_dir)
 
     # Remove the temporary cloned repo
-    shutil.rmtree('temp_repo')
+    shutil.rmtree("temp_repo")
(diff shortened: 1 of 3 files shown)


flytekit/clis/sdk_in_container/helpers.py
```diff
@@ -63,3 +66,34 @@ def patch_image_config(config_file: Optional[str], image_config: ImageConfig) ->
...
+    - github_url: The URL of the GitHub repository.
+    - branch: The branch of the repository.
+    - src_dir: The directory in the repository to copy.
+    - dest_dir: The local directory to copy into.
+    """
+    # Clone the repo
+    repo = Repo.clone_from(git_url, "temp_repo")
+
+    # Checkout to the branch
+    repo.git.checkout(branch)
```
eapolinario, line 85: you can pass the branch name directly in the call to `Repo.clone_from`.

flytekit/clis/sdk_in_container/helpers.py
```diff
@@ -63,3 +66,34 @@ def patch_image_config(config_file: Optional[str], image_config: ImageConfig) ->
...
+    # Check if source directory exists
+    if not os.path.exists(src_path):
+        print(f"Template named {src_dir} does not exist in the repository.")
+        return
+
+    # Copy the files from the source directory to the destination directory
+    shutil.copytree(src_path, dest_dir)
+
+    # Remove the temporary cloned repo
+    shutil.rmtree("temp_repo")
```
eapolinario, line 99: how about we use a [`TemporaryDirectory`](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory) to hold the directory? This way we don't need to do this book-keeping and it also allows for this command to be run multiple times.

setup.py
```diff
@@ -68,7 +68,6 @@
         "docstring-parser>=0.9.0",
         "diskcache>=5.2.1",
         "cloudpickle>=2.0.0",
-        "cookiecutter>=1.7.3",
```
eapolinario, line 71: ❤️

eapolinario (commented):
This is awesome!

We might think about using a brand new repo to hold the examples (instead of flytekit-python-template).

kumare3:
cc @eapolinario / @neverett do you think you want to check this one in?

PR #1674, opened 2023-06-03
Flyte `flyte://` file system and improve remote file handling

wild-endeavor:
# TL;DR
**Examples of improvements**
- pyflyte run used to use a special way of handling certain objects, now that is centralized to use the typeEngine
- pyflyte run and gate node used different string parsing systems, now that is centralized
- pyflyte run can now use multiple files in a directory locally to create a new remote FlyteDirectory representation
- pyflyte run code has been simplified
- workflow defaults that point to local files/folders are auto-uploaded and includes limited support for dataframes. Also data is maintained using hashes so enables versioning

**New additions**
- pyflyte info added
- pyflyte fetch <flyte-uri> added
- This introduces new flyte-uri concept that is available in the UI and you can retrieve inputs and outputs from any execution (workflow or even a single node) instantaneously as it deeplinks into it.
centralizes fsspec as the way to interact with data

**Planned work**
- Artifacts service that is unblocked and simplified with this 
- improved launchform experience from flyte uris
and many more data interaction improvements

## Details
**What it does?**
* adds a new fsspec file system where `flyte://` is the protocol.  This is not a real filesystem though, so it's only used in `pyflyte run` and `pyflyte register`.
* cleans up code around data persistence.

**The reason the file system is useful:**
* Defaults for workflows that are of offloaded types can get uploaded properly during pyflyte run/register.
* `pyflyte run` currently duplicates a bunch of work that's supposed to be handled by the type engine because it needs to handle remote upload.  This can be cleaned up in the future if the data proxy service can be accessible at a lower level.

**The drawback of this approach:**
* The data proxy layer is not a real filesystem.  Currently you can't know beforehand where the thing you upload will land.  That is, if you think about writing files as a `cp src flyte://dest` command, the destination is ignored.  For this reason, `rpath` in the implementation is always ignored and a placeholder is used instead.  Instead, the `remote_fs` implementation returns the s3 native url as part of the `put` call.
* If your default input is a file/folder/pickle, it works, but if it's a dataframe, it works, but sadly, you may end up with duplicate registration problems, the reasons are as follows,

This is basically an fsspec filesystem that writes to a different location than what is specified.  However, it's not always possible to capture the output where the file was really written to. More concretely, if you do:
```
df.to_parquet("s3://a/b/c.parquet")
```
pandas will use fsspec to fetch the file system, call `open` on it, and then write to the open buffer.  After writing, users expect that s3 location to hold the contents of the dataframe.  With the flyte remote fs, if you called `df.to_parquet("flyte://anything")`, the dataframe contents do not end up there.  Instead they end up in some `s3://` location (assuming s3 is the blob store).

Because of this, calls to `put_data` were updated to read back the `remote_path`, now [returned by](https://github.com/flyteorg/flytekit/pull/1674/files#diff-daae50d27a419e3da4b165892cc1d8761f67990d83fb153891401424d7555e80R455) `put_data`.  (this is why dataframes don't work as default input.  we can of course add another transformer that writes to a memory buffer or file first of course only for pyflyte run/register.)

To promote the behavior, a new function was added to data_persistence `put_raw_data` which writes always to the raw output data prefix.  See the [function](https://github.com/flyteorg/flytekit/pull/1674/files#diff-daae50d27a419e3da4b165892cc1d8761f67990d83fb153891401424d7555e80R237) and [corresponding test](https://github.com/flyteorg/flytekit/pull/1674/files#diff-91e3be485e56787b7ac625ca8cb16adeada42372bc6b2de4f97177f24dd804cfR30).  This PR does not update all the calls yet to the new interface.

The flyte fs also does not support `open`.  The reason is simply that the data proxy call requires that we know the MD5 hash.  This doesn't really make sense in the context of an open call where the data might not even be known yet.  If this wasn't the case, we could've written the pandas transformer to fs.open() and then write directly to the buffer.

Also,
* The old `get_random_remote_path/directory` was removed from data persistence and rewritten to be based on an input fs.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Smaller changes
* Add length to `hash_file`
* Added a limited-use (only Remote uses it for now) local fsspec implementation which overrides the `/` field for proper Windows use.

#### Other thoughts
We also briefly thought about replacing the fsspec s3 filesystem with something that will do the signed urls but this felt more complicated and felt like it interfered too much with fsspec.  Also would be dependent on what blob store admin was configured to use.

Needs: https://github.com/flyteorg/flyteidl/pull/416 and https://github.com/flyteorg/flyteadmin/pull/577
https://github.com/flyteorg/flytestdlib/pull/160 and https://github.com/flyteorg/flyteadmin/pull/587

## Tracking Issue
NA


flytekit/core/data_persistence.py
```diff
@@ -72,10 +74,15 @@ def __init__(
         local_sandbox_dir: Union[str, os.PathLike],
         raw_output_prefix: str,
         data_config: typing.Optional[DataConfig] = None,
+        raw_output_local_staging_suffix: typing.Optional[str] = None,
```
pingsutw, line 77: Do we use it anywhere?

flytekit/core/data_persistence.py
```diff
@@ -219,32 +234,38 @@ def put(self, from_path: str, to_path: str, recursive: bool = False, **kwargs):
             from_path, to_path = self.recursive_paths(from_path, to_path)
         return file_system.put(from_path, to_path, recursive=recursive, **kwargs)
 
-    def get_random_remote_path(self, file_path_or_file_name: typing.Optional[str] = None) -> str:
```
pingsutw, line 222: Should we deprecate it first? Someone might use it in their code.

wild-endeavor: yeah i can add it back... in a revised form.

wild-endeavor:
Discussed this with @EngHabu yesterday.

* It's probably okay to make the `flyte://` url a bijection on to the current [create upload link request](https://github.com/flyteorg/flyteidl/blob/b219c2ab37886801039fda67d913760ac6fc4c8b/protos/flyteidl/service/dataproxy.proto#L30) minus the sha.  Currently that request looks like:
  ```
  message CreateUploadLocationRequest {
    string project = 1;
    string domain = 2;
    string filename = 3;
    google.protobuf.Duration expires_in = 4;
    bytes content_md5 = 5;
    string filename_root = 6;
  }
  ```
  The reason the SHA doesn't play nice is that it fundamentally precludes a class of use-cases - where the user has a stream of data that they want to write to this data store, but the stream isn't known yet, so there's no way to compute the sha (e.g. "record data from this temperature sensor for the next ten minutes").
  * Doing this bullet point will make this flyte remote fs look and feel more like a traditional file system.  We felt like this was an advantage.
* To properly namespace this endpoint from the existing tiny URLs that Admin's already supports we should add a prefix - `flyte://dv1/` (i.e. data, version 1).  The current tiny url endpoint for task outputs is namespaced `flyte://v1`.
* The final url will look something like
  ```
  flyte://dv1/project/domain/filename_root/filename
  ```
  * Since the remote fsspec has a remote object in it, and the remote object has a project and domain, flytekit can potentially support non-absolute paths.  Also the raw output prefix should probably be configured to have the project/domain.
* Admin will have to change as it currently requires a sha.  We will remove the sha from admin and add the following logic.
  * If filename_root is specified (the recent use-case that was added to support folder upload).
    * If the sha is specified, then we will check to see if the file currently exists.  If it does we'll check that the sha matches, if it does then admin will continue, if not then error.  The sha if the file already exists should be returned in a head operation.
      * If the sha is missing then return a different error.
    * If the sha is not specified, then make sure that the file doesn't already exist. If it exists then error.
  * If filename_root is not specified, just use current behavior.
* This scheme is not race-condition proof and will lead to dangling files.  But this preserves the safety of not overwriting files.
* When the Artifact service is complete we can keep track of dangling files since a record will be written to that service before the upload url is returned.

EngHabu:
Excellent point regarding md5 for multi-part uploads... while that won't be a problem because the API doesn't do that, it's the same problem on s3... the ETag you will get on a multi-part-uploaded file is the hash of the hashes of each part...
Otherwise looks good to me

flytekit/remote/remote_fs.py
```diff
@@ -0,0 +1,257 @@
...
+    protocol = "flyte"
+
+    def __init__(
+        self,
+        remote: FlyteRemote,
+        **storage_options,
+    ):
+        super().__init__(**storage_options)
+        self._remote = remote
+        self._local_map: typing.Dict[str, str] = {}
```
wild-endeavor, line 57: we'll need a second map for flyte:// -> native url as well.

wild-endeavor:
our fearless leader @kumare3 will find a way!

kumare3:
@pingsutw / @wild-endeavor / @eapolinario this PR is ready and works in cases I have tested. I think a few more unit tests would be good. But, one problem. take the following example

```python
df = pd.DataFrame({"Name": ["Tom", "Joseph11"], "Age": [20, 22]})

@workflow
def wf(sd: StructuredDataset = StructuredDataset(dataframe=df)) -> StructuredDataset:
    return t1(sd=sd)
```

This currently does not work. With this PR, this will work, but will cause "registration clashes" even for registering the same code again, as the path is not auto de-duped. 
As this does not work today, it should not break any old code, but may break new code.

The problem is because we are using pandas parquet writer, we cannot control the remote path. The remote path is created as a random path on the raw prefix. One solution might be to hash the data value, but this is expensive (though completely ok for the registration case). Identifying the difference between registration vs runtime may be tough.

I am still investigating, but want to make the PR ready for review

flytekit/clis/flyte_cli/main.py
```diff
@@ -1822,7 +1822,7 @@ def fast_register_files(
     version = version if version else digest
     full_remote_path = _get_additional_distribution_loc(additional_distribution_dir, version)
     ctx = FlyteContextManager.current_context()
-    ctx.file_access.put_data(compressed_source, full_remote_path)
+    full_remote_path = ctx.file_access.put_data(compressed_source, full_remote_path)
```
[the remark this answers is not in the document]
kumare3: it is used in the next line to print. But it is overriden.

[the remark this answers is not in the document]
pingsutw: nvm, my bad.

flytekit/clis/sdk_in_container/run.py
```diff
@@ -256,8 +255,14 @@ class RunLevelParams(PyFlyteParams):
     _remote: typing.Optional[FlyteRemote] = None
 
     def remote_instance(self) -> FlyteRemote:
+        # TODO we have to check if the previous remote was for the same data upload location
         if self._remote is None:
-            self._remote = get_remote(self.config_file, self.project, self.domain)
+            # TODO @wild-endeavor - why should the local data upload location be /tmp? what should it be?
+            # Also why do we even copy the local data?
+            data_upload_location = "/tmp"
```
pingsutw, line 262: Do we need to use random paths? [As in this case](https://github.com/flyteorg/flytekit/blob/1c813759b2c56d4821350918ee0aef9a78d86859/flytekit/core/data_persistence.py#L498-L501)

flytekit/clis/version.py
```diff
@@ -0,0 +1,27 @@
...
+Flytekit Version: [cyan]{version}[reset]
+Flyte Backend Endpoint: [cyan]{endpoint}
+"""
+
+
+@click.command("info")
+@click.pass_context
+def info(ctx: click.Context):
+    """
+    Print out information about the current Flyte Python CLI environment - like the version of Flytekit, backend endpoint
```
pingsutw, line 20: ❤️

flytekit/extras/tensorflow/record.py
```diff
@@ -81,11 +81,14 @@ def to_literal(
...
-        remote_path = ctx.file_access.get_random_remote_path()
+        remote_path = ctx.file_access.join(
+            ctx.file_access.raw_output_prefix,
+            ctx.file_access.get_random_string(),
+        )
         local_path = os.path.join(local_dir, "0000.tfrecord")
         with tf.io.TFRecordWriter(local_path) as writer:
             writer.write(python_val.SerializeToString())
-        ctx.file_access.put_data(local_path, remote_path, is_multipart=False)
+        remote_path = ctx.file_access.put_raw_data(local_path)
```
pingsutw, line 91: do we need to override here? I thought type_engine would modify it

flytekit/remote/remote_fs.py
```diff
@@ -0,0 +1,312 @@
...
+            return False
+        self.buffer.seek(0)
+        data = self.buffer.read()
+
+        # h = hashlib.md5()
+        # h.update(data)
+        # md5 = h.digest()
+        # l = len(data)
+        #
+        # headers = {"Content-Length": str(l), "Content-MD5": md5}
```
pingsutw, line 81:
```suggestion
```

pingsutw (commented):
LGTM, some questions in the comment

flytekit/extras/tensorflow/record.py
kumare3: i think we are going to override everywhere. TypeEngine could modify, but i want to be sure, we should do it wherever we can

flytekit/clis/sdk_in_container/run.py
kumare3: done

pingsutw (approved):

Tom-Newton (commented):
Hello. I've been trying to upgrade from flytekit 1.9.1 to 1.10.2 and I think the changes from this PR are causing some issues for me when I try to use a `FlyteFile` as an input to a workflow that I submit using `pyflyte`. 

I've left a couple of comments but in addition to that I'm having issues when re-running the same workflow with the same `FllyteFile`.
```
	details = "file already exists at location [<path>], specify a matching hash if you wish to rewrite"
	debug_error_string = "UNKNOWN:Error received from peer ipv4:172.27.128.127:81 {created_time:"2024-01-22T21:59:07.492837995+00:00", grpc_status:6, grpc_message:"file already exists at location [<path>], specify a matching hash if you wish to rewrite"}"
```
This one confuses me because it looks like its doing the same as before except previously it didn't fail.

flytekit/remote/remote_fs.py
```diff
@@ -0,0 +1,312 @@
...
+    ):
+        """
+        fsspec will call this method to upload a file. If recursive, rpath will already be individual files.
+        Make the request and upload, but then how do we get the s3 paths back to the user?
+        """
+        # remove from kwargs otherwise super() call will fail
+        p = kwargs.pop(_PREFIX_KEY)
+        hashes = kwargs.pop(_HASHES_KEY)
+        # Parse rpath, strip out everything that doesn't make sense.
+        rpath = rpath.replace(f"{REMOTE_PLACEHOLDER}/", "", 1)
```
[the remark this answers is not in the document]
Tom-Newton: Sorry, I don't understand how its opt in. `pyflyte` depends on this if you use `FlyteFile`.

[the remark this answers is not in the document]
wild-endeavor: https://github.com/flyteorg/flytekit/pull/2121 this is the pr i'm working on, but not sure what's going on with the test failures, will have to take a look later this week.

Tom-Newton:
BTW I worked out the `file already exists at location [<path>], specify a matching hash if you wish to rewrite` error. It appears this is because `remote_fs.get_upload_link` sets the `filename_root` argument when it calls `SynchronousFlyteClient.get_upload_signed_url` while the old remote uploads in pyflyte did not.

PR #1676, opened 2023-06-06
Update fg91 runtime overrides PR

bstadlbauer:
# TL;DR
- Added `ExecutionSpec.task_node_runtime_overrides`
- Added `LaunchPlanSpec.task_node_runtime_overrides`
- Added tests throughout. *Note*: Have not added any tests for `tools.translator`
- Removed the `runtime_override_default`

@fg91 This should be all we need on the model side. We will still need to add tests and functionality for high-level functions such as `LaunchPlan.get_or_create(...)` and `pyflyte run ...`. Probably it would be nice if translate the usecases we have in the RFC into unit tests

### DIFF
--- a/flytekit/core/launch_plan.py
+++ b/flytekit/core/launch_plan.py
@@ -117,6 +117,7 @@ class LaunchPlan(object):
         max_parallelism: Optional[int] = None,
         security_context: Optional[security.SecurityContext] = None,
         auth_role: Optional[_common_models.AuthRole] = None,
+        task_node_overrides: Optional[_workflow_model.TaskNodeOverrides] = None,
     ) -> LaunchPlan:
         ctx = FlyteContextManager.current_context()
         default_inputs = default_inputs or {}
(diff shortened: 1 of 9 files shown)


ByronHsu:
https://github.com/flyteorg/flyte/issues/475

ByronHsu:
Regarding the question in the [wg channel ](https://flyte-org.slack.com/archives/C05A4PNMKL1/p1686039149200379)
> How will we deal with task_config overrides? In particular, how can we ensure that the type is not changed? What will the UX for this look like?

I think we can just pass any task config task, marshal as pb struct, and let flyteadmin resolves the tasktype resolve, like what we did in[ flytepropeller](https://github.com/flyteorg/flytepropeller/blob/95a4791f59452845714bb22252c77dee57ac38c1/pkg/controller/nodes/task/handler.go#L325)
```python
TaskNodeOverrides(
   task_config=Tfjob(...)
)
```


> Should we rename TaskNode.with_runtime_overrides to TaskNode.with_runtime_override? As this is a 1:1 mapping? :thinking_face:
Does the `s` mean we can override many things at once? I saw we also name `TaskNodeOverrides`.

PR #1677, opened 2023-06-06
Fix config of user facing execution parameters in spawning elastic tasks

fg91:
# TL;DR

When using `@task(task_config=flytekitplugins.kfpytorch.Elastic())`, the task function is started in a number of worker processes using torch `elastic_launch` (`torchrun`). The processes can be created using fork or spawn which is controlled by the arg `Elastic(start_method=...)`.

When using `fork`, the child process inherits a copy of the parent process' stack including the flyte context and the user facing execution parameters `ctx = flytekit.current_context()`.

When spawning, however, fresh processes are started and the flyte context and the execution parameters are not transferred to the child process currently. This means that within a task with ``@task(task_config=Elastic(start_method="spawn"))` the execution id and the checkpoint cannot be accessed from the execution parameters.

This PR fixes this by setting up the flyte context in the spawned worker processes.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

In the spawned worker processes I call `flytekit.bin.entrypoint.setup_execution` which sets up the flyte context the same way as when a normal python task is started. Raw data prefix and checkpoint pathes are transferred from the parent process.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/plugins/flytekit-kf-pytorch/tests/test_elastic_task.py
+++ b/plugins/flytekit-kf-pytorch/tests/test_elastic_task.py
@@ -8,6 +8,7 @@ import torch.distributed as dist
 from dataclasses_json import dataclass_json
 from flytekitplugins.kfpytorch.task import Elastic
 
+import flytekit
 from flytekit import task, workflow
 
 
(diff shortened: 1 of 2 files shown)


plugins/flytekit-kf-pytorch/tests/test_elastic_task.py
```diff
@@ -65,3 +66,32 @@ def wf(config: Config = Config()) -> typing.Tuple[str, Config, torch.nn.Module,
...
     """
     assert distributed_result == sum([5 + 2 * rank + world_size for rank in range(world_size)])
+
+
+@pytest.mark.parametrize(
+    "start_method,target_exec_id,monkeypatch_exec_id_env_var",
+    [
+        ("spawn", "", False),
+        ("spawn", "f12345678", True),
+        ("fork", "local", False),
```
fg91, line 76: When spawning, the `execution_id.name`, `.project`, `.domain`, ... are set to the default value `""` [here](https://github.com/flyteorg/flytekit/blob/c8433ea2d2ecf362815b1ee63554549ced9acd75/flytekit/bin/entrypoint.py#L203) when the `FLYTE_INTERNAL_EXECUTION_ID`, ... env vars are not set, i.e. during a local execution.

When executing a workflow/task locally, these execution identifiers are normally set to `"local"` which happens [here](https://github.com/flyteorg/flytekit/blob/c8433ea2d2ecf362815b1ee63554549ced9acd75/flytekit/core/context_manager.py#L840). Since the parent processes stack is copied during forking, `"local"` is set when using this start method.

This difference between forking and spawning in a local execution is something that gives me a bit of grief.

I see two options for removing this difference.

* Not set `""` as the default value for execution id name, project, domain, ... in `flytekit.bin.entrypoint.setup_execution`. Would this have any undesired effect?
* Maintain an adapted copy of `setup_execution` here, which would, however, lead to quite some code duplication which wouldn't be nice either.

plugins/flytekit-kf-pytorch/setup.py
```diff
@@ -4,7 +4,7 @@
 
 microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
 
-plugin_requires = ["cloudpickle", "flytekit>=1.6.1"]
+plugin_requires = ["cloudpickle", "flyteidl>=1.5.1", "flytekit>=1.6.1"]
```
fg91, line 7: Also from https://github.com/flyteorg/flytekit/pull/1678/files

cosmicBboy (approved):

pingsutw (approved):

pingsutw (approved):

PR #1690, opened 2023-06-12
Fix task type version in pytorch and mpi task plugin 

yubofredwang:
# TL;DR
Previously the pytorch plugin is missing comments. We would suggest user to not set configurations for individual replica groups as they should essentially be the same for worker and master. The only fields user should set is number of worker and restart policies.

Fix a bug in pytorch and mpijob to use task type version 1.

Should be merged in after: https://github.com/flyteorg/flyteplugins/pull/359

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Smoke tested on the following with corrected backend plugin
1. Elastic Pytorch Job
2. MPI Task
3. Horovod Job


pingsutw (approved):
LGTM, thank you!

PR #1695, opened 2023-06-15
Feat: Allow disabling rich tracebacks via env var

fg91:
# TL;DR
In `flytekit==1.6.0` rich tracebacks were introduced (#1582).

Before, exceptions were shown like this:
Screenshot 2023-06-15 at 15 37 04

After, they are shown like this:

Screenshot 2023-06-15 at 15 36 44

This happens as long as `import flytekit` is called somewhere in the imported code base, also when not directly working with flyte.

---

Style ultimately is a question of taste but several of our engineers have complained about this change:

> When/Why did the stacktrace get so <s>ugly</s> colorful. It takes up way too much space and pycharm cannot create crosslinks to the files anymore for me :see_no_evil: And long lines are truncated

> I feel the same, does not help me

> Can we configure this

> I wanted to complain about this too :smile:

> The tracebacks contain way more (white space) characters, meaning I can paste a lot less of them into chat gpt

Because of this, in this PR I propose to introduce an environment variable which disables rich tracebacks. I chose a naming inspired by `"FLYTE_SDK_LOGGING_LEVEL"` or `"FLYTE_SDK_LOGGING_FORMAT"` in `flytekit/loggers.py`.

Would this change be be acceptable to you @cosmicBboy @kumare3 @eapolinario ?

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
_NA_

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -192,6 +192,7 @@ Documentation
    SourceCode
 
 """
+import os
 import sys
 from typing import Generator
 
@@ -303,4 +304,5 @@ def load_implicit_plugins():
 load_implicit_plugins()
 
 # Pretty-print exception messages
-traceback.install(width=None, extra_lines=0)
+if os.environ.get("FLYTE_SDK_RICH_TRACEBACKS") != "0":
+    traceback.install(width=None, extra_lines=0)


fg91:
> I like the idea, just find the bar name too big. Want to start making smaller feature flags

What do you think about `FLYTE_RICH_TRC` or `FLYTE_RICH_TRACE` instead of `FLYTE_SDK_RICH_TRACEBACKS`?

eapolinario (approved):
I'm ok with more descriptive environment variables if their use is restricted (which is the case of this PR).

PR #1700, opened 2023-06-20
PR into #1679 - Support for generalized container types and clean-up of promise.py and workflow.py in local execution

wild-endeavor:
This is a PR into https://github.com/flyteorg/flytekit/pull/1679
(currently this PR is a bit hard to follow because it merged in master, hopefully that gets easier once it's merged)
**Blocked**
This PR is currently blocked - it's unclear if the current approach works.  Currently the new dataclass transformer in the main PR interprets a dataclass as a literalmap type with a `map_value_type` set to an empty `LiteralType()`.  This makes the compiler think that no bindings are needed.  Since dataclasses are inherently multi-variate, we're at a bit of a impasse. 
And this PR depends on that. If this approach is untenable, most of this PR should be discarded.  (The parts that should be retained are in the Other issues below.)  One possible way around this is to make the new dataclass transformer follow the pattern of the existing dataclass json transformer - just expose a scalar, and only interpret the 

## Background
There's a few places in flytekit core today that blur the line between the type engine and not type engine.  These are:
* [Extraction](https://github.com/flyteorg/flytekit/blob/68ac1f5371de1dac96df3f4e9df7ed983cd1c0d5/flytekit/core/promise.py#L35) of literals from nested promises (`translate_inputs_to_literals`) for local execution
* [Extraction](https://github.com/flyteorg/flytekit/blob/68ac1f5371de1dac96df3f4e9df7ed983cd1c0d5/flytekit/core/workflow.py#L282) of literals from nested promises for local workflow execution (`ensure_literal`)
* Construction of literals in pyflyte run ([one example](https://github.com/flyteorg/flytekit/blob/68ac1f5371de1dac96df3f4e9df7ed983cd1c0d5/flytekit/clis/sdk_in_container/run.py#L179)), there are many others in the file.

This PR attempts to resolve the first two by
* Merging the logic for the first two and moving to the type engine.
* Generalizing what it means to be a container type (see the [test here](https://github.com/flyteorg/flytekit/pull/1700/files#diff-d42cbd665ae9670799e81f8d08da7080bfc1f788fcdf1da8b9361e2922379311R175) to see more clearly what that implies),
  * Remove `isinstance(..., list)` and `isinstance(..., dict)`
  * Adding three new functions to the type transformer:
    * `is_container_type` (returns bool)
    * `flyte_container_type` (returns LiteralCollection or LiteralMap)
    * `traverse` (responsible for appropriately walking the structure of the container and returning one of two generators.
      * `ListGen = typing.Generator[typing.Tuple[typing.Any, typing.Type[T], _type_models.LiteralType], None, None]`
      * `MapGen = typing.Generator[typing.Tuple[str, typing.Any, typing.Type[T], _type_models.LiteralType], None, None]`

(Some code has also been left in place to allow `flyte_container_type` to return Scalar.)

## Other issues
In addition to merging the two very related calls (`ensure_literal` and `extract_value`), some other bugs/issues uncovered:
* the handling of unions in the `extract_value` subroutine was not correct but `ensure_literal` was correct.
* we were not correctly walking bindings when the binding was a list or map.  That is, if you do
  ```
  x = t1()
  y = t2(a={"key1": x})
  ```
  the `t1` node would not have been listed as an upstream node of `n1`


wild-endeavor:
Closing this PR.  Talked about it in the contributor sync on 6/22.  this fundamentally requires multi-variate map types, which we're not going to add to the idl.  Will pull the fixes out to individual flytekit PRs

PR #1704, opened 2023-06-22
Real time deck support

Yicheng-Lu-llll:
# TL;DR

This PR adds real-time deck support.

After this PR, User can use `flytekit.Deck.persist()` to generate the html based on current collected metrics/information. So that user can see the real time deck even the task is running or has already failed. See the below example:
 
```python
import flytekit
from flytekit import Resources, task, workflow
from flytekit.core.utils import timeit

@task(
    disable_deck=False,
    limits=Resources(mem="4Gi", cpu="1"),
)
def t1():
    import time

    for i in range(2):
        # timeit measure the time used in the block and shown in time line deck. See https://github.com/flyteorg/flytekit/pull/1581. 
        # Or you can add information to your own deck. See https://docs.flyte.org/projects/cookbook/en/latest/auto/core/flyte_basics/deck.html.
        with timeit(f"iteration {i}"):
            time.sleep(50)
        flytekit.Deck.persist()


@workflow
def wf():
    t1()


if __name__ == "__main__":
    wf()

```

截屏2023-06-23 22 39 08

截屏2023-06-23 22 39 23

real time deck support includes:
- flytekit: https://github.com/flyteorg/flytekit/pull/1704
- flyteadmin: https://github.com/flyteorg/flyteadmin/pull/579
- flytepropeller: https://github.com/flyteorg/flytepropeller/pull/579
- flyteconsole: https://github.com/flyteorg/flyteconsole/pull/781

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/flytekit/deck/deck.py
+++ b/flytekit/deck/deck.py
@@ -72,6 +72,17 @@ class Deck:
     def html(self) -> str:
         return self._html
 
+    @classmethod
+    def persist(self):
+        task_name = FlyteContextManager.current_context().user_space_params.task_id.name
+        print("persist:", FlyteContextManager.current_context().user_space_params.task_id)
+        print("persist:", task_name)
+        print("persist:", task_name.split(".")[-1])
+        new_user_params = FlyteContextManager.current_context().user_space_params
+        _output_deck(task_name , new_user_params)
+
+
+
 
 class TimeLineDeck(Deck):
     """
--- a/flytekit/deck/deck.py
+++ b/flytekit/deck/deck.py
@@ -75,9 +75,6 @@ class Deck:
     @classmethod
     def persist(self):
         task_name = FlyteContextManager.current_context().user_space_params.task_id.name
-        print("persist:", FlyteContextManager.current_context().user_space_params.task_id)
-        print("persist:", task_name)
-        print("persist:", task_name.split(".")[-1])
         new_user_params = FlyteContextManager.current_context().user_space_params
         _output_deck(task_name , new_user_params)
 
--- a/flytekit/deck/deck.py
+++ b/flytekit/deck/deck.py
@@ -76,9 +76,7 @@ class Deck:
     def persist(self):
         task_name = FlyteContextManager.current_context().user_space_params.task_id.name
         new_user_params = FlyteContextManager.current_context().user_space_params
-        _output_deck(task_name , new_user_params)
-
-
+        _output_deck(task_name, new_user_params)
 
 
 class TimeLineDeck(Deck):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -159,7 +159,7 @@ def _dispatch_execute(
 
     ctx.file_access.put_data(ctx.execution_state.engine_dir, output_prefix, is_multipart=True)
     logger.info(f"Engine folder written successfully to the output prefix {output_prefix}")
-
+    print("entrypoint:", task_def.name.split(".")[-1])
     if not task_def.disable_deck:
         _output_deck(task_def.name.split(".")[-1], ctx.user_space_params)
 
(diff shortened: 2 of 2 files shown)


fg91:
Just realizing we could use this feature to solve [this](https://github.com/flyteorg/flyte/discussions/3666) problem: provide the ability to display links to e.g. wandb runs which are generated at runtime before the task ends

cosmicBboy:
This also solves the problem with [eager mode](https://github.com/flyteorg/flytekit/pull/1579), i.e. we can re-render the Flyte deck of subtasks being kicked-off by the eager workflow in real-time.

flytekit/deck/deck.py
```diff
@@ -72,6 +72,12 @@ def name(self) -> str:
     def html(self) -> str:
         return self._html
 
+    @classmethod
+    def persist(self):
```
cosmicBboy, line 76: I think `self` should be `cls`

Yicheng-Lu-llll: oh, I see. Thank you for pointing out!

fg91:
How does this effort here relate to the discussion in https://github.com/flyteorg/flyte/discussions/3838?
(cc @hamersaw)

hamersaw:
@fg91 yeah, this is one of the issues that is blocked, I would rather merge this with a re-usable flytekit -> flytepropeller communication mechanism.

ketian-indeed:
Sorry to bump up on this PR, but this seems like a pretty useful feature to us. Is there any update on it? Thx!

kumare3:
@ketian-indeed would love to understand the type of visualization you want

ketian-indeed:
@kumare3, thanks for the quick response. I'm thinking of a use case like dumping a link (e.g. Datadog/WandB/MLflow links that are valuable to check while a model training is ongoing) to the Deck when a task starts and being able to see the link on the Deck while the task is running. I believe the Deck is currently shown once the task is completed - please correct me if I'm wrong.

kumare3:
Hi @ketian-indeed we are infact adding support for W&B, Comet and neptune natively into flyte, Check this out. https://github.com/flyteorg/flytekit/pull/2449

Also Datadog you should be able to add already using, 
https://docs.flyte.org/en/latest/user_guide/productionizing/configuring_logging_links_in_the_ui.html#configure-logging

But, i also like the idea of Realtime decks - we will be adding support for this, but not like this PR. slightly different.

ketian-indeed:
Thanks, @kumare3! [Configuring logging links in the UI ](https://docs.flyte.org/en/latest/user_guide/productionizing/configuring_logging_links_in_the_ui.html) seems to be a better solution for some of our use cases. I'll test that out.
Meanwhile, I'm also looking forward to the real-time deck support, which I believe will provide even more flexibility for visualization.

kumare3:
I think that is fair. We will prioritize this work in summer. But please add ideas / use cases here that way it is very easy to ensure we address. If you are open to contributing that would be awesome too.

fg91:
>But, i also like the idea of Realtime decks - we will be adding support for this, but not like this PR. slightly different.

> I think that is fair. We will prioritize this work in summer. But please add ideas / use cases here that way it is very easy to ensure we address. If you are open to contributing that would be awesome too.

[This](https://github.com/flyteorg/flyte/issues/3696#issuecomment-1554444032), [this](https://github.com/flyteorg/flyte/discussions/3666#discussioncomment-6125773), and [this](https://github.com/flyteorg/flyte/discussions/3838) discussion are relevant in this context.

TL;DR multiple use cases are currently blocked by not having a real-time communication channel between a task pod and propeller, e.g. live updates in the eager mode deck or adding a log link dynamically at runtime:

```py
@task
def train():
    run_url = my_experiment_tracking_api.init_run()
    flytekit.current_context().add_log_link(run_url)
```

eapolinario:
Closing this PR as its functionality is being subsumed by https://github.com/flyteorg/flytekit/pull/2779.

PR #1705, opened 2023-06-22
add private registry support

RichhLi:
# TL;DR
Add support for pushing/pulling from private registries when using ImageSpec. I submitted a pr to envd devs that enabled this feature: https://github.com/tensorchord/envd/pull/1680

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
**Example:** 

```python

image_spec = ImageSpec(packages=["numpy"], base_image="private_base_image", registry="private_registry", registry_config="registry_config.json")

def t1() -> str:
    import numpy as np
    print("task 1")
    print(np.array([1,2,3]))
    return("blah1")

@workflow
def wf():
    t1()
```

## Tracking Issue
NA

## Follow-up issue
NA


flytekit/image_spec/image_spec.py
```diff
@@ -26,38 +26,40 @@ class ImageSpec:
         builder: Type of plugin to build the image. Use envd by default.
         source_root: source root of the image.
         env: environment variables of the image.
-        registry: registry of the image.
+        output_registry: registry to push the final image.
```
pingsutw, line 29: Can we not rename it? It will break existing usres

RichhLi: Sure I'll change it. I renamed it because I thought having a "registry" and "private_registries" parameter may cause confusion.

flytekit/image_spec/image_spec.py
```diff
@@ -26,14 +26,15 @@ class ImageSpec:
...
-        registry: registry of the image.
+        registry: registry to push the final image.
         packages: list of python packages to install.
         apt_packages: list of apt packages to install.
         cuda: version of cuda to install.
         cudnn: version of cudnn to install.
         base_image: base image of the image.
         platform: Specify the target platforms for the build output (for example, windows/amd64 or linux/amd64,darwin/arm64
         pip_index: Specify the custom pip index url
+        private_registries: Provide credentials for private registries. Each group is separated by a semicolon. Format: 'registry=my-registry,ca=/etc/config/ca.pem,key=/etc/config/key.pem,cert=/etc/config/cert.pem;registry=my-registry2,ca=/etc/config/ca2.pem,key=/etc/config/key2.pem,cert=/etc/config/cert2.pem'
```
ByronHsu, line 37: `registry_credentials` makes more sense to me

ByronHsu (commented):
Actually lets put this on hold until we decide the interface on the envd side first

flytekit/image_spec/image_spec.py
ByronHsu: or maybe just `credentials`

pingsutw (commented):
LGTM.
one nit: we should set registry_config to `None` when we [calculate the hash](https://github.com/flyteorg/flytekit/blob/dab1eed3b3dc29e2587c2473ede218ab97c904a4/flytekit/image_spec/image_spec.py#L145) from image spec. Therefore, flytekit won't rebuild the image, if we change the config path in the image spec.

RichhLi:
> LGTM. one nit: we should set registry_config to `None` when we [calculate the hash](https://github.com/flyteorg/flytekit/blob/dab1eed3b3dc29e2587c2473ede218ab97c904a4/flytekit/image_spec/image_spec.py#L145) from image spec. Therefore, flytekit won't rebuild the image, if we change the config path in the image spec.

Good idea. I was considering hashing the JSON object itself, but I think hashing JSON isn't reliable because fields can be in different order.

pingsutw (approved):
Thank you @RichhLi

PR #1707, opened 2023-06-22
Detect upstream nodes from container nested promises

wild-endeavor:
# TL;DR
Upstream nodes were not correctly being detected in cases where the inputs were behind a {} or a [].  For example, `n1` here

```
def my_wf(a: int, b: str) -> (int, str):
    x, y = t1(a=a)
    d = t2(a={"key1": b, "key2": y})
    return x, d
```
would not have picked up `n0` as an upstream node because we weren't correctly traversing the structure.

Found while implementing the now rejected generalized container types https://github.com/flyteorg/flytekit/pull/1700/files

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Instead of iterating through the kwargs again, save nodes found while traversing the bindings from the promises.
Also get rid of the mostly duplicate `binding_from_flyte_std`

## Tracking Issue
NA

### DIFF
--- a/tests/flytekit/unit/core/test_type_engine.py
+++ b/tests/flytekit/unit/core/test_type_engine.py
@@ -1710,3 +1710,7 @@ def test_is_annotated(t, expected):
 )
 def test_get_underlying_type(t, expected):
     assert get_underlying_type(t) == expected
+
+
+def test_dict_get():
+    assert DictTransformer.get_dict_types(None) == (None, None)
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -593,25 +593,17 @@ def create_task_output(
     return Output(*promises)  # type: ignore
 
 
-def binding_from_flyte_std(
-    ctx: _flyte_context.FlyteContext,
-    var_name: str,
-    expected_literal_type: _type_models.LiteralType,
-    t_value: Any,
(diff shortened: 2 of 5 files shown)


pingsutw (approved):

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -1710,3 +1710,7 @@ def test_is_annotated(t, expected):
 )
 def test_get_underlying_type(t, expected):
     assert get_underlying_type(t) == expected
+
+
+def test_dict_get():
+    assert DictTransformer.get_dict_types(None) == (None, None)
```
eapolinario, line 1716: why are we adding this?

eapolinario (commented):
Do you think it's worth adding a few tests for nested container types (e.g. a list that contains a dictionary, that contains a list, etc) ?

tests/flytekit/unit/core/test_type_engine.py
wild-endeavor: felt like it was missing, just wanted to be sure it could handle None.

pingsutw (approved):

PR #1631, opened 2023-05-11
Improve task decorator type hints with overload

ringohoffman:
# TL;DR
Functions decorated by `flytekit.task` are not hinted as `PythonFunctionTask` as a result of its return typing, so there are spurious type errors when attempting to register such tasks on `FlyteRemote` objects.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description

Here you can see an explanation of the problem and the fix:

```python
import flytekit
import flytekit.configuration as flyte_config
import flytekit.remote


@flytekit.task
def my_task1() -> int:
    return 0


@flytekit.task()
def my_task2() -> int:
    return 0


@flytekit.task(task_config=5)
def my_task3() -> int:
    return 0


my_task4 = flytekit.task(lambda: 0)

remote = flytekit.remote.FlyteRemote(...)  # type: ignore
serialization_settings = flyte_config.SerializationSettings(...)  # type: ignore

# before
reveal_type(my_task1)  # Type of "my_task1" is "((...) -> Any) | PythonFunctionTask"
remote.register_task(my_task1, serialization_settings)  # error: Argument of type "((...) -> Any) | PythonFunctionTask" cannot be assigned to parameter "entity" of type "PythonTask" in function "register_task"

reveal_type(my_task2)  # Type of "my_task2" is "() -> int"
remote.register_task(my_task2, serialization_settings)  # error: Argument of type "() -> int" cannot be assigned to parameter "entity" of type "PythonTask" in function "register_task"

reveal_type(my_task3)  # Type of "my_task3" is "() -> int"
remote.register_task(my_task3, serialization_settings)  # error: Argument of type "() -> int" cannot be assigned to parameter "entity" of type "PythonTask" in function "register_task"

reveal_type(my_task4)  # Type of "my_task4" is "((...) -> Any) | PythonFunctionTask"
remote.register_task(my_task4, serialization_settings)  # Argument of type "((...) -> Any) | PythonFunctionTask" cannot be assigned to parameter "entity" of type "PythonTask" in function "register_task"


# after
reveal_type(my_task1)  # Type of "my_task1" is "PythonFunctionTask"
remote.register_task(my_task1, serialization_settings)  # OK

[...]

PR #1711, opened 2023-06-25
Improve task type hint

pingsutw:
# TL;DR
Before:
image

After:
image

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -300,7 +300,7 @@ class Task(object):
         vals = [Promise(var, outputs_literals[var]) for var in output_names]
         return create_task_output(vals, self.python_interface)
 
-    def __call__(self, *args, **kwargs) -> Any:
+    def __call__(self, *args, **kwargs) -> Union[Tuple[Promise], Promise, VoidPromise, Tuple, None]:
         return flyte_entity_call_handler(self, *args, **kwargs)  # type: ignore
 
     def compile(self, ctx: FlyteContext, *args, **kwargs):
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -300,7 +300,7 @@ class Task(object):
         vals = [Promise(var, outputs_literals[var]) for var in output_names]
         return create_task_output(vals, self.python_interface)
 
-    def __call__(self, *args, **kwargs) -> Union[Tuple[Promise], Promise, VoidPromise, Tuple, None]:
+    def __call__(self, *args, **kwargs) -> Any:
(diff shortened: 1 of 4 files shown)


kumare3:
Interesting

eapolinario:
Isn't this essentially voiding any guarantee that tools like mypy or pyright could give users?

wild-endeavor (approved):

wild-endeavor (approved):

eapolinario:
mypy still complains about tasks invoked in workflows:
```
❯ cat workflows/example.py
import typing
from flytekit.core.task import task
from flytekit.core.workflow import workflow

@task
def say_hello(name: str) -> str:
    return f"hello {name}!"

@task
def greeting_length(greeting: str) -> int:
    return len(greeting)

@workflow
def wf(name: str) -> typing.Tuple[str, int]:
    greeting = say_hello(name=name)
    greeting_len = greeting_length(greeting=greeting)
    return greeting, greeting_len

❯ mypy workflows/example.py
workflows/example.py:17: error: Incompatible return value type (got "tuple[Promise | VoidPromise | tuple[Any, ...] | None, Promise | VoidPromise | tuple[Any, ...] | None]", expected "tuple[str, int]")  [return-value]
Found 1 error in 1 file (checked 1 source file)
```

eapolinario:
What is the right way to override [`__call__` in `Task`](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/base_task.py#L303-L304) so that it carries the type of the underlying task?

pingsutw:
image

eapolinario:
Nevermind. I tested this by producing a wheel and installing it in a separate venv, but probably messed up on the way. Sorry for the false alarm!

eapolinario (commented):
This is awesome!

Should we also add a `py.typed`?

wild-endeavor:
Related to https://github.com/flyteorg/flytekit/pull/1631

pingsutw:
> Should we also add a py.typed?

I'm not sure if we have fixed all the mypy issues. Will do more testing for mypy and add it.

eapolinario:
> > Should we also add a py.typed?
> 
> I'm not sure if we have fixed all the mypy issues. Will do more testing for mypy and add it.

there's at least one case where this is going to break badly: `NamedTuple`. But that shouldn't stop us from shipping this (i.e. we can document this limitation for the time being).

PR #1714, opened 2023-06-29
Use utf-8 to encode flyte deck

pingsutw:
# TL;DR
Use `utf-8` to encode flyte deck, and catch the error during generating the deck file to prevent the entire task from failing)
Fixes https://github.com/unionai/cloud/issues/3733

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/deck/deck.py
+++ b/flytekit/deck/deck.py
@@ -151,7 +151,8 @@ def _output_deck(task_name: str, new_user_params: ExecutionParameters):
             f.write(_get_deck(new_user_params, ignore_jupyter=True))
         logger.info(f"{task_name} task creates flyte deck html to file://{local_path}")
         if ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION:
-            remote_path = f"{new_user_params.output_metadata_prefix}{os.sep}{DECK_FILE_NAME}"
+            fs = ctx.file_access.get_filesystem_for_path(new_user_params.output_metadata_prefix)
+            remote_path = f"{new_user_params.output_metadata_prefix}{ctx.file_access.sep(fs)}{DECK_FILE_NAME}"
             kwargs: typing.Dict[str, str] = {
                 "ContentType": "text/html",  # For s3
                 "content_type": "text/html",  # For gcs
--- a/flytekit/deck/deck.py
+++ b/flytekit/deck/deck.py
@@ -55,7 +55,6 @@ class Deck:
 
     def __init__(self, name: str, html: Optional[str] = ""):
         self._name = name
-        # self.renderers = renderers if isinstance(renderers, list) else [renderers]
         self._html = html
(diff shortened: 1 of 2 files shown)


flytekit/deck/deck.py
```diff
@@ -147,16 +146,19 @@ def _output_deck(task_name: str, new_user_params: ExecutionParameters):
...
+        logger.info(f"{task_name} task creates flyte deck html to file://{local_path}")
+        if ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION:
+            remote_path = f"{new_user_params.output_metadata_prefix}{os.sep}{DECK_FILE_NAME}"
+            kwargs: typing.Dict[str, str] = {
+                "ContentType": "text/html",  # For s3
+                "content_type": "text/html",  # For gcs
+            }
+            ctx.file_access.put_data(local_path, remote_path, **kwargs)
+    except Exception:
+        logger.error("Failed to write flyte deck html.")
```
wild-endeavor, line 161: if we catch the exception we should print the original error

flytekit/deck/deck.py
```diff
@@ -147,16 +146,19 @@ def _output_deck(task_name: str, new_user_params: ExecutionParameters):
...
-            "ContentType": "text/html",  # For s3
-            "content_type": "text/html",  # For gcs
-        }
-        ctx.file_access.put_data(local_path, remote_path, **kwargs)
+    try:
+        with open(local_path, "w", encoding="utf-8") as f:
+            f.write(_get_deck(new_user_params, ignore_jupyter=True))
+        logger.info(f"{task_name} task creates flyte deck html to file://{local_path}")
+        if ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION:
+            remote_path = f"{new_user_params.output_metadata_prefix}{os.sep}{DECK_FILE_NAME}"
```
wild-endeavor, line 154: os.sep won't work on windows though.

pingsutw: hmm, why? https://docs.python.org/3/library/os.html#os.sep

flytekit/deck/deck.py
pingsutw: updated it

flytekit/deck/deck.py
wild-endeavor: because this is the remote path right?  what if you're on windows?  this will return `\`, which s3 won't accept right?

pingsutw: You're right. updated it, thanks!

eapolinario (approved):

PR #1720, opened 2023-07-05
Carry over hash from literal maps in cache key computation

nicholasjng:
# TL;DR
Until now, the hash of a literal map is not carried over into the representative literal map entering the cache key computation, despite the code clearly stating so.

This commit enables carrying hashes from `LiteralMap`s over by explicitly setting the hashes on the representer literal map.

Also fixes a shadowing bug in the key-value loop over the literal map, which happened since the value argument and the input name were both `literal`.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added
 - [x] Any pending items have an associated Issue (there aren't any I guess)

## Complete description

This came up during internal tests, where it broke existing tests when I changed from a single `Literal` to a `LiteralMap` in a type transformer with an explicitly set hash value.

Minimal reproducer:

```py
# test.py
from flytekit import Literal
from flytekit.core.local_cache import _recursive_hash_placement
from flytekit.models.literals import LiteralMap, Scalar, Primitive

lit = Literal(
    map=LiteralMap(
        literals={
            "hello": Literal(scalar=Scalar(primitive=Primitive(string_value="hello"))),
        }
    ),
    hash="0xffff",
)

if __name__ == "__main__":
    print(lit)
# --> prints <FlyteLiteral map { literals { key: "hello" value { scalar { primitive { string_value: "hello" } } } } } hash: "0xffff">
    print(_recursive_hash_placement(lit))
# --> prints <FlyteLiteral map { literals { key: "hello" value { scalar { primitive { string_value: "hello" } } } } }>
```

This means that the `_recursive_hash_placement` function, which facilitates part of the hash key computation, silently scrubs hash values from input literal maps, which can cause involuntary cache hits if the values inside the map stay the same.

I would be happy for a discussion - maybe the repro can go in as a unit test in a suitable location.

## Tracking Issue
N/A

## Follow-up issue
N/A

### DIFF
--- a/flytekit/core/local_cache.py
+++ b/flytekit/core/local_cache.py
@@ -13,18 +13,17 @@ CACHE_LOCATION = "~/.flyte/local-cache"
 
 
 def _recursive_hash_placement(literal: Literal) -> Literal:
-    if literal.collection is not None:
-        literals = [_recursive_hash_placement(literal) for literal in literal.collection.literals]
+    # Base case, hash gets passed through always if set
+    if literal.hash is not None:
+        return Literal(hash=literal.hash)
+    elif literal.collection is not None:
+        literals = [_recursive_hash_placement(lit) for lit in literal.collection.literals]
         return Literal(collection=LiteralCollection(literals=literals))
     elif literal.map is not None:
         literal_map = {}
-        for key, literal in literal.map.literals.items():
-            literal_map[key] = _recursive_hash_placement(literal)
+        for key, literal_value in literal.map.literals.items():
+            literal_map[key] = _recursive_hash_placement(literal_value)
         return Literal(map=LiteralMap(literal_map))
-
-    # Base case
-    if literal.hash is not None:
-        return Literal(hash=literal.hash)
     else:
         return literal
 
--- a/tests/flytekit/unit/core/test_local_cache.py
+++ b/tests/flytekit/unit/core/test_local_cache.py
@@ -15,12 +15,12 @@ from flytekit.core.base_task import kwtypes
 from flytekit.core.context_manager import FlyteContextManager
 from flytekit.core.dynamic_workflow_task import dynamic
 from flytekit.core.hash import HashMethod
-from flytekit.core.local_cache import LocalTaskCache, _calculate_cache_key
+from flytekit.core.local_cache import LocalTaskCache, _calculate_cache_key, _recursive_hash_placement
 from flytekit.core.task import TaskMetadata, task
 from flytekit.core.testing import task_mock
 from flytekit.core.type_engine import TypeEngine
 from flytekit.core.workflow import workflow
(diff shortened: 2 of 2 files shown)


nicholasjng:
Judging by the block handling the case of literal collections directly above, it will be the same situation for literal collections. 

If the solution presented here is feasible, I can also fix the collection case if you'd like.

eapolinario:
Thanks for working on this. This is indeed a problem that affects the local cache.

Let's add a few unit tests. Also, can you also fix the collections case?

nicholasjng:
a049425b75742cfe7647dc226145fad3bb201848 (hopefully) fixes the collection case and adds a unit test for both the literal map and collection. Let me know your thoughts!

kumare3:
How about remote cache?

nicholasjng:
> How about remote cache?

In what way does the remote cache behave differently than the local one?

eapolinario:
> > How about remote cache?
> 
> In what way does the remote cache behave differently than the local one?

Good catch. Initially I was under the impression that this would only affect the local cache, but that's not the case. The computation of remote cache keys happens [here](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/catalog/hashing.go#L15-L55). It's also affected by a similar bug as the one fixed in this PR.

nicholasjng:
Thanks for the link. I can take that up as well if you want, in a separate PR to the plugins repo. What do you think?

eapolinario:
That would be awesome. Thank you!

flytekit/core/local_cache.py
```diff
@@ -14,13 +14,13 @@
...
+        literals = [_recursive_hash_placement(lit) for lit in literal.collection.literals]
+        return Literal(collection=LiteralCollection(literals=literals), hash=literal.hash)
     elif literal.map is not None:
         literal_map = {}
-        for key, literal in literal.map.literals.items():
-            literal_map[key] = _recursive_hash_placement(literal)
-        return Literal(map=LiteralMap(literal_map))
+        for key, literal_value in literal.map.literals.items():
+            literal_map[key] = _recursive_hash_placement(literal_value)
+        return Literal(map=LiteralMap(literal_map), hash=literal.hash)
```
eapolinario, line 23: now that I'm looking at this again, we should simplify this and apply the same idea that we use in the case of the remote cache, i.e. if the hash in the literal we don't have to recurse and can simply return an empty literal containing the hash.

nicholasjng: yeah, that is a good idea. I changed the loop structure to incorporate this idea, let's see what the tests say.

eapolinario (approved):
This looks great!

PR #1727, opened 2023-07-08
Throw warning for nested @Task functions

oliverhu:
# TL;DR
Address https://github.com/flyteorg/flyte/issues/3734 that nested Tasks/Workflows is not warned. 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description

In order to detect entities that are nested inside `@task`:
1. During local execution, create a new `FlyteContext` with `LOCAL_TASK_EXECUTION` for `@task`,and `LOCAL_WORKFLOW_EXECUTION` if it is `@workflow`.
2. If `flyte_entity_call_handler` is invoked inside `LOCAL_TASK_EXECUTION` or `TASK_EXECUTION`, that means it is called inside `@task`, warn the users of nested `@tasks/@workflows`.

Created a helper function `is_local_execution()` to shorthand the checks.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3734

## Follow-up issue
The __call__() function in remote_callable and reference_entity duplicates and can be replaced with `flyte_entity_call_handler` as well in the future.

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -1028,9 +1028,10 @@ def flyte_entity_call_handler(
                     return None
             return cast(LocallyExecutable, entity).local_execute(child_ctx, **kwargs)
     else:
+        mode = cast(LocallyExecutable, entity).local_execution_mode()
         with FlyteContextManager.with_context(
             ctx.with_execution_state(
-                ctx.new_execution_state().with_params(mode=ExecutionState.Mode.LOCAL_WORKFLOW_EXECUTION)
+                ctx.new_execution_state().with_params(mode=mode)
             )
         ) as child_ctx:
             cast(ExecutionParameters, child_ctx.user_space_params)._decks = []
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -963,6 +963,9 @@ class LocallyExecutable(Protocol):
     def local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], Promise, VoidPromise, None]:
         ...
 
+    def local_execution_mode(self) -> ExecutionState.Mode:
+        ...
+
 
 def flyte_entity_call_handler(
     entity: SupportsNodeCreation, *args, **kwargs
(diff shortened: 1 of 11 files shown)


pingsutw (commented):
Others LGTM

pingsutw (approved):

flytekit/core/promise.py
```diff
@@ -996,22 +999,34 @@ def flyte_entity_call_handler(
...
+                    if len(output_names) == 0:
+                        return VoidPromise(entity.name)
+                    vals = [Promise(var, None) for var in output_names]
+                    return create_task_output(vals, cast(SupportsNodeCreation, entity).python_interface)
+                else:
+                    return None
+            return cast(LocallyExecutable, entity).local_execute(child_ctx, **kwargs)
     else:
         with FlyteContextManager.with_context(
             ctx.with_execution_state(
```
ByronHsu, line 1032:
```


@task
def inner_task():
    print(123)

@task
def nested_task():
    inner_task()
```

```
pyflyte run nested_task.py nested_task    
```

`pyflyte run` a nested task inside a workflow works fine, but running a nested task directly doesn't prompt the warning.

I think we may have to change here to

```
    else:
        mode = cast(LocallyExecutable, entity).local_execution_mode()
        with FlyteContextManager.with_context(
            ctx.with_execution_state(
                ctx.new_execution_state().with_params(mode=mode)
            )
        ) as child_ctx:
            cast(ExecutionParameters, child_ctx.user_space_params)._decks = []
            result = cast(LocallyExecutable, entity).local_execute(child_ctx, **kwargs)

```

Otherwise, the nested task will be labled as local_workflow_execution, but it should be local_task_execution instead. It works for me locally after the change.

ByronHsu:
Also, I am curious why can't we just use `TASK_EXECUTION` for the locally executed tasks?

ByronHsu:
Also, i would prefer to throw exceptions instead of merely logging errors because a nested task is illegal when running remotely?

flytekit/core/promise.py
oliverhu: Good catch, updating

ByronHsu:
Just checked with @pingsutw. The nested workflow can actually run fine remotely. To keep backward compatibility, let's just stick with logger.

ByronHsu (dismissed):

pingsutw (approved):

pingsutw (approved):

pingsutw:
merged, thanks @oliverhu @ByronHsu

fellhorn:
There seems to be a bug in this implementation as it throws a warning for valid dynamic workflows:

```
from flytekit import workflow, dynamic, task

@task
def my_task():
    print("foo")

@dynamic
def outer():
    my_task()


@workflow
def workflow(
):
    outer()

if __name__ == "__main__":
    workflow()
```

prints:

```
{"asctime": "2023-08-29 11:57:57,824", "name": "flytekit", "levelname": "ERROR", "message": "You are not supposed to nest @Task/@Workflow inside a @Task!"}
foo
```

oliverhu:
Let's follow up in https://github.com/flyteorg/flyte/issues/3995

PR #1731, opened 2023-07-11
fix Image Spec requirements.txt hash

RichhLi:
# TL;DR
Converting the hash object to string will include the memory address (which differs in every execution), so two identical requirements.txt file will have different hashes. Instead, use hexdigest() for unique hash value.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added


pingsutw (approved):

RichhLi:
One of the checks failed but it doesn't seem related to my code change?

pingsutw:
Yes, that's a flaky test. I'll merge it shortly

PR #1757, opened 2023-07-19
Add is none function

pingsutw:
# TL;DR
Add support `is_none` in the condition

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
import random
import typing
from flytekit import conditional, task, workflow


@task
def coin_toss(seed: int) -> typing.Optional[bool]:
    r = random.Random(seed)
    if r.random() < 0.5:
        return True
    return None


@task
def failed() -> int:
    return -1


@task
def success() -> int:
    return 0


@workflow
def wf(seed: int = 5) -> int:
    result = coin_toss(seed=seed)
    return conditional("test").if_(result.is_none()).then(success()).else_().then(failed())


if __name__ == '__main__':
    wf()
```

image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3514

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -85,7 +85,9 @@ def translate_inputs_to_literals(
     return result
 
 
-def get_primitive_val(prim: Primitive) -> Any:
+def get_primitive_val(prim: Optional[Primitive] = None) -> Any:
+    if prim is None:
+        return None
     for value in [
         prim.integer,
         prim.float_value,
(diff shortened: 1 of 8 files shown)


flytekit/core/promise.py
```diff
@@ -167,6 +181,8 @@ def eval(self) -> bool:
 
         if isinstance(self.rhs, Promise):
             rhs = self.rhs.eval()
+        elif self.rhs.scalar.none_type:
```
wild-endeavor, line 184: why is this only added for rhs?  should we also add it for lhs?

pingsutw: updated it

flytekit/models/core/condition.py
```diff
@@ -134,7 +134,7 @@ def from_flyte_idl(cls, pb2_object):
 
 
 class Operand(_common.FlyteIdlEntity):
-    def __init__(self, primitive=None, var=None):
+    def __init__(self, primitive=None, var=None, scalar=None):
         """
         Defines an operand to a comparison expression.
         :param flytekit.models.literals.Primitive primitive:
```
wild-endeavor, line 140: can you add scalar to the params list

wild-endeavor (approved):

flytekit/models/core/condition.py
pingsutw: updated

wild-endeavor (approved):

PR #1762, opened 2023-07-26
Remove dependency on responses

honnix:
# TL;DR
Remove dependency on responses

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
As reported in https://github.com/flyteorg/flyte/issues/3895, there is no need to depend on `responses` package.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3895

## Follow-up issue
Closes https://github.com/flyteorg/flyte/issues/3895

### DIFF
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -964,7 +964,6 @@ pyyaml==6.0.1
     #   papermill
     #   pre-commit
     #   ray
-    #   responses
     #   sphinx-autoapi
     #   uvicorn
     #   vaex-core
@@ -1012,7 +1011,6 @@ requests==2.31.0
     #   papermill
     #   ray
     #   requests-oauthlib
-    #   responses
     #   sphinx
     #   sphinxcontrib-youtube
     #   tensorboard
@@ -1023,8 +1021,6 @@ requests-oauthlib==1.3.1
     # via
     #   google-auth-oauthlib
     #   kubernetes
-responses==0.23.1
-    # via flytekit
 rfc3339-validator==0.1.4
     # via
     #   jsonschema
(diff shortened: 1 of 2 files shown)


honnix:
I was not able to rebuild all the requirements file for plugins because I got this type of error for each every plugin:

```
  The conflict is caused by:
      The user requested flytekitplugins-dask 0.0.0+develop (from /some/full/path/to/flytekit-dask)
      The user requested flytekitplugins-dask 0.0.0+develop (from .)
```

In the requirements.in file, I'm not sure why there is both `.` and the egg.

honnix:
The failure doesn't seem to relate to this change.

eapolinario:
> The failure doesn't seem to relate to this change.

I also believe that they are not related. I kicked off a retry just in case.


> Another thing to note is that I was not able to do `make doc-requirements.txt` on Apple silicon mac because of `pip._internal.exceptions.DistributionNotFound: No matching distribution found for tensorflow==2.8.1`. I ended up with manually editing `doc-requirements.txt`. :(

Yeah, the current state of this is not great. I'm thinking about dockering requirements generation, starting with doc-requirements in https://github.com/flyteorg/flytekit/pull/1764.

eapolinario (approved):
The failures are not related to your change.

eapolinario:
1.9 is being released shortly, so I'd say that it's not necessary. What do you think?

honnix:
@eapolinario Thanks for replying. I have already got help to do a backport and release in #1774 .

PR #1763, opened 2023-07-26
Override RayJob configs

pingsutw:
# TL;DR
WIP

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/plugins/flytekit-ray/flytekitplugins/ray/task.py
+++ b/plugins/flytekit-ray/flytekitplugins/ray/task.py
@@ -28,12 +28,19 @@ class WorkerNodeConfig:
     ray_start_params: typing.Optional[typing.Dict[str, str]] = None
 
 
+# TODO: This config should be added to flyteidl. https://github.com/flyteorg/flyteidl/blob/95e11cca2dac18b727f122bbc4456ea6ab499289/protos/flyteidl/plugins/ray.proto#L8
+@dataclass
+class RayJobConfigOverride:
+    namespace: str
+
+
 @dataclass
 class RayJobConfig:
     worker_node_config: typing.List[WorkerNodeConfig]
     head_node_config: typing.Optional[HeadNodeConfig] = None
     runtime_env: typing.Optional[dict] = None
     address: typing.Optional[str] = None
+    config_override: typing.Optional[RayJobConfigOverride] = None
 
 
 class RayFunctionTask(PythonFunctionTask):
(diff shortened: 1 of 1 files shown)


plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -34,6 +34,8 @@ class RayJobConfig:
     head_node_config: typing.Optional[HeadNodeConfig] = None
     runtime_env: typing.Optional[dict] = None
     address: typing.Optional[str] = None
+    # TODO: This config should be added to flyteidl. https://github.com/flyteorg/flyteidl/blob/95e11cca2dac18b727f122bbc4456ea6ab499289/protos/flyteidl/plugins/ray.proto#L8
+    config_override: typing.Dict[str, str] = None
```
daikeshi, line 38: I wonder if the type needs to be `Dict[str, Any]` or there're other more elegant ways to handle this. 

Configs for head/worker spec in `RayJob` are buried under a few layers, e.g. 

```
rayClusterSpec:
    headGroupSpec:
        template:
            spec:
                resources: ...
```  

You also have different specs for head and worker groups

martinbomio: I also wonder if having this generic config is beneficial vs extending the  `RayJobConfig` with specific options? For example, we could add a `namespace` option to `RayJobConfig`, a resources (or limits) to the `HeadNodeConfig` and `WorkerNodeConfig`, etc.

The reason why I think it would be better to go that direction is because the ray backend plugin will have to know this configurations options explicitly anyways, because it needs to override specific parts of the `RayJob` and `RayCluster`  CRDs

pingsutw: The problem is that `get_config` will be serialized to proto map<string, string>, so we can only use Dict[str, str] here. 
We definitely can extend RayJobConfig, but we need to update [IDL](https://github.com/flyteorg/flyteidl/blob/95e11cca2dac18b727f122bbc4456ea6ab499289/protos/flyteidl/plugins/ray.proto#L8) first here.

pingsutw: RayJobConfig will be serialized to RayJob proto we defined in the IDL, then flytepropeller will unmarshal it.

PR #1784, opened 2023-08-10
Add Azure-specific headers when uploading to blob storage

devictr:
# TL;DR
This PR fixes an issue causing `pyflyte run --remote`  to fail when using Azure. 

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added
 - [] Any pending items have an associated Issue

## Complete description
The PUT call to Azure Blob Storage returned an HTTP 400 Error: 
```
<?xml version="1.0" encoding="utf-8"?>
<Error>
    <Code>MissingRequiredHeader</Code>
    <Message>
        An HTTP header that's mandatory for this request is not specified.
        RequestId:ebd5dfbc-f01e-001d-5074-cb4f26000000
        Time:2023-08-10T10:24:58.6553767Z
    </Message>
    <HeaderName>x-ms-blob-type</HeaderName>
</Error>
```

This is a starting point that unblocks me, but there's a discussion to be had regarding where you would like to handle such cases. Some open questions:

1. Should this be done on the client? And if yes, is that the right place?
2. Or should the response that contains the signed URL also specify headers based on configuration options?

I have tested this locally by tweaking the class and trying to run a workflow

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3942

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -790,20 +790,23 @@ class FlyteRemote(object):
             filename=to_upload.name,
         )
 
+        extra_headers = self.get_extra_headers_for_protocol(upload_location.native_url)
         encoded_md5 = b64encode(md5_bytes)
         with open(str(to_upload), "+rb") as local_file:
             content = local_file.read()
             content_length = len(content)
+            headers = {"Content-Length": str(content_length), "Content-MD5": encoded_md5}
+            headers.update(extra_headers)
             rsp = requests.put(
                 upload_location.signed_url,
                 data=content,
-                headers={"Content-Length": str(content_length), "Content-MD5": encoded_md5},
+                headers=headers,
                 verify=False
                 if self._config.platform.insecure_skip_verify is True
                 else self._config.platform.ca_cert_file_path,
             )
 
-            if rsp.status_code != requests.codes["OK"]:
+            if rsp.status_code not in (requests.codes["OK"], requests.codes["created"]):
                 raise FlyteValueException(
                     rsp.status_code,
                     f"Request to send data {upload_location.signed_url} failed.",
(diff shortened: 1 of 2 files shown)


flytekit/remote/remote.py
```diff
@@ -790,20 +790,23 @@ def upload_file(
...
                 data=content,
-                headers={"Content-Length": str(content_length), "Content-MD5": encoded_md5},
+                headers=headers,
                 verify=False
                 if self._config.platform.insecure_skip_verify is True
                 else self._config.platform.ca_cert_file_path,
             )
 
-            if rsp.status_code != requests.codes["OK"]:
+            if rsp.status_code not in (requests.codes["OK"], requests.codes["created"]):
```
eapolinario, line 809: why do we have to check this value as well?

devictr: Ah sorry, forgot to specify in my PR description. 
Azure returns an HTTP 201. So this was throwing an exception even though the request was successful

devictr:
Maybe I should add a test on the entire `upload_file` function that also mocks the put call and checks that the headers are set. Might do that tomorrow

flytekit/remote/remote.py
```diff
@@ -1925,3 +1928,9 @@ def launch_backfill(
             return remote_wf
 
         return self.execute(remote_wf, inputs={}, project=project, domain=domain, execution_name=execution_name)
+
+    @staticmethod
+    def get_extra_headers_for_protocol(native_url):
+        if native_url.startswith("abfs://"):
+            return {"x-ms-blob-type": "BlockBlob"}
```
pingsutw, line 1935: In the case of blockblob type, the return code will become 200, won't it?
https://stackoverflow.com/questions/67015459/why-does-azure-blob-storage-return-201-but-not-upload

devictr: Seems like it returns a 201 https://learn.microsoft.com/en-us/rest/api/storageservices/put-blob?tabs=azure-ad#status-code

devictr: That matches what happened when I tested it

flytekit/remote/remote.py
devictr: Done

wild-endeavor (approved):

pingsutw (approved):
Thank you @devictr

PR #1787, opened 2023-08-11
Fg91/feat/proxy authentication

fg91:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/configuration/internal.py
```diff
@@ -64,6 +64,13 @@ class Credentials(object):
     This command is executed to return a token using an external process.
     """
 
+    PROXY_COMMAND = ConfigEntry(
```
fg91, line 67: TODO: do we need the legacy config entry or do we only support this in yaml config?

flytekit/clients/grpc_utils/auth_interceptor.py
```diff
@@ -61,7 +61,7 @@ def intercept_unary_unary(
         fut: grpc.Future = continuation(updated_call_details, request)
         e = fut.exception()
         if e:
-            if e.code() == grpc.StatusCode.UNAUTHENTICATED:
+            if e.code() == grpc.StatusCode.UNAUTHENTICATED or e.code() == grpc.StatusCode.UNKNOWN:
```
fg91, line 64: When unauthenticated, GCP Identity Aware Proxy unfortunately doesn't send 401.

flytekit/configuration/internal.py
wild-endeavor: i think this is the correct declaration.  Let's keep it.  But what is this?  what does this do?

fg91:
@wild-endeavor I merged #1795 into this one as requested 👍

tests/flytekit/unit/clients/auth/test_authenticator.py
```diff
@@ -77,13 +84,11 @@ def test_client_creds_authenticator(mock_requests):
...
     )
 
-    response = MagicMock()
-    response.status_code = 200
-    response.json.return_value = json.loads("""{"access_token": "abc", "expires_in": 60}""")
-    mock_requests.post.return_value = response
     authn.refresh_credentials()
     expected_scopes = static_cfg_store.get_client_config().scopes
+
     assert authn._creds
```
fg91, line 90: The order of mocking vs instantiation of the authenticator matters now.
I added the more explicit check `assert authn._creds.access_token == "abc"` because this would fail in case somebody in the future reverses the order again (instantiate the authenticator first, then mock).

tests/flytekit/unit/clients/auth/test_token_client.py
```diff
@@ -71,8 +78,9 @@ def test_poll_token_endpoint(mock_requests):
...
     response.ok = True
     response.json.return_value = {"access_token": "abc", "expires_in": 60}
-    mock_requests.post.return_value = response
+    session.post.return_value = response
     r = DeviceCodeResponse(device_code="x", user_code="y", verification_uri="v", expires_in=1, interval=0)
     t, e = poll_token_endpoint(r, "test.com", "test", http_proxy_url="http://proxy:3000", verify=True)
-    assert t
-    assert e
+
+    assert t == "abc"
```
fg91, line 85: Let's be more specific here. When adapting the magic mocks, I noticed that here one easily gets a `MagicMock` object for `t` which then also is not `None` but is not `"abc"`.

flytekit/configuration/internal.py
fg91: Added an explanation in this PR's description.

wild-endeavor (commented):
i think this looks good.  cc @EngHabu to take a look though.

before merging, @eapolinario and i will run through all the various authentication flows that we have easy access to as well, just as a sanity check.

fg91:
Thanks for taking a look @wild-endeavor 

I added a second subcommand to `flyte-iap` to generate ID tokens for service accounts:

```console
~ flyte-iap --help 
Usage: flyte-iap [OPTIONS] COMMAND [ARGS]...

  Generate ID tokens for GCP Identity Aware Proxy (IAP).

Options:
  --help  Show this message and exit.

Commands:
  generate-service-account-id-token                                         # <- new
                                  Generate a service account ID token...
  generate-user-id-token          Generate a user account ID token for...
```

The subcommand uses either 
* a service account key json file or
* contacts the GCP metadata server to obtain a token when executed in Compute Engine etc.

To generate a token for service accounts, in contrast to generating one for user accounts, no browser needs to be opened interactively. The goal of this subcommand is that CICD pipelines can talk to flyteadmin through IAP. (@corleyma said they will require this.)

With that, `flyte-iap` is complete. 

`flytectl` still doesn't understand the `proxyCommand` in the config but I will work on this next:

```console
❯ flytectl config init
Error:

1 error(s) decoding:

* '' has invalid keys: proxycommand
ERRO[0000]

1 error(s) decoding:

* '' has invalid keys: proxycommand  src="main.go:13"
```

fg91:
> i think this looks good. cc @EngHabu to take a look though.

Taking you, @EngHabu, as suggested by @wild-endeavor 🙏

EngHabu (dismissed):
I'm less sure about reusing this command for flytectl. Requiring python dependency isn't something I would want to require in CI/CD and other automated systems

fg91:
Hi Haytham, I really appreciate the time you’ve invested in looking at this PR, I'm sure you have a busy schedule :pray:

> I'm less sure about reusing this command for flytectl. Requiring python dependency isn't something I would want to require in CI/CD and other automated systems

In our case, we will run `pyflyte register/package` in CICD as well but I agree with your concern that this shouldn't be required. I documented 2 additional ways to obtain an id token for service accounts using 1) GCP's `gcloud` sdk and 2) a bash script from the identity aware proxy documentation. This way, users can choose to not use python.
https://github.com/flyteorg/flytekit/pull/1787/commits/2437b84545d9cb87f9d4aaea3acd2f86401e02db

EngHabu (approved):

wild-endeavor:
tested this with pkce, cc, and device flow btw, all works

wild-endeavor (approved):

fg91:
> tested this with pkce, cc, and device flow btw, all works

Thanks for testing this @wild-endeavor! 🙏 
I also implemented the same logic for flytectl/the admin client in idl. Everything works, just need to polish a bit and adapt/add tests. Will ping you there as well then.

jtyberg:
I'm trying to get all Flyte clients to pass the additional `proxy-authorization` header with each request. 

I can use the flytekit-identity-aware-proxy plugin (with `proxyCommand` in the Flyte config) successfully from a FlyteRemote script, but  I'm seeing this error when trying to use `flytectl`. 

```
bin/flytectl get projects                                                 

Error: 

1 error(s) decoding:

* '' has invalid keys: proxycommand
ERRO[0000] 

1 error(s) decoding:

* '' has invalid keys: proxycommand  src="main.go:13"
```

It seems the issue is that `flytekit` and `flytectl` are stuck at different versions of `flyteidl`? (this [PR](https://github.com/flyteorg/flytectl/pull/464) never made it in).

fg91:
> I'm trying to get all Flyte clients to pass the additional `proxy-authorization` header with each request. 
> 
> 
> 
> I can use the flytekit-identity-aware-proxy plugin (with `proxyCommand` in the Flyte config) successfully from a FlyteRemote script, but  I'm seeing this error when trying to use `flytectl`. 
> 
> 
> 
> ```
> 
> bin/flytectl get projects                                                 
> 
> 
> 
> Error: 
> 
> 
> 
> 1 error(s) decoding:
> 
> 
> 
> * '' has invalid keys: proxycommand
> 
> ERRO[0000] 
> 
> 
> 
> 1 error(s) decoding:
> 
> 
> 
> * '' has invalid keys: proxycommand  src="main.go:13"
> 
> ```
> 
> 
> 
> It seems the issue is that `flytekit` and `flytectl` are stuck at different versions of `flyteidl`? (this [PR](https://github.com/flyteorg/flytectl/pull/464) never made it in).
> 
> 


The most recent versions of flytectl (released after the move to the monorepo) use the flyteidl admin client which supports proy auth, try upgrading to e.g. >= 0.8.18 please.

jtyberg:
> The most recent versions of flytectl (released after the move to the monorepo) use the flyteidl admin client which supports proxy auth, try upgrading to e.g. >= 0.8.18 please.

I apologize, I should have posted my `flytectl` version, which is the `v0.8.18` that you recommend.

Installed using `curl -sL https://ctl.flyte.org/install | bash`.

```
bin/flytectl version

 A new release of flytectl is available: 0.8.18 → v0.8.22 
To upgrade, run: flytectl upgrade 
https://github.com/flyteorg/flytectl/releases/tag/v0.8.22 

{
  "App": "flytectl",
  "Build": "0a0cbce",
  "Version": "0.8.18",
  "BuildTime": "2024-06-03 10:50:36.469871 -0400 EDT m=+0.045492501"
}%                                                                                                                      
```

`v0.8.18` still shows the error when I add `proxyCommand` to the `flytectl` config.
```
bin/flytectl version

Error: 

1 error(s) decoding:

* '' has invalid keys: proxycommand
```

fg91:
> I apologize, I should have posted my `flytectl` version, which is the `v0.8.18` that you recommend.

Sorry, you are right about `v0.8.18` (was on my phone). Can you please try [`v0.8.23`](https://github.com/flyteorg/flyte/releases/tag/flytectl%2Fv0.8.23)? The install script pointed to the archived flytectl repo until [this](https://github.com/flyteorg/flyte/pull/5405) fix a few days ago.

jtyberg:
> > I apologize, I should have posted my `flytectl` version, which is the `v0.8.18` that you recommend.
> 
> Sorry, you are right about `v0.8.18` (was on my phone). Can you please try [`v0.8.23`](https://github.com/flyteorg/flyte/releases/tag/flytectl%2Fv0.8.23)? The install script pointed to the archived flytectl repo until [this](https://github.com/flyteorg/flyte/pull/5405) fix a few days ago.

Ah, [v0.8.23](https://github.com/flyteorg/flyte/releases/tag/flytectl%2Fv0.8.23) did it. I can now use the `proxyCommand` field in my Flyte config.yaml.
```
bin/flytectl get projects                    

 ----- ------ ----------------- 
| ID  | NAME | DESCRIPTION     |
 ----- ------ ----------------- 
| myproject | myproject  | myproject description |
 ----- ------ ----------------- 
1 rows
```

Thanks for this! (And apologies for opening an old thread).

PR #1790, opened 2023-08-13
Async file sensor

pingsutw:
# TL;DR
Add a base sensor. People can easily implement a sensor by inhering the base sensor.
Add a sensor engine, which will run all the sensors.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
File sensor
```python
import asyncio
from typing import Optional, TypeVar

from flytekit import FlyteContextManager
from flytekit.sensor.base_sensor import BaseSensor

T = TypeVar("T")


class FileSensor(BaseSensor):
    def __init__(self, name: str, config: Optional[T] = None, **kwargs):
        super().__init__(name=name, sensor_config=config, **kwargs)

    async def poke(self, path: str) -> bool:
        fs = FlyteContextManager.current_context().file_access.get_filesystem_for_path(path, asynchronous=True)
        return await asyncio.to_thread(fs.exists, path)
```

## Tracking Issue

## Follow-up issue
_NA_


kumare3 (commented):
LGTM

flytekit/sensor/file_sensor.py
```diff
@@ -0,0 +1,18 @@
+from typing import Optional, TypeVar
+
+from flytekit import FlyteContextManager
+from flytekit.sensor.base_sensor import BaseSensor
+
+T = TypeVar("T")
+
+
+class FileSensor(BaseSensor):
+    def __init__(self, name: str, config: Optional[T] = None, **kwargs):
```
wild-endeavor, line 10: when would someone use the `config` field?

wild-endeavor (approved):
don't really understand the need for both cloudpickle and jsonpickle.  can you explain?

flytekit/sensor/file_sensor.py
pingsutw: timeout, token, etc.

pingsutw: > don't really understand the need for both cloudpickle and jsonpickle. can you explain?

I use `jsonpickle` because pickle is not JSON serializable.  In translator, we use `json.dumps` to searlzie config, so I have to use `jsonpickle`.

https://github.com/flyteorg/flytekit/blob/36fe5813d83477f91056f61bd234e9dd232ec4a2/flytekit/models/task.py#L463

PR #1793, opened 2023-08-13
Fix [BUG] Module 'PIL' has no attribute Image

Future-Outlier:
# TL;DR

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
You can search "return PIL.Image.open(local_path)" in the code, and you will find that it is not enough to just use "import PIL" in the code.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3813

## Follow-up issue
_NA_
OR

### DIFF
--- a/plugins/flytekit-deck-standard/flytekitplugins/deck/renderer.py
+++ b/plugins/flytekit-deck-standard/flytekitplugins/deck/renderer.py
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING, List, Optional, Union
 
+import PIL.Image
+
 from flytekit import lazy_module
 from flytekit.types.file import FlyteFile
 
--- a/doc-requirements.txt
+++ b/doc-requirements.txt
@@ -249,7 +249,7 @@ flask==2.3.2
     # via mlflow
 flatbuffers==23.5.26
     # via tensorflow
-flyteidl==1.5.13
+flyteidl==1.5.14
     # via flytekit
 fonttools==4.41.1
     # via matplotlib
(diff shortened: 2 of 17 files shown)


plugins/flytekit-deck-standard/flytekitplugins/deck/renderer.py
```diff
@@ -1,5 +1,7 @@
 from typing import TYPE_CHECKING, List, Optional, Union
 
+import PIL.Image
```
kumare3, line 3: I do not think we need this? We just use PIL.Image.Image so line 17 should be enough right?

kumare3 (changes requested):
I do not think we should add the import, there is a lazy load already

Future-Outlier:
> I do not think we should add the import, there is a lazy load already

Just found that you are right, thanks a lot for the review!

Future-Outlier:
Just contact the author of this pull request, I think maybe the bug doesn't exist.
https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-deck-standard/tests/test_renderer.py#L60-L72
Our test has covered the case he've mentioned.

plugins/flytekit-deck-standard/flytekitplugins/deck/renderer.py
Future-Outlier: Yes, you are probably right.

PR #1794, opened 2023-08-14
Remove ref in output

wild-endeavor:
# TL;DR
This removes the `ref` property in the default NamedTuple object that gets used when a user want to name outputs.  This is done so that some can use an output named "ref" if they wanted to.  Previously it was getting co-opted by this property.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The property iterates through the list of outputs, and returns the first `NodeOutput` it sees, and the node gets extracted from that.  Instead of doing this, just store the Node object in the Case object, upon construction.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3923

### DIFF
--- a/flytekit/core/condition.py
+++ b/flytekit/core/condition.py
@@ -289,6 +289,18 @@ class Case(object):
         self, p: Union[Promise, Tuple[Promise]]
     ) -> Optional[Union[Condition, Promise, Tuple[Promise], VoidPromise]]:
         self._output_promise = p
+        self._output_node = None
+        if isinstance(p, Promise):
+            if not p.is_ready:
+                self._output_node = p.ref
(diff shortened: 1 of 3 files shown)


eapolinario (dismissed):

pingsutw (commented):
nit: linting error

pingsutw (approved):

honnix:
Thanks for fixing this. I'm wondering whether it makes sense to have reserved prefix/suffix for this type of internal properties? A simple `_` prefix might do it.

PR #1795, opened 2023-08-15
Feat: Add plugin for generating GCP IAP ID tokens via external command

fg91:
# TL;DR

See https://github.com/flyteorg/flyte/issues/3965

This PR adds a plugin called `flytekit-identity-aware-proxy` that provides a CLI called `flyte-iap` which performs an OAuth 2.0 flow with `https://accounts.google.com`. The CLI will be used by `flytekit` and `flytectl` as an external command to create ID tokens to authenticate with [GCP Identity Aware Proxy (IAP)](https://cloud.google.com/iap) protecting Flyte.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

Still need to fix tests.

## Complete description

Flytekit already provides logic to perform OAuth 2.0 flows (see [here](https://github.com/flyteorg/flytekit/blob/bf4f24b17bc5e317d7b492bf73e42074a01c1f32/flytekit/clients/auth/auth_client.py#L171)):

```python
class AuthorizationClient(metaclass=_SingletonPerEndpoint):
    """
    Authorization client that stores the credentials in keyring and uses oauth2 standard flow to retrieve the
    credentials. NOTE: This will open an web browser to retreive the credentials.
    """
```

This authorization client is currently only used by the [`PKCEAuthenticator`](https://github.com/flyteorg/flytekit/blob/bf4f24b17bc5e317d7b492bf73e42074a01c1f32/flytekit/clients/auth/authenticator.py#L88) which is used by flyte's clis to authenticate with flyteadmin.

However, in contrast to what is claimed in the doc string, the `AuthorizationClient` does not perform a standard oauth2 flow but rather a PKCE flow. [Here](https://github.com/flyteorg/flytekit/blob/bf4f24b17bc5e317d7b492bf73e42074a01c1f32/flytekit/clients/auth/auth_client.py#L217), for instance, a code challenge is generated which is part of the pkce flow but not the standard oauth2 flow.

This PR slightly refactors the `AuthorizationClient` by moving pkce-related things like code challenge generation into the `PKCEAuthenticator` (without actually changing the pkce auth flow).

The `AuthorizationClient` is then re-used to perform a standard OAuth 2.0 flow with `https://accounts.google.com` in the new IAP plugin.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3965

## Follow-up issue
_NA_


plugins/flytekit-identity-aware-proxy/flytekitplugins/identity_aware_proxy/cli.py
```diff
@@ -0,0 +1,161 @@
...
+                    "redirect_uri": self.redirect_uri,
+                },
+                # See https://cloud.google.com/iap/docs/authentication-howto#refresh_token
+                refresh_access_token_params={
+                    "client_secret": self.client_secret,
+                    "audience": self.audience,
+                },
+            )
+
+    def refresh_credentials(self):
```
fg91, line 72: Same as in PKCE Authenticator in flytekit.

plugins/flytekit-identity-aware-proxy/flytekitplugins/identity_aware_proxy/cli.py
```diff
@@ -0,0 +1,167 @@
...
+    except NotFound as e:
+        raise click.BadParameter(e.message)
+    payload = response.payload.data.decode("UTF-8")
+    return payload
+
+
+@click.group()
+def cli():
+    """Generate ID tokens for GCP Identity Aware Proxy (IAP)."""
+    pass
```
fg91, line 105: Having a command group, even with only 1 subcommand currently, could allow to add a command to generate id tokens for service accounts in the future by adding:

```python
@cli.command()
def generate_sa_id_token():
    """Generate a service account ID token for proxy-authentication/authorization with GCP Identity Aware Proxy."""
    raise NotImplementedError
```

wild-endeavor (commented):
first off, thank you for doing this!

second, let's get this merged down into 1787 so we have just one thing to look at for this repo.  I'm not sure about how to release though.  Everything looks reasonable, but I think we should get this out as a beta release first and try it.  what do you think @eapolinario?  There's enough auth providers and auth types out there that i think it merits all the flows being tested prior to a release.

fg91:
>  I think we should get this out as a beta release first and try it. what do you think @eapolinario? There's enough auth providers and auth types out there that i think it merits all the flows being tested prior to a release.

I agree 👍

PR #1798, opened 2023-08-17
Enable SecretsManager.get to load and return bytes

ysysys3074:
# TL;DR
If the secrets are a base-64 encoded string, `SecretsManager.get` will raise exception. So add `encoded_mode` to the function here to fix and help to load and return bytes.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -349,7 +349,9 @@ class SecretsManager(object):
         """
         return self._GroupSecrets(item, self)
 
-    def get(self, group: str, key: Optional[str] = None, group_version: Optional[str] = None, encode_mode: str = "r") -> str:
+    def get(
+        self, group: str, key: Optional[str] = None, group_version: Optional[str] = None, encode_mode: str = "r"
+    ) -> str:
         """
         Retrieves a secret using the resolution order -> Env followed by file. If not found raises a ValueError
         """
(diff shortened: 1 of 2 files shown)


flytekit/core/context_manager.py
```diff
@@ -349,7 +349,9 @@ def __getattr__(self, item: str) -> _GroupSecrets:
         """
         return self._GroupSecrets(item, self)
 
-    def get(self, group: str, key: Optional[str] = None, group_version: Optional[str] = None) -> str:
+    def get(
+        self, group: str, key: Optional[str] = None, group_version: Optional[str] = None, encode_mode: str = "r"
```
eapolinario, line 353: what if we make this an enum (containing only 2 options: "r" or "rb")?

ysysys3074: good idea, I will add in next commit

ByronHsu: i think using string is ok. It is consistent with what we pass to `with open(fpath, <here>)` . Enum seems a bit overkilled

eapolinario: Fair enough. At the very least put a description in the function docstring, @ysysys3074 ?

wild-endeavor: you can do a literal type if you want.

ysysys3074: @eapolinario added in new commit

eapolinario (approved):

eapolinario (approved):

PR #1802, opened 2023-08-20
Async agent delete function for while loop case

Future-Outlier:
# TL;DR
As title.
I forgot to handle the case the function in while loop.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
If you use the current async delete function, it will face problem when you try to cancel the job in the waiting loop.
image
This is because the code here: https://github.com/flyteorg/flytekit/blob/master/flytekit/extend/backend/base_agent.py#L190-L199
It didn't give any space for user to use the default event loop, which is used by asyncio.run, so I need to create a new event loop.
## Tracking Issue
https://github.com/flyteorg/flytekit/pull/1800

## Follow-up issue
_NA_
OR


flytekit/extend/backend/base_agent.py
```diff
@@ -211,5 +211,14 @@ def signal_handler(
         signum: int,
         frame: FrameType,
     ) -> typing.Any:
-        agent.delete(context, resource_meta)
+        if agent.asynchronous:
+            try:
+                new_loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(new_loop)
```
pingsutw, line 217: How did you test it? you shouldn't be able to run the new event loop while another loop is running

pingsutw (commented):
you could add add `signal_handler` to the event loop. https://stackoverflow.com/questions/23313720/asyncio-how-can-coroutines-be-used-in-signal-handlers

Future-Outlier:
> you could add add `signal_handler` to the event loop. https://stackoverflow.com/questions/23313720/asyncio-how-can-coroutines-be-used-in-signal-handlers

Really useful advice!
I will improve it now.

flytekit/extend/backend/base_agent.py
Future-Outlier: I tested it in the wrong way, I will improve it no, thanks a lot!

Future-Outlier:
Due to the unique mechanism for asyncio in Python, I will use an event loop from head to toe in the program.
Here is a demo and explain video about my signal handler and delete function in base_agent.py, please take a look, thanks a lot!
https://youtu.be/N8lNQ9e6gIM

flytekit/extend/backend/base_agent.py
```diff
@@ -158,6 +159,8 @@ class AsyncAgentExecutorMixin:
     Task should inherit from this class if the task can be run in the agent.
     """
 
+    aysnc_delete = False
```
pingsutw, line 162: could we rename it to `is_canceled`?

Future-Outlier: Yes, thanks a lot!

flytekit/extend/backend/base_agent.py
```diff
@@ -192,12 +206,17 @@ def execute(self, **kwargs) -> typing.Any:
...
+                    res = loop.run_until_complete(agent.async_get(dummy_context, metadata))
                 else:
                     res = agent.get(dummy_context, metadata)
                 state = res.resource.state
                 logger.info(f"Task state: {state}")
 
+        if agent.asynchronous:
+            loop.close()
+            if self.is_canceled:
+                sys.exit(1)
```
pingsutw, line 218: we should put these codes in the while loop since `loop.run_until_complete(agent.async_get(dummy_context, metadata))` may always return `running` when using the sensor.

Future-Outlier: Done, please review, thanks a lot!

Future-Outlier:
@pingsutw 

I've recently updated the async agent solution in this pull request: [GitHub PR #1802](https://github.com/flyteorg/flytekit/pull/1802/files#diff-2944080a47998a2b23f992855b3a532bda59633aa1eaf11b812363ee191543c7R212-R217).

I've noticed a potential issue: if the user interrupts the program (Ctrl + C) after the `while` loop has successfully completed but before `loop.close()` is executed, the program doesn't terminate as one might expect. Is this behavior intended, or should we address it?

Future-Outlier:
Thanks for @pingsutw's revision, the refactor is really amazing, I've learned a lot!

pingsutw (dismissed):

Future-Outlier:
@pingsutw  Just added a test, please review, thanks a lot!

pingsutw (approved):

PR #1806, opened 2023-08-22
Batch upload flyte directory

pingsutw:
# TL;DR
s3fs will try to read all the files in the directory into memory by default. we can specify the batch size for `s3fsFileSystem.put`, and it will dramatically reduce memory consumption.

Before: 
image
After (batch size = 100):
image

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
from typing_extensions import Annotated

import os
from memory_profiler import profile
from flytekit import BatchSize
directory = "/tmp/test"


@task
@profile
def generate_files(num_files: int, filesize_mb: int) -> Annotated[FlyteDirectory, BatchSize(100)]:
    os.mkdir(directory)
    for i in range(num_files):
        file_path = os.path.join(directory, f"file_{i}.txt")
        with open(file_path, "w") as f:
            f.write(str(os.urandom(filesize_mb * 1024 * 1024)))
    return FlyteDirectory(directory, remote_directory="s3://flyte-benchmark/test/")


@task
def get_files(d: Annotated[FlyteDirectory, BatchSize(100)]):
    d._downloader()


@workflow
def ton_of_files_wf(num_files: int = 1000, filesize_mb: int = 8):
    d = generate_files(num_files=num_files, filesize_mb=filesize_mb)
    get_files(d=d)


if __name__ == '__main__':
    print(ton_of_files_wf())
```

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -59,8 +59,12 @@ class BatchSize:
     """
     This is used to annotate a FlyteDirectory when we want to download/upload the contents of the directory in batches. For example,
 
+    @task
     def t1(Annotated[FlyteDirectory, BatchSize(10)]) -> Annotated[FlyteDirectory, BatchSize(100)]:
         ...
+
+    Flytekit will still download all files within the directory. It reads 10 files into memory, writes them to
+     files, and then clears the memory. Subsequently, it proceeds to read another set of 10 files. Similarly, for upload.
     """
 
     def __init__(self, val: int):
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -64,9 +64,9 @@ class BatchSize:
         ...
         return FlyteDirectory(...)
 
-    In the above example flytekit will download all files from the input `directory` in chunks of 10, i.e. first it 
-    downloads 10 files, loads them to memory, then writes those 10 to local disk, then it loads the next 10, so on 
-    and so forth. Similarly, for outputs, in this case flytekit is going to upload the resulting directory in chunks of 
+    In the above example flytekit will download all files from the input `directory` in chunks of 10, i.e. first it
+    downloads 10 files, loads them to memory, then writes those 10 to local disk, then it loads the next 10, so on
+    and so forth. Similarly, for outputs, in this case flytekit is going to upload the resulting directory in chunks of
     100.
     """
 
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -55,6 +55,27 @@ T = typing.TypeVar("T")
 DEFINITIONS = "definitions"
 
 
+class BatchSize:
+    """
+    Flyte-specific object used to wrap the hash function for a specific type
+    """
+
+    def __init__(self, val: int):
+        self._val = val
+
+    @property
+    def val(self) -> int:
+        return self._val
+
+
+def get_batch_size(t: Type) -> Optional[int]:
+    if is_annotated(t):
+        for annotation in get_args(t)[1:]:
+            if isinstance(annotation, BatchSize):
+                return annotation.val
+    return None
+
+
 class TypeTransformerFailedError(TypeError, AssertionError, ValueError):
     ...
 
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -57,7 +57,10 @@ DEFINITIONS = "definitions"
 
 class BatchSize:
     """
-    Flyte-specific object used to wrap the hash function for a specific type
+    This is used to annotate a FlyteDirectory when we want to download/upload the contents of the directory in batches. For example,
+
+    def t1(Annotated[FlyteDirectory, BatchSize(10)]) -> Annotated[FlyteDirectory, BatchSize(100)]:
+        ...
     """
 
     def __init__(self, val: int):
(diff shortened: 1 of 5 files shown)


flytekit/core/type_engine.py
```diff
@@ -55,6 +55,27 @@
 DEFINITIONS = "definitions"
 
 
+class BatchSize:
+    """
+    Flyte-specific object used to wrap the hash function for a specific type
```
eapolinario, line 60: nit: fix the docstring.

flytekit/types/directory/types.py
```diff
@@ -358,7 +363,7 @@ def to_literal(
         if should_upload:
             if remote_directory is None:
                 remote_directory = ctx.file_access.get_random_remote_directory()
-            ctx.file_access.put_data(source_path, remote_directory, is_multipart=True)
+            ctx.file_access.put_data(source_path, remote_directory, is_multipart=True, batch_size=batch_size)
```
eapolinario, line 366: do we have to worry about which version `batch_size` was introduced?

flytekit/core/type_engine.py
```diff
@@ -55,6 +55,30 @@
...
+    def val(self) -> int:
+        return self._val
+
+
+def get_batch_size(t: Type) -> Optional[int]:
+    if is_annotated(t):
+        for annotation in get_args(t)[1:]:
+            if isinstance(annotation, BatchSize):
+                return annotation.val
+    return None
```
pingsutw, line 79: should we set the default batch size to 100 or something else? default behavior is to read all the files in the directory.

pingsutw: cc @wild-endeavor

eapolinario: according to the [docs](https://github.com/fsspec/filesystem_spec/blob/master/fsspec/asyn.py#L218-L221) `fsspec` already tries to find a reasonable batch size if we pass `batch_size=None`.

flytekit/types/directory/types.py
eapolinario: sync'd offline and this parameter has been present for the past 2 years. Our [lower-bound constraint](https://github.com/flyteorg/flytekit/blob/master/setup.py#L46) in `setup.py` is more than enough.

flytekit/core/type_engine.py
pingsutw: nice, even better. didn't see it, thanks!

flytekit/core/type_engine.py
```diff
@@ -55,6 +55,34 @@
...
+class BatchSize:
+    """
+    This is used to annotate a FlyteDirectory when we want to download/upload the contents of the directory in batches. For example,
+
+    @task
+    def t1(Annotated[FlyteDirectory, BatchSize(10)]) -> Annotated[FlyteDirectory, BatchSize(100)]:
+        ...
+
+    Flytekit will still download all files within the directory. It reads 10 files into memory, writes them to
+     files, and then clears the memory. Subsequently, it proceeds to read another set of 10 files. Similarly, for upload.
```
eapolinario, line 67:
```suggestion
    @task
    def t1(directory: Annotated[FlyteDirectory, BatchSize(10)]) -> Annotated[FlyteDirectory, BatchSize(100)]:
        ...
        return FlyteDirectory(...)

    In the above example flytekit will download all files from the input `directory` in chunks of 10, i.e. first it 
    downloads 10 files, loads them to memory, then writes those 10 to local disk, then it loads the next 10, so on 
    and so forth. Similarly, for outputs, in this case flytekit is going to upload the resulting directory in chunks of 
    100.
```

eapolinario (dismissed):

flytekit/types/directory/types.py
```diff
@@ -322,6 +323,10 @@ def to_literal(
 
         remote_directory = None
         should_upload = True
+        batch_size = get_batch_size(python_type)
+
+        if is_annotated(python_type):
```
wild-endeavor, line 328: what is this?

eapolinario: this is just we can call `get_format` and still get the actual extension of the FlyteDirectory if it's set. We could move this to `get_format`, right? Also, we should probably add a couple of unit tests to verify this.

flytekit/types/directory/types.py
```diff
@@ -379,8 +384,12 @@ def to_python_value(
         # For the remote case, return an FlyteDirectory object that can download
         local_folder = ctx.file_access.get_random_local_directory()
 
+        batch_size = get_batch_size(expected_python_type)
+        if is_annotated(expected_python_type):
```
wild-endeavor, line 388: and this

flytekit/types/directory/types.py
pingsutw: removed it

flytekit/types/directory/types.py
pingsutw: removed

flytekit/types/directory/types.py
wild-endeavor: but like... shouldn't this be handled at the type engine layer?

pingsutw: I remove the code that checks if it's annotated. I just found that it still works if FlyteDirecrory is annotated. 
```python
flyte_dir = Annotated[FlyteDirectory, BatchSize(100)]
print(flyte_dir.extension()0 # it still works
```

eapolinario (approved):

wild-endeavor (approved):

kumare3:
How will users find about this. Cc @zeryx fyi

PR #1812, opened 2023-08-30
Dynamic workflow should not throw nested task warning

oliverhu:
# TL;DR
Fix https://github.com/flyteorg/flyte/issues/3995. Created a new execution state for dynamic execution, we don't throw warnings if we execute tasks inside a @dynamic execution. Ideally we should create a separate type for @dynamic. Validated in local Flyte cluster.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue
[https://github.com/flyteorg/flyte/issues/<number>](https://github.com/flyteorg/flyte/issues/3995)

## Test
With the fix
``` 
❯   pyflyte run --image localhost:30000/flytekit:2 --remote example.py workflow
```
image

Without the fix
```
❯ pyflyte run  --remote example.py workflow
```
image

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -488,6 +488,9 @@ class ExecutionState(object):
         # or propeller.
         LOCAL_TASK_EXECUTION = 3
 
+        # This is the mode that is used to indicate a dynamic task
+        DYNAMIC_TASK_EXECUTION = 4
+
     mode: Optional[ExecutionState.Mode]
     working_dir: Union[os.PathLike, str]
     engine_dir: Optional[Union[os.PathLike, str]]
--- a/flytekit/core/python_function_task.py
+++ b/flytekit/core/python_function_task.py
@@ -202,7 +202,12 @@ class PythonFunctionTask(PythonAutoContainerTask[T]):  # type: ignore
         else:
             cs = ctx.compilation_state.with_params(prefix="d")
 
-        with FlyteContextManager.with_context(ctx.with_compilation_state(cs)):
+        updated_ctx = ctx.with_compilation_state(cs)
+        if self.execution_mode == self.ExecutionBehavior.DYNAMIC:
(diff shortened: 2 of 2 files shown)


fellhorn:
Thanks @oliverhu for immediately fixing this!

I tested your changes with our dynamic workflows and can confirm it works for both local execution as well as execution in the cluster :tada:

ByronHsu (commented):
nit: i think we should name it as [dynamic workflow](https://docs.flyte.org/projects/cookbook/en/v0.3.66/auto/core/control_flow/dynamics.html) instead of dynamic task. Otherwise, lgtm

ByronHsu:
After a closer look, I realize the dynamic is actually a task under the hood, so it makes sense

ByronHsu (approved):

oliverhu:
Glad this works and thanks for the reviews! Please don't merge yet, want to verify a few more scenarios.

oliverhu:
I think it should be ok to merge. @pingsutw @ByronHsu 

I wanted to check if this will break other places which check `state == ExecutionState.Mode.TASK_EXECUTION` while the task is `DYNAMIC_TASK_EXECUTION`. But it seems like DYNAMIC_TASK is strictly a task that assembles a workflow in its body, so that situation won't exist.

pingsutw:
Thanks @oliverhu. LGTM

> I wanted to check if this will break other places which check state == ExecutionState.Mode.TASK_EXECUTION while the task is DYNAMIC_TASK_EXECUTION. But it seems like DYNAMIC_TASK is strictly a task that assembles a workflow in its body, so that situation won't exist.

yes, correct

fg91:
Thanks for fixing this so quickly @oliverhu 🙏

PR #1822, opened 2023-09-07
Chatgpt Agent

Future-Outlier:
# TL;DR
- Create Base class for all external API tasks, a base task for all plugins with `DoTask` function

- Create an `TaskExecutor` class, a base agent for all plugins with `DoTask` function

- Create a ChatGPT plugin by OpenAI API

- Support local execution for do task function

- Do refactor for base agent

- Add 2 variable in `base_agent.py`. `ASYNC_PLUGIN` and `SYNC_PLUGIN`. They are use in `runtime flavor` from `TaskMetadata`

- Use the `runtime_flavor` in `TaskMetadata` (It hasn't been used before in flytekit)

- Change the usecase of  `runtime_flavor` by adding comments

Change the usecase to the plugin routing mechanism 

- Add `runtime_flavor` paramters in all agent task and webapi task
Here's the list.
 
plugins/flytekit-airflow
plugins/flytekit-aws-athena
plugins/flytekit-aws-batch
plugins/flytekit-bigquery
plugins/flytekit-mmcloud
plugins/flytekit-snowflake
plugins/flytekit-spark
flytekit/sensor

I get the list from
1. https://github.com/flyteorg/flyte/tree/master/flyteplugins/go/tasks/plugins/webapi
2. use `find . | grep agent.py` under the `flytekit` folder
3. sensor is a special case.

Note: ChatGPT doensn't need it because the `external api task` has done it for it.

- Add test for the above changes

- Finish above without introducing breaking changes

- Currently, ChatGPT timeout works when retrying the second time in an API call, which means that if the query takes you 40seconds, you have to wait 40seconds.
This is not ideal, we will improve it when a new server integrates into Flyte Backend System. Currently, we can set the timeout settings in the config-map.

## Additional Information
Note: the timeout of the ChatGPT API will check the logic when retry, 
so I highly recommend users change the timeout of the `DoTask` function to 40 seconds, 
which will probably not exceed the limit, and we will introduce a `api_task_server` to fix this situation.

Note: If we use aiohttp to prevent the timeout by setting the timeout time, it is doable, 
but in my current experiment, `openai.acreate` is faster than using aiohttp (I haven't figure it out).
Here is the proof:
https://hackmd.io/Fmx3DHGtQBSqYydcrdF_Cg

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### Configuration Example
```yaml
tasks:
  task-plugins:
    enabled-plugins:
      - agent-service
    default-for-task-types:
      api_task: agent-service
plugins:
  agent-service:
    supportedTaskTypes:
      - api_task
    defaultAgent:
      endpoint: "dns:///localhost:8000"
      insecure: true
      timeouts:
        DoTask: 100s
```
### Run in Remote Environment
You can test the agent with Dockerfile.
```Dockerfile
FROM python:3.9-slim-buster
USER root
WORKDIR /root
ENV PYTHONPATH /root
RUN apt-get install git -y
RUN pip install -U git+https://github.com/Future-Outlier/flytekit.git@0ecebdfc3a018d888748e4d9adf715c72c761180#subdirectory=plugins/flytekit-openai-chatgpt

RUN pip install -U git+https://github.com/Future-Outlier/flyte.git@b879137d24fe6bfb60d6d1182cd8d9b3f32335cc#subdirectory=flyteidl

RUN pip install -U git+https://github.com/Future-Outlier/flytekit.git@0ecebdfc3a018d888748e4d9adf715c72c761180
```
```python
from flytekitplugins.chatgpt import ChatGPTTask


chatgpt_job = ChatGPTTask(
    name="chatgpt",
    config={
        "openai_organization": "org-NayNG68kGnVXMJ8Ak4PMgQv7",
        "chatgpt_conf": {
            "model": "gpt-3.5-turbo",
            # 'messages': [{'role': 'user', 'content': 'Say this is a test!'}],
            "temperature": 0.7,
        },
    },
)

@task()
def t1(s: str) -> str:
    s = "Repsonse: " + s
    return s

@workflow
def wf() -> str:
    message = chatgpt_job(message="hi")
    return t1(s=message)

if __name__ == "__main__":
    print(wf())
 ```
 
1.  start your grpc agent server
 ```
 pyflyte serve
 ```
2. execute it in remote environment
 ```
 pyflyte run --remote --image futureoutlier/flytekit:chatgpt-v2 chatgpt_example.py wf
 ```

## Screen Shot
image
Note: You can find that the ChatGPT task is an `api_task`. 

## How it works at high level
image

## Tracking Issue

### CODE REVIEW

**Review by fg91 — CHANGES_REQUESTED** [review]
What do you think of moving this into a plugin?
I'm a bit worried that a chat gpt integration into flytekit itself is a scope creep. Other task types for specific providers like aws or gcp bigquery are not integrated into the core library to keep it lean but are their own package.

**Comment by Future-Outlier** [comment]
> What do you think of moving this into a plugin? I'm a bit worried that a chat gpt integration into flytekit itself is a scope creep. Other task types for specific providers like aws or gcp bigquery are not integrated into the core library to keep it lean but are their own package.

Thanks for the good question.
I think the reason is because of the file sensor feature here. 
https://github.com/flyteorg/flytekit/pull/1790
I have discussed this with @pingsutw before, and we think that the dispatcher can be used for lots of cases such as [LangChain](https://www.langchain.com/), [OpenAPI](https://www.google.com/url?sa=t&rct=j&q=&esrc=s&source=web&cd=&ved=2ahUKEwinvfGP7tGBAxUEH3AKHamGCPQQFnoECB0QAQ&url=https%3A%2F%2Fwww.openapis.org%2F&usg=AOvVaw0Jhd_W-LB-7SQU3Ph6H7Kg&opi=89978449) and other do-task cases.

**Comment by pingsutw** [comment]
We can add `BaseDispatcher` into flytekit, but ChatGPT task should definitely added to the plugin.

**Comment by fg91** [comment]
> We can add `BaseDispatcher` into flytekit, but ChatGPT task should definitely added to the plugin.

This proposal sounds good to me 👍

**Comment by pingsutw** [comment]
btw, @fg91 If you come up with a better name than dispatcher, let us know.  We are not good at naming things 😂

**Comment by fg91** [comment]
> btw, @fg91 If you come up with a better name than dispatcher, let us know. We are not good at naming things 😂

At the example of this plugin: Do I understand it correctly that instead of launching a task pod in a workflow to send the request to the OpenAi API, a long living agent pod makes the respective request asynchronously?

**Comment by Future-Outlier** [comment]
> > btw, @fg91 If you come up with a better name than dispatcher, let us know. We are not good at naming things 😂
> 
> At the example of this plugin: Do I understand it correctly that instead of launching a task pod in a workflow to send the request to the OpenAi API, a long living agent pod makes the respective request asynchronously?

Yes, you are right!

**Comment by fg91** [comment]
> > > btw, @fg91 If you come up with a better name than dispatcher, let us know. We are not good at naming things 😂
> > 
> > 
> > At the example of this plugin: Do I understand it correctly that instead of launching a task pod in a workflow to send the request to the OpenAi API, a long living agent pod makes the respective request asynchronously?
> 
> Yes, you are right!

How about one of the following more descriptive (but also longer) names: `ExternalApiRequestDispatcher`, `RequestDispatcher`, `ExternalApiTask`, `AsyncRequestTask`, `ApiRequestTask`?

**Comment by Future-Outlier** [comment]
> > > > btw, @fg91 If you come up with a better name than dispatcher, let us know. We are not good at naming things 😂
> > > 
> > > 
> > > At the example of this plugin: Do I understand it correctly that instead of launching a task pod in a workflow to send the request to the OpenAi API, a long living agent pod makes the respective request asynchronously?
> > 
> > 
> > Yes, you are right!
> 
> How about one of the following more descriptive (but also longer) names: `ExternalApiRequestDispatcher`, `RequestDispatcher`, `ExternalApiTask`, `AsyncRequestTask`, `ApiRequestTask`, `AgentApiRequestTask`?
> 
> I personally tend towards a name that contains "task" as that is a familiar concept in Flyte whereas "dispatcher" is new.

I think `ExternalApiTask` or `AgentExternalApiTask` is great!
@pingsutw how do you think?

File: plugins/flytekit-chatgpt/setup.py [hunk 1]
```diff
@@ -0,0 +1,36 @@
...
+    install_requires=plugin_requires,
+    license="apache2",
+    python_requires=">=3.8",
+    classifiers=[
+        "Intended Audience :: Science/Research",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
pingsutw, line 28: nit: we can support python 3.11, right

Future-Outlier: Yes, thanks a lot!

flytekit/core/base_task.py
```diff
@@ -128,7 +129,9 @@ def to_taskmetadata_model(self) -> _task_model.TaskMetadata:
         return _task_model.TaskMetadata(
             discoverable=self.cache,
             runtime=_task_model.RuntimeMetadata(
-                _task_model.RuntimeMetadata.RuntimeType.FLYTE_SDK, __version__, "python"
+                _task_model.RuntimeMetadata.RuntimeType.FLYTE_SDK,
+                __version__,
+                "sync_plugin" if self.use_sync_plugin else "python",
```
honnix, line 134: This seems to be a quite fundamental change. It is a bit strange there is `sync_plugin` but no `async_plugin`.

Future-Outlier: I think the intention of this change is for using `sync_plugin` in flyteplugins, `async_plugin` will be used by default.
Maybe we can change the variable to `use_api_task` or other variables?
How do you think?
cc @pingsutw

Future-Outlier: Solved it, now we can specify it by lots of str value!

flytekit/core/base_task.py
```diff
@@ -105,6 +105,7 @@ class TaskMetadata(object):
     retries: int = 0
     timeout: Optional[Union[datetime.timedelta, int]] = None
     pod_template_name: Optional[str] = None
+    runtime_flavor: str = "python"
```
pingsutw, line 108: could we not add this here for now? let's only update the agent first.

flytekit/extend/backend/base_agent.py
```diff
@@ -135,7 +158,8 @@ def convert_to_flyte_state(state: str) -> State:
     Convert the state from the agent to the state in flyte.
     """
     state = state.lower()
-    if state in ["failed", "timedout", "canceled"]:
+    # timedout is weird but correct, you can refer here: https://github.com/databricks/databricks-sdk-py/pull/407
```
pingsutw, line 161:
```suggestion
    # timedout is the state of Databricks job. https://docs.databricks.com/en/workflows/jobs/jobs-2.0-api.html#runresultstate
```

tests/flytekit/unit/core/test_task_metadata.py
```diff
@@ -0,0 +1,60 @@
...
+    tm = TaskMetadata(timeout=3600)
+    assert isinstance(tm.timeout, datetime.timedelta)
+
+
+def test_retry_strategy():
+    tm = TaskMetadata(retries=5)
+    assert tm.retry_strategy.retries == 5
+
+
+def test_to_taskmetadata_model():
```
pingsutw, line 34:
```suggestion
def test_to_task_metadata_model():
```

flytekit/tools/translator.py
```diff
@@ -163,6 +162,8 @@ def get_serializable_task(
     settings: SerializationSettings,
     entity: FlyteLocalEntity,
 ) -> TaskSpec:
+    from flytekit import PythonFunctionTask
```
pingsutw, line 165: could we revert it?

flytekit/tools/translator.py
```diff
@@ -731,6 +732,8 @@ def get_serializable(
         raise Exception(f"Non serializable type found {type(entity)} Entity {entity}")
 
     if isinstance(entity, TaskSpec) or isinstance(entity, WorkflowSpec):
+        from flytekit import SourceCode
```
pingsutw, line 735: ditto

flytekit/extend/backend/task_executor.py
```diff
@@ -0,0 +1,45 @@
...
+from flytekit.core.external_api_task import TASK_CONFIG_PKL, TASK_MODULE, TASK_NAME, TASK_TYPE
+from flytekit.core.type_engine import TypeEngine
+from flytekit.extend.backend.base_agent import AgentBase, AgentRegistry
+from flytekit.models.literals import LiteralMap
+from flytekit.models.task import TaskTemplate
+
+T = typing.TypeVar("T")
+
+
+class TaskExecutor(AgentBase):
```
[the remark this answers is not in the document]
Future-Outlier: No problem!

flytekit/core/external_api_task.py
```diff
@@ -0,0 +1,73 @@
...
+        self._config = config
+
+    @abstractmethod
+    async def do(self, **kwargs) -> DoTaskResponse:
+        """
+        Initiate an HTTP request to an external service such as OpenAI or Vertex AI and retrieve the response.
+        """
+        raise NotImplementedError
+
+    def get_custom(self, settings: SerializationSettings = None) -> Dict[str, Any]:
```
kumare3, line 64: why do we have to use pickle?
cc @pingsutw do you know? is this to maintain simply python coversion? This is potentially dangerous as it may break across python versions.

flytekit/extend/backend/task_executor.py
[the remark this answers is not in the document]
kumare3: @pingsutw I do not like this name, this is confusing. Isnt this the SyncAgentBase?

flytekit/core/external_api_task.py
Future-Outlier: pickle is faster, we've talked about this before and Kevin told me that it might be ok.
Will there be multiple python versions usecases?

flytekit/extend/backend/task_executor.py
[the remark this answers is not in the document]
Future-Outlier: Yes, you are right, we should create a new class called "SyncAgentBase".
We can use `Union` to handle both async agent and sync agent in `base_agent.py`.

flytekit/core/external_api_task.py
kumare3: we cannot control backend and flytekit version

flytekit/extend/backend/task_executor.py
```diff
@@ -0,0 +1,54 @@
...
+
+    This class is meant to be subclassed when implementing plugins that require
+    an external API to perform the task execution. It provides a routing mechanism
+    to direct the task to the appropriate handler based on the task's specifications.
+    """
+
+    def __init__(self):
+        super().__init__(task_type=TASK_TYPE, asynchronous=True)
+
+    async def async_create(
```
kumare3, line 30: i would change this to SyncAgentBase as follows

```python
class SyncAgentBase(AgentBase):

      @final
      async def async_create( self,
        context: grpc.ServicerContext,  io_ctx: IOContext, task_template: TaskTemplate, 
    ) -> CreateTaskResponse:
          do(context, output_prefix, task_template, inputs)   

    def async async_do(context, io_ctx, task_template):
        python_interface_inputs = {
            name: TypeEngine.guess_python_type(lt.type) for name, lt in task_template.interface.inputs.items()
        }

        native_inputs = {}
        if inputs:
            native_inputs = TypeEngine.literal_map_to_kwargs(ctx, inputs, python_interface_inputs)

        meta = task_template.custom

        task_module = importlib.import_module(name=meta[TASK_MODULE])
        task_def = getattr(task_module, meta[TASK_NAME])
        config = jsonpickle.decode(meta[TASK_CONFIG_PKL]) if meta.get(TASK_CONFIG_PKL) else None
        return task_def(TASK_TYPE, config=config).execute(**native_inputs)
            
    async def execute(**kwargs):
         raise NotImplementedError()

```

Checkout the final decorator - https://docs.python.org/3.8/library/typing.html#typing.final

flytekit/core/external_api_task.py
Future-Outlier: you are right, I fix it!
Thank you very much for your advice!

kumare3:
no changes here yet?

Future-Outlier:
> no changes here yet?

I will update it later.
I originally thought that only a certain part needed to be changed.
I will update the `syncAgentBase` class, xd

Future-Outlier:
image

Future-Outlier:
I am considering the following changes.
1. rename `api_task` to `sync_agent_task`.
2. rename `task_executor` to `sync_agent_executor`.
3. rename `external_api_task` to `sync_agent_task`

Future-Outlier:
This PR will be derprecated.
We can refer to this PR for generic sync agent executor in the future.

PR #1825, opened 2023-09-11
[flytekit] Support attribute access on promises

ByronHsu:
See https://github.com/flyteorg/flyteidl/pull/439

## Tracking Issue

https://github.com/flyteorg/flyte/issues/3864

## Testing

In the following workflow:

- `basic_workflow` contains trivial examples to access output attributes
- `failed_workflow` contains examples that causes exception (e.g. out of range)
- `advanced_workflow` contains examples with more complex attribute access

```python
from typing import Dict, List, NamedTuple

from flytekit import WorkflowFailurePolicy, task, workflow


@dataclass_json
@dataclass
class foo:
    a: str


@task
def t1() -> (List[str], Dict[str, str], foo):
    return ["a", "b"], {"a": "b"}, foo(a="b")


@task
def t2(a: str) -> str:
    print("a", a)
    return a


@task
def t3() -> (Dict[str, List[str]], List[Dict[str, str]], Dict[str, foo]):
    return {"a": ["b"]}, [{"a": "b"}], {"a": foo(a="b")}


@task
def t4(a: List[str]):
    print("a", a)


@task
def t5(a: Dict[str, str]):
    print("a", a)


@workflow
def basic_workflow():
    l, d, f = t1()
    t2(a=l[0])
    t2(a=d["a"])
    t2(a=f.a)


@workflow(
    # The workflow doesn't fail when one of the nodes fails but other nodes are still executable
    failure_policy=WorkflowFailurePolicy.FAIL_AFTER_EXECUTABLE_NODES_COMPLETE
)
def failed_workflow():
    # This workflow is supposed to fail due to exceptions
    l, d, f = t1()
    t2(a=l[100])
    t2(a=d["b"])
    t2(a=f.b)


@workflow
def advanced_workflow():
    dl, ld, dd = t3()
    t2(a=dl["a"][0])
    t2(a=ld[0]["a"])
    t2(a=dd["a"].a)

    t4(a=dl["a"])
    t5(a=ld[0])
```

### Local execution

```
$ pyflyte run test_workflow.py basic_workflow   

a a
a b
a b

$ pyflyte run test_workflow.py failed_workflow

Failed with Exception Code: USER:PromiseAttributeResolveError
Underlying Exception: Failed to resolve attribute path [100] in promise Resolved(o0=<FlyteLiteral collection { literals { scalar { primitive { string_value: "a" } } } literals { scalar { primitive { string_value: "b" } } } }>), index 100 out of range 2
Encountered error while executing workflow 'test_workflow.failed_workflow':
  Error encountered while executing 'failed_workflow':
  Failed to resolve attribute path [100] in promise Resolved(o0=<FlyteLiteral collection { literals { scalar { primitive { string_value: "a" } } } literals { scalar { primitive { string_value: "b" } } } }>), index 100 out of range 2

$ pyflyte run test_workflow.py advanced_workflow
a b
a b
a b
a ['b']
a {'a': 'b'}
```

### Remote Execution

basic

failed

advanced


eapolinario (commented):
Looking good. Will take another pass tomorrow morning.

Thanks for adding those great unit tests!

flytekit/core/promise.py
```diff
@@ -710,13 +840,13 @@ def __repr__(self):
 
 
 class NodeOutput(type_models.OutputReference):
-    def __init__(self, node: Node, var: str):
+    def __init__(self, node: Node, var: str, attr_path: List[PromiseAttribute] = []):
```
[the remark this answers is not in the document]
ByronHsu: Wow! This is probably the most interesting thing I learned about python recently. Thanks for the catch!

flytekit/core/promise.py
```diff
@@ -303,6 +365,8 @@ def __init__(self, var: str, val: Union[NodeOutput, _literals_models.Literal]):
         self._var = var
         self._promise_ready = True
         self._val = val
+        self._ref = None
```
wild-endeavor, line 368: Is there a reason you added this @ByronHsu?  This is causing the linting errors (I think).

wild-endeavor: well this and the Optional below.

flytekit/core/promise.py
```diff
@@ -343,7 +407,7 @@ def val(self) -> _literals_models.Literal:
         return self._val
 
     @property
-    def ref(self) -> NodeOutput:
+    def ref(self) -> Optional[NodeOutput]:
```
wild-endeavor, line 410: I think it's okay to leave this as before... exceptions don't need to be in the type hint.  so before, the logic was basically, if you have a ref, then it's a node output.  If you don't, and you try to call .ref, then you get an AttributeError.

flytekit/core/promise.py
ByronHsu: I workaround this by disabling the type check for `ref`. Changing to `Optional[NodeOutput]` causes so many errors on other files, so I think disabling it might be easier.

wild-endeavor (approved):

PR #1833, opened 2023-09-13
Eager local entrypoint and support for offloaded types

cosmicBboy:
# TL;DR

This PR extends the `@eager` decorator to support running the workflow locally but executing tasks/subworkflows remotely. Also adds support and tests for preserving offloaded types.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description

`@eager(local_entrypoint=True)` will treat the eager workflow as a locally executed task that invokes the configured `remote` object for tasks and subworkflows called inside of it. It also adds support and tests for offloaded types: structured datasets, files, and directories, such that they are not immediately downloaded and can be downloaded/iterated over in the eager workflow function body.

This PR also fixes some formatting issues with the eager workflow flyte deck.

## Tracking Issue
NA

## Follow-up issue
_NA_


flytekit/experimental/eager_function.py
```diff
@@ -296,11 +312,12 @@ def get_io(dict_like):
 @asynccontextmanager
 async def eager_context(
     fn,
-    remote: FlyteRemote,
+    remote: Optional[FlyteRemote],
```
pingsutw, line 315: nit
```suggestion
    remote: Optional[FlyteRemote] = None,
```

flytekit/experimental/eager_function.py
```diff
@@ -127,7 +128,7 @@ async def __call__(self, **kwargs):
                 "If you need to use a subworkflow, use a static @workflow or nested @eager workflow."
             )
 
-        if self.ctx.execution_state.is_local_execution():
+        if not self.local_entrypoint and self.ctx.execution_state.is_local_execution():
```
pingsutw, line 131: Maybe we just check if the remote is `None`?

||local execution | remote execution | 
|----| ----| ---|
|remote| Use flytekit remote to run the task | Use flytekit remote to run the task |
|no remote| Run the task function locally | Run the task in the driver (eager) pod |

cosmicBboy: I think we'd want to run the AsyncEntity locally even if `remote is None` right? e.g.

```
# wf.py
@eager(remote=remote)
def eager_wf(): ...
```

```
pyflyte run wf.py eager_wf
```
should run locally, even if the user has configured a remote object.

It's only when they say `local_entrypoint=True` that the behavior changes such that it runs remotely on the configured remote.

flytekit/experimental/eager_function.py
cosmicBboy: the default `None` value isn't necessary no? The type annotation only means that the two valid inputs are `FlyteRemote` type or `None` type.

pingsutw: if we use optional, we must set the default value to None. https://docs.python.org/3/library/typing.html#typing.Optional

flytekit/experimental/eager_function.py
pingsutw:
```python
@eager(remote=remote, local_entrypoint=False)
def wf()
   ...
```
qq: In the local execution, what happens in this case?, will we also use flytekit remote to execute the task?

flytekit/experimental/eager_function.py
cosmicBboy: If `Optional[T]` is equivalent to `X | None` or `Union[X, None]`, would we expect to always specify a default argument?

```
def fn(arg: X | None = None): ...

def fn(arg: Union[X, None] = None): ...
```

This doesn't seem correct. I think this should be totally valid, no?

```
def fn(arg: X | None): ...

def fn(arg: Union[X, None]): ...
```

All it means is value of type `X` or `None` are valid inputs.

flytekit/experimental/eager_function.py
cosmicBboy: ### Case 1:

Just specifying a `remote` shouldn't change any of the local behavior: it'll just run locally as async functions and ignore the `remote object.

```
@eager(remote=remote, local_entrypoint=False)
# pyflyte run: runs the eager workflow locally, runs the tasks/subworkflows locally
# pyflyte run --remote: runs the eager workflow remotely, runs the tasks/subworkflows remotely
```

### Case 2:

`local_entrypoint=True` just forces the eager workflow's tasks/subworkflows to run remotely (while running the eager workflow itself locally) as a debugging/development tool.
```
@eager(remote=remote, local_entrypoint=True)  # runs eager workflow
# pyflyte run: runs the eager workflow locally, but forces tasks and subworkflows to run remotely
# pyflyte run --remote: runs the 
```

flytekit/experimental/eager_function.py
cosmicBboy: image

I think all this is saying is that if `None` is a default argument, an explicit `Optional[X]` is encouraged but not necessary.

flytekit/experimental/eager_function.py
cosmicBboy: also happy to change the name `local_entrypoint` to something clearer. The use case is: "I want to run the eager workflow locally, but run the tasks/subworkflows remotely to that I can debug it, e.g. get access to intermediary results"

pingsutw: Thanks for clarifying. I mean, we can change the default behaviour, like
```python
@eager(remote=remote)
# pyflyte run: runs the eager workflow locally, but forces tasks and subworkflows to run remotely
# pyflyte run --remote: runs the eager workflow remotely, runs the tasks/subworkflows remotely

@eager()
# If remote isn't set, 
# pyflyte run: runs the eager workflow locally, runs the tasks/subworkflows locally
```

pingsutw: yes, if we want to add a new arg, it will be better to change `local_entrypoint` to something clearer.

cosmicBboy: any suggestions?

cosmicBboy: > ```
> @eager(remote=remote)
> # pyflyte run: runs the eager workflow locally, but forces tasks and subworkflows to run remotely. If fail to read > remote config, Fallback to local execution,
> # pyflyte run --remote: runs the eager workflow remotely, runs the tasks/subworkflows remotely
> ```

The problem with this is that I have to keep changing my code if I want to run completely locally vs. completely remotely. Think of `remote=remote` as just configuration, similar to the configuration that you give to `@task`. I'd like to run my eager workflow locally and remotely by just changing the `pyflyte run` call (adding `--remote` flag for remote execution).

The `local_entrypoint=True` flag is exclusively for the hybrid case of running the eager workflow locally but the tasks/subworkflows remotely.

If it makes the use case clearer we can just introduce a wrapper decorator to isolate this use case:

```
# this will always just run locally and you can't run this on a remote Flyte cluster
@local_entrypoint(remote=remote)
async def wf():
   ...
```

(btw I just took the naming from [modal](https://modal.com/docs/guide/ex/hello_world#running-it), open to suggestions for other names)

pingsutw: Got it. Rather than a decorator, I prefer `@eager(remote=remote, local_entrypoint=True)`.
I'm not good at naming. Maybe we merge it first and then rename it once someone finds a better name.

pingsutw (approved):

flytekit/experimental/eager_function.py
cosmicBboy: Cool, will make sure to document this well on flytesnacks

PR #1842, opened 2023-09-19
Better Azure blob storage support 

Tom-Newton:
# TL;DR
Support Azure blob storage for storing metadata and raw data including structured datasets, without interrupting other standard Azure authentication. Also ensure `storage_options` are set consistently for all uses of fsspec. 

## Type
Its debatable. The original github issue was a bug for fixing Azure authentication but it could be argued that Azure blob storage support is a new feature. 
 - [x] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested - Not really sure what is expected but I have used these change to deploy real workflows. 
 - [x] Unit tests added - I added a bit of new coverage, e.g. initialising the `AzureBlobFilesystem`.
 - [x] Code documentation added - There are code comments, but I think ideally there would be a page in the documentation for Azure. 
 - [x] Any pending items have an associated Issue - There aren't any

## Complete description
Added a new component to `DataConfig` for Azure similar to the ones for S3 and GCS. This allows configuring the flytekit's  storage account without disrupting other Azure auth that may be used by flyte task implementations. This is done by setting the environment variables `FLYTE_AZURE_STORAGE_ACCOUNT_NAME`, `FLYTE_AZURE_STORAGE_ACCOUNT_KEY`, `FLYTE_AZURE_TENANT_ID`, `FLYTE_AZURE_CLIENT_ID`, `FLYTE_AZURE_CLIENT_SECRET`. You can set these using a pod template so that authentication to your flyte storage account works across your whole flyte deployment, inrespective of what authentication each task uses. 

Use `DataConfig` to construct fsspec storage options in `data_persistence.py`. We reuse the same function for fetchign `storage_options` elsewhere inlcuding for encode and decode of structured datasets. This ensures local, S3, GCS and Azure work wherever fsspec is used. 

You should specify `configmap.core.propeller.rawoutput-prefix` to be in the format `abfs://<container-name>/<path-within-container>`. 

## Tracking Issue
Related discussion was on https://github.com/flyteorg/flyte/issues/3962 but technically that issues was about workload identity which was mostly fixed by https://github.com/flyteorg/flytekit/pull/1813

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/configuration/internal.py
+++ b/flytekit/configuration/internal.py
@@ -57,6 +57,14 @@ class GCP(object):
     GSUTIL_PARALLELISM = ConfigEntry(LegacyConfigEntry(SECTION, "gsutil_parallelism", bool))
 
 
+class AZURE(object):
+    SECTION = "azure"
+    ACCOUNT_NAME = ConfigEntry(LegacyConfigEntry(SECTION, "account_name"))
+    ACCOUNT_KEY = ConfigEntry(LegacyConfigEntry(SECTION, "account_key"))
+    CLIENT_ID = ConfigEntry(LegacyConfigEntry(SECTION, "account_key"))
+    CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "access_key_id"))
+
+
 class Credentials(object):
     SECTION = "credentials"
     COMMAND = ConfigEntry(LegacyConfigEntry(SECTION, "command", list), YamlConfigEntry("admin.command", list))
--- a/flytekit/configuration/internal.py
+++ b/flytekit/configuration/internal.py
@@ -61,8 +61,9 @@ class AZURE(object):
     SECTION = "azure"
     ACCOUNT_NAME = ConfigEntry(LegacyConfigEntry(SECTION, "account_name"))
     ACCOUNT_KEY = ConfigEntry(LegacyConfigEntry(SECTION, "account_key"))
-    CLIENT_ID = ConfigEntry(LegacyConfigEntry(SECTION, "account_key"))
-    CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "access_key_id"))
+    TENANT_ID = ConfigEntry(LegacyConfigEntry(SECTION, "tenant_id"))
+    CLIENT_ID = ConfigEntry(LegacyConfigEntry(SECTION, "client_id"))
+    CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "client_secret"))
 
 
 class Credentials(object):
--- a/flytekit/configuration/internal.py
+++ b/flytekit/configuration/internal.py
@@ -59,8 +59,8 @@ class GCP(object):
 
 class AZURE(object):
     SECTION = "azure"
-    ACCOUNT_NAME = ConfigEntry(LegacyConfigEntry(SECTION, "account_name"))
-    ACCOUNT_KEY = ConfigEntry(LegacyConfigEntry(SECTION, "account_key"))
+    STORAGE_ACCOUNT_NAME = ConfigEntry(LegacyConfigEntry(SECTION, "storage_account_name"))
+    STORAGE_ACCOUNT_KEY = ConfigEntry(LegacyConfigEntry(SECTION, "storage_account_key"))
     TENANT_ID = ConfigEntry(LegacyConfigEntry(SECTION, "tenant_id"))
     CLIENT_ID = ConfigEntry(LegacyConfigEntry(SECTION, "client_id"))
     CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "client_secret"))
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -66,21 +66,15 @@ def azure_setup_args(azure_cfg: configuration.AzureBlobStorageConfig, anonymous:
 
     if azure_cfg.account_name:
         kwargs["account_name"] = azure_cfg.account_name
-
     if azure_cfg.account_key:
         kwargs["account_key"] = azure_cfg.account_key
-
     if azure_cfg.client_id:
         kwargs["client_id"] = azure_cfg.client_id
-
     if azure_cfg.client_secret:
         kwargs["client_secret"] = azure_cfg.client_secret
-
     if azure_cfg.tenant_id:
         kwargs["tenant_id"] = azure_cfg.tenant_id
-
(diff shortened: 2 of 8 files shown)


kumare3:
cc @fiedlerNr9

flytekit/core/data_persistence.py
```diff
@@ -106,28 +141,15 @@ def data_config(self) -> DataConfig:
...
-            return fsspec.filesystem(protocol, **s3kwargs)  # type: ignore
-        elif protocol == "gs":
-            if anonymous:
-                kwargs["token"] = _ANON
-            return fsspec.filesystem(protocol, **kwargs)  # type: ignore
-        elif protocol == "abfs":
-            kwargs["anon"] = False
-            return fsspec.filesystem(protocol, **kwargs)  # type: ignore
-
-        # Preserve old behavior of returning None for file systems that don't have an explicit anonymous option.
```
Tom-Newton, line 126: From the type hint on `get_filesystem_for_path` it looks like we assume `get_filesystem` never returns `None` so I thought probably best to delete this. If anyone thinks its important I'm happy to add it back.

flytekit/configuration/internal.py
```diff
@@ -57,6 +57,15 @@ class GCP(object):
     GSUTIL_PARALLELISM = ConfigEntry(LegacyConfigEntry(SECTION, "gsutil_parallelism", bool))
 
 
+class AZURE(object):
+    SECTION = "azure"
+    STORAGE_ACCOUNT_NAME = ConfigEntry(LegacyConfigEntry(SECTION, "storage_account_name"))
```
Tom-Newton, line 62: I would have added yaml configs too but I wasn't really sure about how to namespace them given the options that already exist for S3. I feel like there should be separate sections in the yaml config for azure, google and aws but the aws stuff is currently just `storage.connection.endpoint`.

flytekit/core/data_persistence.py
devictr: @wild-endeavor I think you wanted to delete this already? So this seems like a good idea

Tom-Newton:
Hmm... it looks like a couple of my new tests are failing on the windows builds. 
```
E           Expected: to_parquet('abfs://container/parquet_df/00000', coerce_timestamps=<ANY>, allow_truncated_timestamps=<ANY>, storage_options={'account_name': 'accountname_from_storage_options'})
E           Actual: to_parquet('abfs://container/parquet_df\\00000', coerce_timestamps='us', allow_truncated_timestamps=False, storage_options={'account_name': 'accountname_from_storage_options'})
```
Asserting on the path was not really the primary purpose of this test so I will probably just make the assertion just on the `storage_options`. I suspect this would work because blob storage filenames can be arbitrary strings but normally blob storage uses forward slashes `/`.

Tom-Newton:
Hmm... it looks like a couple of my new tests are failing on the windows builds. 
```
E           Expected: to_parquet('abfs://container/parquet_df/00000', coerce_timestamps=<ANY>, allow_truncated_timestamps=<ANY>, storage_options={'account_name': 'accountname_from_storage_options'})
E           Actual: to_parquet('abfs://container/parquet_df\\00000', coerce_timestamps='us', allow_truncated_timestamps=False, storage_options={'account_name': 'accountname_from_storage_options'})
```
Asserting on the path was not really the primary purpose of this test so I will probably just make the assertion just on the `storage_options`. I suspect this would work because blob storage filenames can be arbitrary strings but normally blob storage uses forward slashes `/`.

flytekit/configuration/internal.py
kumare3: we are trying to stay away from configs, as they are not easily portable

Tom-Newton: To clarify, do you mean stay away from the yaml config in favour of the environment variable config, or more generally avoid configs?

Tom-Newton:
There are 3 test suites that timeout out after 6 hours. It looks like they all failed in the `Cache pip` stage so I don't think these are related to my changes.

flytekit/configuration/internal.py
wild-endeavor: Hey @Tom-Newton - ketan's just saying that we don't need to make them yaml configs because the yaml format was designed for cross compatibility with `flytectl`'s yaml format.  Unf. though, flytectl will crap out if it sees something in the config file that it doesn't recognize (on the list of things to eventually add).

And we shouldn't add it to flytectl, it doesn't belong there.

One more point to note is that while these are needed to make the azure connection work, these are not great to tote around in a config file.  you should have these set by propeller in the default-env-vars config for the k8s plugin.  If you store these in a text file, they become not portable because in addition to the task definition, you somehow have to carry around this file.

flytekit/core/data_persistence.py
```diff
@@ -61,6 +61,41 @@ def s3_setup_args(s3_cfg: configuration.S3Config, anonymous: bool = False):
...
+        kwargs["client_secret"] = azure_cfg.client_secret
+    if azure_cfg.tenant_id:
+        kwargs["tenant_id"] = azure_cfg.tenant_id
+    kwargs[_ANON] = anonymous
+    return kwargs
+
+
+def get_fsspec_storage_options(
+    protocol: str, data_config: typing.Optional[DataConfig] = None, anonymous: bool = False, **kwargs
+) -> typing.Dict:
```
wild-endeavor, line 83: is this always [str, str]?

wild-endeavor (commented):
just some nits

flytekit/core/data_persistence.py
Tom-Newton: Its `Dict[str, Any]`. I made it more precise.

flytekit/core/data_persistence.py
```diff
@@ -61,6 +61,41 @@ def s3_setup_args(s3_cfg: configuration.S3Config, anonymous: bool = False):
     return kwargs
 
 
+def azure_setup_args(azure_cfg: configuration.AzureBlobStorageConfig, anonymous: bool = False):
```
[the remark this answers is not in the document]
Tom-Newton: Added `Dict[str, Any]`. I also made the same addition to the s3 version of this function above.

flytekit/configuration/internal.py
Tom-Newton: Thanks for clarifying :+1:  Personally I thought env vars made more sense for this but the naming of `LegacyConfigEntry` made me think the project preference would have been yaml configs.

wild-endeavor (approved):

PR #1843, opened 2023-09-20
[Extended Resources] GPU Accelerators 

jeevb:
# TL;DR
Adds support for specifying extended resources (e.g. accelerator type - gpu device, partition size etc.) to allocate to a task.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Added support for populating [`ExtendedResources`](https://github.com/flyteorg/flyte/pull/4172) from the `@task` decorator or node `.with_overrides()`. Currently only supports the ability to specify the GPU accelerator to use for a specific task node. The following syntaxes are valid when specifying a task:

No preference of GPU accelerator to use:
```
@task(limits=Resources(gpu="1"))
def my_task() -> None:
    ...
```

Schedule on a specific GPU accelerator:
```
from flytekit.extras.accelerators import NvidiaTeslaT4


@task(
    limits=Resources(gpu="1"),
    accelerator=NvidiaTeslaT4,
)
def my_task() -> None:
    ...
```

Schedule on a Multi-instance GPU (MIG) accelerator with no preference of partition size:
```
from flytekit.extras.accelerators import NvidiaTeslaA100


@task(
    limits=Resources(gpu="1"),
    accelerator=NvidiaTeslaA100,
)
def my_task() -> None:
    ...
```

Schedule on a Multi-instance GPU (MIG) accelerator with a specific partition size:
```
from flytekit.extras.accelerators import NvidiaTeslaA100


@task(
    limits=Resources(gpu="1"),
    accelerator=NvidiaTeslaA100.with_partition_size(NvidiaTeslaA100.partition_sizes.PARTITION_1G_5GB),
)
def my_task() -> None:
    ...
```

Schedule on an unpartitioned Multi-instance GPU (MIG) accelerator:
```
from flytekit.extras.accelerators import NvidiaTeslaA100


@task(
    limits=Resources(gpu="1"),
    accelerator=NvidiaTeslaA100.with_partition_size(None),
)
def my_task() -> None:
    ...
```

An override can also be specified for the GPU accelerator to use, as follows:
```
from flytekit.extra.accelerators import NvidiaTeslaA100, NvidiaTeslaT4


@task(
    limits=Resources(gpu="1"),
    accelerator=NvidiaTeslaT4,
)
def my_task() -> str:
    return "hello"

@workflow
def my_wf() -> str:
    return my_task().with_overrides(
        accelerator=NvidiaTeslaA100.with_partition_size(NvidiaTeslaA100.partition_sizes.PARTITION_1G_5GB)
    )
```

## Tracking Issue

## Follow-up issue
_NA_
OR


flytekit/extras/accelerators.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def __init__(self, device: str, partition_sizes: Type[PART]) -> None:
+        self._device = device
+        self._partition_sizes = partition_sizes
+
+    @property
+    def partition_sizes(self) -> Type[PART]:
+        return self._partition_sizes
+
+    def with_partition_size(self: MIG, partition_size: Optional[PART]) -> MIG:
```
eapolinario, line 49: can we change the typehint for `partition_size` to be `PART`? If the user is going through the trouble of caling `with_partition_size` it's reasonable to assume that they have a real value. wdyt?

tests/flytekit/unit/models/test_workflow_closure.py
```diff
@@ -58,6 +61,7 @@ def test_workflow_closure():
             {},
             {},
         ),
+        extended_resources=_core_task.ExtendedResources(gpu_accelerator=NvidiaTeslaT4.to_flyte_idl()),
```
eapolinario, line 64: why are we changing this?

eapolinario (commented):
one minor comment, otherwise LGTM.

flytekit/core/base_task.py
```diff
@@ -24,6 +24,8 @@
 from dataclasses import dataclass
 from typing import Any, Coroutine, Dict, Generic, List, Optional, OrderedDict, Tuple, Type, TypeVar, Union, cast
 
+from flyteidl.core import tasks_pb2 as _core_task
```
wild-endeavor, line 27: let's not use the _ anymore?  i'm actively trying to remove it from the code base.

flytekit/core/node.py
```diff
@@ -4,6 +4,8 @@
 import typing
 from typing import Any, List
 
+from flyteidl.core import tasks_pb2 as _core_task
```
wild-endeavor, line 7: ditto

tests/flytekit/unit/models/test_workflow_closure.py
jeevb: Probably not super necessary, but just wanted to make sure that this test covers the full `TaskTemplate` object.

flytekit/core/base_task.py
jeevb: Will clean up. Are you using `tasks_pb2` directly, or still aliasing as `core_task`?

flytekit/extras/accelerators.py
jeevb: `.with_partition_size(None)` is still valid, for explicitly requesting an unpartitioned GPU.

kumare3:
@jeevb can you remind me why can we not do
```python
@task(resources=Resources(gpu= NvidiaTeslaA100("1"))
   ...
```
Is this because of requests and limits?

Also can we drop
`NvidiaTesla -> Simply keep A100` I think this is well understood, lets be less verbose

Also this is too verbose `NvidiaTeslaA100.with_partition_size(NvidiaTeslaA100.partition_sizes.PARTITION_1G_5GB)`
Can we make it
`A100.partition(A100.partitions.5GB)` (what is the use of 1G?)

jeevb:
> @jeevb can you remind me why can we not do
> 
> ```python
> @task(resources=Resources(gpu= NvidiaTeslaA100("1"))
>    ...
> ```
> 
> Is this because of requests and limits?

In a way, yes. It introduces room for confusion when specifying required GPUs - would a request of T4(“1”) and a limit of A100(“1”) be valid? We could work around it by adding validations that GPU types match in requests and limits. In addition, we’ll have to break up the “Resources” object into count and GPU type at serialization time anyway.

> Also can we drop `NvidiaTesla -> Simply keep A100` I think this is well understood, lets be less verbose

👍 

> Also this is too verbose `NvidiaTeslaA100.with_partition_size(NvidiaTeslaA100.partition_sizes.PARTITION_1G_5GB)` Can we make it `A100.partition(A100.partitions.5GB)` (what is the use of 1G?)

See: https://cloud.google.com/kubernetes-engine/docs/how-to/gpus-multi

Each partition size definition combines the compute: 1G, and memory: 5GB. “1g.5gb” is the format that Nvidia uses. While GCP doesn’t support it currently, “3g.20gb” and “4g.20gb” are both [valid](https://docs.nvidia.com/datacenter/tesla/mig-user-guide/index.html#a100-profiles) for the A100.

An argument can be made that it would be better to be consistent with existing semantics instead of creating new ones, even if they may be more verbose. Note GCP’s use of the “gke-gpu-partition-size” label instead of “gke-gpu-partition”, for instance. That isn’t necessarily stopping us from doing it differently (both seem plausible), but it might come with the risk of introducing additional cognitive burden to users.

Also, “A100.partition” and “A100.partitions” might be too close. I can see if “A100(A100.partitions.X)” is feasible perhaps. Thoughts?

jeevb:
@kumare3: Went ahead and made the change, so the MIG definition is now: `A100` (unspecified partition size), `A100(A100.partitions.PARTITION_1G_5GB)` (explicit partition size), `A100(None)` (explicit unpartitioned). Wdyt?

wild-endeavor (approved):

kumare3:
One question, what if we don't support partitioning - what happens if a user selects it - or thinking about it another way - what is the reason a user chooses a partitioned (specific) gpu. The reason to ask is to see if we can arrive with something so
Simple that it is obvious

jeevb:
> One question, what if we don't support partitioning - what happens if a user selects it - or thinking about it another way - what is the reason a user chooses a partitioned (specific) gpu. The reason to ask is to see if we can arrive with something so
> Simple that it is obvious

To explicitly on a larger partition. If the cluster only has a single GPU node group, the extra partition specification is not needed. Users need an escape hatch to schedule on specific GPU configs.

jeevb:
Final defs:
* `A100` (unspecified partition size)
* `A100.partition_1g_5gb` (explicit partition size)
* `A100.unpartitioned` (explicit unpartitioned)

wild-endeavor (approved):

PR #1846, opened 2023-09-22
[WIP] Fix: Allow elastic tasks to be recoverable

fg91:
# TL;DR

For some failures, users might want to retry a flyte task, for other failures not.
Flyte tasks can be marked as retriable by raising a `FlyteRecoverableException`.

Torch's `elastic_launch` (which is used by flytekit's `Elastic` task type) always raises a `ChildFailedError` even if in the worker process e.g. a `FlyteRecoverableException` was raised.

This means that Flyte's retry mechanism cannot be used for user errors in elastic tasks.

Elastic tasks offer `torchrun`'s retry mechanism by specifying `max_retries`. However, this retry mechanism (which doesn't restart the pod but the worker processes within the pod) will retry *all* exceptions which is not always desireable.

This PR makes elastic tasks work with Flyte's retry mechanism by propagating `FlyteRecoverableException` up from the worker processes.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -249,9 +249,12 @@ def spawn_helper(
         prev_checkpoint=checkpoint_src,
     ):
         fn = cloudpickle.loads(fn)
+
         try:
             return_val = fn(**kwargs)
         except Exception as e:
+            # See explanation in `create_recoverable_error_file` why we check
+            # for recoverable errors here in the worker processes.
             if isinstance(e, FlyteRecoverableException):
                 create_recoverable_error_file()
             raise
@@ -369,6 +372,8 @@ class PytorchElasticFunctionTask(PythonFunctionTask[Elastic]):
                 try:
                     return_val = self._task_function(**kwargs)
                 except Exception as e:
+                    # See explanation in `create_recoverable_error_file` why we check
+                    # for recoverable errors here in the worker processes.
                     if isinstance(e, FlyteRecoverableException):
                         create_recoverable_error_file()
                     raise
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -298,26 +298,15 @@ class PytorchElasticFunctionTask(PythonFunctionTask[Elastic]):
         """
         Execute the task function using torch distributed's `elastic_launch`.
 
-        Error handling in elastic training tasks is complicated by two timing issues:
-        1. Torch's `elastic_launch`, in each worker group (pod), fails with the first exception raised in any of
-            the worker processes in this worker group (processes within the pod).
-        2. The pods belonging to a Flyte Elastic task write a single `error.pb` into blob storage, causing a
-            race condition as one pod might overwrite the error file of another pod.
-
-        To determine whether an elastic training task is recoverable, we introduce the following convention:
-
-        If the first exception raised in the worker group (pod) with index 0 is a `FlyteRecoverableException`,
-        we consider the failure recoverable. Otherwise, we consider the failure non-recoverable.
-
         Returns:
             The result of (global) rank zero.
 
         Raises:
-            FlyteRecoverableException: If the first exception raised in the worker group with index 0 is a
-                `FlyteRecoverableException`.
-            RuntimeError: If the first exception raised in the worker group with index 0 is not a
-                `FlyteRecoverableException`.
-            IgnoreOutputs: Always raised by default, even if the task is successful, in any worker group with index > 0.
+            FlyteRecoverableException: If the first exception raised in the local worker group is or
+                inherits from `FlyteRecoverableException`.
+            RuntimeError: If the first exception raised in the local worker group is not and does not
+                inherit from `FlyteRecoverableException`.
+            IgnoreOutputs: Raised when the task is succesfull in any worker group with index > 0.
         """
         try:
             from torch.distributed.launcher.api import LaunchConfig, elastic_launch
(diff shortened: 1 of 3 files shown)


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -246,8 +249,12 @@ def spawn_helper(
         prev_checkpoint=checkpoint_src,
     ):
         fn = cloudpickle.loads(fn)
-        return_val = fn(**kwargs)
-
+        try:
+            return_val = fn(**kwargs)
```
fg91, line 253: This is the invocation of the actual task function in the worker processes.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -289,11 +296,28 @@ def __init__(self, task_config: Elastic, task_function: Callable, **kwargs):
...
     def _execute(self, **kwargs) -> Any:
         """
-        This helper method will be invoked to execute the task.
+        Execute the task function using torch distributed's `elastic_launch`.
+
+        Error handling in elastic training tasks is complicated by two timing issues:
+        1. Torch's `elastic_launch`, in each worker group (pod), fails with the first exception raised in any of
+            the worker processes in this worker group (processes within the pod).
+        2. The pods belonging to a Flyte Elastic task write a single `error.pb` into blob storage, causing a
+            race condition as one pod might overwrite the error file of another pod.
```
fellhorn, line 305: How does this work currently? If multiple workers fail it seems like it still only displays the error message of the first one which failed.

fg91: [This](https://github.com/flyteorg/flytekit/blob/3f87154d0356053483fe123ee6e4b4dbbae0a4f8/flytekit/bin/entrypoint.py#L169C25-L169C25) is the part in the pods' entrypoint where the error file is uploaded. 
See [here](https://github.com/flyteorg/flytekit/blob/3f87154d0356053483fe123ee6e4b4dbbae0a4f8/flytekit/core/data_persistence.py#L309) and then [here](https://github.com/flyteorg/flytekit/blob/3f87154d0356053483fe123ee6e4b4dbbae0a4f8/flytekit/core/data_persistence.py#L209).

fg91: For me, an open question is whether the error file can be overwritten by other pods.

But even if not, there is definitely a race condition which pod get's to write its file first.

fg91: > How does this work currently? If multiple workers fail it seems like it still only displays the error message of the first one which failed.

So yes, a single one is displayed. But I think we don't have a guarantee it's the first one.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
kumare3: lets add a comment here

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
kumare3: This is an interesting catch, we could write error files from all pods. But will all pods write and error file, if so we can and let the plugin collate it?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: Will do 👍

fellhorn:
TODO @fg91 
Remember to remove https://github.com/flyteorg/flytekit/blame/de8a5fe7f7e249785c28edfb0ca206e0f12b85f0/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L365

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: This would be the cleanest solution.
All pods will write an error file unless they just disappear because of e.g. a preemption. 

How would this look like in practice?

* Would we add an optional function to the plugin interface [here](https://github.com/flyteorg/flyteplugins/blob/8b417f4029e7b2965cdc667c75304a1528c5a880/go/tasks/pluginmachinery/k8s/plugin.go#L88) to give plugins the option to customize the interpretation of error file(s)?
* On the python side, in [the entrypoint](https://github.com/flyteorg/flytekit/blob/de8a5fe7f7e249785c28edfb0ca206e0f12b85f0/flytekit/bin/entrypoint.py#L120), would we check whether the "task class" wants to customize how the respective error file is called so that the different workers don't overwrite the file?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: https://github.com/flyteorg/flytekit/pull/1846/commits/5ebce626a6866a071a75fa34f8a1c8dfa491cd5f

fg91:
> TODO @fg91 Remember to remove https://github.com/flyteorg/flytekit/blame/de8a5fe7f7e249785c28edfb0ca206e0f12b85f0/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L365

Thanks for the headsup, removed https://github.com/flyteorg/flytekit/pull/1846/commits/fe619e1d56da31230e234cc2406a751f01744dcc

fg91:
@fellhorn given [Ketan's proposal](https://github.com/flyteorg/flytekit/pull/1846#discussion_r1338046086) to properly resolve the race condition which worker pod get's to write the `error.pb`, I removed the logic that always makes worker group 0 write the file regardless of whether the first exception occured in another worker group.

So this PR now only tackles the problem of propagating the recoverable exception to the agent process. Could you pls take another look?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: Given @kumare3's proposal to properly resolve the race condition which worker pod get's to write the `error.pb`, I removed the logic that always makes worker group 0 write the file regardless of whether the first exception occured in another worker group.

So this PR now only tackles the problem of propagating the recoverable exception to the agent process.

kumare3 (approved):

PR #1860, opened 2023-09-29
Add update_execution func in remote

ericwudayi:
# TL;DR
As Title

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/flytekit/clients/friendly.py
+++ b/flytekit/clients/friendly.py
@@ -560,6 +560,21 @@ class SynchronousFlyteClient(_RawSynchronousFlyteClient):
             .id
         )
 
+    def update_execution(self, id, state = None, tags: typing.List[str] = None):
+        """
+        This will update an execution for the given execution spec.
+        :param flytekit.models.core.identifier.WorkflowExecutionIdentifier id:
+        :param state: Optional state to update the execution to.
+        :param list[str] tags: Optional tags to update the execution with.
+        """
+        return super(SynchronousFlyteClient, self).update_execution(
+                _execution_pb2.ExecutionUpdateRequest(
+                    id = id.to_flyte_idl(),
+                    state = state,
+                    tags = tags,
+                )
+            )
+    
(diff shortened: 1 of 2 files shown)


wild-endeavor (commented):
can we add a type hint for [state](https://github.com/flyteorg/flyteidl/blob/85bfc8ff1c36b2d37a4fcc68a2be1911a9fc3940/protos/flyteidl/admin/execution.proto#L378)?  Not sure if there's a model object or something that should be used.  Can also use a more user-friendly state in remote.py and only use the idl object in friendly.py.

ericwudayi:
Sure, in flyteidl the state is type of ExecutionState.
```
enum ExecutionState {
    // By default, all executions are considered active.
    EXECUTION_ACTIVE = 0;

    // Archived executions are no longer visible in the UI.
    EXECUTION_ARCHIVED = 1;
}
```

How about to use the _int_  in remote.py and use _ExecutionState_ in friendly.py

wild-endeavor:
Oh no enums are nice, let's just use it in both places.

ericwudayi:
Use enum class make use-case not so straight-forward. User need to declare ExecutionState when update. But it is very clear.
```
from ... import ExecutionState
remote.update_execution(
   ..., state=ExecutionState.EXECUTION_ARCHIVED, tags=["..."])
```

On the other side, if we use  typing.Literal. It is easy to use, but they need to look into the definition to see the meaning of value. 
```
def update_execution(
        self,
        ...
        state: typing.Literal[0, 1] = None,
        tags: typing.List[str] = None,
    ) -> FlyteWorkflowExecution:
        """Update a workflow execution entity from flyte admin.
        ...
        :param state: state to be updated. 0 = Active, 1 = Archived.
        :param tags: List of tags to be updated.
        ...
        """
```
Which one do you think is better?

PR #1863, opened 2023-09-29
Add driver and executor pod template params to Spark config

andrewwdye:
# TL;DR
Add driver and executor pod template params to Spark config

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
This change adds pairs of pod template and pod template name params for the driver and executor pods. This mirrors the interface we provide for container tasks generally.

TODO: once we get consensus on the api I will populate the SparkJob proto and add tests

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4105

### DIFF
--- a/plugins/flytekit-spark/flytekitplugins/spark/models.py
+++ b/plugins/flytekit-spark/flytekitplugins/spark/models.py
@@ -5,6 +5,7 @@ from flyteidl.plugins import spark_pb2 as _spark_task
 from google.protobuf import json_format
 from google.protobuf.struct_pb2 import Struct
 
+from flytekit.core.pod_template import PodTemplate
 from flytekit.exceptions import user as _user_exceptions
 from flytekit.models import common as _common
 
@@ -28,6 +29,10 @@ class SparkJob(_common.FlyteIdlEntity):
         databricks_conf: Dict[str, Dict[str, Dict]] = {},
         databricks_token: Optional[str] = None,
         databricks_instance: Optional[str] = None,
+        driver_pod_template: Optional["PodTemplate"] = None,
+        driver_pod_template_name: Optional[str] = None,
+        executor_pod_template: Optional["PodTemplate"] = None,
+        executor_pod_template_name: Optional[str] = None,
     ):
         """
         This defines a SparkJob target.  It will execute the appropriate SparkJob.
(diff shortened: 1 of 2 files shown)


plugins/flytekit-spark/flytekitplugins/spark/models.py
```diff
@@ -28,6 +29,10 @@ def __init__(
         databricks_conf: Dict[str, Dict[str, Dict]] = {},
         databricks_token: Optional[str] = None,
         databricks_instance: Optional[str] = None,
+        driver_pod_template: Optional["PodTemplate"] = None,
```
wild-endeavor, line 32: as much as we want to get rid of model files, since these are there, i say we stick to them for now.  There's a [K8sPod model class](https://github.com/flyteorg/flytekit/blob/0e82799e3a701075012a6da76f4a4f5846d5c969/flytekit/models/task.py#L870) that I think we should use here since the [idl is going](https://github.com/flyteorg/flyteidl/pull/450/files#diff-394e059361a84ee9e734be9a23262106290623951064a5bb44ab92b6fb86cbe6R39) to use that object too.

plugins/flytekit-spark/flytekitplugins/spark/task.py
```diff
@@ -25,12 +26,20 @@ class Spark(object):
...
+        driver_pod_template_name: Name of the existing PodTemplate resource to use for the driver pod.
+        executor_pod_template: PodTemplate to use for the executor pods.
+        executor_pod_template_name: Name of the existing PodTemplate resource to use for the executor pods.
     """
 
     spark_conf: Optional[Dict[str, str]] = None
     hadoop_conf: Optional[Dict[str, str]] = None
     executor_path: Optional[str] = None
     applications_path: Optional[str] = None
+    driver_pod_template: Optional["PodTemplate"] = None
```
wild-endeavor, line 39: This is okay to be `PodTemplate` though.  And we can translate from `PodTemplate` to the model `K8sPod` using this [same logic](https://github.com/flyteorg/flytekit/blob/0e82799e3a701075012a6da76f4a4f5846d5c969/flytekit/core/python_auto_container.py#L205).  This translation would happen in the `get_custom` function below.

wild-endeavor (commented):
does this make any sense?

andrewwdye:
Closing until we get consensus on the proto interface for multi-role pod template overrides. See comment [here](https://github.com/flyteorg/flyte/pull/4179#issuecomment-1753579133).

PR #1869, opened 2023-10-04
[WIP] Rust Data Persistence

troychiu:
# TL;DR
Develop a Rust data persistence to improve the performance of uploading/downloading files to/from S3.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
**This PR is still in progress and is created for more discussion.**
In this PR, I developed a data persistence layer with Rust and PyO3(https://pyo3.rs/v0.19.2/) to improve the performance of uploading/downloading files to/from S3.

Below is the experiment result. As you can see, the improvement is huge.
put_local
get_local
put_s3
(Due to the limited upload bandwidth of my network, there is no obvious improvement of putting files to s3.) 
get_s3

**In my opinion, the performance gain mainly comes from:**
1. simultaneously processing requests of small chunks of file with async I/O
2. using rust

**Follow these steps to use the RustS3FileSystem:**
1. install packages and build the rust code
```
make setup
pip install maturin
cd flytekit/core/rust
maturin develop
```
2. export s3 config
```
export AWS_ACCESS_KEY_ID=minio
export AWS_SECRET_ACCESS_KEY=miniostorage
```
3. use RustS3FileSystem class 
```
from flytekit.core.rust.fs import RustS3FileSystem
import asyncio
s3kwargs = {'client_kwargs': {'endpoint_url': 'http://localhost:30002'}}
rustfs = RustS3FileSystem(**s3kwargs)
lpath, rpath = 'path/to/local/file', 'path/to/remote/file'
loop = asyncio.get_event_loop()
task = loop.create_task(rustfs._put_file(lpath=lpath, rpath=rpath))
loop.run_until_complete(task)
```

**TODO Items**
1. Finish the s3 config read according to fsspec's implementation since it should support specifying s3 access key in codes as well as cli export
2. Change rust put_file and get_file to async function so that we can avoid using asyncio.to_thread in fs.py
3. Finish the TODO part in fs.py according to fsspec's implementation
4. Check the memory usage
5. Production build
6. Integrate into flytekit

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4081

## Follow-up issue
_NA_
OR


flytekit/core/rust/fs.py
```diff
@@ -0,0 +1,26 @@
+import asyncio
+from s3fs import S3FileSystem
+from fsspec.callbacks import _DEFAULT_CALLBACK
+import rustfs
+
+class RustS3FileSystem(S3FileSystem):
```
kumare3, line 6: should we re-implement the entire s3Filesystem? it seems s3Filesystem is inefficient? 
Isn't this multi-part effectively serialized because of the `await` https://github.com/fsspec/s3fs/blob/a14c31a382c50996aff37107245768a13efef0cb/s3fs/core.py#L1179?

kumare3: ohh i see, you have overriden __put_file. 
Can you test it against a pure python implementation that does not `await` every chunk, but awaits the futures after all chunks of enqueued?

flytekit/core/rust/src/lib.rs
```diff
@@ -0,0 +1,208 @@
...
+                            .path(lpath)
+                            .offset(chunk_index * CHUNK_SIZE)
+                            .length(Length::Exact(this_chunk))
+                            .build()
+                            .await
+                            .unwrap();
+
+                        // Chunk index needs to start at 0, but part numbers start at 1.
+                        let part_number = (chunk_index as i32) + 1;
+                        let upload_part_res = client.upload_part()
```
kumare3, line 111: how are errors handled?
If you see S3FS handles errors
https://github.com/fsspec/s3fs/blob/a14c31a382c50996aff37107245768a13efef0cb/s3fs/core.py#L108-L128

can you test with no s3 connection or kill the connection in between, how does that work
or in case we get a slowdown or some other retryable error

flytekit/core/rust/fs.py
kumare3: NM i think I am wrong, cc @EngHabu I see that s3fs uses the aiobotocore, and I think upload_part in aiobotocore returns a stream once established

wild-endeavor (commented):
let me think about how to add this.  would love to get this merged this in asap but opt-in.

flytekit/core/rust/src/lib.rs
troychiu: Based on what I know, the error in rust will be converted to Python error, but I will spent some time looking into it.

Of course I can test that. I will post the result later

flytekit/core/rust/src/lib.rs
```diff
@@ -0,0 +1,208 @@
+use std::io::SeekFrom;
```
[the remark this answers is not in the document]
troychiu: Yes it can be called s3fs.rs. I will change it.

Let me check the link!

flytekit/core/rust/fs.py
troychiu: I think s3fs reads the chunks one by one because it await for every read. https://github.com/fsspec/s3fs/blob/a14c31a382c50996aff37107245768a13efef0cb/s3fs/core.py#L1234,

troychiu: But one thing we can test is what you mentioned, using pure python implementation and awaits the futures after all chunks are enqueued.

troychiu: FYI, ttps://github.com/fsspec/s3fs/issues/796

troychiu:
### Memory usage
Get a 256 MB object
s3fs
s3fs_get_s3_mem
Rust
rustfs_get_s3_mem
Put a 256 MB object
s3fs
s3fs_put_s3_mem
Rust
rustfs_put_s3_mem

As you can see, Rust uses more memory in get and less memory in put. However, this still depends on the quality of implementation, so I will keep looking into whether there is anything that I can optimize.

troychiu:
I have implemented a pure python implementation which downloads and uploads files with asyncio. The performance is pretty close to the Rust version, but it's more easy to be integrated into the codebase. 

Below are some experiment result.
local with rtt=20 get/put (since my internet bandwidth is limited, I use local network with rtt=20 to simulate the internet)
rtt_get
rtt_put
memory usage:
put 256MB
mem_put_s3fs-async
get 256MB
mem_get_s3fs-async

flytekit/core/rust/fs.py
troychiu: Hi @kumare3, I have finished the pure python implementation. Please check the latest comment.

troychiu:
Hi @kumare3, which way do you think is better? Python or Rust?
You can check last comment for more information.

wild-endeavor:
my vote is to merge this in as optional code, with a disclaimer saying that we're still playing around with this.  and then cut an alpha release of it and test under a variety of circumstances.  would be nice to be able to add some monitoring to this as well.  What's in the test file you're using the produce the charts?  is there any telemetry we can emit without adding to the overhead?  (we should add for current default implementation as well)

What steps are remaining for cleaning up the pr?  also do you know how to bundle multiple arch's in a pypi library?  we'll have to build multiple rust binaries right?

EngHabu:
Thank you, @troychiu for the better (than s3fs) Python implementation!
What would be the way to let people choose this implementation over s3fs? an ENV var? or just installing a package would trigger the behavior?

EngHabu:
In case it wasn't clear, my vote is to go with the asyncio version as well!

kumare3:
My vote would be to make this as a plugin that can override default s3 implementation and we can try and use it with a few folks - see how it scales and then we can merge it in main code if needed.

Eventually I like the idea of rust just to minimize dependencies in python

troychiu:
@wild-endeavor @EngHabu @kumare3, thank you for the insightful discussion.

I've consulted with @pingsutw and we've decided to proceed as follows:
1. Both the Python and Rust versions will be developed as data persistence plugins. This ensures that the original data persistence won't be overridden by default.
2. Users will have the flexibility to choose between the original persistence and our plugins. The default will change only if they actively install and specify one of our plugins.
3. I'll begin with integrating the Python version, given its simplicity. Following that, I'll tackle the Rust integration.

Your feedback will be invaluable as we move forward.

troychiu:
> my vote is to merge this in as optional code, with a disclaimer saying that we're still playing around with this. and then cut an alpha release of it and test under a variety of circumstances. would be nice to be able to add some monitoring to this as well. What's in the test file you're using the produce the charts? is there any telemetry we can emit without adding to the overhead? (we should add for current default implementation as well)
> 
> What steps are remaining for cleaning up the pr? also do you know how to bundle multiple arch's in a pypi library? we'll have to build multiple rust binaries right?

The test files that I was using are random files that I created. What do you mean telemetry we can emit? Could you elaborate more?
For the Python version, the only step will be making it a plugin.
For the Rust version, there are a few things left
1. More comprehensive error handling
2. S3 config setting
3. Create a workflow to build multiple rust binaries for different arch. (What you mentioned)

PR #1873, opened 2023-10-05
Fixes Auth issue, where refresh token is expired

kumare3:
# TL;DR
Fixes error caused when refresh token is expired. This is intermittent.

Problem of type
```
❯ pyflyte run --remote  test2.py wf --n 10
Running Execution on Remote.
E1004 23:00:53.300669000 8616083776 hpack_parser.cc:853]               Error parsing metadata: error=invalid value key=content-type value=text/html
E1004 23:00:53.790464000 8616083776 hpack_parser.cc:853]               Error parsing metadata: error=invalid value key=content-type value=text/html
E1004 23:00:53.863833000 8616083776 hpack_parser.cc:853]               Error parsing metadata: error=invalid value key=content-type value=text/html
E1004 23:00:53.937017000 8616083776 hpack_parser.cc:853]               Error parsing metadata: error=invalid value key=content-type value=text/html
Failed with Exception Code: SYSTEM:Unknown
Underlying Exception: no refresh token available with which to refresh authorization credentials
```

`Underlying Exception: no refresh token available with which to refresh authorization credentials`

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/flytekit/clients/auth/keyring.py
+++ b/flytekit/clients/auth/keyring.py
@@ -68,11 +68,14 @@ class KeyringStore:
 
     @staticmethod
     def delete(for_endpoint: str):
-        try:
-            _keyring.delete_password(for_endpoint, KeyringStore._access_token_key)
-            _keyring.delete_password(for_endpoint, KeyringStore._refresh_token_key)
-            _keyring.delete_password(for_endpoint, KeyringStore._id_token_key)
-        except PasswordDeleteError as e:
-            logging.debug(f"Id token not found in key store, not deleting. Error: {e}")
-        except NoKeyringError as e:
-            logging.debug(f"KeyRing not available, tokens will not be cached. Error: {e}")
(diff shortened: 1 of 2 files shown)


flytekit/clients/auth/keyring.py
```diff
@@ -71,9 +71,8 @@ def delete(for_endpoint: str):
         try:
             _keyring.delete_password(for_endpoint, KeyringStore._access_token_key)
```
wild-endeavor, line 72: should we just do three separate try/catches?  to make sure we always clean up all three things?

kumare3: Makes sense. Fixed.

wild-endeavor (approved):

PR #1878, opened 2023-10-06
Dep: Pin min required flytekit version in IAP plugin

fg91:
# TL;DR
The flytekit identity aware proxy plugin introduced in https://github.com/flyteorg/flytekit/pull/1787 requires changes in flytekit's authentication logic which were introduced in the same PR. At the moment, the flytekit version is not pinned in the new plugin's `setup.py` though. In this PR I pin it to `>=1.10`, the next release that will contain those changes to the auth logic.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added


pingsutw (approved):
Thanks!

PR #1894, opened 2023-10-15
add default project domain in execute launch plan

troychiu:
# TL;DR
1. add default project domain in execute launch plan
2. add name as a parameter
## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
When I executed a launch plan with FlyteRemote, it didn't run with default project domain of the remote but needed specified project domain. So, I fixed it with if there is no specified project domain, use default project domain of the remote.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4249

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -1232,6 +1232,7 @@ class FlyteRemote(object):
                 version=version,
                 project=project,
                 domain=domain,
+                name=name,
                 execution_name=execution_name,
                 execution_name_prefix=execution_name_prefix,
                 options=options,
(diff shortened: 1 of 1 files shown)


wild-endeavor (commented):
should we just add this to the top level def execute instead?

troychiu:
> should we just add this to the top level def execute instead?

Hi @wild-endeavor, do you mean we can resolve the identifiers before actually classify different types of entities?
For example, move the `resolved_identifiers = self._resolve_identifier_kwargs(entity, project, domain, name, version)`
to https://github.com/flyteorg/flytekit/blob/74d63d7265383c2af07952e835bcebf95b2d0942/flytekit/remote/remote.py#L1158

pingsutw (approved):

PR #1899, opened 2023-10-18
Fix bug in polars parquet decoding handler

danpf:
# TL;DR
When trying to use `flytekitplugins-polars` I would get the error:
```
Message:
    Could not open Parquet input source '<Buffer>': Parquet file size is 0 bytes
User error.
```

Checking out the code, the `PolarsDataFrameToParquetEncodingHandler` actually places the parquet file at '{location}/00000', but it tries to read it at '{location}'

I think this is the easy solution, but maybe `PolarsDataFrameToParquetEncodingHandler` should be returning the '{location}/00000' instead of the parent directory? 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
+++ b/plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
@@ -1,3 +1,4 @@
+import io
 import typing
 
 import pandas as pd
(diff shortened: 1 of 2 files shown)


pingsutw (commented):
could you update this [test](https://github.com/flyteorg/flytekit/blob/da8d03118b8f5962a254512ac3a27bfd3f00680d/plugins/flytekit-polars/tests/test_polars_plugin_sd.py#L73)? To test the decoder, we should have a task that produces structured dataset, and pass it to the download stream task.

danpf:
> could you update this [test](https://github.com/flyteorg/flytekit/blob/da8d03118b8f5962a254512ac3a27bfd3f00680d/plugins/flytekit-polars/tests/test_polars_plugin_sd.py#L73)? To test the decoder, we should have a task that produces structured dataset, and pass it to the download stream task.

@pingsutw I'm not exactly sure what you mean by this.  But I added a very basic roundtrip style test, and I merged with master.

After looking at the code more closely, I believe that this current implementation is more in line with what was intended.  I also suspect that using pandas in the tests was unintentional (but i could be wrong), so I changed that (but it works either way)

Also, for the html representation of the data, I wonder if pl.DataFrame._repr_html_ might be a better fit?

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
```diff
@@ -40,23 +40,32 @@ def encode(
         structured_dataset_type: StructuredDatasetType,
     ) -> literals.StructuredDataset:
         df = typing.cast(pl.DataFrame, structured_dataset.dataframe)
-
+        random_string = ctx.file_access.get_random_string()
```
kumare3, line 43: why do we need the random string here?

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
```diff
@@ -40,23 +40,32 @@ def encode(
...
+                remote_dir,
+                random_string,
+            )
+        else:
+            remote_dir = ctx.file_access.get_random_remote_directory()
+            sd_uri = ctx.file_access.join(
+                remote_dir,
+                random_string,
             )
         ctx.file_access.upload_directory(local_dir, remote_dir)
```
kumare3, line 67: i actually want to get rid of this dual writes, one to local directory and then to remote. can we simply write to remote as a stream?

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
danpf: I changed this back to the default `00000` parquet filename

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
danpf: I've updated the code to write to a BytesIO stream. Let me know if that is in line with what you were thinking!

pingsutw (approved):

pingsutw:
thank you @danpf

PR #1907, opened 2023-10-21
Return None type instead of exception

honnix:
# TL;DR
Return None type instead of exception for list subtype

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
More details in https://github.com/flyteorg/flyte/issues/4270

## Tracking Issue
Fixes https://github.com/flyteorg/flyte/issues/4270

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1108,6 +1108,16 @@ class ListTransformer(TypeTransformer[T]):
         """
         Return the generic Type T of the List
         """
+        if (sub_type := ListTransformer.get_sub_type_or_none(t)) is not None:
+            return sub_type
+
+        raise ValueError("Only generic univariate typing.List[T] type is supported.")
+
+    @staticmethod
+    def get_sub_type_or_none(t: Type[T]) -> Optional[Type[T]]:
+        """
+        Return the generic Type T of the List, or None if the generic type cannot be inferred
+        """
         if hasattr(t, "__origin__"):
             # Handle annotation on list generic, eg:
             # Annotated[typing.List[int], 'foo']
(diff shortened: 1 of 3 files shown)


flytekit/core/type_engine.py
```diff
@@ -1090,6 +1090,16 @@ def get_sub_type(t: Type[T]) -> Type[T]:
 
         raise ValueError("Only generic univariate typing.List[T] type is supported.")
 
+    @staticmethod
+    def get_sub_type_or_none(t: Type[T]) -> Optional[Type[T]]:
```
honnix, line 1094: I chose to add a new method because `get_sub_type` is used in many places where they assume it never returns `None`.

honnix:
Not sure what happened to CI. The failure is not related to the change and it also happens to another PR.

kumare3 (commented):
We do not have support for list of unknown types- what does none mean?

honnix:
> We do not have support for list of unknown types- what does none mean?

It is sort of unknown due to lack of python interface. The use case is https://github.com/flyteorg/flytekit/blob/1b5692adcee561449afe94310c804e1acf5cfd42/flytekit/remote/remote_callable.py#L34. `DictTransformer` handles this by returning `None`, so the type will be inferred by looking at the elements themselves https://github.com/flyteorg/flytekit/blob/1b5692adcee561449afe94310c804e1acf5cfd42/flytekit/core/promise.py#L615

As https://github.com/flyteorg/flyte/issues/4270 describes, binding currently does not work for list type in a remote entity.

honnix:
Hi. Can someone take a look at this PR again? Thank you.

pingsutw (approved):

adrianloy:
Thanks for fixing this, was about to open a PR to address it. Nice to see its already on main!

PR #1917, opened 2023-10-24
BranchNode local execution(#4080)

ericwudayi:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


ericwudayi:
Actually, I run all the workflow testing file by replacing the execution function with newer one, `execution_with_graph`. I pass almost all the tests except for following nine tests. I take a deep dig in following cases, and they can be summarized into three cases.
1. Failing condition on BranchNode. --> also failed in remote. (It would not show the failing message now) 
2. Type checking. --> cannot check using node-based execution.
```
@workflow
def wf():
  return 1

```
3. Regex error. --> missing the line cannot be recovered by node-based execution.

Screenshot 2023-10-25 at 9 37 16 AM

samhita-alla:
@eapolinario @pingsutw @wild-endeavor can one of you please review?

flytekit/core/gate.py
```diff
@@ -127,6 +134,9 @@ def local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], Pr
     def local_execution_mode(self):
         return ExecutionState.Mode.LOCAL_TASK_EXECUTION
 
+    def __call__(self, *args: object, **kwargs: object) -> Union[Tuple[Promise], Promise, VoidPromise, Tuple, None]:
```
kumare3, line 137: why do we have to do this, this is already in the base task type right?

ericwudayi: that's right. however for now, gate does not inherit base task type. so I add __call__ instead.

tests/flytekit/unit/core/test_workflows_simulate_remote.py
```diff
@@ -0,0 +1,194 @@
...
+
+default_img = Image(name="default", fqn="test", tag="tag")
+serialization_settings = flytekit.configuration.SerializationSettings(
+    project="project",
+    domain="domain",
+    version="version",
+    env=None,
+    image_config=ImageConfig(default_image=default_img, images=[default_img]),
+)
+
```
pingsutw, line 21: could we add more tests here?
1. dynamic task inside a workflow
2. gate node inside a workflow
3. map task inside a workflow

flytekit/core/condition.py
```diff
@@ -35,27 +35,30 @@ def name(self):
 
     # Output Node or None
     def __call__(self, **kwargs):
-        # Eval every ConjunctionExpression and ComparisonExpression
-        self._cs.eval_by_kwargs(**kwargs)
         for c in self._cs.cases:
-            # c._expr = _update_promise(c.expr, kwargs)
-
+            c._expr = _update_promise(c.expr, kwargs)
```
ericwudayi, line 39: This line directly change the expression of conditional section. Would cause error when we call Branchnode Twice. I provide a little preseudo code on when it would failed.


```
@workflow
def branching(x: int) -> int:
    return conditional("test").if_(x == 2).then(t()).else_().then(wf1())

@workflow
def call_branching_twice(x: int, y:int) -> Tuple[int, int]:
    return branching(x=x), branching(x=y) # This would cause error because expression is overwrited
```

tests/flytekit/unit/core/test_workflows_simulate_remote.py
ericwudayi: Sure, i would do that.

flytekit/core/condition.py
pingsutw: will take a look

flytekit/core/workflow.py
```diff
@@ -195,6 +195,8 @@ def __init__(
         self._nodes: List[Node] = []
         self._output_bindings: List[_literal_models.Binding] = []
         self._docs = docs
+        # Create a map that holds the outputs of each node.
+        self._intermediate_node_outputs: Dict[Node, Dict[str, Promise]] = {GLOBAL_START_NODE: {}}
```
ericwudayi, line 199: I found this line may also have the node execution independency issue. If we execute the same node twice, this intermediate_node_output is not None before the second execution. Not sure if this would have any problem, maybe we can clear the state on the begining of execution_node?

PR #1922, opened 2023-10-28
Add vscode plugin to enable interactive debugging

troychiu:
# TL;DR
Add vscode plugin to enable interactive debugging

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [X] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
The Flytekit VSCode plugin offers an easy solution for users to run tasks within an interactive VSCode server, compatible with any image and any python task types (e.g. tfjob, pytorchjob, rayjob, etc). This plugin provides a `@vscode` decorator, which users can put within `@task` and user function.

In summary, `@vscode` decorator modifies a container to run a VSCode server:
    1. Overrides the user function with a VSCode setup function.
    2. Download vscode server from remote to local.
    3. Launches and monitors the VSCode server.
    4. Terminates after user specified duration.

## End to end test
1. Build the image with Dockerfile.dev `docker build --push . -f Dockerfile.dev -t localhost:30000/flytekit:dev --build-arg PYTHON_VERSION=3.8`
2. Write a task with `@vscode` decorator and run the workflow on remote `pyflyte run --remote --image localhost:30000/flytekit:dev vscode_task.py wf --a 30 --b 10`

vscode_task.py: 
```Python
from flytekitplugins.vscode import vscode

@task
@vscode
def t(a: int, b: int) -> int:
    return a + b

@workflow
def wf(a: int, b: int) -> int:
    out = t(a=a, b=b)
    return out
``` 
3. forward a local port to the pod `kubectl port-forward -n flytesnacks-development fd314dee47a304bffa75-n0-0 8080:8080`
4. access the server by opening a web browser and navigating to `localhost:8080`

Screenshot:
e2e

## Tracking Issue

## Follow-up issue
1. shut down the server based on the idle time instead of a fix duration
2. support vscode plugin in the code server
3. automatically generate launch.json to support **one click to run** feature
4. detect whether a task run in local mode
5. Have a second entry point defined in the `launch.json` which not only runs the task but also uploads the results and terminates the pod so that the workflow can then continue with the subsequent tasks.


plugins/flytekit-vscode/flytekitplugins/vscode/vscode.py
```diff
@@ -0,0 +1,170 @@
...
+    code_server_tar_path = download_file(code_server_remote_path, DOWNLOAD_DIR)
+
+    # Extract the tarball
+    with tarfile.open(code_server_tar_path, "r:gz") as tar:
+        tar.extractall(path=DOWNLOAD_DIR)
+
+    code_server_dir_path = os.path.join(DOWNLOAD_DIR, code_server_dir_name)
+
+    code_server_bin = os.path.join(code_server_dir_path, "bin", "code-server")
+
```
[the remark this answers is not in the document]
troychiu: Good suggestion, thanks!

troychiu:
I have added a test but I am not sure if there is any better way to test the plugin. Welcome any suggestion!

plugins/flytekit-vscode/README.md
```diff
@@ -0,0 +1,40 @@
+# Flytekit VSCode Plugin
+
+The Flytekit VSCode plugin offers an easy solution for users to run tasks within an interactive VSCode server, compatible with any image. `@vscode` is a decorator which users can put within @task and user function. With `@vscode`, the task will install vscode dependencies and run a vscode server instead of the user defined functions.
```
ByronHsu, line 3:
```suggestion
The Flytekit VSCode plugin offers an easy solution for users to run tasks within an interactive VSCode server, compatible with any image and any python task types (e.g. tfjob, pytorchjob, rayjob, etc). `@vscode` is a decorator which users can put within @task and user function. With `@vscode`, the task will install vscode dependencies (skip if they already exist)  and run a vscode server instead of the user defined functions.
```

plugins/flytekit-vscode/README.md
```diff
@@ -0,0 +1,40 @@
...
+    ...
+```
+
+## User Guide
+1. Run the decorated task on the remote. For example: `pyflyte run --remote [PYTHONFILE] [WORKFLOW|TASK] [ARGS]...`
+2. Once the code server is prepared, you can forward a local port to the pod. For example: `kubectl port-forward -n [NAMESPACE] [PODNAME] 8080:8080`.
+3. You can access the server by opening a web browser and navigating to `localhost:8080`.
+
+VSCode example screenshot:
+<img src="https://raw.githubusercontent.com/flyteorg/flytekit/master/plugins/flytekit-vscode/example.png">
```
ByronHsu, line 28: create `docs/` under flytekit-vscode folder and move `example.png` inside
you can refer the image by ``

plugins/flytekit-vscode/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Build Custom Image with VSCode Plugin
+Users have the option to create a custom image by including the following lines in their Dockerfile.
```
ByronHsu, line 31:
```suggestion
If users want to skip the vscode downloading process at runtime, they have the option to create a custom image with vscode by including the following lines in their Dockerfile.
```

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
```diff
@@ -0,0 +1,180 @@
...
+    # Extract the tarball
+    with tarfile.open(code_server_tar_path, "r:gz") as tar:
+        tar.extractall(path=DOWNLOAD_DIR)
+
+    code_server_dir_path = os.path.join(DOWNLOAD_DIR, code_server_dir_name)
+
+    code_server_bin_dir = os.path.join(code_server_dir_path, "bin")
+
+    # Add the directory of code-server binary to $PATH
+    os.environ["PATH"] = code_server_bin_dir + os.pathsep + os.environ["PATH"]
```
ByronHsu, line 104: is `os.join()` better than manually using `+`?

plugins/flytekit-vscode/README.md
```diff
@@ -0,0 +1,40 @@
...
+```
```
ByronHsu, line 32:
```suggestion
```Dockerfile
```

ByronHsu:
Also, add the testing section in the PR description, and attach an e2e test in the flytekit sandbox

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
```diff
@@ -0,0 +1,180 @@
...
+    code_server_dir_name: Optional[str] = "code-server-4.18.0-linux-amd64",
+    pre_execute: Optional[Callable] = None,
+    post_execute: Optional[Callable] = None,
+):
+    """
+    vscode decorator modifies a container to run a VSCode server:
+    1. Overrides the user function with a VSCode setup function.
+    2. Download vscode server and plugins from remote to local.
+    3. Launches and monitors the VSCode server.
+    4. Terminates after
```
ByronHsu, line 125: after X seconds?

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
```diff
@@ -0,0 +1,180 @@
...
+    code_server_remote_path: Optional[
+        str
+    ] = "https://github.com/coder/code-server/releases/download/v4.18.0/code-server-4.18.0-linux-amd64.tar.gz",
+    # The untarred directory name may be different from the tarball name
+    code_server_dir_name: Optional[str] = "code-server-4.18.0-linux-amd64",
+    pre_execute: Optional[Callable] = None,
+    post_execute: Optional[Callable] = None,
+):
+    """
+    vscode decorator modifies a container to run a VSCode server:
```
ByronHsu, line 121: Add this vscode behavior to PR description for reviewer to easily understand the context

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
troychiu: Do you mean os.path.join()? In this case, I think we cannot use os.path.join() because we are going to concatenate $PATH with new path by ":" (os.pathsep) instead of combining two path to one path. For example, if $PATH is /usr/bin:/local/bin and code_server_bin_dir is /tmp/bin, then the result will be /tmp/bin:/usr/bin:/local/bin.

ByronHsu: good point!

plugins/flytekit-vscode/flytekitplugins/vscode/__init__.py
```diff
@@ -0,0 +1,13 @@
+"""
+.. currentmodule:: flytekitplugins.vscode
+
+This package contains things that are useful when extending Flytekit.
```
fg91, line 4: Is this sentence from the plugin template?

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
```diff
@@ -0,0 +1,178 @@
...
+def download_file(url, target_dir="."):
+    """
+    Download a file from a given URL using fsspec.
+
+    Parameters:
+    - url (str): The URL of the file to download.
+    - target_dir (str, optional): The directory where the file should be saved. Defaults to current directory.
+
+    Returns:
+    - str: The path to the downloaded file.
```
fg91, line 44: Minor nit pick:
Google style docstrings (which this doc string appears to be the closest to) would ommit the `-` in front of the parameters and return value.

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
```diff
@@ -0,0 +1,178 @@
...
+def vscode(
+    _task_function: Optional[Callable] = None,
+    server_up_seconds: Optional[int] = DEFAULT_UP_SECONDS,
+    port: Optional[int] = 8080,
+    enable: Optional[bool] = True,
+    code_server_remote_path: Optional[
+        str
+    ] = "https://github.com/coder/code-server/releases/download/v4.18.0/code-server-4.18.0-linux-amd64.tar.gz",
+    # The untarred directory name may be different from the tarball name
+    code_server_dir_name: Optional[str] = "code-server-4.18.0-linux-amd64",
```
fg91, line 114: Minor:
These could be defined in a constants file.

fg91:
I haven't tried it yet (will do so in the next days) but the code is clear.

I think the proposal to generate a `launch.json` which already has the correct command to start the specific task in the workflow is very nice, would you be willing to include this?

---

When the user runs the task in the interactive VSCode server, how is it handled that then the vscode decorator is ignored and the task function is executed instead?

---

From what I can see, the user can currently terminate the workflow or just wait in order to stop the interactive vscode task.

As a user, I would want to have a second entry point defined in the `launch.json` which not only runs the task but also uploads the results and terminates the pod so that the workflow can then continue with the subsequent tasks.

What are your thoughts on this idea?

fg91:
> compatible with any image and any python task types (e.g. tfjob, pytorchjob, rayjob, etc).

How would this look like e.g. with a pytorch job? Wouldn't the current implementation start the vscode server in every pod in the worker group? And in a torch distributed process group with sync barriers, wouldn't one immediately cause a deadlock if "different workers different things"?

ByronHsu:
> How would this look like e.g. with a pytorch job? Wouldn't the current implementation start the vscode server in every pod in the worker group? And in a torch distributed process group with sync barriers, wouldn't one immediately cause a deadlock if "different workers different things"?

Great point. For ray job, the vscode runs on the head, and because ray submits ray tasks as protos to the worker, there is no sync issue.

However, for pytorch, tfjob, and mpijob, every worker looks at the code on its own container, if users only change on the head (master), the change does not reflect the worker, which will be an issue. To resolve this issue:

Option 1: write a command in pyflyte to sync the code to workers.
Option 2: users switch back to one single pod with multi GPUs to debug. 

I feel option 1 might be overkill because, for us, users can usually pinpoint the issue by running with a single pod many GPUs. Also, the solution for this could be included in the upcoming PR. We can add this plugin as an experimental feature first.

plugins/flytekit-vscode/flytekitplugins/vscode/__init__.py
troychiu: This sentence is in every plugin so I assume it's a kind of a template.

troychiu:
> I haven't tried it yet (will do so in the next days) but the code is clear.
> 
> I think the proposal to generate a `launch.json` which already has the correct command to start the specific task in the workflow is very nice, would you be willing to include this?
> 
> When the user runs the task in the interactive VSCode server, how is it handled that then the vscode decorator is ignored and the task function is executed instead?
> 
> From what I can see, the user can currently terminate the workflow or just wait in order to stop the interactive vscode task.
> 
> As a user, I would want to have a second entry point defined in the `launch.json` which not only runs the task but also uploads the results and terminates the pod so that the workflow can then continue with the subsequent tasks.
> 
> What are your thoughts on this idea?

Hi @fg91, thank you for all the awesome idea!
1. Generating a launch.json which allows one click feature is awesome. I am working on it and will create a separate PR to include this feature!
2. In runtime, we can get the execution state/mode from the flytekit context manager (https://github.com/flyteorg/flytekit/blob/e41ec1eecf0e1bbfccab2e075cf6807860fc51bf/flytekit/core/context_manager.py#L471). When we get local execution, we can simply return the function instead of vscode function. A separate PR will be created.
3. Do you mean when users have done the debugging, they can trigger another launch.json and we will actually run the task and consequently trigger the subsequent workflow as if there is no vscode plugin? I think it is an awesome idea that we should include :)

Thanks again for these awesome idea!

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
troychiu: Good catch. Thank you!

plugins/flytekit-vscode/flytekitplugins/vscode/decorator.py
troychiu: Thank you

fg91:
> However, for pytorch, tfjob, and mpijob, every worker looks at the code on its own container, if users only change on the head (master), the change does not reflect the worker, which will be an issue. To resolve this issue:
> Option 1: write a command in pyflyte to sync the code to workers. Option 2: users switch back to one single pod with multi GPUs to debug.

I agree that option 2 is more than enough. Let's just remove pytorch, tfjob, ... from the description/readme ..

fg91:
> 2. In runtime, we can get the execution state/mode from the flytekit context manager (https://github.com/flyteorg/flytekit/blob/e41ec1eecf0e1bbfccab2e075cf6807860fc51bf/flytekit/core/context_manager.py#L471
>    ). When we get local execution, we can simply return the function instead of vscode function. A separate PR will be created.

Makes sense 👍 

> 3. Do you mean when users have done the debugging, they can trigger another launch.json and we will actually run the task and consequently trigger the subsequent workflow as if there is no vscode plugin? I think it is an awesome idea that we should include :)

Yes, this is exactly what I meant! (Multiple entrypoints defined in a single `launch.json` instead of multiple jsons which I think doesn't work)

ByronHsu (approved):

ByronHsu:
LGTM now!

ByronHsu:
LGTM! Let's move on to the follow up works

ByronHsu (approved):

ByronHsu (dismissed):

plugins/flytekit-vscode/setup.py
```diff
@@ -0,0 +1,36 @@
...
+    install_requires=plugin_requires,
+    license="apache2",
+    python_requires=">=3.8",
+    classifiers=[
+        "Intended Audience :: Science/Research",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
pingsutw, line 28: are we able to support python 3.11?

pingsutw (commented):
Need to add vscode plugin to this GA workflow. https://github.com/flyteorg/flytekit/blob/81e56c0060d112b7bcb84a6c6ed2cb3399ec9e76/.github/workflows/pythonbuild.yml#L94

plugins/flytekit-vscode/setup.py
troychiu: Yes I think so! Thank you.

ByronHsu (approved):

pingsutw (approved):

plugins/flytekit-vscode/README.md
```diff
@@ -0,0 +1,41 @@
...
+pip install flytekitplugins-vscode
+```
+
+## Task Example
+```python
+from flytekit import task
+from flytekitplugins.vscode import vscode
+
+@task
+@vscode
```
wild-endeavor, line 17: did you experiment at all with `@task + @vscode`?

wild-endeavor:
Thanks for the contribution @troychiu and @ByronHsu.  This is awesome.
* Echo the sentiment of being able to see the original command and be able to kick it off and then quit.  But I feel like users should be able to do this two ways.
  * I think it would be nice to be able to run pyflyte itself.
  * Somehow have another py file or ipynb file that's auto generated that triggers just the task.  Maybe it can be an ipynb file where the steps of downloading the fast register tar file, untarring, downloading the inputs, loading the module anew and running the task, uploading outputs, are all different cells.  or just have all this in one py file that the user can set breakpoints on.  Kinda like an interactive entrypoint.py.

Actually question - if this is run with fast register, is the updated code already copied before vscode launches.

Just to document here - when you first started this byron i briefly mentioned it might be possible to think about similarities between vscode and fast register.  they kinda share a similar pattern.  like in both cases, you have a task, but before the task runs, you need to do something.  in the fast register case you need to download and untar a tar file.  in this case you need to download and start a vscode server.  Basically I'm wondering if there's an elegant way to think about the pre and post-processing of tasks.  so that tasks effectively look like
```
@fast
@vscode
@task
def t(): ...
```
or maybe the order is inverted.  does that make any sense?  nothing to do right now ofc.  just rambling.

plugins/flytekit-vscode/README.md
troychiu: Do you mean the end to end test of `@task + @vscode`. If yes, then I have done it. You can check the **End to end test** section of the PR description.

ByronHsu: What does that mean? Could you give an example

PR #1923, opened 2023-10-28
Use communicate in shell task

eapolinario:
# TL;DR
Use a different method to wait for child task in a ShellTask

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The current implementation of `ShellTask` might hit the deadlock described in https://docs.python.org/3/library/subprocess.html#subprocess.Popen.wait if the child process ends up writing more than the pipe buffer size (which in most Linux systems sits around 65KB, or 16 pages).

The fix is to use [Popen.communicate](https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate), which waits for the process to terminate and returns a tuple `(stdout_data, stderr_data)`. The one downside of this is that it keeps both stdout and stderr in memory. We might expose a flag to use the slower method of using temporary files to hold the child process's stdout and stderr in the future.

The test I added demonstrates the deadlock with the old code.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4313

## Follow-up issue
_NA_
OR

### DIFF
--- a/dev-requirements.in
+++ b/dev-requirements.in
@@ -7,6 +7,7 @@ mock
 pytest
 pytest-asyncio
 pytest-cov
+pytest-timeout
 mypy
 pre-commit
 codespell
--- a/flytekit/extras/tasks/shell.py
+++ b/flytekit/extras/tasks/shell.py
@@ -113,14 +113,14 @@ def _run_script(script) -> typing.Tuple[int, str, str]:
     """
     process = subprocess.Popen(script, stdout=subprocess.PIPE, stderr=subprocess.PIPE, bufsize=0, shell=True, text=True)
 
-    # print stdout so that long-running subprocess will not appear unresponsive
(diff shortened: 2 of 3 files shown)


kumare3 (dismissed):

eapolinario:
The test I added is failing on windows with this error message:
```
gen_script = '#!/bin/bash&&&&set -uex&&&&for i in $(seq 1 200000)&&do&&  echo "This is an error message" >&2&&done&&&&echo "This is the output of the program"'
```

This is the script I'm trying to run, but why are those '&' being added there?

wild-endeavor (approved):

pingsutw (approved):

PR #1925, opened 2023-10-30
Sanitize URIs for pathlib.Path constructors

mark-thm:
# TL;DR
In newer versions of fsspec, local path URIs seem to always have the `file://` protocol specified, but `pathlib.Path`'s constructor takes filesystem paths, not URIs, so flytekit is broken for the latest version of fsspec.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Replace every instance of `Path(uri)` with a wrapper function that sanitizes the `file://` prefix away before invoking the constructor.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4326

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/types/structured/basic_dfs.py
+++ b/flytekit/types/structured/basic_dfs.py
@@ -30,6 +30,9 @@ else:
 
 T = TypeVar("T")
 
+def _path_from_uri(path: str) -> Path:
+    return Path(path.removeprefix("file://"))
+
 
 def get_pandas_storage_options(
     uri: str, data_config: DataConfig, anonymous: bool = False
@@ -53,7 +56,7 @@ class PandasToCSVEncodingHandler(StructuredDatasetEncoder):
     ) -> literals.StructuredDataset:
         uri = typing.cast(str, structured_dataset.uri) or ctx.file_access.get_random_remote_directory()
         if not ctx.file_access.is_remote(uri):
-            Path(uri).mkdir(parents=True, exist_ok=True)
+            _path_from_uri(uri).mkdir(parents=True, exist_ok=True)
         path = os.path.join(uri, ".csv")
         df = typing.cast(pd.DataFrame, structured_dataset.dataframe)
         df.to_csv(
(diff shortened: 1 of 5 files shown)


kumare3:
Thank you for the PR, cc @wild-endeavor can you PTAL?

eapolinario (commented):
Mind removing the upper-bound in https://github.com/flyteorg/flytekit/blob/master/setup.py#L44 ?

mark-thm:
I'd happily take some pointers for why a seemingly unrelated test fails only on Windows.

eapolinario:
looking, @mark-thm.

mark-thm:
I rebased on top of master to see if that'll resolve things

mark-thm:
Possible to get the CI jobs permitted?

eapolinario (dismissed):

mark-thm:
@eapolinario any guidance on how to resolve the remaining failures here?

eapolinario:
@mark-thm , the two tests that are failing use `Path` directly to create directories:
- https://github.com/flyteorg/flytekit/blob/f6ae51c70f67ba9dcad2faab76d0d4a9086db614/tests/flytekit/unit/core/test_flyte_directory.py#L57
- https://github.com/flyteorg/flytekit/blob/f6ae51c70f67ba9dcad2faab76d0d4a9086db614/tests/flytekit/unit/experimental/test_eager_workflows.py#L237

I have a feeling that something changed in how fsspec handles paths with non-native separators in `fsspec==2023.10.0` that's breaking those tests in windows. Can you change how those directories are defined in the tests?

mark-thm:
The Path constructors in the test itself are fine, I think there's a deeper issue within some of Flyte's core where the fsspec inclusion of the protocol by default for local files seems to expose a different bug. I'm unfortunately not familiar enough with the Flyte internals to figure out where that's happening, but best guess is it's something with FlyteDirectory. Unfortunately I don't have a windows machine to debug on, or I'd just step through the code to figure out where things are going wrong =/

eapolinario (approved):

mark-thm:
I think this is waiting for another CI approval, when you get a chance

mark-thm:
Think I'm ready to give up here -- the original/small change fixed my issue when I patched locally but I don't understand the broader problems caused by the upgrade. It'd be great to not have tight bounds on some of these libraries as they hold back other dep upgrades.

PR #1926, opened 2023-10-30
Docker plugin

danpf:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

This is a quick pass at a docker / buildkit builder for flytekit. Basically, this plugin provides an alternative interface to ImageSpec which will allow the usage of preexisting Dockerfiles that may be too complex to recreate with envd.

ImageSpec itself seems very tailored to envd, so I'm not sure if this is the best way to go about adding this functionality.

Because of this, It will likely require a deeper discussion with the team before merging

- [ ] Need to create requirements.txt file correctly (getting weird errors)
- [ ] Test the buildkit_build_extra_output option
- [ ] Check if it Is it possible to test this further with the current infrastructure


plugins/flytekit-docker/flytekitplugins/docker/image_builder.py
```diff
@@ -0,0 +1,81 @@
...
+            raise RuntimeError("Image spec dockerfile cannot be None")
+        if image_spec.platform is None:
+            raise RuntimeError("Image spec platform cannot be None")
+        if image_spec.source_root is None:
+            raise RuntimeError("Image spec source_root cannot be None")
+        env_image_id = os.environ.get(_F_IMG_ID)
+        if env_image_id:
+            click.secho(f"Skipping nested build of {image_spec.name}", fg="blue")
+
+            def horrible_horrible_hack() -> str:
```
kumare3, line 40: what?

plugins/flytekit-docker/tests/test_files/example_pyflyte_script.py
```diff
@@ -0,0 +1,36 @@
...
+from flytekit import dynamic, task
+from flytekit.image_spec import ImageSpec
+
+PARENT_DIR = Path(__file__).parent
+DOCKERFILE_DIR = PARENT_DIR / "dockerfiles"
+
+
+biopython_img = ImageSpec(
+    name="flytekit-biopython",
+    builder="docker",
```
kumare3, line 13: interesting

kumare3 (commented):
this is pretty cool, just dont like the horrible_horrible_hack haha

danpf:
Yea I just wanted to get something that works by touching the least amount possible.  I have an idea of what needs to happen, but I wanted to get some feedback before I dumped a bunch more time into it or pushed the scope wider.

I'll probably keep plugging away at this till I'm happy with it, but any direction / thoughts would be greatly appreciated so I don't spend too much time in the wrong places.


Briefly:
- Should ImageSpec be abstracted to ImageSpecBase + ImageSpecEnvD + ImageSpecDockerfile?  
  - I'm not a huge fan of this solution, I think it would be nice to not have the plugins installed on the node
  - The current way is a bit too kitchen-drawer for me.
 
- if non-python files are going to be considered during hashing - that has to be managed delicately as it is not what flyte currently does without copy-all

- horrible_horrible_hack is because the Dockerfile isn't packaged with the script and sent to the remote. So when the remote calls .build_image and tries to determine a hash, it either crashes (folder doesn't exist) or it determines a new one and thus returns a new image_name.  This causes the builder to try and build the image again and then crash because docker doesn't exist.  To prove to myself it could be done I just used this hack but as previously mentioned I don't want it to be merged.

flytekit/remote/remote.py
```diff
@@ -700,10 +700,17 @@ def _serialize_and_register(
...
+                        # Add the image version to the workflow hash
+                        # TODO: does settings.version need to be updated?
+                        tmp_hash = cp_entity.template.container.image.split(":")[-1].split(".")[0]
+                        c_version = f"{version}{tmp_hash}"
                 ident = self.raw_register(
                     cp_entity,
                     settings=settings,
-                    version=version,
+                    version=c_version,
                     create_default_launchplan=create_default_launchplan,
```
fg91, line 714: Could you please explain why this is required?
My naive expectation after reading the PR description is that the plugin "should do the`docker build -t <some tag> .` part" that has to come before `pyflyte run --image <some tag>`. Doing this manually wouldn't require modifying the version here, right?

danpf: The version hash is built based on the python code - which doesn't include the Dockerfile if it changes.  This means that if you `pyflyte run` the same workflow twice but the container hash changes due to a change in the dockerfile, you errors similar to this:
```
RPC Failed, with Status: StatusCode.INVALID_ARGUMENT
        details: task with different structure already exists with id resource_type:TASK project:"flytesnacks" domain:"development" name:"example_pyflyte_script.t1" version:"LMxQheYT9myAmWklNHYD6A=="
        ```

File: plugins/flytekit-docker/flytekitplugins/docker/image_builder.py [hunk 4]
```diff
@@ -0,0 +1,81 @@
...
+            output_image_params = f"type=image,name={image_spec.image_name()}"
+            if image_spec.registry:
+                output_image_params = f"{output_image_params},push=true"
+            if image_spec.buildkit_build_extra_output:
+                output_image_params = f"{output_image_params},{image_spec.buildkit_build_extra_output}"
+            command += [
+                "--output",
+                output_image_params,
+            ]
+            self.execute_command(command)
```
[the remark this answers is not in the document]
fg91: Or don't users pass the tag to `ImageSpec` at all and instead its calculated with [`calculate_hash_from_image_spec`](https://github.com/danpf/flytekit/blob/44f42cfa4f656ed67a525a2611b510dab0d90981/flytekit/image_spec/image_spec.py#L169)?

tests/flytekit/unit/cli/pyflyte/test_run.py
```diff
@@ -274,9 +274,9 @@ def test_list_default_arguments(wf_path):
 )
 
 ic_result_4 = ImageConfig(
```
fg91, line 276: Question:
Are these tags calculated by [`calculate_hash_from_image_spec`](https://github.com/danpf/flytekit/blob/44f42cfa4f656ed67a525a2611b510dab0d90981/flytekit/image_spec/image_spec.py#L169)?

flytekit/image_spec/image_spec.py
```diff
@@ -25,7 +25,7 @@ class ImageSpec:
         name: name of the image.
         python_version: python version of the image. Use default python in the base image if None.
         builder: Type of plugin to build the image. Use envd by default.
-        source_root: source root of the image.
+        source_root: source root of the image, or the dockerfile context path.
```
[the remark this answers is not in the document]
fg91: You wrote:

> I'm not a huge fan of this solution, I think it would be nice to not have the plugins installed on the node
The current way is a bit too cluttered for me as it somewhat defeats the purpose of having the plugin abstraction.

I agree with it being to cluttered and defeating the purpose of having a plugin abstraction. Maybe the `register` method of the `ImageBuildEngine` should receive an accompanying `ImageSpec` class for each `ImageSpecBuilder`?
Could you please point me to where the `ImageSpec` is needed at runtime? Maybe we can avoid this by running the relevant code only when the flyte context tells us we are in registration mode?

I think @pingsutw should have the ultimate say in this.

plugins/flytekit-docker/flytekitplugins/docker/image_builder.py
```diff
@@ -0,0 +1,81 @@
...
+        if env_image_id:
+            click.secho(f"Skipping nested build of {image_spec.name}", fg="blue")
+
+            def horrible_horrible_hack() -> str:
+                return env_image_id
+
+            image_spec.image_name = horrible_horrible_hack
+            return
+
+        # Inject _F_IMG_ID to stop recursive builds
```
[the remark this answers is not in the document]
fg91: You say:

> horrible_horrible_hack is because the Dockerfile isn't packaged with the script and sent to the remote. So when the remote calls .build_image and tries to determine a hash, it either crashes (folder doesn't exist) or it determines a new one and thus returns a new image_name. This causes the builder to try and build the image again and then crash because docker doesn't exist. 

With `when the remote calls .build_image and tries to determine a hash`, do you mean that this happens in the task pod or during the registration process on the user's machine? Sorry for my confusion.

tests/flytekit/unit/cli/pyflyte/test_run.py
danpf: yes, it changed because the interface changed (dockerfile... etc)

plugins/flytekit-docker/flytekitplugins/docker/image_builder.py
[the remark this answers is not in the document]
danpf: the --tag is set via `output_image_params = f"type=image,name={image_spec.image_name()}"`
ie via: https://github.com/danpf/flytekit/blob/44f42cfa4f656ed67a525a2611b510dab0d90981/flytekit/image_spec/image_spec.py#L169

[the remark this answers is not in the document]
danpf: to be more explicit - the user does not put the tag it is generated. you could add more tags via -t, but not sure how that works with buildx export actions.

plugins/flytekit-docker/tests/test_files/dockerfiles/Dockerfile.biopython
```diff
@@ -0,0 +1,26 @@
+from ghcr.io/flyteorg/flytekit:py3.10-1.10.0
+ARG HWARG=EARTH
+RUN pip install --no-cache-dir biopython==1.81
```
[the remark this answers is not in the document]
danpf: we could use anything.   flytekit by default has a ton already installed, so I just picked something that I think would never ever be installed by default hah

flytekit/image_spec/image_spec.py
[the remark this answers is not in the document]
pingsutw: For now, source root is always set by flytekit, and it's the first ancestor from source_path that does not contain a `__init__.py` file. Do you want to set it by yourself?

flytekit/remote/remote.py
pingsutw: yeah, I think workflow and task version will become different

pingsutw: hash is calculated [here](https://github.com/flyteorg/flytekit/blob/085bc645749de66eb1deae5ccd40284b38b821e9/flytekit/remote/remote.py#L779), we can add the dockerfile to the zip file. Hence, flytekit will register a new version of workflow once you update the dockerfile.

plugins/flytekit-docker/flytekitplugins/docker/image_builder.py
```diff
@@ -0,0 +1,81 @@
...
+            if line.decode().strip() != "":
+                click.secho(line_out, fg="blue")
+
+        if p.returncode != 0:
+            _, stderr = p.communicate()
+            raise Exception(f"failed to run command {command} with error {stderr}")
+
+    def build_image(self, image_spec: ImageSpec):
+        if image_spec.dockerfile is None:
+            raise RuntimeError("Image spec dockerfile cannot be None")
```
pingsutw, line 31: Let me fix the dynamic task issue first.

flytekit/image_spec/image_spec.py
```diff
@@ -36,7 +36,11 @@ class ImageSpec:
         base_image: base image of the image.
         platform: Specify the target platforms for the build output (for example, windows/amd64 or linux/amd64,darwin/arm64
         pip_index: Specify the custom pip index url
-        registry_config: Specify the path to a JSON registry config file
+        registry_config: Specify the path to an envD JSON registry config file
+
+        dockerfile: Specify the path to a Dockerfile
+        buildkit_build_extra_output: Any custom arguments for the docker buildx build --output argument
```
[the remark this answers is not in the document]
pingsutw: btw, It's possible to use `buildkit_build_extra_output` and `build_args`, but not dockerfile. Could we raise an exception if people specify the dockerfile when they use envd plugin?

plugins/flytekit-docker/tests/test_image_spec.py
```diff
@@ -0,0 +1,51 @@
...
+
+@mock.patch("flytekitplugins.docker.image_builder.DockerfileImageSpecBuilder.execute_command")
+@mock.patch("flytekit.image_spec.image_spec.calculate_hash_from_image_spec")
+def test_image_spec(calculate_hash_from_image_spec, execute_command, test_data_dir: Path):
+    dockerfile_path = test_data_dir / "Dockerfile.biopython"
+    image_spec = ImageSpec(
+        name="test-biopython",
+        builder="docker",
+        dockerfile=str(dockerfile_path),
+        source_root=str(dockerfile_path.parent),
```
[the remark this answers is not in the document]
danpf: In this current implementation I am using `source_root` to define the docker build context.  I think this implementation is clearly confusing though.  When/if we move it to flytekit/core I think we will have separate ImageSpec/ImageSpecBuilders for each builder, and then I won't have to overload arguments like this

flytekit/remote/remote.py
danpf: a problem is that we would have to also hash the docker build context directory as well

fg91: Or we could do it [here](https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/flytekit/remote/remote.py#L837)?

```py
@staticmethod
    def _version_from_hash(
        md5_bytes: bytes,
        serialization_settings: SerializationSettings,
        *additional_context: str,
    ) -> str:
    """
    ...
    :param additional_context: This is for additional context to factor into the version computation,
          meant for objects (like Options for instance) that don't easily consistently stringify.
```

`pyflyte run` calls [`register_script`](https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/flytekit/remote/remote.py#L920) [under the hood](https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/flytekit/clis/sdk_in_container/run.py#L490), which does:

```py
if version is None:
    # The md5 version that we send to S3/GCS has to match the file contents exactly,
    # but we don't have to use it when registering with the Flyte backend.
    # For that add the hash of the compilation settings to hash of file
    version = self._version_from_hash(md5_bytes, serialization_settings)
```

If the user on the other hand [executes `pyflyte register`](https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/flytekit/clis/sdk_in_container/register.py#L174), [under the hood we do](https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/flytekit/tools/repo.py#L246):

```py
 if not version and fast:
        version = remote._version_from_hash(md5_bytes, serialization_settings, service_account, raw_data_prefix)  # noqa
        click.secho(f"Computed version is {version}", fg="yellow")
```

danpf: The tag isn't generated/parsed until the end of register_script (which uses the previously defined version to create some entities)  I guess you could do it in 2 stages & with a junk version initially, and then change the version on the resulting entities?

<details>
  <summary>build image backtrace</summary>
 
  
  ```
> /Users/dan.farrell/tmp/git/flytekit/flytekit/image_spec/image_spec.py(160)build()                                                                                                                      
-> if image_spec.builder not in cls._REGISTRY:                                                                                                                                                           
(Pdb) bt                                                                                                                                                                                                 
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/bin/pyflyte(8)<module>()                                                                                                               
-> sys.exit(main())                                                                                                                                                                                      
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(1157)__call__()                                                                             
-> return self.main(*args, **kwargs)                                                                                                                                                                     
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/rich_click/rich_group.py(21)main()                                                                        
-> rv = super().main(*args, standalone_mode=False, **kwargs)                                                                                                                                             
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(1078)main()                                                                                 
-> rv = self.invoke(ctx)                                                                                                                                                                                 
  /Users/dan.farrell/tmp/git/flytekit/flytekit/clis/sdk_in_container/utils.py(124)invoke()                                                                                                               
-> return super().invoke(ctx)                                                                                                                                                                            
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(1688)invoke()                                                                               
-> return _process_result(sub_ctx.command.invoke(sub_ctx))                                                                                                                                               
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(1688)invoke()                                                                               
-> return _process_result(sub_ctx.command.invoke(sub_ctx))                                                                                                                                               
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(1688)invoke()                                                                               
-> return _process_result(sub_ctx.command.invoke(sub_ctx))                                                                                                                                               
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(1434)invoke()                                                                               
-> return ctx.invoke(self.callback, **ctx.params)                                                                                                                                                        
  /Users/dan.farrell/.local/share/micromamba/envs/tflyte-subwf-01/lib/python3.11/site-packages/click/core.py(783)invoke()                                                                                
-> return __callback(*args, **kwargs)                                                                                                                                                                    
  /Users/dan.farrell/tmp/git/flytekit/flytekit/clis/sdk_in_container/run.py(490)_run()                                                                                                                   
-> remote_entity = remote.register_script(                                                                                                                                                               
  /Users/dan.farrell/tmp/git/flytekit/flytekit/remote/remote.py(932)register_script()                                                                                                                    
-> return self.register_task(entity, serialization_settings, version)                                                                                                                                    
  /Users/dan.farrell/tmp/git/flytekit/flytekit/remote/remote.py(735)register_task()                                                                                                                      
-> ident = self._serialize_and_register(entity=entity, settings=serialization_settings, version=version)                                                                                                 
  /Users/dan.farrell/tmp/git/flytekit/flytekit/remote/remote.py(691)_serialize_and_register()                                                                                                            
-> _ = get_serializable(m, settings=serialization_settings, entity=entity, options=options)                                                                                                              
  /Users/dan.farrell/tmp/git/flytekit/flytekit/tools/translator.py(696)get_serializable()                                                                                                                
-> cp_entity = get_serializable_task(settings, entity)                                                                                                                                                   
  /Users/dan.farrell/tmp/git/flytekit/flytekit/tools/translator.py(180)get_serializable_task()                                                                                                           
-> container = entity.get_container(settings)                                                                                                                                                            
  /Users/dan.farrell/tmp/git/flytekit/flytekit/core/python_auto_container.py(177)get_container()                                                                                                         
-> return self._get_container(settings)                                                                                                                                                                  
  /Users/dan.farrell/tmp/git/flytekit/flytekit/core/python_auto_container.py(188)_get_container()                                                                                                        
-> image=get_registerable_container_image(self.container_image, settings.image_config),                                                                                                                  
  /Users/dan.farrell/tmp/git/flytekit/flytekit/core/python_auto_container.py(257)get_registerable_container_image()
-> ImageBuildEngine.build(img)
> /Users/dan.farrell/tmp/git/flytekit/flytekit/image_spec/image_spec.py(160)build()
-> if image_spec.builder not in cls._REGISTRY:
(Pdb)
  ```
  
</details>

plugins/flytekit-docker/flytekitplugins/docker/image_builder.py
pingsutw: I fixed it here. https://github.com/flyteorg/flytekit/pull/1967

PR #1927, opened 2023-10-30
Do not cache packages in images

eapolinario:
# TL;DR
Disable caching of packages in images

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Pass the `--no-cache-dir` flag in invocation of `pip install` in dockerfiles. This will have a bigger impact in the agent image, since currently we are using about ~800MB:
```
 docker run --rm -it ghcr.io/flyteorg/flyteagent:1.10.0 du -kh --max-depth=1 /root/.cache
813M    /root/.cache/pip
813M    /root/.cache
``` 

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,11 @@ ARG DOCKER_IMAGE
 RUN apt-get update && apt-get install build-essential -y
 
 # Pod tasks should be exposed in the default image
-RUN pip install -U flytekit==$VERSION \
+RUN pip install --no-cache-dir -U flytekit==$VERSION \
 	flytekitplugins-pod==$VERSION \
 	flytekitplugins-deck-standard==$VERSION \
-	scikit-learn
+	scikit-learn \
+	&& :
 
 RUN useradd -u 1000 flytekit
 RUN chown flytekit: /root
--- a/Dockerfile.agent
+++ b/Dockerfile.agent
@@ -8,7 +8,7 @@ ARG VERSION
 RUN apt-get update && apt-get install build-essential -y
 
 RUN pip install prometheus-client
-RUN pip install -U flytekit==$VERSION \
+RUN pip install --no-cache-dir -U flytekit==$VERSION \
   flytekitplugins-bigquery==$VERSION \
   flytekitplugins-airflow==$VERSION \
   flytekitplugins-mmcloud==$VERSION \


pingsutw (approved):

Dockerfile
```diff
@@ -13,10 +13,11 @@ ARG DOCKER_IMAGE
 RUN apt-get update && apt-get install build-essential -y
 
 # Pod tasks should be exposed in the default image
-RUN pip install -U flytekit==$VERSION \
+RUN pip install --no-cache-dir -U flytekit==$VERSION \
 	flytekitplugins-pod==$VERSION \
 	flytekitplugins-deck-standard==$VERSION \
-	scikit-learn
+	scikit-learn \
+	&& :
```
jeevb, line 20: Why is this necessary?

jeevb (approved):

Dockerfile
kumare3: Great catch should not be needed

Dockerfile.agent
```diff
@@ -8,7 +8,7 @@ ARG VERSION
 RUN apt-get update && apt-get install build-essential -y
 
 RUN pip install prometheus-client
```
jeevb, line 10: Missed one. Why is this separate anyway?

jeevb:
build-essential adds an insignificant overhead. It should only be needed for building wheels, and not needed in the final image. Also apt cache is not removed.

Dockerfile
eapolinario: Adding `&& :` is a no-op. This helps in maintaining git history (if we want to remove `scikit-learn` or install another package you don't have to add another symbol to that line).

jeevb: Ah cool!

Dockerfile.agent
pingsutw: since we don't want to add `prometheus-client` to the default flytekit dependencies.

eapolinario:
> build-essential adds an insignificant overhead. It should only be needed for building wheels, and not needed in the final image. Also apt cache is not removed. We can clean this up in a subsequent PR.

we need `build-essential` for this exact reason. We build wheels for some of our dependencies, e.g. `pyspark` in the case of the agents. I also did a quick test removing the apt cache and we could shave another 20-30MB.

PR #1931, opened 2023-11-03
Sort `install_requires` dependencies

bstadlbauer:
# TL;DR
Pretty much what it says on the can. Sorted all dependencies in the `install_requires` section to make it easier to diff them with the ones in the https://github.com/conda-forge/flytekit-feedstock.
Sorting in the feedstock happened as part of https://github.com/conda-forge/flytekit-feedstock/pull/64


bstadlbauer:
Not quite sure why there are build failures (and why they only appear on some platforms). Happy to look into that in case that's not expected

setup.py
```diff
@@ -28,53 +28,52 @@
         ]
     },
     install_requires=[
-        "googleapis-common-protos>=1.57",
-        "flyteidl>=1.10.0",
-        "wheel>=0.30.0,<1.0.0",
-        "pandas>=1.0.0,<2.0.0",
-        "pyarrow>=4.0.0,<11.0.0",
+        "adlfs",
```
pingsutw, line 31: nit: Could we add a comment here? like
```suggestion
        # Please maintain an alphabetical order in the following list
        "adlfs",
```

pingsutw: btw, fixed the unit tests here. https://github.com/flyteorg/flytekit/pull/1934

bstadlbauer: Thanks for fixing the unit tests 👍 

I've added the suggested line in https://github.com/flyteorg/flytekit/pull/1931/commits/c7cbfff418530a2d7fc6f21d2bc47facdf9c512a

pingsutw: @bstadlbauer thank you. one nit: there is a lint error, could you run `make fmt`, and push it again.

pingsutw (approved):

PR #1945, opened 2023-11-09
Enable cache in shell task

pingsutw:
# TL;DR
Create task metadata and pass it to `PythonInstanceTask`.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
from flytekit import task, workflow, ImageSpec, dynamic, kwtypes
from flytekit.extras.tasks.shell import ShellTask, OutputLocation

new_flytekit = "git+https://github.com/flyteorg/flytekit.git@master"
image = ImageSpec(apt_packages=["git"], packages=[new_flytekit], registry="pingsutw")


cache_repro = ShellTask(
    name="cache_repro",
    debug=True,
    cache=True,
    cache_version="1.0",
    script=
    """
    echo "Hello World1, {inputs.msg}" > {outputs.o}
    """,
    inputs=kwtypes(msg=str),
    output_locs=[OutputLocation(var="o", var_type=FlyteFile, location='/tmp/repro.txt')],
)


def t1(a: int) -> int:
    print(a)
    return a + 2


def t2(f: FlyteFile) -> FlyteFile:
    print(f)
    return f


@workflow
def wf():
    f = cache_repro(msg="hello")
    t2(f=f)
    t1(a=3)


if __name__ == "__main__":
    wf()
```
image

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/extras/tasks/shell.py
+++ b/flytekit/extras/tasks/shell.py
@@ -7,6 +7,7 @@ import typing
 from dataclasses import dataclass
 
 import flytekit
+from flytekit import TaskMetadata
 from flytekit.core.context_manager import ExecutionParameters
 from flytekit.core.interface import Interface
 from flytekit.core.python_function_task import PythonInstanceTask
@@ -129,6 +130,9 @@ class ShellTask(PythonInstanceTask[T]):
     def __init__(
         self,
         name: str,
+        cache: bool = False,
+        cache_serialize: bool = False,
+        cache_version: str = "",
         debug: bool = False,
         script: typing.Optional[str] = None,
         script_file: typing.Optional[str] = None,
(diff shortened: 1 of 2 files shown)


flytekit/core/python_function_task.py
```diff
@@ -63,12 +64,63 @@ def __init__(
...
+            deprecated=deprecated,
+            timeout=timeout,
+        )
+        super().__init__(
+            name=name,
+            task_config=task_config,
+            task_type=task_type,
+            task_resolver=task_resolver,
+            metadata=_metadata,
+            **kwargs,
```
kumare3, line 122: we should ensure that metadata does not exist in kwargs, as that would break things. 
the thing is many folks use metadata directly! 
Another problem with this is that if we change the metadata it will break as we will have to keep adding new things.

kumare3: People are already using metadata. This will fail weirdly.

My recommendation add these fields to InstanceTask and then check for either these or metadata.
Also `metadata` field has higher precedence

PR #1949, opened 2023-11-09
Added subproc_raise public facing function for calling shell commands

pryce-turner:
# TL;DR
ShellTask works great in simple situations but falls short when the DAG or data structures get complicated.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
A lot of bioinformatics tools are based on Java/Perl/C++ and therefore require using ShellTask. Over the course of developing the variant discovery demo I started with using ShellTask almost exclusively since it's the first-class abstraction for doing that sort of thing. However, I quickly ran into how limiting it was and switched almost exclusively to the function in this PR to enable the flexibility that I need. 

For example, I'm using a list of inputs that's not necessarily known at compile time, requiring the use of dynamic workflows. ShellTask works fine in these, but the interface is so rigid that I couldn't repackage the outputs into the requisite dataclasses after processing, without using a new python task specifically for that purpose. Using a regular python task here allows me to unpack and repack dataclasses before and after executing the shell command. I also tried using ShellTasks within eager so I could operate on those outputs within the workflow context, but eager doesn't support them yet. Long story short, ShellTask works great in simple situations but falls short when the DAG or data structures get complicated. I know there's already _run_script but that's marked private/internal so I didn't think that would be appropriate.

I marked this as a draft to start a conversation around this functionality. I think this would be a good option with few drawbacks but would like to solicit feedback.

## Follow-up issue
_NA_


flytekit/extras/tasks/shell.py
```diff
@@ -33,6 +34,41 @@ class OutputLocation:
     location: typing.Union[os.PathLike, str]
 
 
+def subproc_raise(command: List[str]) -> Tuple[str, str]:
```
pingsutw, line 37: Does it mean user will call this function in their python function task?

pingsutw (commented):
nit: lint error, could you run `make lint`, and push it again

pryce-turner:
Hmm I'm a bit confused bc that lint error is in a file I didn't touch, hasn't been touched in months, and raises the same error on master... How does anyone get this to pass? 😅

pingsutw:
nvm, I found that it's not related your PR, I'll fix it, then merge your PR.

pingsutw (approved):

pingsutw (dismissed):

pingsutw (approved):

PR #1955, opened 2023-11-13
Fix remote file system (get/put)

pingsutw:
# TL;DR
Failed to serialize the flytefile due to below error:
```bash
[3/3] currentAttempt done. Last Error: SYSTEM::Traceback (most recent call last):

      File "/usr/local/lib/python3.10/site-packages/flytekit/exceptions/scopes.py", line 165, in system_entry_point
        return wrapped(*args, **kwargs)
      File "/usr/local/lib/python3.10/site-packages/flytekit/core/base_task.py", line 662, in dispatch_execute
        literals_map, native_outputs_as_map = self._output_to_literal_map(native_outputs, exec_ctx)
      File "/usr/local/lib/python3.10/site-packages/flytekit/core/base_task.py", line 561, in _output_to_literal_map
        raise TypeError(msg)

Message:

    Failed to convert outputs of task 'test.t1' at position 0:
  'list' object has no attribute 'startswith'

SYSTEM ERROR! Contact platform administrators.
```

That's because asyncFileSystem (s3fs) return [None] when calling `file_system.put` [here](https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/flytekit/core/data_persistence.py#L274)

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
from flytekit import task, ImageSpec, workflow

new_flytekit = "git+https://github.com/flyteorg/flytekit.git@c49ceb779cb4617053004c94d4bd71b0622c63b2"
image = ImageSpec(apt_packages=["git"], packages=[new_flytekit], registry="pingsutw")


@task(container_image=image)
def t1() -> FlyteFile:
    tmp_file = "/tmp/abc"
    with open(tmp_file, "w") as f:
        f.write("wow")
    return tmp_file


@workflow()
def wf():
    t1()


if __name__ == '__main__':
    wf()
```
image

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -271,7 +271,11 @@ class FileAccessProvider(object):
                     self.strip_file_header(from_path), self.strip_file_header(to_path), dirs_exist_ok=True
                 )
             from_path, to_path = self.recursive_paths(from_path, to_path)
-        return file_system.put(from_path, to_path, recursive=recursive, **kwargs)
+        dst = file_system.put(from_path, to_path, recursive=recursive, **kwargs)
+        if isinstance(dst, (str, pathlib.Path)):
+            return dst
+        else:
+            return to_path
 
     def put_raw_data(
         self,
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -248,7 +248,10 @@ class FileAccessProvider(object):
                     self.strip_file_header(from_path), self.strip_file_header(to_path), dirs_exist_ok=True
                 )
             print(f"Getting {from_path} to {to_path}")
-            return file_system.get(from_path, to_path, recursive=recursive, **kwargs)
+            dst = file_system.get(from_path, to_path, recursive=recursive, **kwargs)
+            if isinstance(dst, (str, pathlib.Path)):
+                return dst
+            return to_path
         except OSError as oe:
             logger.debug(f"Error in getting {from_path} to {to_path} rec {recursive} {oe}")
             file_system = self.get_filesystem(get_protocol(from_path), anonymous=True)
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -558,7 +558,7 @@ class PythonTask(TrackedInstance, Task, Generic[T]):
                     key = k if k != f"o{i}" else i
                     msg = f"Failed to convert outputs of task '{self.name}' at position {key}:\n  {e}"
                     logger.error(msg)
-                    raise TypeError(msg)
+                    raise TypeError(msg) from e
 
         return _literal_models.LiteralMap(literals=literals), native_outputs_as_map
 
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -558,7 +558,7 @@ class PythonTask(TrackedInstance, Task, Generic[T]):
                     key = k if k != f"o{i}" else i
                     msg = f"Failed to convert outputs of task '{self.name}' at position {key}:\n  {e}"
                     logger.error(msg)
-                    raise TypeError(msg) from e
+                    raise TypeError(msg)
 
         return _literal_models.LiteralMap(literals=literals), native_outputs_as_map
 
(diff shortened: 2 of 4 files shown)


pingsutw:
cc @kumare3 @wild-endeavor @eapolinario

flytekit/core/data_persistence.py
```diff
@@ -248,7 +248,10 @@ def get(self, from_path: str, to_path: str, recursive: bool = False, **kwargs):
                     self.strip_file_header(from_path), self.strip_file_header(to_path), dirs_exist_ok=True
                 )
             print(f"Getting {from_path} to {to_path}")
-            return file_system.get(from_path, to_path, recursive=recursive, **kwargs)
+            dst = file_system.get(from_path, to_path, recursive=recursive, **kwargs)
```
wild-endeavor, line 251: what does dst stand for?

wild-endeavor (dismissed):
is there a unit test we can add?

wild-endeavor:
@eapolinario basically with the [pr that went in](https://github.com/flyteorg/flytekit/pull/1674) we kinda added a new API to the api without being able to make it explicit.

Basically, when an fsspec filesystem now handles get/put, it has the option of returning a value, and that value is now also used by flytekit [as the uri in most places](https://github.com/flyteorg/flytekit/pull/1674/files#diff-a193385ec66e2421a616dc8e7ec6847506d0ec92fda83449ada3af02ac950d19R92).  This was useful because the flyte fs is not a real fs, so users don't actually pick the to_path that they want to write to.  instead where it was actually written is [returned](https://github.com/flyteorg/flytekit/pull/1674/files#diff-670201510934240ac68131c20c81526d05d42d326bcb335d19a08e3b5a2b647cR187) by the filesystem.

In this case however, the s3 filesystem was returning `[None]`, so the [replacement mechanism](https://github.com/flyteorg/flytekit/pull/1674/files#diff-68624e201f644896b015de1a5716eabef9b8b5a106caba1d9f21c51c0708684bR91) was failing.  (the replacement mechanism is there because in oss flyte at least, even if the user opts in to write to the flyte remote fs, you don't want to store `flyte://` as the uri.  you still want to store the underlying s3/gcs path).

flytekit/core/data_persistence.py
wild-endeavor: maybe add a test like https://github.com/flyteorg/flytekit/blob/38c76876dfe7fc2c62536ca6a195bce8a56c6270/tests/flytekit/unit/core/test_type_engine.py#L203, just making a dummy transformer, if that's easy to do.  if not, just add the below as a comment in both places.

flytekit/core/data_persistence.py
```diff
@@ -248,7 +248,10 @@ def get(self, from_path: str, to_path: str, recursive: bool = False, **kwargs):
...
+            if isinstance(dst, (str, pathlib.Path)):
```
eapolinario, line 252: IIUC, `[None]` is a guard value returned by `fsspec`, right? If that's the case, why don't we handle those explicitly?

wild-endeavor: oh really?  didn't see that.  link me?

wild-endeavor: and yes we should, but the action should still be to return the to_path.

eapolinario: I don't think it's part of the spec and I fear this is specific to `s3fs`. Doing what we're doing in the PR is probably safer (as it might apply to other fsspec-compliant implementations).

pingsutw: AsyncFileSystem.put or AsyncFileSystem.get always return a list.
- https://github.com/fsspec/filesystem_spec/blob/294248578994de67f8da00692de6872f1352ce33/fsspec/asyn.py#L574
- https://github.com/fsspec/filesystem_spec/blob/294248578994de67f8da00692de6872f1352ce33/fsspec/asyn.py#L242-L256

flytekit/core/data_persistence.py
pingsutw: yes, since fsspec uses dst as well

eapolinario (approved):

PR #1961, opened 2023-11-15
change the wait

wild-endeavor:
# TL;DR
Try a different syntax for the publish wait.

### DIFF
--- a/.github/workflows/pythonpublish.yml
+++ b/.github/workflows/pythonpublish.yml
@@ -57,10 +57,11 @@ jobs:
             echo "No tagged version found, exiting"
             exit 1
           fi
-          LINK="https://pypi.org/project/flytekitplugins-pod/${VERSION}"
+          LINK="https://pypi.org/project/flytekitplugins-pod/${VERSION}/"
           for i in {1..60}; do
-            if curl -L -I -s -f ${LINK} >/dev/null; then
-              echo "Found pypi"
+            result=$(curl -L -I -s -f ${LINK})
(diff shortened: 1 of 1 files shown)


pingsutw (approved):

.github/workflows/pythonpublish.yml
```diff
@@ -57,10 +57,11 @@ jobs:
             echo "No tagged version found, exiting"
             exit 1
           fi
-          LINK="https://pypi.org/project/flytekitplugins-pod/${VERSION}"
+          LINK="https://pypi.org/project/flytekitplugins-pod/${VERSION}/"
           for i in {1..60}; do
-            if curl -L -I -s -f ${LINK} >/dev/null; then
-              echo "Found pypi"
+            result=$(curl -L -I -s -f ${LINK})
```
eapolinario, line 62: What's the difference between these two?

eapolinario (commented):
I don't think this fixes the underlying issue? We should have a 5 minute sleep before trying this as a last resort.

.github/workflows/pythonpublish.yml
wild-endeavor: nothing - but it prints the link next.  and explicitly check $? - not sure if that's what the if was doing before.

eapolinario: This was doing this already, but this code you wrote is more clear to the uninitiated. I like it.

eapolinario (approved):

PR #1966, opened 2023-11-16
MAINT Removes GitPython dependency

thomasjpfan:
This PR removes GitPython dependency and adds more unit test to test the update function.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin
 - [x] Maintenance

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
`GitPython` was only used to read the remote url. This configuration is located in `.git/config`, which we can parse directly.

## Tracking Issue
Towards flyteorg/flyte#4418

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -6,6 +6,7 @@ but in Python object form.
 from __future__ import annotations
 
 import base64
+import configparser
 import functools
 import hashlib
 import os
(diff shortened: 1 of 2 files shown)


flytekit/remote/remote.py
```diff
@@ -147,14 +148,25 @@ def _get_git_repo_url(source_path):
     Get git repo URL from remote.origin.url
     """
     try:
-        from git import Repo
+        git_config = source_path / ".git" / "config"
+        if not git_config.exists():
+            raise ValueError(f"{source_path} is not a git repo")
```
thomasjpfan, line 153: I added tests to check these code paths. It's strange that codecov is not picking it up.

eapolinario (approved):
This is a great start. I fear that we're going to have to support more exoteric url formats (e.g. explicit ssh urls instead of the nice `git@`).

eapolinario:
just to expand on my previous comment, here are the tests for url parsing in git itself: https://github.com/git/git/blob/77bd3ea9f54f1584147b594abc04c26ca516d987/t/t0110-urlmatch-normalization.sh. Parsing urls is itself a can of worms. Having basic support for ssh + https/http covers more than the 90% of the cases. We can always expand later if necessary.

eapolinario (approved):

PR #1967, opened 2023-11-16
Do not check if image builder exists in the dynamic task

pingsutw:
# TL;DR
There is an error when using imageSpec for dynamic task.
image

The builder should only be checked if the image does not exist 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
```python
from flytekit import ImageSpec, task, dynamic, workflow

new_flytekit = "git+https://github.com/flyteorg/flytekit.git@90c7750b99d53b44dd92633fb8225b3bed51fb21"
repro_img = ImageSpec(
    name="repro",
    apt_packages=["curl", "git"],
    packages=[new_flytekit],
    registry="pingsutw",
)


@task(container_image=repro_img)
def foo():
    print("foo")


@dynamic(container_image=repro_img)
def bar():
    foo()


@workflow
def wf():
    bar()


if __name__ == '__main__':
    wf()

```

## Tracking Issue

## Follow-up issue
_NA_

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -147,13 +147,13 @@ class ImageBuildEngine:
 
     @classmethod
     def build(cls, image_spec: ImageSpec):
-        if image_spec.builder not in cls._REGISTRY:
-            raise Exception(f"Builder {image_spec.builder} is not registered.")
         img_name = image_spec.image_name()
         if img_name in cls._BUILT_IMAGES or image_spec.exist():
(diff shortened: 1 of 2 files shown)


eapolinario (commented):
Should we add that test you added in the description as a real test?

pingsutw:
will add that example to the integration test I'm working on

eapolinario (approved):

PR #1969, opened 2023-11-16
MAINT Remove natsort dependency

thomasjpfan:
This PR removes the natsort dependency.

## Type
 - [x] Maintenance

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
From reading the example, the prefix `__FLYTE_CMD_` is always followed by an integer:

https://github.com/flyteorg/flytekit/blob/28612676b3820c73fb0ae8c4f2f4457dedf16651/plugins/flytekit-aws-sagemaker/scripts/flytekit_sagemaker_runner.py#L22-L27

Thus, we can cast the string into an integer for sorting purposes.

## Tracking Issue
Towards https://github.com/flyteorg/flyte/issues/4418

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -300,8 +300,6 @@ mypy-extensions==1.0.0
     # via
     #   mypy
     #   typing-inspect
-natsort==8.4.0
-    # via flytekit
 nodeenv==1.8.0
     # via pre-commit
 numpy==1.23.5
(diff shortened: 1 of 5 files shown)


eapolinario (approved):
This integration hasn't received a lot of love in a while. I'm not even sure it still works. Having `natsort` as part of flytekit just because of this is even crazier. Thanks for fixing it! ❤️

PR #1978, opened 2023-11-20
Removes docker-image-py dependency

thomasjpfan:
This PR removes the `docker-image-py` dependency.

## Type
 - [x] Housekeeping

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Instead of using `docker-image-py`, we can use the `docker` library, to parse the tag.

## Tracking Issue
Towards https://github.com/flyteorg/flyte/issues/4418

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -111,8 +111,6 @@ distlib==0.3.7
     # via virtualenv
 docker==6.1.3
     # via flytekit
-docker-image-py==0.1.12
-    # via flytekit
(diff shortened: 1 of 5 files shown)


eapolinario (approved):
Thank you! 

`docker-image-py` is another case of a library that was covering a much larger number of use cases than its counterpart in `docker`. For our usage though, I have a feeling that `docker-image-py` is not needed.

PR #1985, opened 2023-11-21
Run serialization test individually

pingsutw:
## Tracking issue
_NA_

## Docs link
_NA_

## Describe your changes
- Add a new command `test-serialization` in the makefile. 
- Add a new GA workflow to run `test-serialization`

The reason to do this because
1. It's hard to find the specific test log in the [github action](https://github.com/flyteorg/flytekit/actions/runs/6944772256/job/18907513537?pr=1976) since we run 20000+ tests in the build test
2. It always takes 10-20 mins to run the unit test locally. `test-serialization` takes 10~15 mins to run. sometimes I want to run all the unit-tests except `test-serialization`

In test-serialization, we only test `to_flyte_idl`, `from_flyte_idl`. In most cases, it won't break when the user changes the code. Therefore, I think it's fine to run it individually.

### DIFF
--- a/.github/workflows/pythonbuild.yml
+++ b/.github/workflows/pythonbuild.yml
@@ -14,7 +14,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build:
+  unit-test:
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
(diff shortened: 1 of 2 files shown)


eapolinario (approved):
That's a good idea!

PR #1987, opened 2023-11-21
Simplify changes that require ongoing flyteidl PRs

eapolinario:
## Tracking issue


thomasjpfan (commented):
This would work out of the box if `flyteidl`'s development version was `1.10.1+develop` or any future version. For example, `flask` uses `3.0.1.dev`: https://github.com/pallets/flask/blob/d61198941adcb191ddb591f08d7d912e40bde8bc/pyproject.toml#L3

What is the reason for using `0.0.0+develop` as the development version?

eapolinario:
> This would work out of the box if `flyteidl`'s development version was `1.10.1+develop` or any future version. For example, `flask` uses `3.0.1.dev`: https://github.com/pallets/flask/blob/d61198941adcb191ddb591f08d7d912e40bde8bc/pyproject.toml#L3
>
This means we'd have to bump that future version as part of releasing flyteidl, right? Do other projects use a version far in the future (or maybe a major version) as the development version?
 
> What is the reason for using `0.0.0+develop` as the development version?
No particular reason, this is just a placeholder we use to inject the correct version at release time.

thomasjpfan:
> Do other projects use a version far in the future (or maybe a major version) as the development version?

For other projects, I usually see the minor version bumped for development.

> This means we'd have to bump that future version as part of releasing flyteidl, right?

Yes, that is correct. During the release process, the development version get bumped to the next version + `.dev`. The simplest procedure is to have the actual version in the code and committed. Then instead of injecting the version, we check that the version is consistent with the git tag.

thomasjpfan:
It looks like `flyteidl` does bug fixes on the development branch. In that case, I would have the future version be a bug fix bump, i.e. `+0.0.1`.

PR #1994, opened 2023-11-23
Introduce class decorator to flyin vscode for url exposure

ByronHsu:
## Tracking issue

## Why?

We want to serialize **flyin info** to task template, so the backend can detect and generate a URL on **flyte console**. By doing so, users don't need port-forward if they properly setup ingress.

## What?

Inspired by the [idea](https://docs.google.com/document/d/14XOGKlyahIAoPMJzSOrd4J_zNQN4Q6ouCB811V7pPsc) from @kumare3 
1. Introduce class decorator to pass config from decorator to task template. The class can be **reused by mlflow**.
2. Convert vscode from function decorator to class decorator and attach **extra config** on it
3. Name some member variables in vscode with `_` as the prefix to avoid **conflict with base_task** (`pre_execute`, `task_function`, etc). Otherwise, base_task will be overridden and cause unexpected behavior. The conflict is because we [wrap](https://github.com/flyteorg/flytekit/blob/a6f9991952f6f95a2de99cf187e0778f3abff4e8/flytekit/core/task.py#L287) the task instance with the task function to reserve function metadata.  
 
## Screenshots

conducted e2e test

```python
from flytekitplugins.flyin import vscode, VscodeConfig, DEFAULT_CODE_SERVER_EXTENSIONS

IMAGE="localhost:30000/flytekit:0.0.2"

@task(
    container_image=IMAGE,
)
@vscode
def t():
    print("123123")
    ...



# this vscode task will be killed within 10 secs
@task(
    container_image=IMAGE,
)
@vscode(
    max_idle_seconds=10,
)
def t_short_live():
    ...



# this vscode task will download default extension + vim extension
config_with_vim = VscodeConfig(
    extension_remote_paths=DEFAULT_CODE_SERVER_EXTENSIONS+["https://open-vsx.org/api/vscodevim/vim/1.27.0/file/vscodevim.vim-1.27.0.vsix"]
)

@task(
    container_image=IMAGE,
)
@vscode(
    vscode_config=config_with_vim
)
def t_vim():
    ...


@workflow
def wf():
    t()
    t_short_live()
    t_vim()

if __name__ == "__main__":
    t()
```
image


flytekit/core/tracker.py
```diff
@@ -190,7 +190,8 @@ def foo_inner():
 
     In the above example `foo_inner` is the local function or a nested function.
     """
-    return func.__code__.co_flags & inspect.CO_NESTED != 0
+
+    return hasattr(func, "__code__") and (func.__code__.co_flags & inspect.CO_NESTED != 0)
```
ByronHsu, line 194: Function decorated by class dec doesn't have `__code__`

flytekit/tools/translator.py
```diff
@@ -202,6 +203,17 @@ def get_serializable_task(
                 pod = entity.get_k8s_pod(settings)
                 entity.reset_command_fn()
 
+    entity_config = entity.get_config(settings)
+    if entity_config is None:
+        entity_config = {}
```
pingsutw, line 208: nit: 
```suggestion
    entity_config = entity.get_config(settings)
    if entity_config is None:
        entity_config = {}
```
```suggestion
    entity_config = entity.get_config(settings) or {}
```

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -167,63 +172,77 @@ def download_vscode(vscode_config: VscodeConfig):
...
+VSCODE_TYPE_VALUE = "vscode"
+
+
+class vscode(ClassDecorator):
+    def __init__(
+        self,
+        fn: Optional[Callable] = None,
+        max_idle_seconds: Optional[int] = MAX_IDLE_SECONDS,
+        port: Optional[int] = 8080,
+        enable: Optional[bool] = True,
```
pingsutw, line 184: if the default value is not None, we don't need to set it optional. https://docs.python.org/3/library/typing.html#typing.Optional

pingsutw (commented):
Overall LGTM

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
ByronHsu: done

pingsutw (approved):

PR #1998, opened 2023-11-27
[flyin] Use current context logger instead of root logger to show info level logs

ByronHsu:
## Tracking issue

## Why?
Originally, we used the root logger imported from flytekit. However, it doesn't show the info level logs. The default level is a warning.

## What?

Switch to use the logger from flytekit current_context, which is used to print logs during execution. By default, it uses [user-space logger](https://github.com/flyteorg/flytekit/blob/a6f9991952f6f95a2de99cf187e0778f3abff4e8/flytekit/core/context_manager.py#L868), whose level is [INFO](https://github.com/flyteorg/flytekit/blob/a6f9991952f6f95a2de99cf187e0778f3abff4e8/flytekit/loggers.py#L78), so it can also print logs >= INFO level.

The usage is simple. We call `logger = flytekit.current_context().logging` to get the logger.

## Test
```python
IMAGE="localhost:30000/flytekit:0.0.4"

@task(
    container_image=IMAGE,
)
@vscode
def t():
    print("123123")
    ...
```
image

### DIFF
--- a/plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
+++ b/plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
@@ -10,9 +10,7 @@ from functools import wraps
 from typing import Callable, List, Optional
 
 import fsspec
-
-from flytekit.loggers import logger
-
+import flytekit
 from .constants import (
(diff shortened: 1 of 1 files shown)


Future-Outlier:
I think it's great!

wild-endeavor (approved):

PR #1997, opened 2023-11-26
Run task first and detect local execution

troychiu:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4284

## Describe your changes
This PR introduces two enhancements to the Flyin VSCode decorator:

1. Conditional VSCode Server Launching: The VSCode server launch behavior is now context-sensitive:
- Local Execution: When a task is executed locally (using commands like python xxx.py or pyflyte run xxx), the VSCode server will not be initiated.
- Remote Execution: In contrast, when a task is run remotely (using pyflyte run --remote ...), the VSCode server will be launched.

2. New Argument - run_task_first: A new argument has been added to the decorator:

- Functionality: When set to True, this argument ensures that the user's task is executed first.
- Server Launching Condition: The VSCode server will only be launched if the user's task execution fails.

## Check all the applicable boxes 

- [X] I updated the documentation accordingly.
- [X] All new and existing tests passed.
- [X] All commits are signed-off.

PR #1976, opened 2023-11-19
feat: add jupyter notebook decorator

jasonlai1218:
# TL;DR
add jupyter notebook in `flytekit-vscode/flytekitplugins` 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
For the convenience of developers debugging:
1. `vscode`
2. `jupyter notebook`

## Tracking Issue
NA

## Follow-up issue
https://github.com/flyteorg/flyte/issues/4284

PR #2000, opened 2023-11-27
Flyin - interactive debugging

troychiu:
## Tracking issue

## Why
After users launch VSCode server and fix their failed tasks, users may want to rerun the tasks with the original inputs to make sure they work fine. In this PR, I provide a way to let users effortless rerun their tasks with original inputs. Also, it is easy to partially change the original inputs.

## What
In runtime, vscode decorator will generate a python script.
```python
# This file is auto-generated by flyin

from vscodedemo1 import t1
from flytekitplugins.flyin import get_task_inputs

if __name__ == "__main__":
    inputs = get_task_inputs(
        task_module_name="vscodedemo1",
        task_name="t1",
        context_working_dir="/tmp/flyte-67ykhnsp/sandbox/local_flytekit",
    )
    # You can modify the inputs! Ex: inputs['a'] = 5
    print(t1(**inputs))
```

And a launch.json
```json
{
    "version": "0.2.0",
    "configurations": [
        {
            "name": "Interactive Debugging",
            "type": "python",
            "request": "launch",
            "program": "/root/_flyin_interactive_debug.py",
            "console": "integratedTerminal",
            "justMyCode": true
        }
    ]
}
```

## End to end usage
With the following workflow, user might run `pyflyte run --remote demo.py wf --a 10 --b 0`. Since t1 will return a // b, there  will be an exception (devision by zero) and the vscode server will be launched because run_task_first is set to true.
(run_task_first was introduced in this pr https://github.com/flyteorg/flytekit/pull/1997. When run_task_first is set to true, the decorator would first run user's task, and only launch vscode server if the task fails)

```python
# demo.py

@task()
@vscode(run_task_first=True)
def t1(a: int, b: int) -> int:
    return a // b

@workflow
def wf(a: int, b: int) -> int:
    out = t1(a=a, b=b)
    return out
```

After connecting to vscode server, users can see _flyin_interactive_debug.py.
```python
# This file is auto-generated by flyin

from vscodedemo1 import t1
from flytekitplugins.flyin import get_task_inputs

if __name__ == "__main__":
    inputs = get_task_inputs(
        task_module_name="vscodedemo1",
        task_name="t1",
        context_working_dir="/tmp/flyte-67ykhnsp/sandbox/local_flytekit",
    )
    # You can modify the inputs! Ex: inputs['a'] = 5
    print(t1(**inputs))
```

With _flyin_interactive_debug.py, users can debug the task with original input. For example, they can handle the case of b being equal to 0.
```python
# demo.py

@task()
@vscode(run_task_first=True)
def t1(a: int, b: int) -> int:
    if b == 0:
        return 0
    return a // b

@workflow
def wf(a: int, b: int) -> int:
    out = t1(a=a, b=b)
    return out
```
Then run the _flyin_interactive_debug.py to make sure the task works fine.

With launch.json, users even can leverage VSCode's `Run and Debug` (set breakpoint...)
## Screenshots
截圖 2023-11-29 上午3 28 29
截圖 2023-11-28 上午3 12 04
截圖 2023-11-28 上午3 11 56


ByronHsu (commented):
Very impressive and solid implementation. Great work!!

ByronHsu:
Also, do we want to create launch.json as well? we can do it in the later pr. this pr already contains lots of changes

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -185,6 +220,7 @@ def vscode(
     vscode decorator modifies a container to run a VSCode server:
     1. Overrides the user function with a VSCode setup function.
     2. Download vscode server and extension from remote to local.
+    3. Prepare the interactive debugging Python script.
     3. Launches and monitors the VSCode server.
     4. Terminates if the server is idle for a set duration.
```
pingsutw, line 225:
```suggestion
    3. Prepare the interactive debugging Python script.
    4. Launches and monitors the VSCode server.
    5. Terminates if the server is idle for a set duration.
```

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/constants.py
```diff
@@ -23,3 +23,5 @@
 # The path is hardcoded by code-server
 # https://coder.com/docs/code-server/latest/FAQ#what-is-the-heartbeat-file
 HEARTBEAT_PATH = os.path.expanduser("~/.local/share/code-server/heartbeat")
+
+INTERACTIVE_DEBUGGING_FILE_NAME = "interactive_debug.py"
```
ByronHsu, line 27: maybe we should make the name a bit special so users know it is generated. how about `_generated_interactive_debug.py`

ByronHsu:
After playing with it a bit, the overall experience is awesome! I am able to change inputs modify code and reflect correctly. but it is quite unintuitive that I have to manually call `python interactive_debugging.py` to trigger the script. If we can have launch.json, I can just simply click the side bar to start the script. Can you also add launch.json generation code?

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -171,6 +172,40 @@ def download_vscode(vscode_config: VscodeConfig):
...
+    working_dir = FlyteContextManager.current_context().execution_state.working_dir
+
+    # Copy the user's Python file to the working directory.
+    shutil.copy(f"{task_function.__module__}.py", os.path.join(working_dir, f"{task_function.__module__}.py"))
+
+    # Generate a Python script
+    task_module_name, task_name = task_function.__module__, task_function.__name__
+    file_content = f"""from {task_module_name} import {task_name}
+from flytekitplugins.flyin import get_interactive_debugging_inputs
+
```
ByronHsu, line 193:
```suggestion

"""
This file is auto-generated by flyin 
"""
```

plugins/flytekit-flyin/flytekitplugins/flyin/__init__.py
```diff
@@ -11,8 +11,11 @@
    VscodeConfig
    DEFAULT_CODE_SERVER_REMOTE_PATH
    DEFAULT_CODE_SERVER_EXTENSIONS
+   jupyter
```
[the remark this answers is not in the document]
troychiu: Just found this PR https://github.com/flyteorg/flytekit/pull/1976 forgot to add it and fixed it. Should I remove it?

troychiu:
> After playing with it a bit, the overall experience is awesome! I am able to change inputs, modify code and reflect correctly. but it is quite unintuitive to manually call `python interactive_debugging.py` to trigger the script. If we can have launch.json, I can just simply click the side bar to start the script. Can you also add launch.json generation code? The json is simple and easy to generate
> 
> image

Definitely! Nice suggestion. Thank you @ByronHsu .

plugins/flytekit-flyin/flytekitplugins/flyin/utils.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Args:
+        task_module_name (str): The name of the Python module containing the task function.
+        task_name (str): The name of the task function within the module.
+        context_working_dir (str): The directory path where the input file and module file are located.
+
+    Returns:
+        dict: A dictionary containing the task inputs, converted into Python types and structures.
+    """
+    local_inputs_file = os.path.join(context_working_dir, "inputs.pb")
```
wild-endeavor, line 48: isn't there similar code in entrypoint.py?  we should reuse that as much as possible.

wild-endeavor (commented):
i think interactive debugging is a good name, clearer than local task, but have a few questions.

wild-endeavor:
Questions & thoughts:
* I don't really understand the `run_task_first=True` argument to the decorator makes it seem that the task should be run first, before it drops into vscode.  Right?  That means there should be an exception waiting for me that's been caught.  How do I print the exception?  If that's not what's happening we should rename the arg.
* It's not clear what should be happening in this case that's different than the normal operation of the vscode plugin.  Is it just the generation of the file?
* Thoughts on the `_flyin_interactive_debug`:
  * let's call it something else, without the underscore. how about `flyin_interactive_entrypoint.py`
  * I think the file should be structured so that:
    * without user changes, if the user just runs it, it proceeds normally.  That is, if we turn this feature on, go to vscode, and run the generated script without changing anything, the task should proceed as if the `@vscode` decorator was never used.  Is this possible?
    * it should mirror and use as much code from entrypoint.py as possible.
* Isn't it a bit dangerous to allow the user to modify the code and then run it?  Doesn't this break all sorts of guarantees around Flyte?  It'd be nice if there were a top level flag at the task execution layer that somehow indicates to the rest of the system that this output is non-production.

troychiu:
Hi @wild-endeavor , thank you for the great suggestions.
1. run_task_first is in this PR https://github.com/flyteorg/flytekit/pull/1997. If we set it to true, the decorator would first run the task and only launch vscode server if the task fails. https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py#L218
2. Yes this PR would not affect the decorator behavior in runtime. It only generate files that users can use to debug.
3. flyin_interactive_entrypoint.py is good!


> without user changes, if the user just runs it, it proceeds normally.

Yes it currently behaves like you mentioned since we will detect if it is a local execution https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py#L215
That is, user do not have to change anything when they use the generated file.

> it should mirror and use as much code from entrypoint.py as possible.

No problem, let me look into it.

> Isn't it a bit dangerous to allow the user to modify the code and then run it? Doesn't this break all sorts of guarantees around Flyte? It'd be nice if there were a top level flag at the task execution layer that somehow indicates to the rest of the system that this output is non-production.

I think this will be a problem but not in this PR, since this PR only generate files that users can use to debug in VSCode server. However, another feature that we are working on is resuming the workflow in VSCode. And in that one, we should handle the thing you mentioned.

plugins/flytekit-flyin/flytekitplugins/flyin/utils.py
troychiu: After looking into entrypoint.py https://github.com/flyteorg/flytekit/blob/master/flytekit/bin/entrypoint.py#L87, I found it's somehow hard to reuse the code since the behavior is a little bit different. Do you have any suggestion? Thank you.

troychiu:
Added the explanation of run_task_first to the PR description!

plugins/flytekit-flyin/flytekitplugins/flyin/__init__.py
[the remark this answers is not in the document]
ByronHsu: all good. let's keep it here

plugins/flytekit-flyin/flytekitplugins/flyin/utils.py
ByronHsu: i think it's ok to add the utils here. this usage is not found in the core flytekit

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -171,6 +173,61 @@ def download_vscode(vscode_config: VscodeConfig):
         execute_command(f"code-server --install-extension {p}")
 
 
+def prepare_interactive_python(task_function):
+    """
+    1. Copy the user's Python file to the working directory. This ensures that the inputs.pb can be loaded, even when the user switches the task interface.
```
ByronHsu, line 178:
```suggestion
    1. Copy the original task file to the context working directory. This ensures that the `inputs.pb` can be loaded because loading requires the original task interface. By doing so, even if users change the task interface in the user code, we can use the copied task file to load the inputs as native python objects.
```

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -171,6 +173,61 @@ def download_vscode(vscode_config: VscodeConfig):
...
+
+    context_working_dir = FlyteContextManager.current_context().execution_state.working_dir
+
+    # Copy the user's Python file to the working directory.
+    shutil.copy(f"{task_function.__module__}.py", os.path.join(context_working_dir, f"{task_function.__module__}.py"))
+
+    # Generate a Python script
+    task_module_name, task_name = task_function.__module__, task_function.__name__
+    python_script = f"""\"\"\"
+This file is auto-generated by flyin
```
ByronHsu, line 193: maybe change 
```
"""
"""
```

to `# ...`?
so we don't have the ugly escape char

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -171,6 +173,61 @@ def download_vscode(vscode_config: VscodeConfig):
...
+
+    # Generate a launch.json
+    launch_json = {
+        "version": "0.2.0",
+        "configurations": [
+            {
+                "name": "Interactive Debugging",
+                "type": "python",
+                "request": "launch",
+                "program": f"/root/{INTERACTIVE_DEBUGGING_FILE_NAME}",
```
ByronHsu, line 216: `/root/` cannot be hardcoded. it should be the current working dir

plugins/flytekit-flyin/flytekitplugins/flyin/utils.py
```diff
@@ -0,0 +1,56 @@
...
+        ImportError: If the module cannot be loaded from the provided path, an ImportError is raised.
+    """
+    spec = importlib.util.spec_from_file_location(module_name, path)
+    if spec is not None:
+        module = importlib.util.module_from_spec(spec)
+        sys.modules[module_name] = module
+        spec.loader.exec_module(module)
+        return module
+    else:
+        raise ImportError(f"Module at {path} could not be loaded")
```
ByronHsu, line 33:
```suggestion
        raise ImportError(f"Module {module_name} at {path} could not be loaded")
```

plugins/flytekit-flyin/flytekitplugins/flyin/vscode_lib/decorator.py
```diff
@@ -171,6 +173,61 @@ def download_vscode(vscode_config: VscodeConfig):
...
+    task_module_name, task_name = task_function.__module__, task_function.__name__
+    python_script = f"""\"\"\"
+This file is auto-generated by flyin
+\"\"\"
+
+from {task_module_name} import {task_name}
+from flytekitplugins.flyin import get_task_inputs
+
+if __name__ == "__main__":
+    inputs = get_task_inputs("{task_module_name}", "{task_name}", "{context_working_dir}")
```
ByronHsu, line 200: let's specify the argument name, so it is more clear to users

plugins/flytekit-flyin/flytekitplugins/flyin/utils.py
troychiu: this module_name is "The name you want to assign to the imported module." So actually you can assign any name to the module that you are going to import.

ByronHsu: ah i see! that makes sense then

wild-endeavor (approved):

PR #2004, opened 2023-11-28
Fix: Detection of wrong type transformer for multi-inheritance enum

fg91:
## Describe your changes

Flytekit's type engine has support for string enums but not for e.g. int enums:

```py
class TestEnum(Enum):
    A = auto()
    B = auto()

@task
def foo(e: TestEnum) -> str:
    return str(e)

@workflow
def wf() -> str:
    return foo(e=TestEnum.A)
```

This fails with:

```console
    raise TypeTransformerFailedError("Only EnumTypes with value of string are supported")
```

If one changes the definition of the enum as follows, everything works:

```py
class TestEnum(Enum):
    A = "a"
    B = "b"
```

Unfortunately, in python [it is rather common to use this syntax](https://stackoverflow.com/questions/58608361/string-based-enum-in-python/58608362#58608362):

```py
class TestEnum(str, Enum):
    A = auto()
    B = auto()
```

With this syntax, the detection of the correct type transformer unfortunately fails: When walking the inheritance hierarchy, we find a transformer that matches *the first base class* (see [here](https://github.com/flyteorg/flytekit/blob/a7596019dfe76970587cf4171abb0145b1559c9d/flytekit/core/type_engine.py#L822)).

Unfortunately, the first base class that is detected is `str` which gives us the `SimpleTransformer` defined for `str` [here](https://github.com/flyteorg/flytekit/blob/a7596019dfe76970587cf4171abb0145b1559c9d/flytekit/core/type_engine.py#L1824). The reason is that the str transformer is registered before the enum transformer.

Our example then fails [in the `SimpleTransformer`](https://github.com/flyteorg/flytekit/blob/a7596019dfe76970587cf4171abb0145b1559c9d/flytekit/core/type_engine.py#L233) with:

```console
flytekit.core.type_engine.TypeTransformerFailedError: Expected value of type <class 'str'> but got '1' of type <enum 'TestEnum'>
...
AssertionError: Error encountered while executing 'wf':
  Failed to Bind variable e for function test.foo.
```

## Note to reviewers

While this problem doesn't occur when one doesn't use multi-inheritance, unfortunately this syntax appears to be rather common in python and if users do use it, it is rather difficult for them to understand where the error comes from. It requires detailed understanding of the way the type transformer is determined for a given type.

Alternatives I considered:

* We could just register the `EnumTransformer` before the string transformer but this is a lot more implicit and easy to break in the future.
* Instead of supporting the multi-inheritance syntax, I tried catching it and raising a user oriented exception that this is not supported. However, we never enter the `EnumTransformer` where I feel this should be handled but instead immediately run into the error in the `SimpleTransformer` - where I feel this shouldn't be handled.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -919,6 +919,7 @@ class TypeEngine(typing.Generic[T]):
 
             python_type = args[0]
 
+        # Step 2
         # this makes sure that if it's a list/dict of annotated types, we hit the unwrapping code in step 2
         # see test_list_of_annotated in test_structured_dataset.py
         if (
@@ -930,7 +931,7 @@ class TypeEngine(typing.Generic[T]):
         ) and python_type in cls._REGISTRY:
             return cls._REGISTRY[python_type]
 
-        # Step 2
+        # Step 3
         if hasattr(python_type, "__origin__"):
             # Handling of annotated generics, eg:
             # Annotated[typing.List[int], 'foo']
@@ -942,7 +943,7 @@ class TypeEngine(typing.Generic[T]):
 
             raise ValueError(f"Generic Type {python_type.__origin__} not supported currently in Flytekit.")
 
-        # Step 3
+        # Step 4
         # To facilitate cases where users may specify one transformer for multiple types that all inherit from one
         # parent.
         if inspect.isclass(python_type) and issubclass(python_type, enum.Enum):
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
```diff
@@ -757,6 +757,83 @@ def guess_python_type(self, literal_type: LiteralType) -> Type[T]:
         raise ValueError(f"Transformer {self} cannot reverse {literal_type}")
 
 
+class EnumTransformer(TypeTransformer[enum.Enum]):
```
fg91, line 760: I didn't change this transformer, just had to move it up so that it can be used in TypeEngine.

flytekit/core/type_engine.py
```diff
@@ -838,6 +919,7 @@ def get_transformer(cls, python_type: Type) -> TypeTransformer[T]:
 
             python_type = args[0]
 
+        # Step 2
```
fg91, line 922: Scope creep: Referring to the algorithm description in the doc string, the numbering was off here.

eapolinario (approved):

PR #2005, opened 2023-11-28
Enable python 3.12

eapolinario:
## Describe your changes

Python 3.12 has been out for a couple months. Time to enable it in time for the next flytekit release.


eapolinario:
tensorflow does not run on 3.12 yet, being tracked in https://github.com/tensorflow/tensorflow/issues/62003

pingsutw:
Maybe we could skip those tensorflow tests first. like https://github.com/flyteorg/flytekit/blob/a34ab93532f32a5c28ff65f72a7a7212e986e77c/tests/flytekit/unit/cli/pyflyte/test_run.py#L307-L311

Dockerfile
```diff
@@ -1,5 +1,5 @@
 ARG PYTHON_VERSION
-FROM python:${PYTHON_VERSION}-slim-buster
+FROM python:${PYTHON_VERSION}-slim-bookworm
```
pingsutw, line 2: just curious, why do we change it to `bookworm`

pingsutw (approved):

Dockerfile
eapolinario: there's no `3.12-slim-buster` ([only an RC published 6 months ago](https://hub.docker.com/layers/library/python/3.12-rc-slim-buster/images/sha256-bef98ea2e06e1bc3c860b8a3ab089fe6063acb013aada5050a998cc04da368f0?context=explore)). 

Also, debian buster hit no longer getting security, as per https://www.debian.org/releases/buster/. Bookworm is the latest debian version, which the official python images use as [base image](https://hub.docker.com/layers/library/python/slim-bookworm/images/sha256-c127a8c4aca8a5d3ac3a333cbab4c082c7ddbd0891441cc4e30d88dc351f1ce5?context=explore).

pingsutw (approved):

PR #2022, opened 2023-12-05
Add various executions phases which are available in flyteidl, but not in flytekit

mdjong1:
## Tracking issue
None afaik

So that we can access all possible phases returned by Flyte within `flytekit`.

Also suggesting a refactoring to `enum_to_string()` in the [second commit](https://github.com/flyteorg/flytekit/pull/2022/commits/c3ed4e6f0d7fa7bbb3fe72dddffd909949347e7f). Happy to drop that if it's deemed unnecessary.

## How was this patch tested?

I ran `make test` locally and it ran through. I didn't add any more tests because as much as I could find everything would fall apart already if I messed up `enum_to_string()`, so expecting that that is still functional after this PR :)


pingsutw (approved):
LGTM, thank you. it becomes much simpler

PR #2024, opened 2023-12-05
Fix union transformer

pingsutw:
Failed to convert  a literal to python value`union[dataclass, dataclass]`.
```python
from typing import Union

from flytekitplugins.flyin import vscode
from flytekit import task


@dataclass
class Cat(DataClassJSONMixin):
    name: str
    age: int


@dataclass
class Yarn(DataClassJSONMixin):
    color: str
    length: int

@task
def t1(num: int) -> Union[Cat, Yarn]:
    return Cat(name="Mittens", age=num)


if __name__ == '__main__':
    print(t1(num=1))
```

When converting literal to python value, union transformer should catch any kind of exception. 
https://github.com/flyteorg/flytekit/blob/5c6802c97388c40d0431e4fc308e2ed651528952/flytekit/core/type_engine.py#L1513-L1514. Therefore, if one of transformers in the union fails, it will try next one.

## How was this patch tested?

Run above workflow locally

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1510,7 +1510,7 @@ class UnionTransformer(TypeTransformer[T]):
                         )
                     res_tag = trans.name
                     found_res = True
-            except (TypeTransformerFailedError, AttributeError) as e:
+            except Exception as e:
                 logger.debug(f"Failed to convert from {lv} to {v}", e)
 
         if found_res:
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
```diff
@@ -1455,12 +1456,15 @@ def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], exp
                 res_type = _add_tag_to_type(trans.get_literal_type(t), trans.name)
                 if found_res:
                     # Should really never happen, sanity check
```
eapolinario, line 1458: can you remove the comment?

kumare3 (approved):

eapolinario (approved):

PR #2039, opened 2023-12-11
Enable the flytekit to be pluggable

thomasjpfan:
Alternative to https://github.com/flyteorg/flytekit/pull/1941

This PR makes `pyflyte` pluggable by external libraries to adjust the CLI commands and update the remote.

This PR adds a `FlytekitPlugin` to have the pluggable CLI API be in one place.  For this initial PR, it gives the minimum API surface for configuring the API.

Implementation-wise, the external plugins are loaded through entry points and have the group name: `flytekit.configuration.plugin`.

## How was this patch tested?

- Existing unit tests were updated to into account the changes.
- The Plugin API is tested with it's own unit test.

### DIFF
--- a/flytekit/configuration/plugin.py
+++ b/flytekit/configuration/plugin.py
@@ -23,7 +23,7 @@ from click import Command
 from importlib_metadata import entry_points
 
 from flytekit.configuration import Config, get_config_file
-from flytekit.loggers import cli_logger
+from flytekit.loggers import logger
 from flytekit.remote import FlyteRemote
 
 
@@ -49,10 +49,10 @@ class FlytekitPlugin:
         cfg_file = get_config_file(config)
         if cfg_file is None:
             cfg_obj = Config.for_sandbox()
-            cli_logger.info("No config files found, creating remote with sandbox config")
+            logger.info("No config files found, creating remote with sandbox config")
         else:  # pragma: no cover
             cfg_obj = Config.auto(config)
-            cli_logger.info(f"Creating remote with config {cfg_obj}" + (f" with file {config}" if config else ""))
+            logger.info(f"Creating remote with config {cfg_obj}" + (f" with file {config}" if config else ""))
         return FlyteRemote(
             cfg_obj, default_project=project, default_domain=domain, data_upload_location=data_upload_location
         )
@@ -73,10 +73,10 @@ def _get_plugin_from_entrypoint():
 
     if len(plugins) >= 2:
         plugin_names = [p.name for p in plugins]
-        cli_logger.info(f"Multiple plugins seen for {group}: {plugin_names}")
+        logger.info(f"Multiple plugins seen for {group}: {plugin_names}")
 
     plugin_to_load = plugins[0]
-    cli_logger.info(f"Loading plugin: {plugin_to_load.name}")
+    logger.info(f"Loading plugin: {plugin_to_load.name}")
     return plugin_to_load.load()
 
 


tests/flytekit/unit/cli/pyflyte/test_plugin.py
```diff
@@ -0,0 +1,28 @@
...
+
+from flytekit.clis.sdk_in_container.plugin import get_cli_plugin, PyFlyteCLIPlugin
+
+
+@patch("flytekit.clis.sdk_in_container.plugin.entry_points")
+def test_get_plugin_default(entry_points):
+    entry_points.side_effect = lambda *args, **kwargs: []
+
+    default_plugin = get_cli_plugin()
+    assert default_plugin is PyFlyteCLIPlugin
```
cosmicBboy, line 11: can we add a test that creates a custom CLIPlugin that modifies the `main` entrypoint via `configure_pyflyte_cli`?

cosmicBboy (dismissed):

flytekit/configuration/plugin.py
```diff
@@ -0,0 +1,57 @@
...
+        cli_logger.info(f"Multiple plugins seen for flytekit.plugin: {plugin_names}")
+
+    plugin_to_load = plugins[0]
+    cli_logger.info(f"Loading plugin: {plugin_to_load.name}")
+    return plugin_to_load.load()
+
+
+# Ensure that plugin is always configured to FlytekitPlugin during pytest runs
+# Set USE_DEFAULT_FLYTEKIT_PLUGIN_FOR_TESTING=0 for testing other plugins
+if "pytest" in sys.modules and os.environ.get("USE_DEFAULT_FLYTEKIT_PLUGIN_FOR_TESTING", "1") == "1":
```
thomasjpfan, line 54: I added a `USE_DEFAULT_FLYTEKIT_PLUGIN_FOR_TESTING` here so that plugins can set `USE_DEFAULT_FLYTEKIT_PLUGIN_FOR_TESTING=0` to allow the plugin to be loaded.

thomasjpfan (commented):
I moved the `FlytekitPlugin` to `flytekit.configuration` because it is "configuration flytekit".

flytekit/configuration/plugin.py
thomasjpfan: I can see an issue with this approach when a user writes code that imports `pytest` during runtime. If they do, then plugins will not be enabled.

thomasjpfan:
@cosmicBboy The [monodocs failure in the CI](https://github.com/flyteorg/flytekit/actions/runs/7200897101/job/19615851271) is unrelated to this PR, but looks real.

thomasjpfan:
In https://github.com/flyteorg/flytekit/pull/2039/commits/9b90029b71232340e3378407abffb780596df9ca I have pytest patch the plugin variable. This way if another plugin gets loaded, the tests still run with the first-party `FlytekitPlugin`.

thomasjpfan:
I updated the solution to https://github.com/flyteorg/flytekit/pull/2039, which I am happy with. This allows pytest to configure the loaded plugin on demand.

flytekit/configuration/plugin.py
```diff
@@ -0,0 +1,74 @@
+"""Defines a plugin API allowing other libraries to modify the behavior of flytekit.
+
+Libraries can register by defining an object that follows the same API as FlytekitPlugin
+and providing entry pont with the group name "flytekit.plugin". In `setuptools`,
```
cosmicBboy, line 4: typo: entry point

flytekit/configuration/plugin.py
```diff
@@ -0,0 +1,74 @@
...
+you can specific them with:
+
+```python
+setup(entry_points={
+    "flytekit.plugin": ["my_plugin=..."]
```
cosmicBboy, line 9: is the `...` meant to be the fully qualified class name?

flytekit/configuration/plugin.py
```diff
@@ -0,0 +1,74 @@
...
+
+from click import Command
+from importlib_metadata import entry_points
+
+from flytekit.configuration import Config, get_config_file
+from flytekit.loggers import cli_logger
+from flytekit.remote import FlyteRemote
+
+
+class FlytekitPlugin:
```
cosmicBboy, line 30: I wonder if this is a good use case for [Protocols](https://docs.python.org/3/library/typing.html#typing.Protocol) so we can type-check the API of libraries that use this plugin. Maybe overkill for now

thomasjpfan: If the API surface gets bigger, a protocol makes it easier to figure out what to implement. (Although, I hope it does not get too big)

I added the protocol in https://github.com/flyteorg/flytekit/pull/2039/commits/dde46d9126aa936b9ec74ec63fa9f140c6bdba79

flytekit/configuration/plugin.py
thomasjpfan: Yea, I updated the `...` with more detail.

pingsutw (approved):
LGTM, could you resolve the merge conflict?

eapolinario (approved):

pingsutw (approved):

PR #2042, opened 2023-12-12
Flyin - fix vscode task path

troychiu:
## Tracking issue

Current implementation of vscode decorator wrongly assumes that _current working dir_ == the directory of task function module. For example, the working dir of imagespec is `/`. Therefore, when we run
```python
    shutil.copy(
        f"{task_function.__module__}.py",
        os.path.join(context_working_dir, f"{task_function.__module__}.py"),
    )
```
wf.py will not be found because it is actually at /root/wf.py.

1. Use inspect.getsourcefile(self.fn) to get the path of the source of the task function.
2. In `exit_handler`, `prepare_interactive_python`, `prepare_resume_task_python`, and `prepare_launch_json`, use this path instead of _current working dir_, which we wrongly assumed before.


ByronHsu (approved):
Awesome! LGTM

pingsutw (approved):

PR #2046, opened 2023-12-13
Flyin - VSCode default directory

troychiu:
## Tracking issue

Previously, when users accessed the VSCode server by navigating to `http://localhost:8080` in their web browser, they were not directed to a specific directory. This could be confusing, especially for those who are not familiar with VSCode, as they might not see their code files immediately.

The proposed change sets a default directory for the VSCode server. Now, when users connect to the server without setting the folder argument such as `http://localhost:8080`, VSCode server will automatically open the directory where the task function's source files are located, which is `http://localhost:8080/?folder=/root`.
截圖 2023-12-14 上午12 07 45

### DIFF
--- a/flytekit/models/literals.py
+++ b/flytekit/models/literals.py
@@ -8,10 +8,11 @@ from google.protobuf.struct_pb2 import Struct
 from flytekit.exceptions import user as _user_exceptions
 from flytekit.models import common as _common
 from flytekit.models.core import types as _core_types
-from flytekit.models.types import Error, StructuredDatasetType
+from flytekit.models.types import Error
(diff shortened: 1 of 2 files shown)


flytekit/models/literals.py
```diff
@@ -8,11 +8,10 @@
 from flytekit.exceptions import user as _user_exceptions
 from flytekit.models import common as _common
 from flytekit.models.core import types as _core_types
-from flytekit.models.types import Error
```
ByronHsu, line 11: nit: revert the change?

ByronHsu (commented):
Thanks for the quick improvement!

pingsutw (approved):

PR #2047, opened 2023-12-14
Fix: Improve error handling in workflow compilation when output binding fails

fg91:

```py
@workflow
def wf() -> tuple[str, str, str]:
    return "foo", "bar", None
```

When registering this workflow, one gets the following obvious error:

```console
~ pyflyte register test_wf.py
...
Python value cannot be None, expected <class 'str'>/<FlyteLiteral simple: STRING>
```

However, the exception does not hint which workflow/output is causing the error. In a large workflow with numerous sub workflows and dozens of tasks, it took one of our ML engineers quite some time to figure out where exactly the `return` was missing. They asked whether we could make this simpler.

I propose to catch errors raised by `binding_from_python_std` in the `PythonFunctionWorkflow`'s `compile` function just as is done e.g. [here](https://github.com/flyteorg/flytekit/blob/c559e2649f9ed0de02e9fe0151d491bb7dde664e/flytekit/core/promise.py#L1089).
This way, finding the error would have been a no-brainer.

## How was this patch tested?

For the example workflow above, the error message now is:

```console
~ pyflyte register test_wf.py
...
Failed to bind output o2 for function test_wf.wf: Python value cannot be None, expected <class 'str'>/<FlyteLiteral simple: STRING>
```

### DIFF
--- a/flytekit/core/workflow.py
+++ b/flytekit/core/workflow.py
@@ -756,7 +756,7 @@ class PythonFunctionWorkflow(WorkflowBase, ClassStorageTaskResolver):
                 )
                 bindings.append(b)
             except Exception as e:
-                raise AssertionError(f"Failed to bind output {output_names[0]} for function {self.name}.") from e
+                raise AssertionError(f"Failed to bind output {output_names[0]} for function {self.name}: {e}") from e
         elif len(output_names) > 1:
             if not isinstance(workflow_outputs, tuple):
                 raise AssertionError("The Workflow specification indicates multiple return values, received only one")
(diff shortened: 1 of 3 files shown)


flytekit/core/workflow.py
```diff
@@ -746,14 +746,17 @@ def compile(self, **kwargs):
...
+                b, _ = binding_from_python_std(
+                    ctx,
+                    output_names[0],
+                    self.interface.outputs[output_names[0]].type,
+                    workflow_outputs,
+                    t,
+                )
+                bindings.append(b)
+            except Exception as e:
+                raise AssertionError(f"Failed to bind output {output_names[0]} for function {self.name}: {e}") from e
```
kumare3, line 759: @wild-endeavor should these be AssertionErrors? or FlyteExceptions?

kumare3:
one comment else lgtm

flytekit/core/workflow.py
fg91: I used `AssertionError` in analogy to how it is done [here](https://github.com/flyteorg/flytekit/blob/c559e2649f9ed0de02e9fe0151d491bb7dde664e/flytekit/core/promise.py#L1089) but I'm happy to change to `FlyteException` if you prefer.

fg91: Let's do it like this https://github.com/flyteorg/flytekit/pull/2047/commits/50ac42fbde0f56cf55aa84251db1caf45b348ae5?

pingsutw (approved):

PR #2053, opened 2023-12-14
Logger improvements

kumare3:
## Tracking issue

With this change, flytekit now has 2 loggers, one for framework and one for user code.
Logger is default initialized based on environment. This logger is usually a json logger.

But, when a user runs any pyflyte command, the logger is switched (upgraded) to a rich logger. this can be turned off, but the logging is way more readable. the format is also adjusted

using `pyflyte -vvv` or `pyflyte -v` changes verbosity level of the log

```
 pyflyte -vv run test2.py wf --n 10
```

Screenshot 2023-12-14 at 1 03 54 PM

The PR also simplifies the logging setup. Now, everywhere in the framework, `logger` should be used. There is work to be done to make the logging less verbose and more meaningful, that will be done as a follow up

Better logging when using the CLI

Logger setup for flytekit

## How was this patch tested?
locally and remotely

### DIFF
--- a/flytekit/loggers.py
+++ b/flytekit/loggers.py
@@ -97,7 +97,7 @@ def initialize_global_loggers():
 def upgrade_to_rich_logging(
     console: typing.Optional["rich.console.Console"] = None, log_level: typing.Optional[int] = None
 ):
-    formatter = logging.Formatter(fmt="[%(name)s] %(message)s")
+    formatter = logging.Formatter(fmt="%(message)s")
     handler = logging.StreamHandler()
     if os.environ.get(LOGGING_RICH_FMT_ENV_VAR) != "0":
         try:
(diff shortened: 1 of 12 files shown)


cosmicBboy (approved):

flytekit/loggers.py
```diff
@@ -17,77 +18,129 @@
...
+    Initializes the global loggers to the default configuration.
+    """
     handler = logging.StreamHandler()
+    handler.setLevel(logging.DEBUG)
+    formatter = logging.Formatter(fmt="[%(name)s] %(message)s")
+    if os.environ.get(LOGGING_FMT_ENV_VAR, "json") == "json":
+        formatter = jsonlogger.JsonFormatter(fmt="%(asctime)s %(name)s %(levelname)s %(message)s")
+    handler.setFormatter(formatter)
 
-handler.setLevel(logging.DEBUG)
```
honnix, line 54: This is unfortunately a breaking change because we use this `handler` for further customization. This is the second time this part breaks :(

eapolinario: @honnix , sorry for the breaking change. The intent of this change was to simplify the logging infrastructure. 

Can you talk about what sort of customizations you are applying to this? We might simply expose that as an escape hatch.

honnix: No worries. We have adapted accordingly. So there are two needs:

* we need to customize the json formatter to remap a field (levelname) to satisfy stackdriver logging 
* we use the same handler to set all existing loggers (things from other libs) so we can make sure everything is json logged, not just flytekit and user space loggers

PR #2071, opened 2023-12-26
Generalize subproc exec

pryce-turner:

The previous subproc_execute was too restrictive and wouldn't allow additional kwargs to subprocess.run.

This adds the option to pass arbitrary kwargs, with sensible defaults that can be overridden. Tests were also added.

## How was this patch tested?

2 new unit tests


kumare3 (approved):

pryce-turner:
Rebasing after a merge pulled in and signed off a bunch of other commits.... closing to do this properly.

PR #2075, opened 2023-12-29
Inputs outputs wrapper

EngHabu:
## Tracking issue

Related: https://github.com/flyteorg/flyte/pull/4298

Allow flytekit to pass additional information in outputs that can be propagated to the UI/flyteadmin and used in subsequent tasks..

### DIFF
--- a/flytekit/models/literals.py
+++ b/flytekit/models/literals.py
@@ -700,6 +700,40 @@ class LiteralMap(_common.FlyteIdlEntity):
         return cls({k: Literal.from_flyte_idl(v) for k, v in pb2_object.literals.items()})
 
 
+class OutputData(_common.FlyteIdlEntity):
+    def __init__(self, outputs: LiteralMap):
+        self._outputs = outputs
+
+    @property
+    def outputs(self):
+        return self._outputs
+
+    def to_flyte_idl(self):
+        return _literals_pb2.OutputData(outputs=self.outputs.to_flyte_idl())
+
+    @classmethod
+    def from_flyte_idl(cls, pb2_object: _literals_pb2.OutputData):
+        return cls(outputs=LiteralMap.from_flyte_idl(pb2_object.outputs))
+
+
+class InputData(_common.FlyteIdlEntity):
+    def __init__(self, inputs: LiteralMap):
+        self._inputs = inputs
+
+    @property
+    def inputs(self):
+        return self._inputs
+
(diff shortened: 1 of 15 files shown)


flytekit/remote/remote.py
```diff
@@ -1892,9 +1892,12 @@ def _get_input_literal_map(self, execution_data: ExecutionDataResponse) -> liter
             with self.remote_context() as ctx:
                 tmp_name = os.path.join(ctx.file_access.local_sandbox_dir, "inputs.pb")
                 ctx.file_access.get_data(execution_data.inputs.url, tmp_name)
-                return literal_models.LiteralMap.from_flyte_idl(
-                    utils.load_proto_from_file(literals_pb2.LiteralMap, tmp_name)
-                )
+                proto_type, value = utils.load_one_proto_from_file(tmp_name, literals_pb2.InputData,
```
kumare3, line 1895: interesting that we had to only change here,
did you update the getData method in flyteadmin, if so, this will need an additional if
https://github.com/flyteorg/flytekit/blob/452a2ecde0bf4f9ddd1ed25dc8b446f1ad0e3e7d/flytekit/remote/remote.py#L273

flytekit/models/literals.py
```diff
@@ -700,6 +700,40 @@ def from_flyte_idl(cls, pb2_object):
         return cls({k: Literal.from_flyte_idl(v) for k, v in pb2_object.literals.items()})
 
 
+class OutputData(_common.FlyteIdlEntity):
```
kumare3, line 703: maybe you should add a new property called 
```
@property
def literals() -> Dict[str, Literal]:
  ...
```
This is to mimic - https://github.com/flyteorg/flytekit/blob/452a2ecde0bf4f9ddd1ed25dc8b446f1ad0e3e7d/flytekit/models/literals.py#L681

Otherwise, things like this will be missed IMO 
https://github.com/flyteorg/flytekit/blob/452a2ecde0bf4f9ddd1ed25dc8b446f1ad0e3e7d/flytekit/remote/remote.py#L1880

you have handled it fine here as you convert to literalmap

kumare3:
i think this PR needs a few rounds of testing

EngHabu:
I don't think it's complete, I agree... I only tested inputs/outputs for an execution... if you have scenarios in mind for testing, that would help a lot..

PR #2080, opened 2024-01-02
Support mashumaro DataClassORJSONMixin

quinten-flwls:

The mashumaro `DataClassORJSONMixin` class works like the mashumaro `DataClassJSONMixin` class. Additionally, it allows for serializing dataclasses with a wider variety of attributes types, including:
- [numpy](https://numpy.org/)
- [datetime](https://docs.python.org/3/library/datetime.html#datetime.datetime)
- [date](https://docs.python.org/3/library/datetime.html#datetime.date)
- [time](https://docs.python.org/3/library/datetime.html#datetime.time)
- [uuid.UUID](https://docs.python.org/3/library/uuid.html#uuid.UUID)

Currently, flytekit does not yet support the `DataClassJSONMixin` class and throws an error when classes derived from `DataClassORJSONMixin` are serialized.

This PR introduces support for classes derived from the mashumaro `DataClassORJSONMixins` class.

## How was this patch tested?

Added DataClassORJSONMixin instances in `test_type_engine.py`

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -61,7 +61,7 @@ TITLE = "title"
 try:
     from mashumaro.mixins.orjson import DataClassORJSONMixin
 
-    DataClassJSONMixin = DataClassJSONMixin | DataClassORJSONMixin
+    DataClassJSONMixin = typing.Union[DataClassJSONMixin, DataClassORJSONMixin]
 except ModuleNotFoundError:
     pass
 
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -325,11 +325,11 @@ class DataclassTransformer(TypeTransformer[object]):
 
     def __init__(self):
         super().__init__("Object-Dataclass-Transformer", object)
-        self.serializable_classes = [DataClassJSONMixin, DataClassJsonMixin]
+        self._serializable_classes = [DataClassJSONMixin, DataClassJsonMixin]
         try:
             from mashumaro.mixins.orjson import DataClassORJSONMixin
 
-            self.serializable_classes.append(DataClassORJSONMixin)
+            self._serializable_classes.append(DataClassORJSONMixin)
         except ModuleNotFoundError:
             pass
 
@@ -473,7 +473,7 @@ class DataclassTransformer(TypeTransformer[object]):
         return _type_models.LiteralType(simple=_type_models.SimpleType.STRUCT, metadata=schema, structure=ts)
 
     def is_serializable_class(self, class_: Type[T]) -> bool:
-        return any(issubclass(class_, serializable_class) for serializable_class in self.serializable_classes)
+        return any(issubclass(class_, serializable_class) for serializable_class in self._serializable_classes)
 
     def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
         if isinstance(python_val, dict):
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -58,13 +58,6 @@ T = typing.TypeVar("T")
 DEFINITIONS = "definitions"
 TITLE = "title"
 
-try:
-    from mashumaro.mixins.orjson import DataClassORJSONMixin
-
-    DataClassJSONMixin = typing.Union[DataClassJSONMixin, DataClassORJSONMixin]
-except ModuleNotFoundError:
-    pass
-
 
 class BatchSize:
     """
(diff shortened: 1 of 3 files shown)


flytekit/core/type_engine.py
```diff
@@ -24,6 +23,7 @@
 from google.protobuf.struct_pb2 import Struct
 from marshmallow_enum import EnumField, LoadDumpOptions
 from mashumaro.mixins.json import DataClassJSONMixin
+from mashumaro.mixins.orjson import DataClassORJSONMixin
```
thomasjpfan, line 26: I prefer not to add `orjson` as a hard dependency for `flytekit`. Can you write the PR in way where `orjson` is a soft dependency?

thomasjpfan (commented):
Thank you for the PR!

For the CI to run, we need to add `orjson` as a developer dependency in https://github.com/flyteorg/flytekit/blob/master/dev-requirements.in

As for testing, can you add a test that checks the expected behavior for your NumPy use case?

quinten-flwls:
> Thank you for the PR!
> 
> For the CI to run, we need to add `orjson` as a developer dependency in https://github.com/flyteorg/flytekit/blob/master/dev-requirements.in
> 
> As for testing, can you add a test that checks the expected behavior for your NumPy use case?

Hi Thomas, thank you for your feedback! I will incorporate it. As for dev dependendies, we need both orjson and mashumaro 3.11+. What is the best way for me to specify a minimal version in the dev dependencies?

Fatal1ty:
👋 Hi, author of mashumaro here.

> Mashumaro 3.11 introduced the DataClassORJSONMixin class

> As for dev dependendies, we need both orjson and mashumaro>=3.11. What is the best way for me to specify a minimal version in the dev dependencies?

To avoid all possible unwanted changes made on false conclusions, I’d like to make a correction. `DataClassORJSONMixin` was added earlier in 3.1 (see [release notes](https://github.com/Fatal1ty/mashumaro/releases/tag/v3.1)). I hope it’s not that important in case of this pull request and I apologize for bothering you.

flytekit/core/type_engine.py
quinten-flwls: Great idea, orjson is a soft dependency now

quinten-flwls:
> 👋 Hi, author of mashumaro here.
> 
> > Mashumaro 3.11 introduced the DataClassORJSONMixin class
> 
> > As for dev dependendies, we need both orjson and mashumaro>=3.11. What is the best way for me to specify a minimal version in the dev dependencies?
> 
> To avoid all possible unwanted changes made on false conclusions, I’d like to make a correction. `DataClassORJSONMixin` was added earlier in 3.1 (see [release notes](https://github.com/Fatal1ty/mashumaro/releases/tag/v3.1)). I hope it’s not that important in case of this pull request and I apologize for bothering you.

Hi @Fatal1ty! Thank you for reaching out and reminding me that `DataClassORJSONMixin` does not require mashumaro 3.11. That is encredibly useful information indeed. I updated the PR description.

The reason why I had mashumaro 3.11 in mind is because we internally use `json_encode` and `json_decode` to process `DataClassORJSONMixin` instances. But that is completely unrelated to this PR.

thomasjpfan (commented):
Thanks for the update!

> wider variety of attributes, including numpy arrays.

Does your NumPy array use case work out of the box after this PR? If so, can you add a test where a NumPy array is an attribute?

quinten-flwls:
> Thanks for the update!
> 
> > wider variety of attributes, including numpy arrays.
> 
> Does your NumPy array use case work out of the box after this PR? If so, can you add a test where a NumPy array is an attribute?

Thank you for reviewing! Unfortunately, Numpy array attributes do not work out of the box. However, there are other attribute types that do. I added a datetime attribute as an example in the tests. I hope that suffices to demonstrate the use case. Let me know if it does not!

flytekit/core/type_engine.py
```diff
@@ -325,6 +325,13 @@ class Test(DataClassJsonMixin):
 
     def __init__(self):
         super().__init__("Object-Dataclass-Transformer", object)
+        self.serializable_classes = [DataClassJSONMixin, DataClassJsonMixin]
```
thomasjpfan, line 328: Can we make this attribute private by adding an underscore? 
(It's "private" by convention)

```suggestion
        self._serializable_classes = [DataClassJSONMixin, DataClassJsonMixin]
```

And update it in all other places.

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -2464,6 +2497,15 @@ class Datum(DataClassJSONMixin):
...
     assert datum_mashumaro.y.value == pv.y
 
+    lt = TypeEngine.to_literal_type(DatumMashumaroORJSON)
+    now = datetime.datetime.now()
+    datum_mashumaro_orjson = DatumMashumaroORJSON(5, Color.RED, now)
+    lv = transformer.to_literal(ctx, datum_mashumaro_orjson, DatumMashumaroORJSON, lt)
+    gt = transformer.guess_python_type(lt)
+    pv = transformer.to_python_value(ctx, lv, expected_python_type=gt)
+    assert datum_mashumaro_orjson.x == pv.x
+    assert datum_mashumaro_orjson.y.value == pv.y
```
thomasjpfan, line 2507: Can we also check the value for the new datetime attribute?

```suggestion
    assert datum_mashumaro_orjson.y.value == pv.y
    assert datum_mashumaro_orjson.z.isoformat() == pv.z
```

thomasjpfan (commented):
Minor nits, but otherwise LGTM

flytekit/core/type_engine.py
eapolinario: done.

eapolinario (approved):

quinten-flwls:
Thank you for the feedback and the help!

PR #2087, opened 2024-01-03
Constraint the full fsspec family

honnix:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/4665

Described in https://github.com/flyteorg/flyte/issues/4665

Use the same constraint for all packages from fsspec family

## How was this patch tested?

Through CI probably.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ readme = { file = "README.md", content-type = "text/markdown" }
 requires-python = ">=3.8,<3.13"
 dependencies = [
     # Please maintain an alphabetical order in the following list
-    "adlfs",
+    "adlfs>=2023.3.0,<=2023.9.2",
     "click>=6.6,<9.0",
     "cloudpickle>=2.0.0",
     "cookiecutter>=1.7.3",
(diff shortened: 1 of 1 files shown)


pingsutw (approved):

mark-thm:
@honnix curious if you have a sense of what it'd take to increase the upper range here? I tried (and failed) to bump it for flyte but couldn't decipher the windows errors.

honnix:
@mark-thm I think you will need to bump all those packages to a matching version. They are usually matched on "20xx.x" level but there seems to be exceptions.

Tom-Newton:
I would be very interested in bumping the upper range. There is actually no version of `adlfs` in the range which doesn't suffer from [this bug](https://github.com/fsspec/adlfs/issues/389). The latest working version is `2022.11.2` and its still broken on `main`. https://github.com/fsspec/adlfs/pull/454 should fix it. 

It seems to me like its largely the user's responsibility to ensure that fsspec packages are aligned. I don't see how it can be done on a flyte level without restricting to exactly one version.

honnix:
Yeah, the fsspec package versions are tricky.

Tom-Newton:
Personally I think ai would be in favour of removing all the upper limits, so that users have control. 

I'm definitely going to want to use a new version of adlfs once https://github.com/fsspec/adlfs/issues/389 gets resolved.

Tom-Newton:
I just discovered that `adlfs` puts a constraint on the version of `fsspec`. https://github.com/fsspec/adlfs/blob/18ea349d8a286f8bbeb8055127a9d58fe4cde0d9/pyproject.toml#L30
I think this makes sense to ensure that fsspec and adlfs are kept in sync, without forcing users to use one specific version that happens to be broken.

Tom-Newton:
It looks like https://github.com/flyteorg/flytekit/pull/2143 has addressed this problem

PR #2106, opened 2024-01-14
Option to disable local caching

bstadlbauer:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/4395

While local caching is useful in some cases, it has often led to unexpected errors for users not being aware. Especially in large codebases this can lead to unexpected test failures, etc.

This PR introduces a new config section `local` with one value `cachen_enabled` (which defaults to true):
```
[local]
cache_enabled = True
```

## How was this patch tested?
Added a test which confirms that setting `FLYTE_LOCAL_CACHE_ENABLED=false` prevents the local use of the cache.

## Docs link
Changed in https://github.com/flyteorg/flytesnacks/pull/1425

https://flyte--1425.org.readthedocs.build/projects/cookbook/en/1425/auto_examples/development_lifecycle/task_cache.html#how-does-local-caching-work

### DIFF
--- a/flytekit/configuration/internal.py
+++ b/flytekit/configuration/internal.py
@@ -66,7 +66,7 @@ class AZURE(object):
     CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "client_secret"))
 
 
-class LOCAL(object):
+class Local(object):
     SECTION = "local"
     CACHE_ENABLED = ConfigEntry(LegacyConfigEntry(SECTION, "cache_enabled", bool))
 
--- a/flytekit/configuration/internal.py
+++ b/flytekit/configuration/internal.py
@@ -66,6 +66,11 @@ class AZURE(object):
     CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "client_secret"))
 
 
+class LOCAL(object):
+    SECTION = "local"
+    CACHE_ENABLED = ConfigEntry(LegacyConfigEntry(SECTION, "cache_enabled", bool))
+
+
 class Credentials(object):
     SECTION = "credentials"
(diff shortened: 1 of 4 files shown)


flytekit/configuration/internal.py
```diff
@@ -66,6 +66,11 @@ class AZURE(object):
     CLIENT_SECRET = ConfigEntry(LegacyConfigEntry(SECTION, "client_secret"))
 
 
+class LOCAL(object):
```
kumare3, line 69: Local

kumare3 (commented):
Ideally we should also make pyflyte run --(option for invalidating cache) work in local mode

flytekit/configuration/internal.py
bstadlbauer: Thanks, fixed in https://github.com/flyteorg/flytekit/pull/2106/commits/836398ab5382056bd7436857f48b4e63ffe2af9a

bstadlbauer:
@kumare3 
> Ideally we should also make pyflyte run --(option for invalidating cache) work in local mode

That sounds reasonable. I've poked around a little and found `pyflyte run --overwrite-cache`. Given that that's a slightly different usecase (cache overwrite/invalidation vs. cache disabling) would you want me to open a different PR? Also do you think it would ok to do a full `LocalTaskCache.clear()`? Or should we invalidate per-task (which might need another config or some other plumbing)

kumare3:
Aah it should be invalidate per cache key only

bstadlbauer:
@kumare3 I've opened https://github.com/flyteorg/flyte/issues/4727 to track the `pyflyte run --overwrite-cache` behavior (as that's likely more work). Could we still merge this PR before?

fg91:
(Dropping a message to be notified once this is merged <3)

pingsutw (approved):

PR #2110, opened 2024-01-17
try removing multiple imports to resolve fast-mode issue

novahow:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4716_

Please refer to may example in https://github.com/flyteorg/flyte/issues/4716#issuecomment-1893423240

When we have a python file that contains workflows, tasks, and launch plans and try to run that in fast mode, we get an error:
``AssertionError: The cached values aren't the same as the current call arguments``

This happens because when executing the file, the launch_plan's ``get_or_create`` method is invoked by [the following snippet](https://github.com/flyteorg/flytekit/blob/80ca660e1c22bf014b4cf7335fd362360436690f/flytekit/core/tracker.py#L75) multiple times, thus the ``LaunchPlan.CACHE`` stores a workflow. After that when we're really creating a plan with ``standard_scale_launch_plan = LaunchPlan.get_or_create(``, the workflow of the plan is different from the workflow of the cached plan, which I haven't figured out the reason. 

My workaround is removing ``exec_module`` in the try block and ``import_module`` in the exception block so that ``get_or_create`` won't be invoked multiple times


pingsutw (commented):
cc @cosmicBboy mind taking a look

cosmicBboy (approved):

pingsutw (approved):

PR #2113, opened 2024-01-18
[Bug] Map task caching failures

pvditt:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4731_

[Legacy map tasks](https://github.com/flyteorg/flytekit/blob/7996c2ebb6892341d31e6efbd7c7f0f8ca2c9b34/flytekit/core/map_task.py#L95) and [new map tasks](https://github.com/flyteorg/flytekit/blob/7996c2ebb6892341d31e6efbd7c7f0f8ca2c9b34/flytekit/core/array_node_map_task.py#L79) include the collection interface as part of task names to support partially binding different parameters of the same mappable task. This introduces issues when collection interface includes types that produces non-deterministic outputs such as guaranteed cache misses. 

Example where the output contains a memory address causing for guaranteed cache misses:
```
@task()
def process_data(index: int) -> Annotated[pd.DataFrame, Annotated[pd.DataFrame, HashMethod(hash_pandas_dataframe)]]:
    ...
```

The same can be done for inputs.

Utilize an ordered list of the bounded input names in favor of the entire interface. This is similar to [this](https://github.com/flyteorg/flytekit/blob/7996c2ebb6892341d31e6efbd7c7f0f8ca2c9b34/flytekit/core/interface.py#L278) but disregards any Typed values, opting to only keep the keys/names.

Note, this will cause for all cache enabled map tasks to have cache misses on first runs as task names will all change.

### Setup process

Confirmed that intputs/outputs w/ memory address don't impact caching
```
def hash_pandas_dataframe(df: pd.DataFrame) -> str:
    return str(pd.util.hash_pandas_object(df))


@task()
def mem_output(index: int) -> Annotated[pd.DataFrame, HashMethod(hash_pandas_dataframe)]:
    target = [index] * 3
    df = pd.DataFrame({"target": target})
    return df


@task()
def mem_input(df: Annotated[pd.DataFrame, HashMethod(hash_pandas_dataframe)]) -> int:
    return df["target"].tolist()[0]


@workflow
def caching_df(
    l: List[int] = [1, 2, 3, 4],
) -> None:
    from flytekit.experimental import map_task
    partial_task = functools.partial(mem_output)
    outputs = map_task(partial_task, metadata=TaskMetadata(cache=True, cache_version="2"))(index=l)
    partial_task2 = functools.partial(mem_input)
    map_task(partial_task2, metadata=TaskMetadata(cache=True, cache_version="2"))(df=outputs)
```

Confirmed if partially binding different parameters for the same mapped task still worked:
```
@task
def square_add(i: int, j: int) -> List[int]:
   return [i*i + j]


@workflow
def wf(xs: List[int] = [1, 1, 1], ys: List[int] = [2, 2, 2]):
   square_fixed = functools.partial(square_add, j=10)
   add_fixed = functools.partial(square_add, i=2)
   map_task(square_fixed, metadata=TaskMetadata(cache=True, cache_version="1.0.0"))(i=xs)
   map_task(add_fixed, metadata=TaskMetadata(cache=True, cache_version="1.0.0"))(j=ys)
```

### Screenshots
Screenshot 2024-01-17 at 5 57 40 PM

### DIFF
--- a/flytekit/core/array_node_map_task.py
+++ b/flytekit/core/array_node_map_task.py
@@ -77,8 +77,9 @@ class ArrayNodeMapTask(PythonTask):
             f = actual_task.lhs
         else:
             _, mod, f, _ = tracker.extract_task_module(cast(PythonFunctionTask, actual_task).task_function)
+        sorted_bounded_inputs = ','.join(sorted(self._bound_inputs))
         h = hashlib.md5(
-            f"{collection_interface.__str__()}{concurrency}{min_successes}{min_success_ratio}".encode("utf-8")
+            f"{sorted_bounded_inputs}{concurrency}{min_successes}{min_success_ratio}".encode("utf-8")
         ).hexdigest()
         self._name = f"{mod}.map_{f}_{h}-arraynode"
 
--- a/flytekit/core/map_task.py
+++ b/flytekit/core/map_task.py
@@ -92,7 +92,9 @@ class MapPythonTask(PythonTask):
             f = actual_task.lhs
         else:
             _, mod, f, _ = tracker.extract_task_module(typing.cast(PythonFunctionTask, actual_task).task_function)
-        h = hashlib.md5(collection_interface.__str__().encode("utf-8")).hexdigest()
+
+        sorted_bounded_inputs = ','.join(sorted(self._bound_inputs))
+        h = hashlib.md5(sorted_bounded_inputs.encode("utf-8")).hexdigest()
         name = f"{mod}.map_{f}_{h}"
 
         self._cmd_prefix: typing.Optional[typing.List[str]] = None
--- a/tests/flytekit/unit/core/test_array_node_map_task.py
+++ b/tests/flytekit/unit/core/test_array_node_map_task.py
@@ -187,7 +187,7 @@ def test_inputs_outputs_length():
     assert m.python_interface.inputs == {"a": List[int], "b": List[str], "c": List[float]}
     assert (
         m.name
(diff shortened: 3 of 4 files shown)


wild-endeavor (commented):
can we also update the args also actually?

Another issue is that when flytekit creates the command: https://github.com/flyteorg/flytekit/blob/6279b817d5df76e651c866c6c5ba7e12a65e12b2/flytekit/core/array_node_map_task.py#L390

the bound vars can show up in different orders.  Can we order these first to make sure the args are always the same?

Same for the old map task: https://github.com/flyteorg/flytekit/blob/6279b817d5df76e651c866c6c5ba7e12a65e12b2/flytekit/core/map_task.py#L407

wild-endeavor:
if you sort then this should resolve https://github.com/flyteorg/flyte/issues/4702 as well @pvditt

pvditt:
@wild-endeavor thanks for pointing that out. Will update this PR

flytekit/core/array_node_map_task.py
```diff
@@ -77,8 +77,9 @@ def __init__(
             f = actual_task.lhs
         else:
             _, mod, f, _ = tracker.extract_task_module(cast(PythonFunctionTask, actual_task).task_function)
+        sorted_bounded_inputs = ",".join(sorted(self._bound_inputs))
```
pvditt, line 80: note: could also just turn self._bound_inputs into a sorted list as opposed to set - don't feel too strongly either way

pingsutw (approved):

PR #2115, opened 2024-01-18
[Core Feature] Notification System

Future-Outlier:
## Tracking issue

## How was this patch tested?
```python
from flytekit import Secret, task, workflow
from flytekitplugins.slack import SlackNotifier, SlackConfig
from flytekitplugins.flyin import vscode, VscodeConfig

slack_notifier = SlackNotifier(slack_conf=SlackConfig(
                        channel="demo",
                        secret_group="slack-api",
                        secret_key="token",
                    ))

@task(
    container_image="localhost:30000/flyin-notification:1655",
    secret_requests=[Secret(key="token", group="sendgrid-api"),
                    Secret(key="token", group="slack-api"),   ],
)
@vscode(
    config=VscodeConfig(),
    max_idle_seconds=80,
    warning_seconds_before_termination=60,
    notifier=slack_notifier,
)
def t(a: int, b: int) -> int:
    return a + b

@workflow
def wf(a: int = 5, b: int = 3) -> int:
    out = t(a=a, b=b)
    return out
```

### DIFF
--- a/plugins/flytekit-bigquery/setup.py
+++ b/plugins/flytekit-bigquery/setup.py
@@ -13,7 +13,7 @@ setup(
     version=__version__,
     author="flyteorg",
     author_email="admin@flyte.org",
-    description="This package holds the Slack plugins for flytekit",
+    description="This package holds the Bigquery plugins for flytekit",
     namespace_packages=["flytekitplugins"],
     packages=[f"flytekitplugins.{PLUGIN_NAME}"],
     install_requires=plugin_requires,
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -22,3 +22,9 @@ repos:
     rev: 0.0.9
     hooks:
       - id: check_pdb_hook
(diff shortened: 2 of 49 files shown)


plugins/flytekit-bigquery/setup.py
```diff
@@ -13,7 +13,7 @@
     version=__version__,
     author="flyteorg",
     author_email="admin@flyte.org",
-    description="This package holds the Bigquery plugins for flytekit",
+    description="This package holds the Slack plugins for flytekit",
```
pingsutw, line 16:
```suggestion
    description="This package holds the Bigquery plugins for flytekit",
```

kumare3:
I prefer us building a `@task(notification=Notification(...))` instead of this one off.
Reason is, we can then use the same notification engine we have to send these.

Future-Outlier:
> I prefer us building a `@task(notification=Notification(...))` instead of this one off. Reason is, we can then use the same notification engine we have to send these.

Thank you for the review, I will discuss with @pingsutw next week!

PR #2118, opened 2024-01-19
Fix type hint

pingsutw:
## Tracking issue
Before:
image

After:
image

Get warning in IDE

Update type hint.

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -310,7 +310,7 @@ class Task(object):
         vals = [Promise(var, outputs_literals[var]) for var in output_names]
         return create_task_output(vals, self.python_interface)
 
-    def __call__(self, *args: object, **kwargs: object) -> Union[Tuple[Promise], Promise, VoidPromise, Tuple, None]:
+    def __call__(self, *args, **kwargs) -> Any:
         return flyte_entity_call_handler(self, *args, **kwargs)  # type: ignore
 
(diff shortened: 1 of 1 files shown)


ringohoffman:
What was the error you were seeing?

pingsutw:
image

Future-Outlier (approved):

eapolinario (changes requested):
We shouldn't do this. This effectively removes typing information from tasks.

pingsutw:
> We shouldn't do this. This effectively removes typing information from tasks.

is there any way to fix it? current typing information is wrong.

ringohoffman:
Seems like it got its return type from `flyte_entity_call_handler`:

https://github.com/flyteorg/flytekit/blob/cba830ea0c9844f8079342ae99a2105b507eca2e/flytekit/core/promise.py#L1132-L1134

also there is a `type: ignore` comment on the return of `Task.__call__`. Maybe we can start with making sure this signature is as expected.

It seems like you have another workflow called `train` you are saying returns a `str`, but `Task.__call__` never references the user's declared return type. On which of the paths does it return the result of the underlying function?

pingsutw:
@ringohoffman [here](https://github.com/flyteorg/flytekit/blob/aaf69f825625d67ce3dc2b5a9cbc944dc9b87131/flytekit/core/promise.py#L1200)

PR #2119, opened 2024-01-19
Extends ImageSpec to accept image names from plugin and have priority for plugins

thomasjpfan:

This PR:

- Extends the image spec plugin to allow `ImageSpecBuilder.build_image` to return an image name. This allows image spec plugins to point to images in it's own registry.
- Changes the default builder to `None` and introduce a priority. This allows the default builder to be the one with the higher priority.

The changes are backward compatible since returning an image name in `ImageSpecBuilder.build_image` is optional. If an plugin returns a name, it will be used by `ImageSpec.image_name()` when referencing the same image.

## How was this patch tested?

I added unit tests to check the expected behavior.

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -7,8 +7,7 @@ import typing
 from abc import abstractmethod
 from dataclasses import asdict, dataclass
 from functools import lru_cache
-from heapq import heappush
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Optional, Tuple, Union
 
 import click
 import requests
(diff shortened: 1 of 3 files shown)


flytekit/image_spec/image_spec.py
```diff
@@ -44,7 +45,7 @@ class ImageSpec:
 
     name: str = "flytekit"
     python_version: str = None  # Use default python in the base image if None.
-    builder: str = "envd"
+    builder: Optional[str] = None
```
eapolinario, line 48: won't this change force every existing client to specify "envd" in their uses of `ImageSpec`?

thomasjpfan: If `"envd"` is not set and `flytekitplugins-envd` is installed, then `envd` is still used. As for testing, there is an existing test that does not have `builder` set and it still runs through `envd` :

https://github.com/flyteorg/flytekit/blob/139fa2fa1589bdc93ad1588d3b3e15fd8f55441a/plugins/flytekit-envd/tests/test_image_spec.py#L10-L16

With this PR, the default can switch when another image spec plugin is installed and it gives itself a higher priority. If a user still wants `envd` to be the default, they can reregister `envd` and give it a higher priority:

```python
from flytekitplugins.envd import EnvdImageSpecBuilder
from flytekit.image_spec.image_spec import ImageBuildEngine

ImageBuildEngine.register("envd", EnvdImageSpecBuilder(), priority=20)
```

cosmicBboy (approved):
LGTM. Going to defer to @eapolinario @pingsutw or another Flyte maintainer for second approval before merging.

pingsutw (approved):

eapolinario (approved):

PR #2120, opened 2024-01-19
Pass in env vars in FlyteRemote register_script

knordstrom-muon:

The `FlyteRemote.register_script` function is useful for registering local code to remote during integration tests of workflows. It does not allow the user to pass in environment variables to the serialization though, even though it should be easy to support this.

Make it possible to pass in environment variables using a `dict[str,str]` which can be passed directly to `SerializationSettings`.

## How was this patch tested?

Don't think `register_script` is covered by unit tests at the moment.

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -907,6 +907,7 @@ class FlyteRemote(object):
         :param options: Additional execution options that can be configured for the default launchplan
         :param source_path: The root of the project path
         :param module_name: the name of the module
+        :param envs: Environment variables to be passed to the serialization
         :return:
         """
         if image_config is None:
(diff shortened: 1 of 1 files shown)


pingsutw (approved):

nineunderground:
@knordstrom-muon Sorry, just a quick validation from myself, but this is an implementation only valid for "pyflyte register", but not using "flytectl register". Am I right?
I just checked with latest 1.10.3 and it it is like that to me. With that being said, and If I am correct, is it possible to replicate for flytectl tool?

knordstrom-muon:
Hi @nineunderground -- I am not using this with `pyflyte` or `flytectl` but with `flytekit.remote` directly (so in python code). To achieve the same thing with the CLIs, I have in the past used:

```
pyflyte serialize ... --env ENV_VAR=TEST ...
flytectl register files ...
```

So I think in your case, you could achieve it with `flytectl register` but you would still need `pyflyte` to serialize first.

PR #2122, opened 2024-01-20
Check envd version before building images

pingsutw:
## Tracking issue
NA

This [PR](https://github.com/flyteorg/flytekit/pull/2108) changes the default `destination-dir` to `.`, so flytekit will download the code to the `workdir` specified in the Dockerfile. However, envd<0.3.39 replace the `workdir` to /, and flytekit may not have permission to download a file to `/`. Therefore, we should force people to use envd>0.3.39 to build the image.

image

Check `envd` version before building images. Raise an error if it is lower than 0.3.39.

## How was this patch tested?

```python

image_spec = ImageSpec(
    base_image="ghcr.io/flyteorg/flytekit:py3.11-1.10.2",
    packages=["flytekit>=1.6.0", "numpy", "mypy"],
    registry="pingsutw",
    python_version="3.11"
)


@task(container_image=image_spec, cache_version="1.0", cache=True)
def get_data() -> str:
    return "hello dgx.."


@task(
    container_image=dgx_image_spec
)
def train(data: str) -> str:
    print(data)

    return f"gpu is available: 123"


@workflow
def wf() -> str:
    data = get_data()
    return train(data=data)
```

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -7,10 +7,14 @@ import typing
 from abc import abstractmethod
 from dataclasses import asdict, dataclass
 from functools import lru_cache
+from importlib import metadata
 from typing import List, Optional, Union
 
 import click
 import requests
+from packaging.version import Version
+
+from flytekit.exceptions.user import FlyteAssertion
 
 DOCKER_HUB = "docker.io"
 _F_IMG_ID = "_F_IMG_ID"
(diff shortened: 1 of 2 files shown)


Future-Outlier (approved):
Test it by using the code in the PR.
image

flytekit/image_spec/image_spec.py
```diff
@@ -206,6 +210,16 @@ def build(cls, image_spec: ImageSpec):
             click.secho(f"Image {img_name} not found. Building...", fg="blue")
             if image_spec.builder not in cls._REGISTRY:
                 raise Exception(f"Builder {image_spec.builder} is not registered.")
+            if image_spec.builder == "envd":
+                envd_version = metadata.version("envd")
+                # flytekit v1.10.2+ copies the workflow code to the WorkDir specified in the Dockerfile. However, envd<0.3.39
+                # overwrites the WorkDir when building the image, resulting in a permission issue when flytekit downloads the file.
+                if Version(envd_version) < Version("0.3.39"):
```
eapolinario, line 217: should we also add a lower-bound version in the envd [plugin](https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-envd/setup.py#L7)?

pingsutw: I think we don't need to add that. People can upgrade their `flytekitplugins-envd` without upgrading flytekit. if they didn't upgrade flytekit, they don't need to upgrade envd.

eapolinario (approved):

PR #2124, opened 2024-01-21
Core/show version

novahow:
## Tracking issue
fixes https://github.com/flyteorg/flyte/issues/4707

Show versions and creation time of remote-task/launchplan/workflow

This PR contains 2 parts. 
For the show-versions part, I created a callback like the --help message. The callback is invoked when `--show-versions` is specified. 
```
self._version_option = click.Option(
    [f"--{self.VERSION_DISPLAY}"],
    help="Show the version of the entity",
    is_flag=True,
    default=False,
    callback=self.show_versions,
    expose_value=False,
)
```

```
def show_versions(self, ctx: click.Context, param: str, value: bool):
      ...
```

In `DynamicEntityLaunchCommand._get_params` method, after options from the task is registered, the `show-versions` option was registered with `params.append(self._version_option)`

To fetch the creation time of the task, the `created_at` attribute is added in closures. This ensures that both versions and created_time can be retrieved from `list_tasks_paginated` method.
```
@classmethod
    def from_flyte_idl(cls, p):
        """
        :param flyteidl.admin.workflow_pb2.WorkflowClosure p:
        :rtype: WorkflowClosure
        """
        return cls(compiled_workflow=_compiler_models.CompiledWorkflowClosure.from_flyte_idl(p.compiled_workflow))
        return cls(
            compiled_workflow=_compiler_models.CompiledWorkflowClosure.from_flyte_idl(p.compiled_workflow),
            created_at=p.created_at.ToDatetime().replace(tzinfo=_timezone.utc) if p.HasField("created_at") else None,
        )
```
For the execute with version part, I simply split the entity name with `":"`, and iteratively loop through possible combinations of entity_name and version with the `_looped_fetch_entity ` method. 

## How was this patch tested?

By running `pyflyte run remote-[launchplan|workflow|task] {name} --show-versions` , the versions will be displayed
Code snippet:
```
from math import sqrt
from flytekit import workflow, task, LaunchPlan
from typing import List, Union
import random
import pdb

@task
def standard_deviation(values: List[float], mu: float) -> float:
    variance = sum([(x - mu) ** 2 for x in values])
    return sqrt(variance)

@task
def standard_scale(values: List[float], mu: float, sigma: float) -> List[float]:
    return [(x - mu) / sigma for x in values]


@task
def mean(values: List[float]) -> float:
    return sum(values) / len(values)

@task
def mean2(values: List[float]) -> float:
    return sum(values) / len(values)

@task
def generate_data(num_samples: int, seed: int) -> List[float]:
    random.seed(seed)
    return [random.random() for _ in range(num_samples)]


@workflow
def standard_scale_workflow(values: List[float]) -> List[float]:
    mu = mean(values=values)
    sigma = standard_deviation(values=values, mu=mu)
    return standard_scale(values=values, mu=mu, sigma=sigma)


@workflow
def workflow_with_subworkflow(num_samples: int, seed: int) -> List[float]:
    data = generate_data(num_samples=num_samples, seed=seed)
    return standard_scale_workflow(values=data)


# pdb.set_trace()
print("Compiled")
# LaunchPlan.CACHE.pop("standard_scale_lp", None)
standard_scale_launch_plan = LaunchPlan.get_or_create(
    standard_scale_workflow,
    name="standard_scale_lp",
    default_inputs={"values": [3.0, 4.0, 5.0]}
)

```
Example:

`
pyflyte run remote-task flyteTest.example.mean --show-versions
`

One can then run 
`pyflyte run remote-task flyteTest.example.mean:{version} --inputs ...`

### Screenshots
image
image
image

- [v] All new and existing tests passed.
- [v] All commits are signed-off.


kumare3:
There is one problem with this
`--show-versions` implies you cannot have a task or a workflow with `show_versions` as an input
Example
```python
@task 
def do(show_versions: bool):
    ...
```

which might be ok. But to counter this, I suggested add a subcommand of type
```
pyflyte run remote-task mytask show-versions
```
Note the missing `--`

cc @wild-endeavor what should we do?

If you create a subcommand, like how we do for remote-entities. the table also does not need to be created!

flytekit/clis/sdk_in_container/run.py
```diff
@@ -641,6 +673,41 @@ def invoke(self, ctx: click.Context) -> typing.Any:
...
+                named_entity, sort_by=Sort("created_at", Sort.Direction.DESCENDING)
+            )
+        elif isinstance(entity, FlyteWorkflow):
+            sorted_entities, _ = _remote_instance.client.list_workflows_paginated(
+                named_entity, sort_by=Sort("created_at", Sort.Direction.DESCENDING)
+            )
+        else:
+            raise ValueError(f"Unknown entity type {type(entity)}")
+        formatter = click.HelpFormatter()
+        task_table = []
```
kumare3, line 702: please use rich table - https://rich.readthedocs.io/en/stable/tables.html

novahow:
> can we make `show-versions` or maybe just `versions` a top level command on pyflyte? so just `pyflyte versions` instead of `pyflyte run`?

Thanks for the advice, this sounds more feasible for me, I've modified my PR to implement this behavior.

wild-endeavor:
I can't get this to run locally actually.  running the new command I'm getting
```
  File "/Users/ytong/go/src/github.com/flyteorg/flytekit/flytekit/clis/sdk_in_container/versions.py", line 19, in __init__
    super(click.RichGroup, self).__init__(name, h, entity_name, launcher, **kwargs)
```

related, why is `DynamicEntityVersionCommand` inheriting both `click.RichGroup` and `DynamicEntityLaunchCommand`?  why is this related to launch?

novahow:
@wild-endeavor Hi, sorry for the confusion. `DynamicEntityVersionCommand ` inherits the `DynamicEntityLaunchCommand ` because it needs to call the `_fetch_entity` method of `DynamicEntityLaunchCommand` in `list_commands` method, so that it can find the type of the desired entity. 
Regarding the error, could you please provide the command you used that triggers the error you mentioned?

flytekit/clis/sdk_in_container/run.py
```diff
@@ -656,7 +698,7 @@ class RemoteEntityGroup(click.RichGroup):
     WORKFLOW_COMMAND = "remote-workflow"
     TASK_COMMAND = "remote-task"
 
-    def __init__(self, command_name: str):
+    def __init__(self, command_name: str, h: str):
```
pingsutw, line 701: Do we use `h` anywhere?

novahow: thanks for pointing out, should use h instead, fixed

pingsutw (approved):

jpvotta:
Can we please also test with reference task, reference workflow, and reference launchplan?

flytekit/models/admin/workflow.py
```diff
@@ -1,4 +1,6 @@
 import typing
+from datetime import datetime as _datetime
```
wild-endeavor, line 2: can we not use the `_library` pattern anymore.  Let's make this just `from datetime import datetime`

flytekit/models/admin/workflow.py
```diff
@@ -1,4 +1,6 @@
...
+from datetime import timezone as _timezone
```
wild-endeavor, line 3: ditto

flytekit/models/admin/workflow.py
```diff
@@ -124,16 +128,31 @@ def compiled_workflow(self):
         """
         return self._compiled_workflow
 
+    @property
+    def created_at(self):
```
wild-endeavor, line 132: move type hint to function signature and remove comment

flytekit/models/launch_plan.py
```diff
@@ -1,4 +1,5 @@
 import typing
+from datetime import timezone as _timezone
```
wild-endeavor, line 2:
```suggestion
from datetime import timezone
```

flytekit/models/launch_plan.py
```diff
@@ -349,15 +352,25 @@ def expected_outputs(self):
         """
         return self._expected_outputs
 
+    @property
+    def created_at(self):
+        """
```
wild-endeavor, line 357: ditto

wild-endeavor (commented):
the top command currently shows
image
Instead of every line item showing "Display version of a remote-task." can we show the version?  If that's not easy to do, can we just hide that column completely?

novahow:
> Can we please also test with reference task, reference workflow, and reference launchplan?

Hi, can you elaborate more on how to cover that? Since I'm not really familiar with reference task, thanks

flytekit/models/admin/workflow.py
novahow: I'll just hide that column completely, thanks for the suggestion. Since the version is a list and is shown when you add the task as a subcommand

jpvotta:
> > Can we please also test with reference task, reference workflow, and reference launchplan?
> 
> @jpvotta Hi, can you please elaborate more on how to cover that? Since I'm not really familiar with reference task, thanks

Apologies. I am going to defer to @eapolinario on whether or not this is necessary (guessing it shouldn't be relevant here).

eapolinario:
@novahow , sorry for the delay I was out in the last 2.5 weeks. 

Reference entities are a way of referring to pre-existing Flyte entities. Take a look at https://docs.flyte.org/en/latest/user_guide/productionizing/reference_tasks.html#reference-tasks, https://docs.flyte.org/en/latest/user_guide/productionizing/reference_launch_plans.html, and https://docs.flyte.org/en/latest/api/flytekit/generated/flytekit.reference_workflow.html#flytekit-reference-workflow.

PR #2131, opened 2024-01-25
Adjust tar method to iterate over files/dirs in dir rather than strip…

neilisaur:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4479

Closes #4479

When tar'ing up scripts, there is an additional entry in the tar file of `/`. This causes any extractions to remove the leading `/` and leave an empty entry that attempts to modify times and ownerships on the current directory. When not running as `root` or a consistent user--as is the case with OpenShift's randomized UIDs--the context is not able to extract the tar file successfully.

Example tar file as-is:

```
va@de:tmp/tmp42o0n2ru(⎈|panel/api-pcell:panel)$ tar tvfz script_mode.tar.gz 
tar: Removing leading `/' from member names
drwxrwx--- 0/0               0 1970-01-01 00:00 /
drwxrwx--- 0/0               0 1970-01-01 00:00 workflows/
-rw-rw---- 0/0               0 1970-01-01 00:00 workflows/__init__.py
-rw-rw---- 0/0            2244 1970-01-01 00:00 workflows/ray-example.py
```

Change the method of creation of the script_mode tar file to not include a `/` in the resultant tar file by iterating over the first-level files and directories copied into the temporary `code` directory.

## How was this patch tested?

Tested locally to ensure that the tar generated in temporary directories no longer included the erroneous `/` entry:

New tar file contents:

```
va@de:tmp/tmpes7z_phu(⎈|panel/api-pcell:panel)$ tar tvfz script_mode.tar.gz 
drwxrwx--- 0/0               0 1970-01-01 00:00 workflows/
-rw-rw---- 0/0               0 1970-01-01 00:00 workflows/__init__.py
-rw-rw---- 0/0            2302 1970-01-01 00:00 workflows/ray-example.py
```


wild-endeavor (approved):
🙏🖖🚀

PR #2133, opened 2024-01-26
Adds get_default_image into configuration plugin

thomasjpfan:

This PR allows a third party plugin to override the default images.

This PR extends the configuration plugin API with a `get_default_image` method:

- If plugin returns `None`, then nothing changes and the images are pulled from flytekit's `DefaultImages`.
- If the plugin returns a string, then that string is used as the default image.

## How was this patch tested?

I added unit test to check the plugin's behavior with `DefaultImage`.

### DIFF
--- a/flytekit/configuration/default_images.py
+++ b/flytekit/configuration/default_images.py
@@ -1,6 +1,7 @@
 import enum
 import sys
 import typing
+from contextlib import suppress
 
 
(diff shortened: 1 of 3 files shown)


cosmicBboy (approved):
lgtm

pingsutw (approved):

PR #2141, opened 2024-01-30
Remove asynchronous flag from base agent

pingsutw:
## Tracking issue

Remove the asynchronous flag from the base agent, and use `inspect.iscoroutinefunction` to check if the function is a coroutine function. It will make the code more concise and cleaner

## How was this patch tested?

Run airflow sensor.

### Setup process
```python
from datetime import datetime, timedelta
from pytz import UTC
from airflow.sensors.time_sensor import TimeSensor
from flytekit import workflow

x = (datetime.now(tz=UTC)+timedelta(seconds=20)).time()


@workflow
def wf():
    for i in range(3):
        TimeSensor(task_id=f"time_sensor_{i}", target_time=x)


if __name__ == '__main__':
    wf()
```

image

image


Future-Outlier (commented):
The order should be output prefix, task template, inputs.
(When is not coroutine function, we need to consider the order of the inputs.

https://github.com/flyteorg/flytekit/pull/2141/files#diff-2944080a47998a2b23f992855b3a532bda59633aa1eaf11b812363ee191543c7R60-R62

Future-Outlier (approved):
LGTM, test it by sensor agent.
(from both local and remote scenario)

Future-Outlier:
need 1 more approve to merge this PR.

eapolinario (approved):

PR #2148, opened 2024-01-30
Remove kubernetes from dependencies

pingsutw:
## Tracking issue
NA

we lazy load `kuberentes` [flytekit/flytekit/core/pod_template.py](https://github.com/flyteorg/flytekit/blob/a1636a89daee2301a056b3eba41fe4dcea1257b7/flytekit/core/pod_template.py#L22-L23), so we don't need to add it to the dependency by default.

when using podTemplate, people will import kubernetes in their workflow code. Therefore, if it's not installed, they will see the error in IDE or at compile time.

Remove `Kubernetes` from dependencies

## How was this patch tested?
Unit tests

## Related PRs
NA

## Docs link
NA


eapolinario:
How do you plan to handle pod templates?

pingsutw:
We lazy load `Kubernetes`. We can raise an error in flytekit if `Kubernetes` is not installed while using the pod template.

eapolinario (approved):

flixr:
So this also means that there are no pre-built images available anymore that can use pod templates :-(

PR #2151, opened 2024-01-30
Warn user when overriding requests but not limits

fg91:

```py
@task(
    requests=Resources(
        cpu="3",
        mem="3Gi",
    ),
)
def foo():
    ...



@workflow
def my_wf():
    foo().with_overrides(
        requests=Resources(
            cpu="5",
            mem="5Gi",
        ),
    )
```

As a user, I would expect that the resulting task pod has 5 vCPUs and 5Gi or RAM. Turns out, the task pod has 3 each!

This behaviour is *very* unintuitive, the reason being as follows:

* When one doesn't explicitly set limits on a task, the requests are taken as limits as well.
* When one only overrides the requests, the requests are clamped to the old limits[ in the backend](https://github.com/flyteorg/flyte/blob/ffd3ab2e4784733a446a3440eab066852819f5fe/flyteplugins/go/tasks/pluginmachinery/flytek8s/container_helper.go#L107).

In this PR I add a warning so that at least users are warned to set limit as well.

## How was this patch tested?

Running the above workflow with `pyflyte -v run --remote ...` results in the following new log line:

```console
Requests overridden on node n0 (<FlyteLiteral name: "foo" retries { }>) without specifying limits. Requests are clamped to original limits.
```

The fact that `-v` is required to see this warning unfortunately limits its usefulness. Maybe this warrants a discussion whether this behaviour should be changed.

### DIFF
--- a/flytekit/core/node.py
+++ b/flytekit/core/node.py
@@ -145,6 +145,15 @@ class Node(object):
             limits = kwargs.get("limits")
             if limits and not isinstance(limits, Resources):
                 raise AssertionError("limits should be specified as flytekit.Resources")
+
+            if not limits:
+                logger.warning(
+                    (
+                        f"Requsts overridden on node {self.id} ({self.metadata.short_string()}) without specifying limits. "
+                        "Requests are clamped to original limits."
+                    )
+                )
+
             resources = convert_resources_to_resource_model(requests=requests, limits=limits)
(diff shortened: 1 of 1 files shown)


pingsutw (approved):
Thank you

Future-Outlier (approved):

PR #2170, opened 2024-02-07
Improve Rich Logger

Future-Outlier:
When executing flyekit pod task, we will have context in the container, and we don't want to use the rich logger to print them.

## How was this patch tested?
```python
@task
def say_hello() -> None:
    raise ValueError("This is a test error")
```

### DIFF
--- a/flytekit/loggers.py
+++ b/flytekit/loggers.py
@@ -4,9 +4,9 @@ import typing
 
 from pythonjsonlogger import jsonlogger
 
-from .tools import interactive
-
 from flytekit.core.context_manager import FlyteContextManager
+
+from .tools import interactive
 
 # Note:
 # The environment variable controls exposed to affect the individual loggers should be considered to be beta.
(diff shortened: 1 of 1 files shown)


Future-Outlier:
@pingsutw I am not sure how to test it, my logs with failure scenario are still

Future-Outlier:
@fg91 I've traced the git history and find that you have made this feature before, would you like to take a look?
Thank you!

Future-Outlier:
Image built by this branch
image
Image built by default
image

Future-Outlier:
have tried these 2 examples, still can't reproduce.
```python
@task
def say_hello() -> None:
    raise Exception("This is a test exception")
```

```python
from airflow.sensors.filesystem import FileSensor
from airflow.sensors.time_sensor import TimeSensor
from datetime import datetime, timedelta
from pytz import UTC

@task()
def t1():
    print("flyte")

@workflow
def wf_time_sensor():
    # sensor = FileSensor(task_id="id", filepath="/tmp/1234")
    sensor = TimeSensor(task_id="fire_immediately", target_time=(datetime.now(tz=UTC)+timedelta(seconds=1)).time())
    sensor >> t1()
```

PR #2176, opened 2024-02-07
Fix: Handle overriding of container image in backend

fg91:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4543

The `container_image` task node override, among others, is currently broken (doesn't work with `pyflyte register` and only once with `pyflyte run`, see [tracking issue](https://github.com/flyteorg/flyte/issues/4543).

Don't override container image of `run_entity` in `.with_overrides` but actually in flyteidl's `TaskNodeOverride` so that the container image is overridden in the backend.

## How was this patch tested?

In combination with https://github.com/flyteorg/flyte/pull/4858 I ensured that the resulting pod in the cluster has the correct image. I also adapted the unit tests.

## Related PRs

https://github.com/flyteorg/flyte/pull/4858

### DIFF
--- a/flytekit/core/node.py
+++ b/flytekit/core/node.py
@@ -65,6 +65,7 @@ class Node(object):
         self._outputs = None
         self._resources: typing.Optional[_resources_model] = None
         self._extended_resources: typing.Optional[tasks_pb2.ExtendedResources] = None
+        self._container_image: typing.Optional[str] = None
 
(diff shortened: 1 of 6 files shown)


fg91:
Tests are failing because an idl upgrade is needed once https://github.com/flyteorg/flyte/pull/4858 is merged.

PR #2179, opened 2024-02-08
Raises envd specific error in envd plugin instead of image_spec builder itself

thomasjpfan:
This PR raises envd specific error in envd plugin instead of image_spec builder itself.

Currently, `ImageSpecBuilder` requires `envd` to be installed for the version check. This forces non-envd based image builders to have `envd` installed as well.

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -7,14 +7,10 @@ import typing
 from abc import abstractmethod
 from dataclasses import asdict, dataclass
 from functools import lru_cache
-from importlib import metadata
 from typing import Dict, List, Optional, Tuple, Union
 
 import click
 import requests
-from packaging.version import Version
-
-from flytekit.exceptions.user import FlyteAssertion
 
 DOCKER_HUB = "docker.io"
 _F_IMG_ID = "_F_IMG_ID"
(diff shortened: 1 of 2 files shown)


flytekit/image_spec/image_spec.py
```diff
@@ -228,15 +230,16 @@ def build(cls, image_spec: ImageSpec) -> str:
...
-                    raise FlyteAssertion(
-                        f"envd version {envd_version} is not compatible with flytekit>v1.10.2."
-                        f" Please upgrade envd to v0.3.39+."
-                    )
+            with contextlib.suppress(PackageNotFoundError):
+                if builder == "envd":
+                    envd_version = metadata.version("envd")
+                    # flytekit v1.10.2+ copies the workflow code to the WorkDir specified in the Dockerfile. However, envd<0.3.39
+                    # overwrites the WorkDir when building the image, resulting in a permission issue when flytekit downloads the file.
+                    if Version(envd_version) < Version("0.3.39"):
```
wild-endeavor, line 238: not sure i follow... if the exception is suppressed, why have it at all?

wild-endeavor (approved):
+1 but question.

flytekit/image_spec/image_spec.py
eapolinario: is this to handle the case someone uses `Imagespec` but don't have envd installed? What happens in that case? What should happen?

thomasjpfan: Looking this over, I think this PR is not required anymore.

When `builder=="envd"`, `envd` should be installed at that point.

thomasjpfan:
I do not think this PR is required, so I closed it. Thank you all for reviewing!

REF: https://github.com/flyteorg/flytekit/pull/2179#discussion_r1482378114

PR #2186, opened 2024-02-10
Fix Base Agent Input Bug

Future-Outlier:
## Tracking issue

The input of `create` function in base_agent is wrong, we should fix it.

Before:
image

After:
image


pingsutw (approved):

pingsutw:
thank you

PR #2198, opened 2024-02-18
Fix local overwrite cache

ChungYujoyce:
## Tracking issue
Closes _https://github.com/flyteorg/flyte/issues/4727_

We can't overwrite the cache when using `pyflyte run --overwrite-cache` in local runs.

* Once the argument `--overwrite-cache` is added, we get `run_level_params.overwrite_cache=True` and set the environment variable `FLYTE_LOCAL_CACHE_OVERWRITE=true` accordingly. 
* With `FLYTE_LOCAL_CACHE_OVERWRITE=true`, the task will be executed and overwrite the existing cache.
* The cache `add` function in `LocalTaskCache`  is changed to `set` as the previous one can't update the cache if there's already an existing one.

## How was this patch tested?

Added a test which confirms that setting `FLYTE_LOCAL_CACHE_OVERWRITE=true` can successfully overwrite the existing cache.

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -266,6 +266,7 @@ class Task(object):
 
         # if metadata.cache is set, check memoized version
         local_config = LocalConfig.auto()
+        print(local_config.cache_overwrite)
         if self.metadata.cache and local_config.cache_enabled:
             # TODO: how to get a nice `native_inputs` here?
             logger.info(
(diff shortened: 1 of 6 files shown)


flytekit/core/base_task.py
```diff
@@ -274,8 +274,11 @@ def local_execute(
             )
             outputs_literal_map = LocalTaskCache.get(self.name, self.metadata.cache_version, input_literal_map)
             # The cache returns None iff the key does not exist in the cache
-            if outputs_literal_map is None:
-                logger.info("Cache miss, task will be executed now")
+            if outputs_literal_map is None or local_config.cache_overwrite:
```
pingsutw, line 277: if cache_overwrite is enabled, we still retrieve the output from LocalTaskCache. could we update the code to something like

```python
if cache_overwrite:
    self.sandbox_execute(...)
else:
    outputs_literal_map = LocalTaskCache.get(...)
    ...
```

ChungYujoyce: updated

pingsutw (approved):
LGTM, thank you!

eapolinario (approved):
Thank you.

Don't worry about the failing unit tests in https://github.com/flyteorg/flytekit/actions/runs/7969843181/attempts/2?pr=2198, those are being fixed in https://github.com/flyteorg/flytekit/pull/2199.

PR #2200, opened 2024-02-19
docs(flyteinteractive): Change docker build to docker buildx build

MortalHappiness:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4918_

There is no `--push` flag for the `docker build` command in Linux machine.

Change `docker build` to `docker buildx build`

### DIFF
--- a/plugins/flytekit-flyteinteractive/README.md
+++ b/plugins/flytekit-flyteinteractive/README.md
@@ -29,7 +29,7 @@ def train():
 ```
 
 ## User Guide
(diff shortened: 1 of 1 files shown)


Future-Outlier (commented):
How about use `docker buildx build` instead of `docker build`?
It seems that `docker build` will be deprecated in the future, we will not need to update the doc again in the future.

eapolinario (approved):

pingsutw (approved):

eapolinario:
The move to `buildx` requires a builder to be defined, but the error will be apparent to the user (i.e. it will say something along the lines of a builder being needed), so no need to mention it in the docs.

PR #2204, opened 2024-02-21
Add launch condition column

wild-endeavor:
This relies on the changes to IDL in: https://github.com/flyteorg/flyte/pull/4903/files

* Remove bindings when embeding as query, only one binding is allowed now.
* Introduce a launch plan trigger base protocol, and add the schedule implementation of it (backend code is not ready yet so it'll throw an error for now).
* Replace poorly named additional_metadata with `trigger`.
* Removing some underscores in schedule model.
* Update translator to call to flyte idl on the implementation of the launch plan trigger base, with the python launch plan object itself as an arg.

## How was this patch tested?
End-to-end testing with private backend will be run.


flytekit/core/artifact.py
```diff
@@ -147,7 +147,11 @@ def __init__(
         self.time_partition = time_partition
         self.partitions = partitions
         self.tag = tag
-        self.bindings = bindings
+        print(f"AQ constructor {name} bindings len {len(bindings)}")
```
wild-endeavor, line 150: need to update with set operation

squiishyy (approved):

squiishyy (approved):

pingsutw (approved):

PR #2211, opened 2024-02-23
[WIP] Msgspec Dataclass Type Transformer

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4485


thomasjpfan:
Thank you for your interest in the original issue. With https://github.com/flyteorg/flyte/issues/4485#issuecomment-1962096208, I think it's best to go with [mashumaro](https://github.com/Fatal1ty/mashumaro) directly.

Future-Outlier:
> Thank you for your interest in the original issue. With https://github.com/flyteorg/flyte/issues/4485#issuecomment-1962096208, I think it's best to go with [mashumaro](https://github.com/Fatal1ty/mashumaro) directly.

Thank you, will take a look!

kumare3:
Can we really remove marshmallow- if so let's do it

Future-Outlier:
> Can we really remove marshmallow- if so let's do it 

Thank you, will take a look

Future-Outlier:
@thomasjpfan, @pingsutw 
Please look at the example above, I found that we only need to use `@dataclass`, then we can use msgspec to serialize and deserialize all of them.
I've haven't dealt with compicate cases, for example adding `int`, but maybe `msgspec` is the answer we need, right?

Future-Outlier:
@jcrist Hi, here's a question.
Flyte might have possibility to use `msgspec`, since that I found that if we don't need to inherirt `msgspec.Struct`, we can still use `msgspec.json.encode` and `msgspec.json.decode` to serialize and deserialize dataclasses.
Is there any limit or any potential issues for`msgspec`?
Or more specifically, can you tell us restrictions about `msgspec` here?
https://jcristharif.com/msgspec/supported-types.html

thomasjpfan:
> I just found that lots of its functions are written in C, which will far more faster than python.

It could be faster, but I do not think that is Flyte's bottleneck right now or there is a demand for `msgspec`. Every new feature adds more code for us to maintain. I have experimented with using `msgspec` just to handle dataclasses which works, but I did not open a PR because of the maintenance cost.

> we can still use msgspec.json.encode and msgspec.json.decode to serialize and deserialize dataclasses.

This use case is also supported by mashumaro: https://github.com/Fatal1ty/mashumaro#supported-serialization-formats

```python
from mashumaro.codecs.json import JSONDecoder, JSONEncoder

@dataclass
class MyModel:
    a: int
    b: bool

model = MyModel(a=10, b=False)
encoder = JSONEncoder(MyModel)

encoder.encode(model)
# '{"a": 10, "b": false}'
```

thomasjpfan:
> remove dataclass_json
> remove mixins
> if it doesn't work, I will list reasons and discuss with you.

To clarify, I want to remove the `@dataclass_json` and `DataClassORJSONMixin` requirement for majority of the use cases. If a user decides to use mixins, like in https://github.com/flyteorg/flytekit/pull/2080, that should still continue to work.

Future-Outlier:
> > remove dataclass_json
> > remove mixins
> > if it doesn't work, I will list reasons and discuss with you.
> 
> To clarify, I want to remove the `@dataclass_json` and `DataClassORJSONMixin` requirement for majority of the use cases. If a user decides to use mixins, like in #2080, that should still continue to work.

No problem, thank you!

Future-Outlier:
Just create a PR here.
https://github.com/flyteorg/flytekit/pull/2279

So I will close this PR.

PR #2213, opened 2024-02-25
Agent Log Capture Feature

Future-Outlier:
## Tracking issue

Make every log in the agent server can be shown in flyte console instead of the agent server.

In airflow agent, use `io.StringIO` and "logging" module to capture logging messages, and send them to flytepropeller.

TODO: we should split the feature into 2 functions, set log capture and unset it.
we can move it to `flytekit.extend.backend.utils`, to support all agents.

## How was this patch tested?
In local single binary.

image

```python
if message:
    message = message + "\n" + log_capture_string.getvalue()
else:
    message = log_capture_string.getvalue()
print("message: ", message)
return GetTaskResponse(resource=Resource(phase=cur_phase, message=message))
```

### DIFF
--- a/plugins/flytekit-airflow/flytekitplugins/airflow/agent.py
+++ b/plugins/flytekit-airflow/flytekitplugins/airflow/agent.py
@@ -1,9 +1,9 @@
 import asyncio
+import logging
 import typing
 from dataclasses import dataclass, field
-from typing import Optional
-import logging
 from io import StringIO
+from typing import Optional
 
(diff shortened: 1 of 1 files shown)


Future-Outlier:
I've faced a problem on how to close the airflow logger on the terminal, not yet found a solution.
Also, I am finding ways to show message on flyteconsole in `RUNNING` Phase,
need to tweak some code in flytepropeller.

PR #2223, opened 2024-02-29
Enhance default image to check FLYTE_INTERNAL_IMAGE

ddl-rliu:
## Tracking issue

Provides context for https://flyte-org.slack.com/archives/CP2HDHKE1/p1709158841281599

Some CLI paths like `pyflyte run, pyflyte register` do not use the image specified by the environment variable `FLYTE_INTERNAL_IMAGE` in the manner one might expect

### DIFF
--- a/flytekit/configuration/__init__.py
+++ b/flytekit/configuration/__init__.py
@@ -285,7 +285,7 @@ class ImageConfig(DataClassJsonMixin):
                 images.append(img)
 
         if default_image is None:
-            default_image_str = os.environ.get("FLYTE_INTERNAL_IMAGE", DefaultImages.default_image())
+            default_image_str = DefaultImages.default_image()
             default_image = Image.look_up_image_info(DEFAULT_IMAGE_NAME, default_image_str, False)
         return ImageConfig.create_from(default_image=default_image, other_images=images)
 
--- a/flytekit/configuration/default_images.py
+++ b/flytekit/configuration/default_images.py
@@ -1,4 +1,5 @@
 import enum
+import os
 import sys
 import typing
 from contextlib import suppress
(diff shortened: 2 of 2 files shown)


flytekit/configuration/__init__.py
```diff
@@ -285,7 +285,7 @@ def validate_image(_: typing.Any, param: str, values: tuple) -> ImageConfig:
                 images.append(img)
 
         if default_image is None:
-            default_image_str = os.environ.get("FLYTE_INTERNAL_IMAGE", DefaultImages.default_image())
+            default_image_str = DefaultImages.default_image()
```
ddl-rliu, line 288: This code block is entered when `default_image is None` however for e.g. `pyflyte register` the default_image will always be something like `cr.flyte.org/...` and so the `FLYTE_INTERNAL_IMAGE` env var never takes effect.

samhita-alla: @ddl-rliu `os.environ.get` returns `DefaultImages.default_image()` only if `FLYTE_INTERNAL_IMAGE` doesn't exist, correct?

ddl-rliu: Right! In particular, note that the existence of `FLYTE_INTERNAL_IMAGE` is checked here https://github.com/flyteorg/flytekit/pull/2223/files#diff-e3e9429cda3e0f0d375729117b25595b280937f3a04e90f1efc7ee2c1a164008R38 and so it would not be necessary to doubly-check that env var here in `configuration/__init__.py`

pingsutw (approved):
LGTM, @ddl-rliu could we merge this?

ddl-rliu:
Yes, ready to merge!

ddl-rliu:
@pingsutw Yes, ready to merge! Thanks for the review 🙇

pingsutw (approved):

PR #2239, opened 2024-03-05
set subnode and parent node timeout metadata for arraynode

pvditt:
## Tracking issue
related to looking into this issue:

_https://github.com/flyteorg/flyte/issues/3875_

https://unionai.atlassian.net/browse/FLYTE-204

Currently don't set metadata for either the parent ArrayNode or Subnodes

- enable timeouts to be defined for parent ArrayNodes. Set the timeout in the node's metadata
map_task(timeout= ...)

- set the underlying entity (only task for now) timeout in the target array_node's metadata

Note the subnode task timeouts are not supported yet.

Given:
```
@task(timeout="4h")
def foo(x: int) -> int:
    pass

@workflow
def bar(x: List[int]) -> List[int]:
    maptask(foo, timeout="6h")(x=x)
```

We want:
```
Node:
	NodeMetadata:
		timeout:"6h"
	Target
		ArrayNode
			Node
				NodeMetadata
					timeout: "4h"
				Target
					TaskNode
							"foo"
			Parallelism
			SuccessCriteria
```

### Setup process

ran the map_task workflow in flytesnacks (updated to use array node map tasks) + set timeouts -> confirm appropriate timeout is set on NodeSpec and SubNodeSpec.

```
@task
def multi_input_task(quantity: int, price: float, shipping: float) -> float:
    return quantity * price * shipping


@workflow
def multiple_inputs_map_workflow(list_q: list[int] = [1, 2, 3, 4, 5], p: float = 6.0, s: float = 7.0) -> list[float]:
    partial_task = functools.partial(multi_input_task, price=p, shipping=s)
    return map_task(partial_task, timeout=timedelta(hours=1), metadata=TaskMetadata(timeout=timedelta(minutes=5)))(quantity=list_q)
```

### DIFF
--- a/flytekit/core/array_node_map_task.py
+++ b/flytekit/core/array_node_map_task.py
@@ -91,7 +91,7 @@ class ArrayNodeMapTask(PythonTask):
         self._min_successes: Optional[int] = min_successes
         self._min_success_ratio: Optional[float] = min_success_ratio
         self._collection_interface = collection_interface
-        self._timeout = timeout
+        self._timeout: Optional[timedelta] = timeout
 
         if "metadata" not in kwargs and actual_task.metadata:
             kwargs["metadata"] = actual_task.metadata
--- a/flytekit/core/array_node_map_task.py
+++ b/flytekit/core/array_node_map_task.py
@@ -361,8 +361,9 @@ def map_task(
         successfully before terminating this task and marking it successful.
     :param timeout: If specified, this determines the timeout for the parent ArrayNode
     """
-    return ArrayNodeMapTask(task_function, concurrency=concurrency, min_success_ratio=min_success_ratio,
-                            timeout=timeout, **kwargs)
+    return ArrayNodeMapTask(
+        task_function, concurrency=concurrency, min_success_ratio=min_success_ratio, timeout=timeout, **kwargs
+    )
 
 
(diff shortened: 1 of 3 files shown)


hamersaw:
The UX feels rather confusing to me. IMO the internal tasks `NodeMetadata` should just be populated from the internal `Node` (unsure how this works exactly in flytekit). Your example could be something like:
```
@task(timeout=timedelta(hours=5))
def multi_input_task(quantity: int, price: float, shipping: float) -> float:
    return quantity * price * shipping


@workflow
def multiple_inputs_map_workflow(list_q: list[int] = [1, 2, 3, 4, 5], p: float = 6.0, s: float = 7.0) -> list[float]:
    partial_task = functools.partial(multi_input_task, price=p, shipping=s)
    return map_task(partial_task, timeout=timedelta(hours=1)(quantity=list_q)
````
and this should produce the `ArrayNode` IDL spec you described.

pvditt:
@hamersaw I should've left a better example - the example you left also works. Setting the metadata param in the map_task(...) call takes precedent over metadata set in the task decorator. https://docs.flyte.org/en/latest/user_guide/advanced_composition/map_tasks.html shows the example of setting the mapped task's metadata with the param as opposed to decorator.

Would a "parent_timeout" be better to lead to less confusion for when the metadata is set?

PR #2246, opened 2024-03-07
Adds a FlyteDeck markdown renderer to core

thomasjpfan:

This PR introduces a `MarkdownRender` to core. This makes `FlyteDeck` more useful without needing to install a plugin. 

I added `markdown-it-py` as a direct dependency, but it does not increase our overall dependency count. `markdown-it-py` is a direct dependency of `rich`:

https://github.com/flyteorg/flytekit/blob/329aea8545aa9ea9f903acd7e006f3f12efb9cbb/dev-requirements.txt#L241-L242

And I do not think we will be removing the `rich` dependency.

## How was this patch tested?

I added a unittest to check the renderer. I also ran this workflow built with this PR and the Flyte deck appears.

```python
from flytekit import task, workflow, current_context, Deck

@task(
    enable_deck=True,
    container_image="ghcr.io/thomasjpfan/flytekit:deck-markdown-0.0.1",
)
def write_to_deck():
    from flytekit.deck.renderer import MarkdownRenderer

    context = current_context()
    decks = context.decks

    decks.insert(0, Deck("My deck", MarkdownRenderer().to_html("# Hello world")))


@workflow
def main():
    write_to_deck()
```

### DIFF
--- a/flytekit/deck/renderer.py
+++ b/flytekit/deck/renderer.py
@@ -1,5 +1,6 @@
 from typing import TYPE_CHECKING, Any
 
+from markdown_it import MarkdownIt
 from typing_extensions import Protocol, runtime_checkable
 
 from flytekit import lazy_module
@@ -48,3 +49,10 @@ class ArrowRenderer:
     def to_html(self, df: "pyarrow.Table") -> str:
         assert isinstance(df, pyarrow.Table)
         return df.to_string()
+
+
+class MarkdownRenderer:
+    """Convert a markdown string to HTML and return HTML as a unicode string."""
+
+    def to_html(self, text: str) -> str:
+        return MarkdownIt().render(text)
--- a/flytekit/deck/__init__.py
+++ b/flytekit/deck/__init__.py
@@ -12,7 +12,8 @@ Contains deck renderers provided by flytekit.
 
    Deck
    TopFrameRenderer
+   MarkdownRenderer
 """
 
 from .deck import Deck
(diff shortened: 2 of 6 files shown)


flytekit/deck/renderer.py
```diff
@@ -48,3 +49,10 @@ class ArrowRenderer:
     def to_html(self, df: "pyarrow.Table") -> str:
         assert isinstance(df, pyarrow.Table)
         return df.to_string()
+
+
+class MarkdownRenderer:
```
pingsutw, line 54: we already have this in the [plugin](https://github.com/flyteorg/flytekit/blob/b4de2f845972967746babbeb8be645025d214086/plugins/flytekit-deck-standard/flytekitplugins/deck/renderer.py#L64-L73). if we want to move to core, the onc in the plugin can be removed.

thomasjpfan: I deprecated the previous `MarkdownRenderer` here: [`4e93235` (#2246)](https://github.com/flyteorg/flytekit/pull/2246/commits/4e93235ab3f3177ca93fe52e38964cb1b3a034a4)

Given that `FlyteDeck` is beta, I am okay with removing it from the plugin as well.

What do you prefer?

pingsutw: makes sense, let's deprecate first

pingsutw (approved):

eapolinario (approved):

PR #2258, opened 2024-03-12
Raw Container Task Local Execution

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/3876

We want to support users to run `raw-container` task locally.

Note: I didn't support ~~`datetime.timedelta`~~, `List` and `Dict`.
Since I found that copilot has bug to deal with that, so I will support them until those bugs are fixed.
(This implies that we have to fix flytecopilot for nested types and collections types first)

1.  override `local_execute` function in `class ContainerTask(PythonTask)`
2. use `docker` python library to execute docker command
3. support int, str, float, bool, datetime.datetime, FlyteDirectory and FlyteFile
~~4. I will strip white space because it will be too complicated to support string with white space.~~
~~(datetime.datetime will be affected)~~
5. I've supported FlyteDirectroy and FlyteFile as input, ~~however, in remote cases, flytecopilot will fail, so I think I can support it locally first, and create an issue to support remote cases.~~
6. Support `direct file paths`  and `template-style references` as inputs.
For example, `{{.inputs.infile}}` and `/var/inputs/infile` both are valid inputs.

## How was this patch tested?
1. local execution with 4 examples.
Reference: https://docs.flyte.org/en/latest/user_guide/customizing_dependencies/raw_containers.html#raw-container
Note: `julia` image has error when using arm64, I've tested it by `docker run -it` into the container and found this error.
image

2. local execution with 5 prmitive types as input and output.
3. `FlyteFile` task with input and output
4. `FlyteFile` task with output only
5. `FlyteDirectory` task with input and output
7. `FlyteDirectory` task with output only

Note: I've provided an image on docker hub, you can switch to this branch and test it directly.
image: `futureoutlier/rawcontainer:0320`

### Setup process
```bash
git clone https://github.com/flyteorg/flytekit
gh pr checkout 2258
make setup
pip install -e .
```
```
python raw_container_local_execution.py
```
python example:
```python
import logging
from typing import Tuple, List
import datetime
from flytekit import ContainerTask, kwtypes, workflow, task


logger = logging.getLogger(__file__)

@workflow
def primitive_types(a: int, b: bool, c: float, d: str, e: datetime.datetime, f: datetime.timedelta) \
                            -> Tuple[int, bool, float, str, datetime.datetime, datetime.timedelta]:
    return python_return_same_values(a=a, b=b, c=c, d=d, e=e, f=f)

python_return_same_values = ContainerTask(
    name="python_return_same_values",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    inputs=kwtypes(a=int, b=bool, c=float, d=str, e=datetime.datetime, f=datetime.timedelta),
    outputs=kwtypes(a=int, b=bool, c=float, d=str, e=datetime.datetime, f=datetime.timedelta),
    image="futureoutlier/rawcontainer:0320",
    command=[
        "python",
        "return_same_value.py",
        "{{.inputs.a}}",
        "{{.inputs.b}}",
        "{{.inputs.c}}",
        "{{.inputs.d}}",
        "{{.inputs.e}}",
        "{{.inputs.f}}",
        "/var/outputs",
    ],
)
 

flyte_file_io = ContainerTask(
    name="flyte_file_io",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    inputs=kwtypes(inputs=FlyteFile),
    outputs=kwtypes(out=FlyteFile),
    image="futureoutlier/rawcontainer:0320",
    command=[
        "python",
        "write_flytefile.py",
        "{{.inputs.inputs}}",
        # "/var/inputs/inputs",
        "/var/outputs/out",
    ],
)

flyte_dir_io = ContainerTask(
    name="flyte_dir_io",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    inputs=kwtypes(inputs=FlyteDirectory),
    outputs=kwtypes(out=FlyteDirectory),
    image="futureoutlier/rawcontainer:0320",
    command=[
        "python",
        "write_flytedir.py",
        "{{.inputs.inputs}}",
        # "/var/inputs/inputs",
        "/var/outputs/out",
    ],
)

flyte_dir_out_only = ContainerTask(
    name="flyte_dir_out_only",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    # inputs=kwtypes(inputs=FlyteDirectory),
    outputs=kwtypes(out=FlyteDirectory),
    image="futureoutlier/rawcontainer:0320",
    command=[
        "python",
        "return_flytedir.py",
        "/var/outputs/out",
    ],
)

flyte_file_out_only = ContainerTask(
    name="flyte_file_out_only",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    # inputs=kwtypes(inputs=FlyteDirectory),
    outputs=kwtypes(out=FlyteFile),
    image="futureoutlier/rawcontainer:0320",
    command=[
        "python",
        "return_flytefile.py",
        "/var/outputs/out",
    ],
)

@task
def flyte_file_task() -> FlyteFile:
    with open("./a.txt", "w") as file:
        file.write("This is a.txt file.")
    return FlyteFile(path="./a.txt")

@workflow
def flyte_file_io_wf() -> FlyteFile:
    ff = flyte_file_task()
    return flyte_file_io(inputs=ff)

@task
def flyte_dir_task() -> FlyteDirectory:
    from pathlib import Path
    import flytekit
    import os

    working_dir = flytekit.current_context().working_directory
    local_dir = Path(os.path.join(working_dir, "csv_files"))
    local_dir.mkdir(exist_ok=True)
    write_file = local_dir / "a.txt"
    with open(write_file, "w") as file:
        file.write("This is for flyte dir.")

    return FlyteDirectory(path=str(local_dir))

@workflow
def flyte_dir_io_wf() -> FlyteDirectory:
    fd = flyte_dir_task()
    return flyte_dir_io(inputs=fd)

if __name__ == "__main__":
    print(flyte_dir_io_wf())
    print(flyte_file_io_wf())
    print(primitive_types(a=0, b=False, c=3.0, d="hello", e=datetime.datetime.now(), 
                          f=datetime.timedelta(days=1, hours=3, minutes=2, seconds=3, microseconds=5)))
    print(flyte_dir_out_only())
    print(flyte_file_out_only())
```

Dockerfile
```Dockerfile
# Use the Alpine Linux version of Python 3.9 as the base image
FROM python:3.9-alpine

# Set the working directory in the container
WORKDIR /root

# You can use the docker copy command after building the image to copy test.py into the /app directory of the container
COPY ./write_flytefile.py /root/write_flytefile.py
COPY ./write_flytedir.py /root/write_flytedir.py
COPY ./return_same_value.py /root/return_same_value.py
COPY ./return_flytefile.py /root/return_flytefile.py
COPY ./return_flytedir.py /root/return_flytedir.py


# Specify the command to run when the container starts. Here, we use /bin/sh to start a shell.
CMD ["/bin/sh"]
```

`write_flytefile.py`
```python
import sys
from pathlib import Path

def copy_content_to_output(input_path: Path, output_path: Path):

    content = input_path.read_text()

    output_path.write_text(content)

if __name__ == "__main__":
    if len(sys.argv) > 2:
        input_path = Path(sys.argv[1])
        output_path = Path(sys.argv[2])
        copy_content_to_output(input_path, output_path)
    else:
        print("Usage: script.py <input_path> <output_path>")
```

`write_flytedir.py`
```python
import sys
from pathlib import Path
import shutil

def copy_directory(input_path: Path, output_path: Path):
    if not input_path.exists() or not input_path.is_dir():
        print(f"Error: {input_path} does not exist or is not a directory.")
        return

    try:
        shutil.copytree(input_path, output_path)
        print(f"Directory {input_path} was successfully copied to {output_path}")
    except Exception as e:
        print(f"Error copying {input_path} to {output_path}: {e}")

if __name__ == "__main__":
    if len(sys.argv) > 2:
        input_path = Path(sys.argv[1])
        output_path = Path(sys.argv[2])
        copy_directory(input_path, output_path)
    else:
        print("Usage: script.py <input_directory_path> <output_directory_path>")
```

`return_same_value.py`
```python
import sys

def write_output(output_dir, output_file, v):
    with open(f"{output_dir}/{output_file}", "w") as f:
            f.write(str(v))

def main(*args, output_dir):
    # Generate output files for each input argument
    for i, arg in enumerate(args, start=1):
        # Using i to generate filenames like 'a', 'b', 'c', ...
        output_file = chr(ord('a') + i - 1)
        write_output(output_dir, output_file, arg)

if __name__ == "__main__":
    *inputs, output_dir = sys.argv[1:]  # Unpack all inputs except for the last one for output_dir

    main(*inputs, output_dir=output_dir)
```

`return_flytefile.py`
```python
import sys
from pathlib import Path


def write_output(out_path: Path):
    out_path.write_text("return flyte file content")


if __name__ == "__main__":
    write_output(Path(sys.argv[1]))
```

`return_flytedir.py`
```python
import sys
from pathlib import Path

def write_output(directory_path: Path):
    directory_path.mkdir(parents=True, exist_ok=True)
    
    file_path = directory_path / "test_flyte_dir_file"
    
    file_path.write_text("prove flytedir works")

if __name__ == "__main__":
    if len(sys.argv) > 1:
        write_output(Path(sys.argv[1]))
    else:
        print("Usage: script.py <output_directory_path>")
```

### Screenshots
Support documentation example
image

All examples above
image

## Related PRs
https://github.com/flyteorg/flytekit/pull/1745

### DIFF
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -105,6 +105,7 @@ class ContainerTask(PythonTask):
             import docker
         except ImportError:
             raise ImportError(DOCKER_IMPORT_ERROR_MESSAGE)
+        import datetime
         import os
 
         from flytekit.core.promise import translate_inputs_to_literals
@@ -130,7 +131,7 @@ class ContainerTask(PythonTask):
             logger.error(msg)
             raise type(exc)(msg) from exc
 
-        container_output_dir = "/flyte/raw-container-task/output"
+        container_output_dir = "/flyte/raw-container-task/outputs"
         output_directory = ctx.file_access.get_random_local_directory()
         volume_bindings = {
             output_directory: {
@@ -144,7 +145,7 @@ class ContainerTask(PythonTask):
             for cmd in self._cmd:
                 if cmd.startswith("{{.inputs.") and cmd.endswith("}}"):
                     k = cmd[len("{{.inputs.") : -len("}}")]
-                    commands += str(native_inputs[k]) + " "
+                    commands += str(native_inputs[k]).split()[0] + " "
                 elif cmd == self._output_data_dir:
                     commands += container_output_dir + " "
                 else:
@@ -153,7 +154,7 @@ class ContainerTask(PythonTask):
             for arg in self._args:
                 if arg.startswith("{{.inputs.") and arg.endswith("}}"):
                     k = arg[len("{{.inputs.") : -len("}}")]
-                    commands += str(native_inputs[k]) + " "
+                    commands += str(native_inputs[k]).split()[0] + " "
                 elif arg == self._output_data_dir:
                     commands += container_output_dir + " "
                 else:
@@ -186,11 +187,15 @@ class ContainerTask(PythonTask):
             for k, output_type in self._outputs.items():
                 with open(os.path.join(output_directory, k), "r") as f:
                     output_val = f.read()
+
                 # bool('False') is True, so we need to handle this case
                 if output_type == bool:
                     output_dict[k] = output_val == "True"
+                elif output_type == datetime.datetime:
+                    output_dict[k] = datetime.datetime.fromisoformat(output_val)  # type: ignore
                 else:
                     output_dict[k] = output_type(output_val)
+
         outputs_literal_map = TypeEngine.dict_to_literal_map(ctx, output_dict)
         outputs_literals = outputs_literal_map.literals
         output_names = list(self.interface.outputs.keys())
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -121,6 +121,7 @@ class ContainerTask(PythonTask):
             msg = f"Failed to convert inputs of task '{self.name}':\n  {exc}"
             logger.error(msg)
             raise TypeError(msg) from exc
+
         input_literal_map = _literal_models.LiteralMap(literals=kwargs)
         try:
             native_inputs = self._literal_map_to_python_input(input_literal_map, ctx)
@@ -150,9 +151,21 @@ class ContainerTask(PythonTask):
                     commands += cmd + " "
         if self._args:
             for arg in self._args:
-                cmd += arg + " "
+                if arg.startswith("{{.inputs.") and arg.endswith("}}"):
+                    v = arg[len("{{.inputs.") : -len("}}")]
+                    commands += str(native_inputs[v]) + " "
+                elif arg == self._output_data_dir:
+                    commands += container_output_dir + " "
+                else:
+                    commands += arg + " "
 
         client = docker.from_env()
+
+        # If can't find the image, pull it
+        if client.images.list(filters={"reference": self._image}) == []:
+            logger.info(f"Pulling image:{self._image} for container task:{self.name}")
+            client.images.pull(self._image)
+
         container = client.containers.run(
             self._image,
             command=[
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -143,8 +143,8 @@ class ContainerTask(PythonTask):
         if self._cmd:
             for cmd in self._cmd:
                 if cmd.startswith("{{.inputs.") and cmd.endswith("}}"):
-                    v = cmd[len("{{.inputs.") : -len("}}")]
-                    commands += str(native_inputs[v]) + " "
+                    k = cmd[len("{{.inputs.") : -len("}}")]
+                    commands += str(native_inputs[k]) + " "
                 elif cmd == self._output_data_dir:
                     commands += container_output_dir + " "
                 else:
@@ -152,8 +152,8 @@ class ContainerTask(PythonTask):
         if self._args:
             for arg in self._args:
                 if arg.startswith("{{.inputs.") and arg.endswith("}}"):
-                    v = arg[len("{{.inputs.") : -len("}}")]
-                    commands += str(native_inputs[v]) + " "
+                    k = arg[len("{{.inputs.") : -len("}}")]
+                    commands += str(native_inputs[k]) + " "
                 elif arg == self._output_data_dir:
                     commands += container_output_dir + " "
                 else:
@@ -186,7 +186,11 @@ class ContainerTask(PythonTask):
             for k, output_type in self._outputs.items():
                 with open(os.path.join(output_directory, k), "r") as f:
                     output_val = f.read()
-                output_dict[k] = output_type(output_val)
+                # bool('False') is True, so we need to handle this case
+                if output_type == bool:
+                    output_dict[k] = output_val == "True"
+                else:
+                    output_dict[k] = output_type(output_val)
         outputs_literal_map = TypeEngine.dict_to_literal_map(ctx, output_dict)
         outputs_literals = outputs_literal_map.literals
         output_names = list(self.interface.outputs.keys())
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -1,20 +1,24 @@
 import typing
 from enum import Enum
-from typing import Any, Dict, List, Optional, OrderedDict, Type
+from typing import Coroutine, Dict, List, Optional, OrderedDict, Tuple, Type, Union
 
 from flytekit.configuration import SerializationSettings
 from flytekit.core.base_task import PythonTask, TaskMetadata
 from flytekit.core.context_manager import FlyteContext
 from flytekit.core.interface import Interface
 from flytekit.core.pod_template import PodTemplate
+from flytekit.core.promise import Promise, VoidPromise, create_task_output
 from flytekit.core.python_auto_container import get_registerable_container_image
 from flytekit.core.resources import Resources, ResourceSpec
 from flytekit.core.utils import _get_container_definition, _serialize_pod_spec
 from flytekit.image_spec.image_spec import ImageSpec
+from flytekit.loggers import logger
+from flytekit.models import literals as _literal_models
 from flytekit.models import task as _task_model
 from flytekit.models.security import Secret, SecurityContext
 
 _PRIMARY_CONTAINER_NAME_FIELD = "primary_container_name"
+DOCKER_IMPORT_ERROR_MESSAGE = "Docker is not installed. Please install Docker by running `pip install docker`."
 
 
 class ContainerTask(PythonTask):
@@ -82,6 +86,7 @@ class ContainerTask(PythonTask):
         self._args = arguments
         self._input_data_dir = input_data_dir
         self._output_data_dir = output_data_dir
+        self._outputs = outputs
         self._md_format = metadata_format
         self._io_strategy = io_strategy
         self._resources = ResourceSpec(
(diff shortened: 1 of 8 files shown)


kumare3:
Can we check if docker is installed and then use it please do not add docker dependency

Future-Outlier:
> Can we check if docker is installed and then use it please do not add docker dependency

No problem, will do that.

Future-Outlier:
cc @eapolinario , @pingsutw Could you please help review?
I think I might miss some edge cases.
I will add a unit test after you think is good, thank you!

flytekit/core/container_task.py
```diff
@@ -93,8 +98,90 @@ def __init__(
...
+                "bind": container_output_dir,
+                "mode": "rw",
+            },
+        }
+
+        commands = ""
+        if self._cmd:
+            for cmd in self._cmd:
+                if cmd.startswith("{{.inputs.") and cmd.endswith("}}"):
+                    v = cmd[len("{{.inputs.") : -len("}}")]
```
Future-Outlier, line 145: Is this implementation correct?
I am not sure how users define their inputs.

Future-Outlier:
cc @wild-endeavor If possible, please take a look too, thank you!

Future-Outlier:
Question:
1. can I add it in `dev-requirements.in`? so that we can add unit tests for local execution.
2. are we comfortable with `def local_execute`'s implementation?
I think it will be better to split it into small functions, however, most of implementation is referenced from here.
https://github.com/flyteorg/flytekit/blob/329aea8545aa9ea9f903acd7e006f3f12efb9cbb/flytekit/core/base_task.py#L251-L328
3. Are there any edge cases for container tasks?
I assume every input should be have prefix "{{.inputs." and suffix "}}"

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+            cmd_and_args += self._cmd
+        if self._args:
+            cmd_and_args += self._args
+
+        for cmd in cmd_and_args:
+            k = self._get_key_from_cmd(cmd)
+            if k:
+                if type(native_inputs[k]) in [FlyteFile, FlyteDirectory]:
+                    local_flyte_file_or_dir_path = str(native_inputs[k])
+                    remote_flyte_file_or_dir_path = os.path.join(self._input_data_dir, k.replace(".", "/"))  # type: ignore
```
eapolinario, line 187: can you add a few tests to cover this case? Specifically the use of `k.replace(...)`?

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+                    volume_bindings[local_flyte_file_or_dir_path] = {
+                        "bind": remote_flyte_file_or_dir_path,
+                        "mode": "rw",
+                    }
+                    commands += remote_flyte_file_or_dir_path + " "
+                else:
+                    commands += str(native_inputs[k]).split()[0] + " "
```
eapolinario, line 194: why do we need to split here?

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+        output_directory = ctx.file_access.get_random_local_directory()
+        volume_bindings = {
+            output_directory: {
+                "bind": self._output_data_dir,
+                "mode": "rw",
+            },
+        }
+
+        # Build the command string
+        commands = ""
```
eapolinario, line 175: let's make `commands` an array and use a join (i.e. `" ".join(commands)`) at the end to build the actual command.

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+                "-c",
+                commands,
+            ],
+            volumes=volume_bindings,
+            detach=True,
+        )
+        # Wait for the container to finish the task
+        container.wait()
+        container.stop()
+        container.remove()
```
eapolinario, line 217: we can use `remove=True` in the invocation of `run` instead.

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+                "sh",
+                "-c",
+                commands,
+            ],
+            volumes=volume_bindings,
+            detach=True,
+        )
+        # Wait for the container to finish the task
+        container.wait()
+        container.stop()
```
eapolinario, line 216: this is not neede as the previous call (i.e. the call to `wait`) only returns after the container stops.

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+            else:
+                commands += cmd + " "
+
+        client = docker.from_env()
+        # If we can't find the image, pull it
+        if client.images.list(filters={"reference": self._image}) == []:
+            logger.info(f"Pulling image:{self._image} for container task:{self.name}")
+            client.images.pull(self._image)
+
+        container = client.containers.run(
```
eapolinario, line 204: we should find a way to configure a timeout as well. This could be done in a separate PR, but can you leave a TODO here?

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+            command=[
+                "sh",
+                "-c",
+                commands,
+            ],
+            volumes=volume_bindings,
+            detach=True,
+        )
+        # Wait for the container to finish the task
+        container.wait()
```
eapolinario, line 215: we should find a way to configure a timeout as well. This could be done in a separate PR, but can you leave a TODO here?

eapolinario (commented):
First of all, thanks for you contribution! This will be a massive improvement to the local flytekit use cases.

This is shaping up to be a great PR. I left some comments along the way. But can you make sure to write tests where container tasks are involved in workflows where a combination of regular tasks and container tasks?

Future-Outlier:
> First of all, thanks for you contribution! This will be a massive improvement to the local flytekit use cases.
> 
> 
> 
> This is shaping up to be a great PR. I left some comments along the way. But can you make sure to write tests where container tasks are involved in workflows where a combination of regular tasks and container tasks?

Really great advice!
I'll do it today, thank you!

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+        except Exception as exc:
+            msg = f"Failed to convert inputs of task '{self.name}':\n  {exc}"
+            logger.error(msg)
+            raise type(exc)(msg) from exc
+
+        # Normalize the input and output directories
+        if self._input_data_dir:
+            self._input_data_dir = os.path.normpath(self._input_data_dir)
+        if self._output_data_dir:
+            self._output_data_dir = os.path.normpath(self._output_data_dir)
```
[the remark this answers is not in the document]
Future-Outlier: Yes, copilot will do it, I've tested it, and will ad tests to cover this.

flytekit/core/container_task.py
Future-Outlier: We need to split here because currently, strings such as `datetime.datetime` may contain blank spaces in some cases, causing the input argument to split from 1 to 2.
Perhaps we could consider not supporting `datetime` for now?
Alternatively, users might have to address this edge case on their own.


image

flytekit/core/container_task.py
Future-Outlier: nice advice, thank you

flytekit/core/container_task.py
Future-Outlier: nice advice, thank you

flytekit/core/container_task.py
Future-Outlier: No problem, thank you.

flytekit/core/container_task.py
Future-Outlier: nice catch, thank you

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+        # If we can't find the image, pull it
+        if client.images.list(filters={"reference": self._image}) == []:
+            logger.info(f"Pulling image:{self._image} for container task:{self.name}")
+            client.images.pull(self._image)
+
+        container = client.containers.run(
+            self._image,
+            command=[
+                "sh",
+                "-c",
```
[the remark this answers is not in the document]
Future-Outlier: Nice advice, we should use command directly.

flytekit/core/container_task.py
```diff
@@ -93,8 +99,148 @@ def __init__(
...
+            for k, output_type in self._outputs.items():
+                if output_type in [FlyteFile, FlyteDirectory]:
+                    output_dict[k] = output_type(path=os.path.join(output_directory, k))
+                else:
+                    with open(os.path.join(output_directory, k), "r") as f:
+                        output_val = f.read()
+
+                    # bool('False') is True, so we need to handle this case
+                    if output_type == bool:
+                        output_dict[k] = output_val == "True"
```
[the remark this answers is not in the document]
Future-Outlier: No problem, will do it

flytekit/core/container_task.py
Future-Outlier: update: will use an array as inputs, this could fix cases we need to split strings.

Future-Outlier:
@eapolinario 

These are my updates:
1. Command inputs now use an array, and this can make `str` don't need to be splitted.
2. Added support for `datetime.timedelta`.
3. Implemented regex for input parsing.
4. Updated all related tests.
5. Improved string to boolean conversion: `output_dict[k] = False if output_val.lower() == "false" else True`.


Thank you very much!

Future-Outlier:
Hi, @eapolinario 
Can you help approve this PR?
Companies like AXIS and Tesla might want this feature, thank you!

eapolinario (commented):
This is looking pretty good, but we should increase the test coverage.

Can you add a separate test suite (call it `test_local_raw_container.py` under `tests/flytekit/unit/core`) and add a Dockerfile there and use it to build a local image to be used in tests. Add a few more tests to show that the manipulation of input/output data dir works. Also, make sure to add examples where `FlyteFile` are used. This will help the implementation to catch up with the current state of copilot.

Future-Outlier:
> This is looking pretty good, but we should increase the test coverage.
> 
> 
> 
> Can you add a separate test suite (call it `test_local_raw_container.py` under `tests/flytekit/unit/core`) and add a Dockerfile there and use it to build a local image to be used in tests. Add a few more tests to show that the manipulation of input/output data dir works. Also, make sure to add examples where `FlyteFile` are used. This will help the implementation to catch up with the current state of copilot. 
> 
> 

I will do it today, thank you

Future-Outlier:
> This is looking pretty good, but we should increase the test coverage.
> 
> Can you add a separate test suite (call it `test_local_raw_container.py` under `tests/flytekit/unit/core`) and add a Dockerfile there and use it to build a local image to be used in tests. Add a few more tests to show that the manipulation of input/output data dir works. Also, make sure to add examples where `FlyteFile` are used. This will help the implementation to catch up with the current state of copilot.


Hi, @eapolinario ,

I've implemented your suggestions:
- Added 3 Python files for testing local execution of `FlyteFile`, `FlyteDirectory`, and primitive types within container tasks.
- Created a Dockerfile to include the above test files.
- Introduced 4 tests:
  1. `test_flytefile_wf` for `FlyteFile` I/O,
  2. `test_flytedir_wf` for `FlyteDirectory` I/O,
  3. `test_primitive_types_wf` for various primitive types,
  4. `test_input_output_dir_manipulation` to validate input/output directory handling.

Note: Despite `FlyteDirectory` not being supported by copilot, I've included it for local execution, ready to extend support based on user feedback.

If it looks good to you, can you help approve it?
This is a huge improvement to companies need `ContainerTaskl`, thank you!

flytekit/core/container_task.py
```diff
@@ -93,8 +99,175 @@ def __init__(
...
+                else:
+                    commands.append(str(native_inputs[k]))
+            else:
+                commands.append(cmd)
+
+        client = docker.from_env()
+        # If the specified image is not found locally, pull it from the registry
+        if client.images.list(filters={"reference": self._image}) == []:
+            logger.info(f"Pulling image:{self._image} for container task:{self.name}")
+            client.images.pull(self._image)
```
pingsutw, line 231: Won't `containers.run` pull the image automatically?

flytekit/core/container_task.py
```diff
@@ -93,8 +99,175 @@ def __init__(
...
+        output_dict = {}
+        if self._outputs:
+            for k, output_type in self._outputs.items():
+                if output_type in [FlyteFile, FlyteDirectory]:
+                    output_dict[k] = output_type(path=os.path.join(output_directory, k))
+                else:
+                    with open(os.path.join(output_directory, k), "r") as f:
+                        output_val = f.read()
+
+                    # bool('False') and bool('false') are True, so we need to handle this case
```
pingsutw, line 253:
```suggestion
```

flytekit/core/container_task.py
```diff
@@ -93,8 +99,175 @@ def __init__(
...
+                        output_dict[k] = output_type(output_val)
+
+        outputs_literal_map = TypeEngine.dict_to_literal_map(ctx, output_dict)
+        outputs_literals = outputs_literal_map.literals
+        output_names = list(self.interface.outputs.keys())
+        # Tasks that don't return anything still return a VoidPromise
+        if len(output_names) == 0:
+            return VoidPromise(self.name)
+        vals = [Promise(var, outputs_literals[var]) for var in output_names]
+        return create_task_output(vals, self.python_interface)
```
pingsutw, line 270: We should override `execute()` and return output_dict here. You don't need to create promises manually.

flytekit/core/container_task.py
```diff
@@ -93,8 +99,175 @@ def __init__(
...
+        # Build the command array
+        commands = []
+        cmd_and_args = []
+        if self._cmd:
+            cmd_and_args += self._cmd
+        if self._args:
+            cmd_and_args += self._args
+
+        for cmd in cmd_and_args:
+            k = self._get_key_from_cmd(cmd)
```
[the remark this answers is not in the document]
Future-Outlier: amazing advice, thank you.

flytekit/core/container_task.py
Future-Outlier: nice advice, thank you!!!

flytekit/core/container_task.py
Future-Outlier: Yes, but we want to let users know we are pulling since some images are large, which will spend much time.

eapolinario (approved):
So clean. Thank you!

Future-Outlier:
> So clean. Thank you!

Thank you both too😭😭😭

PR #2264, opened 2024-03-13
Add Python Dependency Deck

jasonlai1218:
In Python Dependency Deck, the version of the Python package currently being used can be displayed, which can make debugging easier.

- Add a class `PythonDependencyDeck` to deck.py
- Add a method `add_deck` to `FlyteContext` class

### Setup process
```
import flytekit

flytekit_version = "git+https://github.com/jasonlai1218/flytekit.git@add-py-deps-deck#commit=b5f49b4070621ba87224c9d223a6f7d710084673"
deck_plugin_version = "git+https://github.com/jasonlai1218/flytekit.git@add-py-deps-deck#commit=b5f49b4070621ba87224c9d223a6f7d710084673#subdirectory=plugins/flytekit-deck-standard"

image_spec = ImageSpec(
    packages=[flytekit_version, deck_plugin_version],
    apt_packages=["git"],
    registry="jasonlai1218",
    env={"FLYTE_SDK_LOGGING_LEVEL": "20"},
)


@task(container_image=image_spec, enable_deck=True)
def python＿dependency_deck() -> None:
    ctx = flytekit.FlyteContextManager.current_context()
    ctx.add_deck(PythonDependencyDeck())


@workflow
def wf() -> None:
    python＿dependency_deck()


if __name__ == "__main__":
    wf()
```

### Screenshots

Screenshot 2024-04-03 at 11 38 04 AM

click copy button and paste result:
> adlfs==2024.2.0
> aiobotocore==2.12.1
> aiohttp==3.9.3
> aioitertools==0.11.0
> aiosignal==1.3.1
> annotated-types==0.6.0
> arrow==1.3.0
> asttokens==2.4.1
> attrs==23.2.0
> azure-core==1.30.1
> azure-datalake-store==0.0.53
> azure-identity==1.15.0
> azure-storage-blob==12.19.1
> binaryornot==0.4.4
> botocore==1.34.51
> ...
> ...
> ...
> websocket-client==1.7.0
> widgetsnbextension==4.0.10
> wordcloud==1.9.3
> wrapt==1.16.0
> yarl==1.9.4
> ydata-profiling==4.7.0
> zipp==3.18.1

## Related PRs
https://github.com/flyteorg/flyteconsole/pull/852


flytekit/deck/deck.py
```diff
@@ -174,3 +174,39 @@ def get_deck_template() -> "Template":
...
+
+        import subprocess
+
+        import pandas as pd
+
+        try:
+            installed_packages = subprocess.check_output(["pip", "freeze"]).decode().split("\n")
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Error occurred while fetching installed packages: {e}")
+            return ""
```
pingsutw, line 204: Have you tried this? It's better to avoid creating a subprocess.
```python
from pip._internal.operations import freeze

x = freeze.freeze()
for dependency in x:
   print(dependency)
```

flytekit/deck/deck.py
```diff
@@ -174,3 +174,39 @@ def get_deck_template() -> "Template":
...
+            warning_info = "Plugin 'flytekit-deck-standard' is not installed. To display python dependency, install the plugin in the image."
+            logger.warning(warning_info)
+            return warning_info
+
+        import subprocess
+
+        import pandas as pd
+
+        try:
+            installed_packages = subprocess.check_output(["pip", "freeze"]).decode().split("\n")
```
thomasjpfan, line 201: To be safe, I usually like adding this:

```suggestion
            installed_packages = subprocess.check_output([sys.executable, "-m", "pip", "freeze"], text=True).split("\n")
```

Also `text=True` handles the encoding.

I think this is okay for pip packages, but `pip freeze` will not show the non-python dependencies for conda users.

thomasjpfan (commented):
Thank you for the PR!

flytekit/deck/deck.py
thomasjpfan: Using `pip._internal` itself seems a dangerous. Since it is internal, their API may break. I would not want to have to pin `pip` versions in `flytekit`.

Also using `pip` adds a direct dependency on `pip` itself. This is usually installed, but we should add it into our `pyproject.toml`.

flytekit/deck/deck.py
```diff
@@ -174,3 +174,39 @@ def get_deck_template() -> "Template":
...
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Error occurred while fetching installed packages: {e}")
+            return ""
+
+        columns_name = ["Library", "Version"]
+        installed_packages = [package.split("==") for package in installed_packages if package]
+        df = pd.DataFrame(installed_packages, columns=columns_name)
+        html = TableRenderer().to_html(df, header_labels=columns_name)
```
eapolinario, line 209: how about we simplify this to:
```suggestion
        try:
            installed_packages = json.loads(subprocess.check_output(["pip", "list", "--format", "json"]))
        except subprocess.CalledProcessError as e:
            logger.error(f"Error occurred while fetching installed packages: {e}")
            return ""

        df = pd.DataFrame(installed_packages)
        html = TableRenderer().to_html(df)
```

flytekit/deck/deck.py
```diff
@@ -174,3 +174,39 @@ def get_deck_template() -> "Template":
...
+
+    def __init__(self, name: Optional[str] = "Python Dependency", html: Optional[str] = ""):
+        super().__init__(name, html)
+
+    @property
+    def html(self) -> str:
+        try:
+            from flytekitplugins.deck.renderer import TableRenderer
+        except ImportError:
+            warning_info = "Plugin 'flytekit-deck-standard' is not installed. To display python dependency, install the plugin in the image."
```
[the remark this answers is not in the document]
jasonlai1218: good advice

flytekit/deck/deck.py
jasonlai1218: updated

jasonlai1218: I like this simple way

flytekit/deck/deck.py
```diff
@@ -174,3 +174,29 @@ def get_deck_template() -> "Template":
...
     )
     return env.get_template("template.html")
+
+
+class PythonDependencyDeck(Deck):
+    """
+    PythonDependencyDeck is a deck that contains information about the library name and version.
+    """
+
+    def __init__(self, name: Optional[str] = "Python Dependency", html: Optional[str] = ""):
```
thomasjpfan, line 184: Nit: Usually there is more than one dependency

```suggestion
    def __init__(self, name: Optional[str] = "Python Dependencies", html: Optional[str] = ""):
```

flytekit/deck/deck.py
```diff
@@ -174,3 +174,29 @@ def get_deck_template() -> "Template":
...
+    @property
+    def html(self) -> str:
+        import json
+        import subprocess
+
+        import pandas as pd
+        from flytekitplugins.deck.renderer import TableRenderer
+
+        try:
+            installed_packages = json.loads(subprocess.check_output(["pip", "list", "--format", "json"]))
```
thomasjpfan, line 196: This is the [recommended way](https://pip.pypa.io/en/stable/user_guide/#using-pip-from-your-program) of calling into `pip`: 

```suggestion
            installed_packages = json.loads(subprocess.check_output([sys.executable, "-m", "pip", "list", "--format", "json"]))
```

EngHabu, line 196: Should that be configurable? not everybody has/uses `pip`

thomasjpfan: I think this PR is already a good first step. If one uses `conda`, `pip` will still pick up all the Python dependencies. The next step is to grab the non-Python `conda` dependencies as well.

I do not think we need to make it configurable. For example, Weights and Bias will automatically pick up all the `pip` dependencies and `conda` dependencies and generate a `requirements.txt` or `environment.yaml`. This allows the user to easily recreate the environment.

thomasjpfan:
From https://github.com/flyteorg/flytekit/pull/2264#discussion_r1524000584, I think it'll be more useful to show `pip freeze` directly and do not table formatting. For me, the use case is to copy out the `requirements.txt`, so I can recreate the environment. If the requirements were rendered as a table, it'll be harder to recreate the environment.

flytekit/deck/deck.py
```diff
@@ -174,3 +174,29 @@ def get_deck_template() -> "Template":
...
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Error occurred while fetching installed packages: {e}")
+            return ""
+        df = pd.DataFrame(installed_packages)
+        html = TableRenderer().to_html(df)
```
eapolinario, line 201: Let's make this html a tad bit more complicated so that we can also solve the reproduction use-case (i.e. copy list of dependencies as a requirements.txt file), by doing something like:

```suggestion
        table = TableRenderer().to_html(df)
        html = f"""
        <!DOCTYPE html>
        <html lang="en">
        <head>
        <meta charset="UTF-8">
        <meta name="viewport" content="width=device-width, initial-scale=1.0">
        <title>Flyte Dependencies</title>
        <script>
        async function copyTable() {
        var table = document.getElementsByTagName('table')[0];
        var requirements_txt = '';

        for (var i = 0, row; row = table.rows[i]; i++) {
            if (i == 0) {
            continue;
            }
            requirements_txt += row.cells[0].innerText+""+row.cells[1].innerText+'\n';
        }

        try {
            await navigator.clipboard.writeText(requirements_txt);
        } catch (err) {
            console.log('Oops, unable to copy');
        }
        }
        </script>
        </head>
        <body>

        <svg width="100px" height="100px" viewBox="0 0 24 24" onclick="copyTable()" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
            <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
                <g fill="#212121" fill-rule="nonzero">
                    <path d="M5.50280381,4.62704038 L5.5,6.75 L5.5,17.2542087 C5.5,19.0491342 6.95507456,20.5042087 8.75,20.5042087 L17.3662868,20.5044622 C17.057338,21.3782241 16.2239751,22.0042087 15.2444057,22.0042087 L8.75,22.0042087 C6.12664744,22.0042087 4,19.8775613 4,17.2542087 L4,6.75 C4,5.76928848 4.62744523,4.93512464 5.50280381,4.62704038 Z M17.75,2 C18.9926407,2 20,3.00735931 20,4.25 L20,17.25 C20,18.4926407 18.9926407,19.5 17.75,19.5 L8.75,19.5 C7.50735931,19.5 6.5,18.4926407 6.5,17.25 L6.5,4.25 C6.5,3.00735931 7.50735931,2 8.75,2 L17.75,2 Z M17.75,3.5 L8.75,3.5 C8.33578644,3.5 8,3.83578644 8,4.25 L8,17.25 C8,17.6642136 8.33578644,18 8.75,18 L17.75,18 C18.1642136,18 18.5,17.6642136 18.5,17.25 L18.5,4.25 C18.5,3.83578644 18.1642136,3.5 17.75,3.5 Z">

        </path>
        {table}

        </body>
        </html>
        """
```


This renders like:
Screenshot 2024-03-13 at 10 11 21 PM

The idea being that if you click on the icon you copy the list of requirements in the format that `requirements.txt` expects. Obviously someone who knows a lot more of html than me can get this to a better state (maybe just turn that into a button that says "Copy in requirements.txt format"?).

eapolinario (commented):
> From [#2264 (comment)](https://github.com/flyteorg/flytekit/pull/2264#discussion_r1524000584), I think it'll be more useful to show `pip freeze` directly and do not table formatting. For me, the use case is to copy out the `requirements.txt`, so I can recreate the environment. If the requirements were rendered as a table, it'll be harder to recreate the environment.

why-not-both-why-not

We can render the nice table and provide a little copy button that copies in the `requirements.txt` style.

flytekit/deck/deck.py
thomasjpfan: > (maybe just turn that into a button that says "Copy in requirements.txt format"?).

I like this option more. The button is a bit too abstract for me.

eapolinario: Not amazing, but if we swap the `<svg>` tag with:
```
<button onclick="copyTable()">
  <span>Copy table as requirements.txt</span>
  </button>
```

then this renders like:
Screenshot 2024-03-14 at 8 47 17 AM

thomasjpfan: That is great! My only nit is to have some margin under the button.

As for implementation, I prefer to not parse the HTML table and include the raw `pip freeze` in the HTML and hide it. Then, the button will just need to read the text. (Also less javascript code for us to maintain)

eapolinario: I put the new version of the html template in https://gist.github.com/eapolinario/dc45c349a582d287c1b2a77ac83a7fce

This assumes that there's a python variable called `requirements_txt` that will be passed to this string.

jasonlai1218: updated

jasonlai1218: I found that when using iframe, `iframe` needs to add the allow attribute
`allow="clipboard-write"`, so that the copy can be successful only after clicking the button.

jasonlai1218: Results pasted into txt file, file content:
> adlfs==2024.2.0
> aiobotocore==2.12.1
> aiohttp==3.9.3
> aioitertools==0.11.0
> aiosignal==1.3.1
> ...
> ...
> ...
> wrapt==1.16.0
> yarl==1.9.4
> ydata-profiling==4.6.5
> zipp==3.18.1

jasonlai1218: PR for adding attributes to iframe: https://github.com/flyteorg/flyteconsole/pull/852

flytekit/deck/deck.py
```diff
@@ -174,3 +174,72 @@ def get_deck_template() -> "Template":
...
+        try:
+            installed_packages = json.loads(
+                subprocess.check_output([sys.executable, "-m", "pip", "list", "--format", "json"])
+            )
+            requirements_txt = ""
+
+            for package_info in installed_packages:
+                package_name = package_info["name"]
+                package_version = package_info["version"]
+                requirements_txt += f"{package_name}=={package_version}\n"
```
eapolinario, line 205: what if we just run `pip freeze` instead?

```suggestion
            requirements_txt = subprocess.check_output(["pip", "freeze"]).decode("utf-8").replace("\\n", "\n")
```

jasonlai1218: updated

flytekit/deck/deck.py
jasonlai1218: updated

flytekit/deck/deck.py
jasonlai1218: updated

flytekit/deck/deck.py
```diff
@@ -174,3 +174,66 @@ def get_deck_template() -> "Template":
...
+        super().__init__(name, html)
+
+    @property
+    def html(self) -> str:
+        import json
+        import subprocess
+        import sys
+
+        import pandas as pd
+        from flytekitplugins.deck.renderer import TableRenderer
```
thomasjpfan, line 194: Can we use `flytekit.deck.MarkdownRenderer` instead? This would make it possible to not require `flytekitplugins-deck-standard` to get the dependency information. `flytekitplugins-deck-standard` is a pretty heavy package because of all it's dependencies.

flytekit/core/context_manager.py
```diff
@@ -741,6 +741,17 @@ def get_deck(self) -> typing.Union[str, "IPython.core.display.HTML"]:  # type:ig
 
         return _get_deck(typing.cast(ExecutionState, self.execution_state).user_space_params)
 
+    def add_deck(self, deck: Deck) -> None:
```
thomasjpfan, line 744: Although this is interesting to have, can this method be removed from this PR? It's orthogonal to adding a Python Dependency Deck and decks are automatically added to the context during `Deck.__init__`:

https://github.com/flyteorg/flytekit/blob/3642ec6771985f504c2b6a9dfd2598a84219ac58/flytekit/deck/deck.py#L58

As for this PR, I'll want to include the deck automatically when `enable_deck=True` and add `PythonDependencyDeck` here:

https://github.com/flyteorg/flytekit/blob/3642ec6771985f504c2b6a9dfd2598a84219ac58/flytekit/core/python_function_task.py#L351

flytekit/deck/deck.py
jasonlai1218: I use `MarkdownRenderer` instead of `TableRenderer`
https://github.com/flyteorg/flytekit/blob/0091f25e8bceb48513271a2e5266535114061485/flytekit/deck/deck.py#L207

Deck:
Screenshot 2024-03-25 at 6 05 34 PM

when the version is too long, the layout becomes confusing

eapolinario: if we use the `MarkdownRenderer` we lose the control over the overall html (so this idea of adding a button that copies the contents of a hidden div won't work). That said, the `TableRenderer` only [depends on pandas](https://github.com/flyteorg/flytekit/blob/d23eb440a9cdb730e3802d7fa93b5aafd0670779/plugins/flytekit-deck-standard/flytekitplugins/deck/renderer.py#L135-L177) and could easily be moved to core.

flytekit/core/context_manager.py
jasonlai1218: Good advice, thanks

jasonlai1218: updated

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,66 @@ def to_html(self, source_code: str) -> str:
...
+
+    def __init__(self, title: str = "Python Dependencies"):
+        self._title = title
+
+    def to_html(self) -> str:
+        import json
+        import subprocess
+        import sys
+
+        from flytekitplugins.deck.renderer import TableRenderer
```
eapolinario, line 104: This is not correct. My [suggestion](https://github.com/flyteorg/flytekit/pull/2264#discussion_r1538289174) was to move `TableRenderer` to deck/renderer.py and then use it here.

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,66 @@ def to_html(self, source_code: str) -> str:
         css = formatter.get_style_defs(".highlight").replace("#fff0f0", "#ffffff")
         html = highlight(source_code, PythonLexer(), formatter)
         return f"<style>{css}</style>{html}"
+
+
+class PythonDependencyRenderer:
+    """
+    PythonDependencyDeck is a deck that contains information about the library name and version.
```
eapolinario, line 93:
```suggestion
    PythonDependencyDeck is a deck that contains information about packages installed via pip.
```

flytekit/core/python_function_task.py
```diff
@@ -354,10 +354,14 @@ def _write_decks(self, native_inputs, native_outputs_as_map, ctx, new_user_param
...
-            from flytekit.deck.renderer import SourceCodeRenderer
+            from flytekit.deck.renderer import PythonDependencyRenderer, SourceCodeRenderer
 
             source_code_deck = Deck("Source Code")
             renderer = SourceCodeRenderer()
             source_code_deck.append(renderer.to_html(source_code))
 
+            python_dependencies_deck = Deck("Python Dependencies")
+            renderer = PythonDependencyRenderer()
+            python_dependencies_deck.append(renderer.to_html())
```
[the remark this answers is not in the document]
thomasjpfan: I am -1 on this PRs approach of moving `TableRenderer` into core which requires pandas.

I would use `flytekit.deck.render.MarkdownRenderer` to render a markdown table, which does not need pandas at all. (`MarkdownRenderer` does not require any extra dependencies)

flytekit/deck/renderer.py
eapolinario: @jasonlai1218 , sorry for flip-flopping on this. One hard requirement that we have is that we shouldn't require pandas in flytekit core. In order to make this happen, let's revert the move of `TableRenderer` from the plugin and use the `MarkdownRenderer` (which is already in core) in  the new `PythonDependencyRenderer`. We'll have to build the markdown as the table, so something like:

```
...
installed_packages = json.loads(
                subprocess.check_output([sys.executable, "-m", "pip", "list", "--format", "json"])
            )

markdown_table = "| Name | Version |\n| --- | --- |\n"

# Append each dictionary's values to the markdown string
for entry in installed_packages:
    markdown_table += f"| {entry['name']} | {entry['version']} |\n"

...

table = MarkdownRenderer().to_html(markdown_table)
...
```

flytekit/core/python_function_task.py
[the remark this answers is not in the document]
eapolinario: @thomasjpfan and I sync'd offline to align on this requirement. I provided a path forward in a [separate](https://github.com/flyteorg/flytekit/pull/2264#discussion_r1543575447) comment.

flytekit/deck/renderer.py
jasonlai1218: It's OK. We want the basic core to be lightweight, but also rich in functionality, so we need to find an optimal balance. It was a great conversation, and I learned a lot. Thanks again.

jasonlai1218: I use `<br>` instead of `\n` , because `\n` not work in html
https://github.com/flyteorg/flytekit/blob/f731757836b2b4ab83946cf52aeb98386d51cdd8/flytekit/deck/renderer.py#L115-L117

result:
Screenshot 2024-03-29 at 12 45 25 PM

and `copy button` is work

jasonlai1218: or maybe we can show this

> adlfs==2024.2.0
> aiobotocore==2.12.1
> aiohttp==3.9.3
> aioitertools==0.11.0
> aiosignal==1.3.1
> annotated-types==0.6.0
> arrow==1.3.0
> asttokens==2.4.1
> attrs==23.2.0
> azure-core==1.30.1
> azure-datalake-store==0.0.53
> azure-identity==1.15.0
> azure-storage-blob==12.19.1
> binaryornot==0.4.4
> botocore==1.34.51

jasonlai1218: I adjusted it again. 
Current results:
Screenshot 2024-03-29 at 2 29 20 PM

tests/flytekit/unit/deck/test_deck.py
```diff
@@ -176,3 +187,15 @@ def test_source_code_renderer():
...
     assert "#fff0f0" not in result
+
+
+def test_python_dependency_renderer():
+    renderer = PythonDependencyRenderer()
+    result = renderer.to_html()
+
+    # Assert that the result includes parts of the python dependency
+    assert "name" in result
+    assert "version" in result
```
thomasjpfan, line 198: According to [the CI](https://github.com/flyteorg/flytekit/actions/runs/8477979862/job/23229643233?pr=2264), `assert "version" in result"` is failing.

flytekit/core/python_function_task.py
```diff
@@ -349,15 +349,20 @@ def dynamic_execute(self, task_function: Callable, **kwargs) -> Any:
...
-            source_code_deck = Deck("Source Code")
-            renderer = SourceCodeRenderer()
-            source_code_deck.append(renderer.to_html(source_code))
+                source_code_deck = Deck("Source Code")
+                renderer = SourceCodeRenderer()
+                source_code_deck.append(renderer.to_html(source_code))
+
+                python_dependencies_deck = Deck("Python Dependencies")
+                renderer = PythonDependencyRenderer()
+                python_dependencies_deck.append(renderer.to_html())
```
thomasjpfan, line 366: Since the `supress` is only for `getsource`:

```python
if self._disable_deck is False:
    from flytekit.deck import Deck

    # These errors are raised if the source code can not be retrieved
    with suppress(OSError, TypeError):
        source_code = inspect.getsource(self._task_function)
        from flytekit.deck.renderer import SourceCodeRenderer

        source_code_deck = Deck("Source Code")
        renderer = SourceCodeRenderer()
        source_code_deck.append(renderer.to_html(source_code))

    python_dependencies_deck = Deck("Python Dependencies")
    renderer = PythonDependencyRenderer()
    python_dependencies_deck.append(renderer.to_html())

```

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,73 @@ def to_html(self, source_code: str) -> str:
...
+                subprocess.check_output([sys.executable, "-m", "pip", "list", "--format", "json"])
+            )
+            requirements_txt = subprocess.check_output(["pip", "freeze"]).decode("utf-8").replace("\\n", "\n").rstrip()
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Error occurred while fetching installed packages: {e}")
+            return ""
+
+        markdown_table = (
+            "<table>\n<tr>\n<th style='text-align:left;'>Name</th>\n<th style='text-align:left;'>Version</th>\n</tr>\n"
+        )
```
thomasjpfan, line 117: Since this is pure HTML, I do not think we need to use `MarkdownRenderer` at all.

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,73 @@ def to_html(self, source_code: str) -> str:
...
+        import subprocess
+        import sys
+
+        from flytekit.loggers import logger
+
+        try:
+            installed_packages = json.loads(
+                subprocess.check_output([sys.executable, "-m", "pip", "list", "--format", "json"])
+            )
+            requirements_txt = subprocess.check_output(["pip", "freeze"]).decode("utf-8").replace("\\n", "\n").rstrip()
```
thomasjpfan, line 110: To be safe:

```suggestion
            requirements_txt = subprocess.check_output([sys.executable, "-m", "pip", "freeze"]).decode("utf-8").replace("\\n", "\n").rstrip()
```

flytekit/core/python_function_task.py
eapolinario: that's a nice touch.

flytekit/deck/renderer.py
eapolinario: good call.

jasonlai1218: updated

flytekit/core/python_function_task.py
jasonlai1218: good advice, thanks

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,77 @@ def to_html(self, source_code: str) -> str:
...
+            )
+            requirements_txt = (
+                subprocess.check_output([sys.executable, "-m", "pip", "freeze"])
+                .decode("utf-8")
+                .replace("\\n", "\n")
+                .rstrip()
+            )
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Error occurred while fetching installed packages: {e}")
+            return ""
```
thomasjpfan, line 118: Having an empty HTML feels like a bad dev experience. I think it's worth having an error message:

```suggestion
            return "Error occurred while fetching installed packages."
```

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,77 @@ def to_html(self, source_code: str) -> str:
...
+            await navigator.clipboard.writeText(requirements_txt.innerText);
+          }} catch (err) {{
+            console.log('Error accessing the clipboard: ' + err);
+          }}
+        }}
+        </script>
+        </head>
+        <body>
+
+        <button onclick="copyTable()">
```
thomasjpfan, line 150: @eapolinario This copy button does not work. I get this error in the console:

Screenshot 2024-04-02 at 3 14 42 PM

Screenshot 2024-04-02 at 3 15 38 PM

If we can not get this to work, I'm okay with removing the copy button and just showing the table.

flytekit/core/python_function_task.py
```diff
@@ -349,15 +349,21 @@ def dynamic_execute(self, task_function: Callable, **kwargs) -> Any:
...
+                from flytekit.deck.renderer import SourceCodeRenderer
+
+                source_code_deck = Deck("Source Code")
+                renderer = SourceCodeRenderer()
+                source_code_deck.append(renderer.to_html(source_code))
 
-            source_code_deck = Deck("Source Code")
-            renderer = SourceCodeRenderer()
-            source_code_deck.append(renderer.to_html(source_code))
+            python_dependencies_deck = Deck("Python Dependencies")
```
thomasjpfan, line 365: Nit: Making this shorter looks better in the UI:

```suggestion
            python_dependencies_deck = Deck("Dependencies")
```


Screenshot 2024-04-02 at 3 21 58 PM

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,77 @@ def to_html(self, source_code: str) -> str:
...
         html = highlight(source_code, PythonLexer(), formatter)
         return f"<style>{css}</style>{html}"
+
+
+class PythonDependencyRenderer:
+    """
+    PythonDependencyDeck is a deck that contains information about packages installed via pip.
+    """
+
+    def __init__(self, title: str = "Python Dependencies"):
```
thomasjpfan, line 96:
```suggestion
    def __init__(self, title: str = "Dependencies"):
```

flytekit/deck/renderer.py
```diff
@@ -86,3 +86,77 @@ def to_html(self, source_code: str) -> str:
...
+          }}
+        }}
+        </script>
+        </head>
+        <body>
+
+        <button onclick="copyTable()">
+          <span>Copy table as requirements.txt</span>
+        </button>
+        <br><br><br>
```
thomasjpfan, line 153:
```suggestion
        <h3>Python Dependencies</h3>
```

flytekit/deck/renderer.py
eapolinario: This is blocked on https://github.com/flyteorg/flyteconsole/pull/852.

eapolinario: That just got merged, so next Flyte release is going to have this.

flytekit/deck/renderer.py
jasonlai1218: agree

flytekit/deck/renderer.py
jasonlai1218: Screenshot 2024-04-03 at 11 38 04 AM

thomasjpfan (approved):
Assuming https://github.com/flyteorg/flytekit/pull/2264/files#r1548469361 will get resolved with a new flyte release, this PR LGTM

eapolinario:
FYI: I pushed a [commit](https://github.com/flyteorg/flytekit/pull/2264/commits/814a952edd8bd7a71bce02f57ca5a04c1a2fe1e6) to handle tests that require `hypothesis` slightly differently. Now we run them via a separate make target (called `unit_test_hypothesis`) where we control some of the hyperparameters like `dealine` and `max_examples` using a profile (as per the [docs](https://hypothesis.readthedocs.io/en/latest/settings.html#hypothesis.settings.load_profile)).

eapolinario (approved):

PR #2266, opened 2024-03-14
Create envd context for users

pingsutw:
## Tracking issue
NA

When pushing an image to the local registry, Users have to create a new envd context by themself.
if they terminate the sandbox, they must switch to another context.
```
envd context create --name flyte-sandbox --builder tcp --builder-address localhost:30003 --use
```

Create a new context for users when they're trying to push an image to the local registry.

## How was this patch tested?
Local

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -140,7 +140,7 @@ class ImageSpec:
 
     def with_commands(self, commands: Union[str, List[str]]) -> "ImageSpec":
         """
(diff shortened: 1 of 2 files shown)


Future-Outlier (approved):
LGTM, it works to me.
image

I've used `envd context rm --name flyte-sandbox` and then built the image.

### after using local registry.
image

### after using docker remote registry.
image

wild-endeavor (approved):

PR #2270, opened 2024-03-15
Migrate over to using datetime.now(timezone.utc)

thomasjpfan:

`datetime.utcnow()` is deprecated:

> DeprecationWarning: datetime.datetime.utcnow() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.now(datetime.UTC).

This PR migrates over to using `datetime.now(timezone.utc)` everywhere. (We can not use `datetime.UTC` yet because it was introduced in requires Python >= 3.11.)

## How was this patch tested?

This PR does not change functionally, thus the existing test should cover the changes.

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -249,7 +249,7 @@ def setup_execution(
             domain=exe_domain,
             name=exe_name,
         ),
-        execution_date=_datetime.datetime.utcnow(),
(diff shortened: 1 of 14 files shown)


pingsutw (dismissed):
LGTM, but integration test is failing

pingsutw (approved):

PR #2279, opened 2024-03-18
Stop requiring users to import `dataclasses_json` or `DataClassJSONMixin` for dataclass

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4486
For a better user experience.

1. use `mashumaro>=3.11`, so that we can use `JSONEncoder` and `JSONDecoder`
2. ~~change `python_val.to_json()` to `JSONEncoder(python_type).encode(python_val)`~~
3. ~~change `expected_python_type.from_json(json_str)` to `JSONDecoder(expected_python_type).decode(json_str)`~~
4. ~~change `return dataclass_json(dataclasses.make_dataclass(schema_name, attribute_list))` to
`return dataclasses.make_dataclass(schema_name, attribute_list)`, 
since we don't need `to_json` method and `from_json` method anymore.~~
5. add tests
6. fix mypy errors
7. change type annotations
8. ~~remove `flytekit-dolt` from CI test, since it doesn't work now and needs to be implemented a new version.~~

9. use `JSONEncoder` and `JSONDecoder` to convert `dataclass` to `json str` and convert `json str` to `dataclass` when 
the user didn't use `dataclasses_json` and `DataClassJSONMixin`.
10. add an encoder registry and a decoder registry to cache `JSONEncoder` and `JSONDecoder` when using `List[dataclass]`
11. add a benchmark test in real case scenario by dynamic workflow and return `List[dataclass]`

## How was this patch tested?
1. unit tests
3. local and remotely with only `dataclass` decorator
4. local and remotely with dataclass inherits from `DataClassJSONMixin`. (for backward compatible)
5.  local and remotely with `dataclass_json`  decorator. (for backward compatible)

Note: you can use `futureoutlier/dataclass:0321` this image to test it.

### Setup process
```bash
python dataclass_example.py
```
```bash
pyflyte run --remote --image localhost:30000/dataclass:0951 dataclass_example.py dataclass_wf --x 10 --y 20
```
```python
import os
import tempfile
from typing import Tuple, List, Optional
import pandas as pd
from flytekit.types.structured import StructuredDataset
# from mashumaro.mixins.json import DataClassJSONMixin

@dataclass
class Datum:
    x: int
    y: str
    z: dict[int, int]
    w: List[int] = None

@task
def stringify(s: int) -> Datum:
    """
    A dataclass return will be treated as a single complex JSON return.
    """
    return Datum(x=s, y=str(s), z={s: str(s)}, w=[s,s,s,s])

@task
def add(x: Datum, y: Datum) -> Datum:
    """
    Flytekit automatically converts the provided JSON into a data class.
    If the structures don't match, it triggers a runtime failure.
    """
    x.z.update(y.z)
    return Datum(x=x.x + y.x, y=x.y + y.y, z=x.z, w=x.w + y.w)

@dataclass
class FlyteTypes:
    dataframe: StructuredDataset
    file: FlyteFile
    directory: FlyteDirectory

@task
def upload_data() -> FlyteTypes:
    """
    Flytekit will upload FlyteFile, FlyteDirectory and StructuredDataset to the blob store,
    such as GCP or S3.
    """
    # 1. StructuredDataset

    # 2. FlyteDirectory
    temp_dir = tempfile.mkdtemp(prefix="flyte-")
    df.to_parquet(temp_dir + "/df.parquet")

    # 3. FlyteFile
    file_path = tempfile.NamedTemporaryFile(delete=False)
    file_path.write(b"Hello, World!")

    fs = FlyteTypes(
        dataframe=StructuredDataset(dataframe=df),
        file=FlyteFile(file_path.name),
        directory=FlyteDirectory(temp_dir),
    )
    return fs


@task
def download_data(res: FlyteTypes):
    assert pd.DataFrame({"Name": ["Tom", "Joseph"], "Age": [20, 22]}).equals(res.dataframe.open(pd.DataFrame).all())
    f = open(res.file, "r")
    assert f.read() == "Hello, World!"
    assert os.listdir(res.directory) == ["df.parquet"]

@workflow
def dataclass_wf(x: int, y: int) -> Tuple[Datum, FlyteTypes]:
    o1 = add(x=stringify(s=x), y=stringify(s=y))
    o2 = upload_data()
    download_data(res=o2)
    return o1, o2

if __name__ == "__main__":
    print(dataclass_wf(x=10, y=20))
```
```Dockerfile
FROM python:3.9-slim-buster
USER root
WORKDIR /root
ENV PYTHONPATH /root
RUN apt-get install git -y

RUN pip install -U git+https://github.com/flyteorg/flytekit.git@30223e45c6b773cb25846f5031f92e4f1f783c33
RUN pip install pandas -U
```

### Screenshots
local execution (with only dataclass decorator)

image

local execution (with `DataClassJSONMixin`)

image

remote execution (with only dataclass decorator)

image

image

remote execution (with `DataClassJSONMixin`)

image

image

remote execution (with `dataclass_json`)

image


flytekit/core/type_engine.py
```diff
@@ -729,8 +720,9 @@ def _fix_dataclass_int(self, dc_type: Type[DataClassJsonMixin], dc: DataClassJso
         # https://developers.google.com/protocol-buffers/docs/reference/google.protobuf#google.protobuf.Value
         # Thus we will have to walk the given dataclass and typecast values to int, where expected.
         for f in dataclasses.fields(dc_type):
-            val = dc.__getattribute__(f.name)
-            dc.__setattr__(f.name, self._fix_val_int(f.type, val))
+            val = getattr(dc, f.name)
+            setattr(dc, f.name, self._fix_val_int(f.type, val))
```
Future-Outlier, line 714: For ignoring mypy error.

Future-Outlier:
cc @thomasjpfan Please take a look, thank you!

Future-Outlier:
I think we can remove all  `DataClassJSONMixin ` class which has also `@dataclass` decorator.
Should we do this in this PR or create another?

image

Future-Outlier:
> I think we can remove all `DataClassJSONMixin ` class which has also `@dataclass` decorator. Should we do this in this PR or create another?
> 
> image

I think it will be ok to open another housekeeping PR to do this since we need to also update relevant tests.
And the change will be super huge and time-consuming for review.

Future-Outlier:
## Test Performance
I use the code above to test the performance compared using this PR (only need `dataclass` decorator), 
(use mashumaro `DataClassJSONMixin`) and using `@dataclass_json`.

```python
def test_speed():
    import time
    start_time = time.time()
    for _ in range(1000):
        dataclass_wf(x=10, y=20)
    end_time = time.time()
    print(f"Time taken: {end_time - start_time}")
```

image

I think the main reason for this performance gap is that we need to create a `JsonEncoder` or a ` JsonDecoder` to serialize and deserialize our dataclasses. However, I still think this PR is worth merging. If we compare them in only 1 iteration, the gap is very small and won't pose a performance issue

Fatal1ty:
> I think the main reason for this performance gap is that we need to create a JsonEncoder or a  JsonDecoder to serialize and deserialize our dataclasses.

I’m sure it is. Creating decoders and encoders is not a cheap operation. I would recommend to use a registry dataclass_type -> decoder(encoder) if this is an issue for you.

Future-Outlier:
> > I think the main reason for this performance gap is that we need to create a JsonEncoder or a  JsonDecoder to serialize and deserialize our dataclasses.
> 
> I’m sure it is. Creating decoders and encoders is not a cheap operation. I would recommend to use a registry dataclass_type -> decoder(encoder) if this is an issue for you.

I love your idea, but in the real case, we will need to repeat register dataclass type encoder and decoder in every pod task node.
So maybe this solution will be better, which means still faster.
However, I really think that it is worth to use `JsonEncoder` and `JsonDecoder`.
The main performance issue is not because of serialization, or in other words, we can optimize other places.

flytekit/core/type_engine.py
```diff
@@ -739,13 +721,9 @@ def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type:
...
             )
-        if not self.is_serializable_class(expected_python_type):
-            raise TypeTransformerFailedError(
-                f"Dataclass {expected_python_type} should be decorated with @dataclass_json or mixin with DataClassJSONMixin to be "
-                f"serialized correctly"
-            )
+
         json_str = _json_format.MessageToJson(lv.scalar.generic)
-        dc = expected_python_type.from_json(json_str)  # type: ignore
+        dc = JSONDecoder(expected_python_type).decode(json_str)
```
thomasjpfan, line 726: Same comment about `to_json` applies here.

flytekit/core/type_engine.py
```diff
@@ -486,14 +472,9 @@ def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], exp
...
-        if not self.is_serializable_class(type(python_val)):
-            raise TypeTransformerFailedError(
-                f"Dataclass {python_type} should be decorated with @dataclass_json or inherit DataClassJSONMixin to be "
-                f"serialized correctly"
-            )
-        self._serialize_flyte_type(python_val, python_type)
 
-        json_str = python_val.to_json()  # type: ignore
+        self._serialize_flyte_type(python_val, python_type)
+        json_str = JSONEncoder(python_type).encode(python_val)
```
thomasjpfan, line 477: If the `expected_python_type` defines a `to_json` I think we should still call it:

```suggestion
        if hasattr(python_val, "to_json"):
            dc = python_val.to_json()
        else:
            # Use JSONEncoder
```

There could be users that define their own custom `to_json` themselves or through a mixin, and we need to make sure that still works.

thomasjpfan (commented):
@Future-Outlier Thank you for looking into the performance! Although, I do not think a running the same workflow 1000 times on the same Python process represents a typical use case.

For local execution, I think a dataclass can be deserialized a few times, so I am okay with the suggestion from https://github.com/flyteorg/flytekit/pull/2279#issuecomment-2012057417 to add a registry. For benchmarking, I'll want to see how this PR performs with a dynamic workflow that spawns 100 tasks, which is more typical use case. I feel like we are store the registry in `DataclassTransformer` itself:

```python
def to_literal(...):
    
    try:
        encoder = self._dataclass_to_encoder[python_type]
    except KeyError:
        encoder = JsonEncoder(python_type)
        self._dataclass_to_encoder = encoder
```


For remote execution, each task is it in it's own container, so they will all have to create their own deserializers or serializers. Can you run a benchmark with dynamic workflow that spawns ~50 tasks with remote? (~50 task because remote will take longer)

For both benchmarks, we can assume that `DataClassJSONMixin` and `dataclass_json` are the same performance-wise, thus we only need to compare `DataClassJSONMixin` and with this PR. 

In either case, I am okay with not requiring `dataclass_json` or the `DataClassJSONMixin`, but I want to make sure we do not introduce a regression. After this PR, users should still be able to use `DataClassJSONMixin` and get the same performance as before.

flytekit/core/type_engine.py
Future-Outlier: Really nice suggestion and I feel that I've learned a lot

eapolinario: Can you make sure to add a few tests to cover this scenario, @Future-Outlier ?

Future-Outlier: No problem, nice advice, thank you so much

Future-Outlier:
@thomasjpfan , @Fatal1ty 
I use this example to test the speed of workflow, it seems that the performance is close : ) 

```python
@task
def create_dataclasses() -> List[Datum]:
    return [Datum(x=1, y="1", z={1: 1}, w=[1,1,1,1])]

@task
def concat_dataclasses(x: List[Datum], y: List[Datum]) -> List[Datum]:
    return x + y

@dynamic
def dynamic_wf() -> List[Datum]:
    all_dataclasses = [Datum(x=1, y="1", z={1: 1}, w=[1,1,1,1])]
    for _ in range(300):
        data = create_dataclasses()
        all_dataclasses = concat_dataclasses(x=all_dataclasses, y=data)
    return all_dataclasses

@workflow
def benchmark_workflow() -> List[Datum]:
    return dynamic_wf()

if __name__ == "__main__":
    import time
    start_time = time.time()
    benchmark_workflow()
    end_time = time.time()
    print(f"Time taken: {end_time - start_time}")
```

image

flytekit/core/type_engine.py
Future-Outlier: I've added tests to cover this scenario and added the registry!
If possible, please take a look again!

Future-Outlier: I've also updated the description, if there's anything unclear, feel free to ping me, thank you all!

Future-Outlier:
> > I think the main reason for this performance gap is that we need to create a JsonEncoder or a  JsonDecoder to serialize and deserialize our dataclasses.
> 
> I’m sure it is. Creating decoders and encoders is not a cheap operation. I would recommend to use a registry dataclass_type -> decoder(encoder) if this is an issue for you.

I've tested it on your advice, and it really reduces the time drastically.
image

flytekit/core/type_engine.py
```diff
@@ -486,14 +474,15 @@ def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], exp
...
+
         self._serialize_flyte_type(python_val, python_type)
 
-        json_str = python_val.to_json()  # type: ignore
+        if hasattr(python_val, "to_json"):
+            json_str = python_val.to_json()
+        else:
+            if not self._encoder.get(python_type):
+                self._encoder[python_type] = JSONEncoder(python_type)
+            json_str = self._encoder[python_type].encode(python_val)
```
thomasjpfan, line 485: Nit: Usually Pythonic code prefers to "Ask for forgiveness"

```suggestion
            try:
                encoder = self._encoder[python_type]
            except KeyError:
                encoder = JSONEncoder(python_type)
                self._encoder[python_type] = encoder
            json_str = encoder.encode(python_val)
```

REF: https://devblogs.microsoft.com/python/idiomatic-python-eafp-versus-lbyl/

flytekit/core/type_engine.py
```diff
@@ -739,13 +729,15 @@ def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type:
...
-            )
+
         json_str = _json_format.MessageToJson(lv.scalar.generic)
-        dc = expected_python_type.from_json(json_str)  # type: ignore
+
+        if hasattr(expected_python_type, "from_json"):
+            dc = expected_python_type.from_json(json_str)  # type: ignore
+        else:
+            if not self._decoder.get(expected_python_type):
+                self._decoder[expected_python_type] = JSONDecoder(expected_python_type)
```
thomasjpfan, line 739: Same here regular `try: except KeyError:`

flytekit/core/type_engine.py
```diff
@@ -720,7 +701,7 @@ def _fix_val_int(self, t: typing.Type, val: typing.Any) -> typing.Any:
 
         return val
 
-    def _fix_dataclass_int(self, dc_type: Type[DataClassJsonMixin], dc: DataClassJsonMixin) -> DataClassJsonMixin:
+    def _fix_dataclass_int(self, dc_type: Type, dc: dataclasses.dataclass) -> dataclasses.dataclass:  # type: ignore
```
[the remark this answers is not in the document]
thomasjpfan: Can `dc_type` be `Type[dataclasses.dataclass]`? (I think this is required for `dataclasses.fields` to work.)

Given how dynamic the code is, I think we have to go with `dc: typing.Any` for now.

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -2499,16 +2497,29 @@ class DatumMashumaro(DataClassJSONMixin):
...
-    class Datum(DataClassJSONMixin):
+    class DatumDataclassJson(DataClassJSONMixin):
         x: int
         y: Color
 
-    transformer = DataclassTransformer()
+    @dataclass
+    class DatumDataclass:
+        x: int
+        y: Color
```
thomasjpfan, line 2507: Can you add another test to see what happens with `typing.Union`? Specifically:

```python
@dataclass
class DatumDataUnion:
    path: typing.Union[str, os.PathLike]
```

If the `path` started out as a `pathlib.Path(...)`, does it deserialize into a string or a `Path` object?

thomasjpfan (commented):
Overall, this looks good. I am okay with the current scope. We can update the other parts of the codebase in follow up PRs.

wild-endeavor:
is this backwards compatible?  serialize with an old flytekit release, with old user code.  then deserialize with new flytekit and new user code.

flytekit/core/type_engine.py
```diff
@@ -486,14 +474,15 @@ def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], exp
...
-            raise TypeTransformerFailedError(
-                f"Dataclass {python_type} should be decorated with @dataclass_json or inherit DataClassJSONMixin to be "
-                f"serialized correctly"
-            )
+
         self._serialize_flyte_type(python_val, python_type)
 
-        json_str = python_val.to_json()  # type: ignore
+        if hasattr(python_val, "to_json"):
+            json_str = python_val.to_json()
```
thomasjpfan, line 481: @wild-endeavor With this check, we are backward compatible. If one used `DataClassJsonMixin` or `@dataclass_json`, then `to_json` is called, which matches the original behavior. XREF: https://github.com/lidatong/dataclasses-json/blob/8512afc0a87053dbde52af0519c74198fa3bb873/dataclasses_json/api.py#L26

@Future-Outlier Can you include a comment here about how this preserves backward compatibility?

flytekit/core/type_engine.py
Future-Outlier: thank you for this nice advice, I will also take notes about it!

tests/flytekit/unit/core/test_type_engine.py
Future-Outlier: I found that TypeTransformer can't handle support for typing.Union[str, os.PathLike] or typing.Union[str, FlyteFile].
(I thought that `os.PathLike` is decided not supported here.
https://docs.flyte.org/en/latest/api/flytekit/generated/flytekit.types.file.FlyteFile.html#flytekit.types.file.FlyteFile.path

Support `typing.Union[str, FlyteFile]` in flyte could be an enhancement.
For more details, I need to trace more code to find the core reason why we can support these cases now, but it definitely is not because of this PR.

flytekit/core/type_engine.py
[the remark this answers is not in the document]
Future-Outlier: You're right!
Have updated your advice, thank you.

flytekit/core/type_engine.py
Future-Outlier: Hi, @wild-endeavor and @thomasjpfan 

In flytekit-python, we use the function `to_json` to serialize dataclasses and `from_json` to deserialize them. Previously, we required users in older flytekit releases to add the`dataclass_json` decorator or inherit from the `DataclassJsonMixin` class.
This was because both provided the necessary methods to serialize (convert a dataclass to bytes) and deserialize (convert bytes back to a dataclass).

In this PR, the `mashumaro` module introduces two classes, `JSONEncoder` and `JSONDecoder`, for serializing and deserializing dataclasses.
These new classes eliminate the reliance on the `to_json` and `from_json` methods.

Initially, we used `if hasattr(python_val, "to_json"):` to check for the method's presence.
Therefore, introducing these changes will not cause any breaking changes.
This means that dataclasses inheriting from `DataclassJsonMixin` will continue to use `to_json` and `from_json` for serialization and deserialization when using this version of flytekit.


to_json REF: https://github.com/flyteorg/flytekit/blob/master/flytekit/core/type_engine.py#L498
from_json REF: https://github.com/flyteorg/flytekit/blob/master/flytekit/core/type_engine.py#L750

thomasjpfan: To clarify, I wanted to have a short comment in the code that states how the "to_json" check, helps preserves backward compatibility.

tests/flytekit/unit/core/test_type_engine.py
thomasjpfan: Hmm, thank you for checking. When I suggested `typing.Union[str, os.PathLike]`, I was thinking about `FlyteFile`. But `FlyteFile` has it's own type transformer, so it's okay.

In any case, I'm okay with the current scope of this PR.

thomasjpfan (dismissed):
I think not requiring `dataclasses_json` or `DataClassJSONMixin` for many use cases is already a net improvement.

LGTM

Future-Outlier:
> I think not requiring `dataclasses_json` or `DataClassJSONMixin` for many use cases is already a net improvement.
> 
> LGTM

Added comments, thank you so much

pingsutw (approved):

PR #2280, opened 2024-03-18
pyflyte run spark task

pingsutw:
## Tracking issue
NA

Since only the Spark driver downloads the workflow code, pyflyte run doesn't work with Spark tasks as the Spark executors also need the workflow code.

Always copy the code into the image if the task is a spark task.

## How was this patch tested?
remote cluster

```bash
pyflyte run --remote flyte-example/test2.py spark_wf
```

### Setup process
```python
import random

from flytekitplugins.spark import Spark

import flytekit
from flytekit import ImageSpec, task, Resources, workflow
from operator import add

custom_image = ImageSpec(
    name="flytekit",
    registry="pingsutw",  # your docker registry
)


def f(_):
    x = random.random() * 2 - 1
    y = random.random() * 2 - 1
    return 1 if x**2 + y**2 <= 1 else 0


@task(
    task_config=Spark(
        spark_conf={
            "spark.driver.memory": "1000M",
            "spark.executor.memory": "1000M",
            "spark.executor.cores": "1",
            "spark.executor.instances": "2",
            "spark.driver.cores": "1",
            "spark.jars": "https://storage.googleapis.com/hadoop-lib/gcs/gcs-connector-hadoop3-latest.jar",
        },
    ),
    limits=Resources(mem="2000M"),
    container_image=custom_image,
)
def hello_spark(partitions: int) -> float:
    print("Starting Spark with Partitions: {}".format(partitions))

    n = 1 * partitions
    sess = flytekit.current_context().spark_session
    count = (
        sess.sparkContext.parallelize(range(1, n + 1), partitions).map(f).reduce(add)
    )

    pi_val = 4.0 * count / n
    return pi_val


@workflow
def spark_wf(partitions: int = 10) -> float:
    return hello_spark(partitions=partitions)


if __name__ == '__main__':
    runner = CliRunner()
    result = runner.invoke(pyflyte.main,
                           ["register",
                            "--non-fast",
                            "--version",
                            "v3",
                            "/Users/kevin/git/flytekit/flyte-example/test2.py",
                            ])
    print(result.output)
```

### Screenshots
image

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -112,14 +112,18 @@ class ContainerTask(PythonTask):
             io_strategy=self._io_strategy.value if self._io_strategy else None,
         )
 
+    def _get_image(self, settings: SerializationSettings) -> str:
+        if settings.fast_serialization_settings is None or not settings.fast_serialization_settings.enabled:
+            if isinstance(self._image, ImageSpec):
+                # Set the source root for the image spec if it's non-fast registration
+                self._image.source_root = settings.source_root
+        return get_registerable_container_image(self._image, settings.image_config)
+
     def _get_container(self, settings: SerializationSettings) -> _task_model.Container:
         env = settings.env or {}
         env = {**env, **self.environment} if self.environment else env
-        if isinstance(self._image, ImageSpec):
-            if settings.fast_serialization_settings is None or not settings.fast_serialization_settings.enabled:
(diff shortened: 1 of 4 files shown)


Future-Outlier (approved):
TEST

eapolinario (approved):

PR #2288, opened 2024-03-20
Setting interruptible on ArrayNode sub node metadata

hamersaw:
## Tracking issue
_NA_

If a task is labeled as `interruptible` that attribute is not persisted when executing within an ArrayNode.

Maintain the `interruptible` setting from the internal task in the sub node metadata.

## How was this patch tested?
locally

## Related PRs
_NA_

## Docs link
_NA_

### DIFF
--- a/flytekit/core/array_node_map_task.py
+++ b/flytekit/core/array_node_map_task.py
@@ -113,9 +113,9 @@ class ArrayNodeMapTask(PythonTask):
 
     def construct_node_metadata(self) -> NodeMetadata:
         # TODO: add support for other Flyte entities
-        interruptible=None
+        interruptible = None
         if self._run_task.metadata:
-            interruptible=self._run_task.metadata.interruptible
+            interruptible = self._run_task.metadata.interruptible
 
         return NodeMetadata(
             name=self.name,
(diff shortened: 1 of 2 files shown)


pingsutw (dismissed):
nit: there is a lint error

pingsutw (dismissed):

flytekit/core/array_node_map_task.py
```diff
@@ -113,8 +113,13 @@ def python_interface(self):
 
     def construct_node_metadata(self) -> NodeMetadata:
         # TODO: add support for other Flyte entities
+        interruptible = None
+        if self._run_task.metadata:
+            interruptible = self._run_task.metadata.interruptible
+
         return NodeMetadata(
             name=self.name,
+            interruptible=interruptible,
```
eapolinario, line 122: Should we also copy other values from task metadata?

eapolinario (commented):
Can you add a few unit tests to https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/core/test_array_node_map_task.py ?

flytekit/core/array_node_map_task.py
hamersaw: This is all kinds of weird. So I added the cache metadata to NodeMetadata to support subworkflows / launchplans. IMO caching should have been set at the Node level, but for task executions its embedded in the TaskTemplate, so not necessary here. 

cc @pvditt can you expand on your other PR (since closed) to update the timeouts here.

wild-endeavor (approved):

PR #2298, opened 2024-03-27
feat: Support pep604 union operator

ggydush:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/4706

Python 3.10 introduced the ability to write union types using | rather than importing from typing (e.g., int | str instead of typing.Union[int, str]), and I believe this should be supported. See [PEP 604](https://peps.python.org/pep-0604/).

## How was this patch tested?
Added unit tests and registered a large code base with the updated Flytekit and tested a workflow on remote. Ran a small test locally on Python 3.8 too to ensure correct error is bubbled up:
```console
│ ❱  5 def process(a: int | None) -> int:                                                                                                                                      │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
```

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -9,6 +9,7 @@ import inspect
 import json
 import json as _json
 import mimetypes
+import sys
 import textwrap
 import typing
 from abc import ABC, abstractmethod
@@ -30,7 +31,7 @@ from typing_extensions import Annotated, get_args, get_origin
 from flytekit.core.annotation import FlyteAnnotation
 from flytekit.core.context_manager import FlyteContext
 from flytekit.core.hash import HashMethod
-from flytekit.core.type_helpers import load_type_from_tag
+from flytekit.core.type_helpers import convert_pep604_union_type, load_type_from_tag
 from flytekit.core.utils import timeit
 from flytekit.exceptions import user as user_exceptions
 from flytekit.interaction.string_literals import literal_map_string_repr
@@ -58,6 +59,11 @@ from flytekit.models.types import LiteralType, SimpleType, StructuredDatasetType
 T = typing.TypeVar("T")
 DEFINITIONS = "definitions"
 TITLE = "title"
+
+if sys.version_info >= (3, 10):
+    from types import UnionType as _UnionType
+else:
+    _UnionType = typing.Union
 
 
 class BatchSize:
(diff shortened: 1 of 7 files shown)


flytekit/core/type_engine.py
```diff
@@ -694,7 +696,7 @@ def _fix_val_int(self, t: typing.Type, val: typing.Any) -> typing.Any:
         if val is None:
             return val
 
-        if get_origin(t) is typing.Union and type(None) in get_args(t):
+        if is_union_type(t) and type(None) in get_args(t):
```
pingsutw, line 699: ditto

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -1794,6 +1802,9 @@ def test_union_of_lists():
 def test_list_of_unions():
```
pingsutw, line 1802: Should we run this test only when Python version >= 3.10?

```python
@pytest.mark.skipif(sys.version_info < (3, 10))
def test_optional_type():
  ...
```

pingsutw: We have another cron job that runs every day. https://github.com/flyteorg/flytekit/blob/a150899c8ac8b49cc7fed28c8771a2c0039f9681/.github/workflows/pythonbuild.yml#L11

ggydush: gtk, yeah I can skip all of the test with `|` syntax

tests/flytekit/unit/core/test_interface.py
```diff
@@ -213,6 +213,18 @@ def z(
     assert not params.parameters["c"].required
     assert params.parameters["c"].default.scalar.none_type == Void()
 
+    def z(a: int | None = None, b: str | None = None, c: typing.List[int] | None = None) -> typing.Tuple[int, str]:
```
pingsutw, line 216: Should we skip this test as well?

pingsutw (approved):

tests/flytekit/unit/core/test_interface.py
eapolinario: we should split this portion of the `test_parameters_and_defaults` test into its own test.

It's not failing in CI now because we only [run](https://github.com/flyteorg/flytekit/blob/master/.github/workflows/pythonbuild.yml#L28-L32) on python 3.12 in PRs. This test would start to fail in the nightly runs.

eapolinario (commented):
This is awesome!

flytekit/core/type_engine.py
```diff
@@ -26,6 +27,7 @@
 from marshmallow_enum import EnumField, LoadDumpOptions
 from mashumaro.mixins.json import DataClassJSONMixin
 from typing_extensions import Annotated, get_args, get_origin
+from typing_inspect import is_union_type
```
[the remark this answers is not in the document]
ggydush: added!

eapolinario (dismissed):

eapolinario (approved):

PR #2299, opened 2024-03-27
Add flytekit-omegaconf plugin

mg515:

Flytekit natively supports serialization of many data types for exchanging information between tasks.

The Flytekit OmegaConf Plugin extends these by the `DictConfig` type from the [OmegaConf package](https://omegaconf.readthedocs.io/) as well as related types that are being used by the [hydra package](https://hydra.cc/) for configuration management.

## Task example
```
from omegaconf import DictConfig
import flytekitplugins.omegaconf  # noqa F401


@dataclass
class MySimpleConf:
    _target_: str = "lightning_module.MyEncoderModule"
    learning_rate: float = 0.0001


@task
def my_task(cfg: DictConfig) -> None:
    print(f"Doing things with {cfg.learning_rate=}")


@workflow
def pipeline(cfg: DictConfig) -> None:
    my_task(cfg=cfg)


if __name__ == "__main__":
    from omegaconf import OmegaConf

    cfg = OmegaConf.structured(MySimpleConf)
    pipeline(cfg=cfg)
```

## How was this patch tested?

Unit tests are added within the same PR.

Additionally, the plugin has been in private use for a while and is tested within multiple projects.

## Acknowledgements
The first version of the plugin was developed roughly at the end of year 2021 by some of the current and former members of [Merantix Momentum](https://www.merantix-momentum.com/). Shoutout to @SebS94 @fg91 @AdemFr @YmirKhang @axkoenig @adrianloy @ThomasWollmann


SebS94 (approved):
Thanks for taking care of this Miha!

One small consideration - would it be more fitting to rename this plugin into flytekit-omegaconf as this is what it actually serialises? Obviously, our main consideration is the combination with hydra, but there is no actual direct dependency here.

mg515:
@SebS94 was having the same thoughts myself as well, could as well do it yeah.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/flytekit_patch.py
```diff
@@ -0,0 +1,64 @@
...
+import inspect
+import logging
+from typing import Type, get_args
+
+from flytekit.core.type_engine import TypeEngine, TypeTransformer, is_annotated
+
+logger = logging.getLogger()
+
+
+def iterate_get_transformers(python_type: Type) -> TypeTransformer:
```
eapolinario, line 11: can you expand on why we need this?

eapolinario (commented):
I did one pass. It's looking pretty good. I left a few comments, but no showstopper. I worry about conflicts with other plugins, like pydantic. We don't necessarily have to fix everything in this PR.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,196 @@
...
+from flytekit.models.types import LiteralType, SimpleType
+from omegaconf import DictConfig, OmegaConf
+
+logger = logging.getLogger(__name__)
+
+T = TypeVar("T")
+NoneType = type(None)
+
+
+def is_flatable(d: typing.Union[DictConfig]) -> bool:
```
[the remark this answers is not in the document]
mg515: Would assume it's a forgotten artefact, changed to `DictConfig` type now

plugins/flytekit-omegaconf/setup.py
```diff
@@ -0,0 +1,39 @@
...
+    install_requires=plugin_requires,
+    license="apache2",
+    python_requires=">=3.8",
+    classifiers=[
+        "Intended Audience :: Science/Research",
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Programming Language :: Python :: 3.8",
+        "Programming Language :: Python :: 3.9",
+        "Programming Language :: Python :: 3.10",
```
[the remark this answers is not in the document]
mg515: Ran tests locally with both of those python versions now, passes through. Will add them to `setup.py`

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,196 @@
...
+                            cfg_dict[key] = transformer.to_python_value(ctx, value_literal, node_type)
+                            break
+                        except Exception:
+                            logger.debug(
+                                f"Serialisation with transformer {type(transformer)} failed:\n"
+                                f"{traceback.format_exc()}\n\n"
+                            )
+                            transformation_logs += (
+                                f"Deserialisation with transformer {type(transformer)} failed:\n"
+                                f"{traceback.format_exc()}\n\n"
```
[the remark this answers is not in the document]
mg515: makes sense

plugins/flytekit-omegaconf/README.md
```diff
@@ -0,0 +1,79 @@
...
+SharedConfig.set_mode(OmegaConfTransformerMode.DictConfig)
+```
+
+```note
+Since the DictConfig is flattened and keys transformed into dot notation, the keys of the DictConfig must not contain
+dots.
+```
+
+```note
+Warning: This plugin overwrites the default serializer for Enum-objects to also allow for non-string-valued enum definitions.
```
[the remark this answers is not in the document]
mg515: My guess would be that natively, only `(str, Enum)` can be serialized.
This transformer enables (de)serialization of enums with any type value.

@SebS94 Anything to add?

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/flytekit_patch.py
mg515: @SebS94 What's the reason again we can't directly identify the right transformer, but need to try them out one by one?

SebS94: The main reason is serialisation of objects with multi-inheritance or Union types. The current flytekit implementation will pick the first matching TypeTransformer and fail if it is unable to serialise the type correctly. Instead we want to try out all viable Transformers and only fail if none of them work.

We encounter these cases quite frequently when building (structured) configs with hydra where nodes will accept different subconfig types.

plugins/flytekit-omegaconf/README.md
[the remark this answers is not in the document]
SebS94: Nothing real to add - this is the main reason. We construct omegaconfs with hydra where quite often the enums are not string valued but still of types that can be serialised with existing Typetransformers.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/flytekit_patch.py
mg515: Thanks, added this information to the docstring

mg515:
Somehow the test builds started to fail after an innocent commit. 

@pingsutw I saw that you tried to restart the checks already, but no success. Any idea what's going on?

plugins/flytekit-omegaconf/README.md
```diff
@@ -0,0 +1,79 @@
...
+from flytekit import task, workflow
+from omegaconf import DictConfig
+
+@dataclass
+class MySimpleConf:
+    _target_: str = "lightning_module.MyEncoderModule"
+    learning_rate: float = 0.0001
+
+@task
+def my_task(cfg: DictConfig) -> None:
```
fg91, line 21: If the user would specify `MySimpleConf` here, the dict config transformer wouldn't be used, right?

plugins/flytekit-omegaconf/README.md
```diff
@@ -0,0 +1,79 @@
...
+`Auto` - This mode will try to deserialize according to the Dataclass mode and fall back to the DictConfig mode if the
+dataclass definition is not available. This is the default mode.
+
+To set the mode either initialise the transformer with the `mode` argument or set the mode of the config directly:
+
+```python
+from flytekitplugins.omegaconf.config import SharedConfig, OmegaConfTransformerMode
+from flytekitplugins.omegaconf import DictConfigTransformer
+
+# Set the mode directly on the transformer
```
fg91, line 59: I find it untypical that users import type transformers to configure their behaviour. Would the user have to register `transformer_slim` in this example with the `TypeEngine` again?

Would it be ok for you to remove/not document this option but use the auto mode by default and only document how it can be overridden by configuring the dict config itself?

plugins/flytekit-omegaconf/README.md
```diff
@@ -0,0 +1,79 @@
...
+transformer_slim = DictConfigTransformer(mode=OmegaConfTransformerMode.DictConfig)
+
+# Set the mode directly in the config
+SharedConfig.set_mode(OmegaConfTransformerMode.DictConfig)
+```
+
+```note
+Since the DictConfig is flattened and keys transformed into dot notation, the keys of the DictConfig must not contain
+dots.
```
fg91, line 68: Would it be caught if the user tried to do that? If yes, maybe we can remove this sentence here.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/config.py
```diff
@@ -0,0 +1,30 @@
+from enum import Enum
+
+
+class OmegaConfTransformerMode(Enum):
+    """
+    Operation Mode indicating whether a (potentially unannotated) DictConfig object or a structured config using the
+    underlying dataclass is returned.
+
+    Note: We define a single shared configs across all transformers as recursive calls should refer to the same config
```
fg91, line 9:
```suggestion
    Note: We define a single shared config across all transformers as recursive calls should refer to the same config
```

plugins/flytekit-omegaconf/README.md
fg91: Or is `SharedConfig` a singleton that is used to control the behaviour globally?

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,194 @@
...
+from flytekit.models.types import LiteralType, SimpleType
+from omegaconf import DictConfig, OmegaConf
+
+logger = logging.getLogger("flytekit")
+
+T = TypeVar("T")
+NoneType = type(None)
+
+
+def is_flatable(d: DictConfig) -> bool:
```
fg91, line 30: Nit (non-blocking):
Should maybe called `is_flattenable`.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,194 @@
...
+        Convert from given python type object ``DictConfig`` to the Literal representation.
+
+        Since the DictConfig type does not offer additional type hints for its nodes, typing information is stored
+        within the literal itself rather than the Flyte LiteralType.
+        """
+        # instead of raising TypeError, raising AssertError so that flytekit can catch it in
+        # https://github.com/flyteorg/flytekit/blob/60c982e4b065fdb3aba0b957e506f652a2674c00/flytekit/core/
+        # type_engine.py#L1222
+        assert isinstance(python_val, DictConfig), f"Invalid type {type(python_val)}, can only serialise DictConfigs"
+        assert is_flatable(
```
fg91, line 80: Mabye it would be better to explicitly raise in case users use `python -O` which strips asserts.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/flytekit_patch.py
fg91: I know this used to be a problem for string enums where the type engine happened to try the string transformer first, found a match, and then picked this one even though from a users perspective the enum transformer would have been a better pick.

I fixed this special case a while ago (see [here](https://github.com/flyteorg/flytekit/blob/94a48ae86d9b6924bd01c106b493fe50a5169ebb/flytekit/core/type_engine.py#L973)).

Are there other cases where this is required or could it be that this logic here predates the fix I mentioned above and might actually not be required anymore now?

plugins/flytekit-omegaconf/README.md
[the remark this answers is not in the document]
fg91: I see why one would in some situations want such an enum transformer that does not require a `(str, Enum)` type but I also wonder whether all users of this omegaconf plugin also want to have the default enum transformer overridden as a side-effect.

Do you think it would be feasible to split out this enum transformer into a separate plugin?

fg91 (commented):
Thank you for opening this PR 🙏 

I talked about this flyte<->hydra integration [at one of the flyte community syncs](https://www.youtube.com/watch?v=tghvVvHJi7s) >2 years ago and multiple people have since asked in the Flyte slack community whether this has been open sourced. So I definitely see a need for this plugin.

I didn't yet review the details of the (de-) serialization but I feel we should first discuss three more high level points:

* The UX of how the “mode” is configured (see [this](https://github.com/flyteorg/flytekit/pull/2299/files#r1595731270) comment).
* Should the enum transformer potentially be split out of this plugin into a separate one? (See [this](https://github.com/flyteorg/flytekit/pull/2299/files#r1595771994) comment).
* Is `iterate_get_transformers` still required (see [this](https://github.com/flyteorg/flytekit/pull/2299/files#r1595763416) comment) and if yes, what are use cases for this? Are these use cases so common that it’s worth having this logic in the open-source version of the plugin or are they maybe so specific to your use cases that this logic should maybe be omitted in the open source version?

plugins/flytekit-omegaconf/README.md
mg515: Yes, correct

mg515:
> Is iterate_get_transformers still required?

I'm not sure at this point. I tried falling back to use `TypeEngine.get_transformer()` and could not break it with the current set of test configs. It makes me believe that maybe iterating was implemented more to accommodate serialization of mixed Enums than anything else.

@SebS94 See commit https://github.com/flyteorg/flytekit/pull/2299/commits/debe2f7c3bf91dd118c1145af9b7a0d17dc6c95b . Can you help me find a small example that would break this implementation? I thought having a nested config would already be enough.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
mg515: addressed in https://github.com/flyteorg/flytekit/pull/2299/commits/4fdb7e54e775cc16c5c2c510fd89a4b29135438f

plugins/flytekit-omegaconf/README.md
mg515: Would prefer to keep it in, despite

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
mg515: agreed, addressed in https://github.com/flyteorg/flytekit/pull/2299/commits/08f9a8092adc6943e13c474250c72ceb36f96a45

plugins/flytekit-omegaconf/README.md
mg515: > Do I understand it correctly that these are two mechanisms that achieve exactly the same thing?

I think the second example configures the mode of the (dict) transformer globally, whereas the first one only configures the specific instance.

Either way, I can't think of a use-case where this would really be useful, I think we can configure it globally across all transformers, the way you suggested. Any objections @SebS94 ?
Will come back to you once I implement it (ran out of time now).

plugins/flytekit-omegaconf/README.md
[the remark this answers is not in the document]
fg91: @mg515 and I discussed this offline and it seems overriding the enum transformer is made obsolete by [this](https://github.com/flyteorg/flytekit/pull/2004) fix I made in the `TypeEngine` a while ago. Since you removed the enum type transformer from this PR, could you please remove this warning here as well?

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,178 @@
...
+            OmegaConf.is_missing(d, k)  # values are either MISSING ...
+            or not isinstance(d[k], DictConfig)  # ... not nested Dictionaries ...
+            or is_flattenable(d[k])
+        )  # or flatable themselves
+        for k in d.keys()
+    )
+
+
+class DictConfigTransformer(TypeTransformer[DictConfig]):
+    def __init__(self, mode: typing.Optional[OmegaConfTransformerMode] = None):
```
fg91, line 43: As discussed [here](https://github.com/flyteorg/flytekit/pull/2299/files#r1660227529), shell we remove the `mode` here and only provide a global mode?

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,178 @@
...
+                    transformer = TypeEngine.get_transformer(node_type)
+                    value_literal = Literal.from_flyte_idl(ParseDict(nested_dict["values"][key], PB_Literal()))
+                    cfg_dict[key] = transformer.to_python_value(ctx, value_literal, node_type)
+
+                    if key not in cfg_dict.keys():
+                        raise ValueError(
+                            f"Could not identify matching transformer for object {nested_dict['values'][key]} of "
+                            f"proposed type {node_type}. This may either indicate that no such transformer "
+                            "exists or the appropriate transformer cannot deserialise this object. Attempted the "
+                            f"following transformers:\n{transformation_logs}"
```
fg91, line 158: @mg515 I need some clarification here, sorry :)

Is it still true in the current version of the PR that the dict config type transformer tries different type transformers for serializing the dict config nodes? When searching for `transformation_logs`, I can only see the variable being set to an empty string and being used in the `VallueError` messages. Can you please point me to where we iterate over different transformers and where `transformation_logs` is being appended to?

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/flytekit_patch.py
fg91: Resolving as enum transformer was removed, see https://github.com/flyteorg/flytekit/pull/2299#issuecomment-2106344696.

kumare3:
@fg91 / @mg515 I would love for this to go in if you folks are into pushing it?

cameronraysmith:
I think it's a good idea to support omegaconf's dictconfig directly because it is a foundational layer, but would just note that you get something like this almost for free if you use [hydra-zen](https://github.com/mit-ll-responsible-ai/hydra-zen) because it already provides a bridge from omegaconf's types to dataclasses.

fg91:
> I think it's a good idea to support omegaconf's dictconfig directly because it is a foundational layer, but would just note that you get something like this almost for free if you use [hydra-zen](https://github.com/mit-ll-responsible-ai/hydra-zen) because it already provides a bridge from omegaconf's types to dataclasses.

Do you have a minimal working example that shows how a `DictConfig` can be passed to a Flyte task using `hydra-zen`? 🙏

cameronraysmith:
@fg91 let's redirect this discussion to https://github.com/flyteorg/flyte/discussions/5576 to avoid getting into details that aren't directly relevant to this particular PR.

I probably should have done this instead of posting https://github.com/flyteorg/flytekit/pull/2299#issuecomment-2241268292 here but I'll leave it for the record.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/type_information.py
```diff
@@ -0,0 +1,124 @@
...
+
+    Args:
+        python_val: A DictConfig
+        key: Key of the node to analyze
+        mode: The mode in which to serialize type information
+
+    Returns:
+        Type - The extracted type
+        str - String representation for (de-)serialisation
+
```
eapolinario, line 59: Reformat this docstring to use the [reST format](http://daouzli.com/blog/docstring.html#id26) (e.g. [[1]](https://github.com/flyteorg/flytekit/blob/aa0b7c70450a5ff65956e4964f48f33649a25233/flytekit/configuration/file.py#L298-L304) or [[2]](https://github.com/flyteorg/flytekit/blob/aa0b7c70450a5ff65956e4964f48f33649a25233/flytekit/core/array_node_map_task.py#L44-L51)).

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/type_information.py
```diff
@@ -0,0 +1,124 @@
...
+def substitute_types(t: typing.Type) -> typing.Type:
+    """
+    Provides a substitute type hint to use when selecting transformers for serialisation.
+
+    Args:
+        t: Original type
+
+    Returns:
+        A corrected typehint
```
eapolinario, line 23: reformat docstring.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
eapolinario: bump.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/listconfig_transformer.py
```diff
@@ -0,0 +1,124 @@
...
+                type_list.append(type_name)
+
+                transformation_logs = ""
+                transformer = TypeEngine.get_transformer(node_type)
+                literal_type = transformer.get_literal_type(node_type)
+                value_list.append(
+                    MessageToDict(transformer.to_literal(ctx, python_val[idx], node_type, literal_type).to_flyte_idl())
+                )
+
+                if len(type_list) != len(value_list):
```
eapolinario, line 79: this check is always false, no?

eapolinario (commented):
All the string manipulation to load modules in runtime seems a bit brittle to me, but won't block the PR on it. Besides that, we really should get on the same page on some of the outstanding @fg91 's comments before proceeding with this change.

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/listconfig_transformer.py
mg515: addressed in https://github.com/flyteorg/flytekit/pull/2299/commits/709d4a4364e94f2e531f533c076b564dc3a26d38

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
```diff
@@ -0,0 +1,178 @@
...
+        type_map = {}
+        value_map = {}
+        for key in python_val.keys():
+            if OmegaConf.is_missing(python_val, key):
+                type_map[key] = "MISSING"
+            else:
+                node_type, type_name = extract_node_type(python_val, key)
+                type_map[key] = type_name
+
+                transformation_logs = ""
```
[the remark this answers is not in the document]
mg515: addressed in https://github.com/flyteorg/flytekit/pull/2299/commits/709d4a4364e94f2e531f533c076b564dc3a26d38

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
mg515: Yeah, the warning is obsolete with the removal of iterating through type transformers, my bad. Addressed now in https://github.com/flyteorg/flytekit/pull/2299/commits/709d4a4364e94f2e531f533c076b564dc3a26d38

plugins/flytekit-omegaconf/flytekitplugins/omegaconf/dictconfig_transformer.py
mg515: Addressed in https://github.com/flyteorg/flytekit/pull/2299/commits/e6e18964ff9d0f53a6722d1320e239f3d3a96545

We provide global mode and a context manager for ability to control it locally still.

plugins/flytekit-omegaconf/README.md
mg515: Addressed in https://github.com/flyteorg/flytekit/pull/2299/commits/e6e18964ff9d0f53a6722d1320e239f3d3a96545

We provide global mode and a context manager for ability to control it locally still.

eapolinario (approved):
Thank you for the persistence. This is looking great!

eapolinario (approved):

cameronraysmith:
Many thanks @mg515!

PR #2303, opened 2024-03-28
Core/fix register remote

novahow:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4936_

According to https://github.com/flyteorg/flyte/issues/4936, some users encountered difficulties when trying to call `register_workflow`. While `register_scripts` can serve as an alternative, setting up the `source_path` and `module_name` can be pretty annoying. This change also enables execution with a relative path. For example:

file structure
image
`test_register.py`
```python=
import typing
from flytekit.tools.script_mode import _find_project_root
import os
import inspect
import pathlib
import pdb

@task
def say_hello(name: str) -> str:
    return f"hello {name}!"


@task
def greeting_length(greeting: str) -> int:
    return len(greeting)


@workflow
def wf(name: str = "union") -> typing.Tuple[str, int]:
    greeting = say_hello(name=name)
    greeting_len = greeting_length(greeting=greeting)
    return greeting, greeting_len



if __name__ == "__main__":
    from flytekit.configuration import Config, SerializationSettings, ImageConfig

    remote = FlyteRemote(
        config=Config.for_sandbox(),
        default_domain="development"
    )


    print(wf)
    remote_wf = remote.register_workflow_script_mode(
        entity=wf,
    )

    wf_exe = remote.execute(entity=remote_wf, inputs={"name": "mike"})
    print(wf_exe.id)
```

You can run `python test_register.py` or `python ../test_register.py` under `func_dir` without worrying about the `source_path`. 

Added an function `register_workflow_script_mode`, which basically leverages `register_scripts`, but automatically handles the `source_path` and `module_name` params. Also resolved the problem of calling `register_script` in the `__main__` block when the registered workflow is in the same file as the `__main__` block. 

## How was this patch tested?
1. unit test
2. remote integration test
3. local testing

**Since we didn't export `FLYTE_PYTHON_PACKAGE_ROOT` and an `__init__.py` resides under our module `register_test`, it raises an error**
image
**After export works fine**
image
**here an `__init__.py` also resides under `issue_register`, without export would raise an error**
image
**execute with relative path, again, an `__init__.py` resides under `issue_register`, without export would raise an error**
image

#### Conclusion: 
when current working dir is inside the module and you want to execute `register_workflow_script_mode`, please run `export FLYTE_PYTHON_PACKAGE_ROOT="."`. While executing in the parent dir of the module works fine. 

##### Results of `python entrypoint.py`
**__init__.py** under `register_test`, need export
image

image

`entrypoint.py`

```python=
from issue_register.test_register import wf as sub_wf
from flytekit import workflow
import pdb
import os


@workflow
def wf(name: str = "union") -> str:
    return sub_wf(name=name)[0]

if __name__ == "__main__":
    from flytekit.configuration import Config, SerializationSettings, ImageConfig, FastSerializationSettings
    import pathlib
    CONFIG = os.environ.get("FLYTECTL_CONFIG", str(pathlib.Path.home() / ".flyte" / "config-sandbox.yaml"))

    remote = FlyteRemote(
        config=Config.auto(config_file=CONFIG),
        default_domain="development"
    )


    remote_sub_wf = remote.register_workflow_script_mode(
        entity=sub_wf,
    )
    remote_wf = remote.register_workflow_script_mode(
        entity=wf,
    )
    from issue_register.func_dir.test_register_in_func import register_wf
    func_remote_wf = register_wf(remote)
    remote.execute(entity=func_remote_wf, inputs={"name": "mike666"}, wait=True)

    subwf_exe = remote.execute(entity=remote_sub_wf, inputs={"name": "mike666"}, wait=True)
    wf_exe = remote.execute(entity=remote_wf, inputs={"name": "mike"}, wait=True)
    # pdb.set_trace()
    print("Func Remote", func_remote_wf.id)
    print("Remote SubWF", subwf_exe.id)
    print("Remote WF", wf_exe.id)
``` 

`test_register_in_func.py`

```python=
from issue_register.test_register import wf


def register_wf(remote):
    from flytekit.configuration import Config, SerializationSettings, ImageConfig, FastSerializationSettings
    import pathlib
    import os


    remote_wf = remote.register_workflow_script_mode(
        entity=wf,
        version="f0.0.1",
    )


    return remote_wf

```

#### test_register.py
image

#### entrypoint.py
image

### DIFF
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -40,7 +40,11 @@ class InstanceTrackingMeta(type):
 
         file = Path(file)
         try:
-            file_relative = file.resolve().relative_to(curdir)
+            # import pdb
+            # pdb.set_trace()
+            _root_dir = os.path.commonpath([file.resolve(), curdir.resolve()])
+            file_relative = Path(os.path.relpath(file.resolve(), _root_dir))
+            curdir = Path(_root_dir)
         except ValueError:
             return None
 
@@ -62,7 +66,7 @@ class InstanceTrackingMeta(type):
         while frame:
             if frame.f_code.co_name == "<module>" and "__name__" in frame.f_globals:
                 if frame.f_globals["__name__"] != "__main__":
-                    return frame.f_globals["__name__"], None
+                    return frame.f_globals["__name__"], frame.f_globals.get("__file__")
 
                 # Try to find the module and filename in the case that we're in the __main__ module
                 # This is useful in cases that use FlyteRemote to load tasks/workflows that are defined
@@ -81,7 +85,7 @@ class InstanceTrackingMeta(type):
         o = super(InstanceTrackingMeta, cls).__call__(*args, **kwargs)
         mod_name, mod_file = InstanceTrackingMeta._find_instance_module()
         o._instantiated_in = mod_name
-        o._module_file = mod_file
+        o._module_file = Path(mod_file).resolve() if mod_file else None
         return o
 
 
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -33,16 +33,14 @@ class InstanceTrackingMeta(type):
 
     @staticmethod
     def _get_module_from_main(globals) -> Optional[str]:
-        curdir = Path.cwd()
         file = globals.get("__file__")
         if file is None:
             return None
 
         file = Path(file)
         try:
-            root_dir = os.path.commonpath([file.resolve(), curdir.resolve()])
+            root_dir = os.path.commonpath([file.resolve(), Path.cwd()])
             file_relative = Path(os.path.relpath(file.resolve(), root_dir))
-            curdir = Path(root_dir)
         except ValueError:
             return None
 
@@ -51,8 +49,8 @@ class InstanceTrackingMeta(type):
         if len(module_components) == 0:
             return None
 
-        # make sure current directory is in the PYTHONPATH.
-        sys.path.insert(0, str(curdir))
+        # make sure /root directory is in the PYTHONPATH.
+        sys.path.insert(0, root_dir)
         try:
             return import_module_from_file(module_name, file)
         except ModuleNotFoundError:
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -40,8 +40,6 @@ class InstanceTrackingMeta(type):
 
         file = Path(file)
         try:
-            # import pdb
-            # pdb.set_trace()
             _root_dir = os.path.commonpath([file.resolve(), curdir.resolve()])
             file_relative = Path(os.path.relpath(file.resolve(), _root_dir))
             curdir = Path(_root_dir)
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -40,9 +40,9 @@ class InstanceTrackingMeta(type):
 
         file = Path(file)
         try:
-            _root_dir = os.path.commonpath([file.resolve(), curdir.resolve()])
-            file_relative = Path(os.path.relpath(file.resolve(), _root_dir))
-            curdir = Path(_root_dir)
+            root_dir = os.path.commonpath([file.resolve(), curdir.resolve()])
+            file_relative = Path(os.path.relpath(file.resolve(), root_dir))
+            curdir = Path(root_dir)
         except ValueError:
             return None
 
@@ -330,6 +330,7 @@ def extract_task_module(f: Union[Callable, TrackedInstance]) -> Tuple[str, str, 
             f = f.task_function
         # If the module is __main__, we need to find the actual module name based on the file path
         inspect_file = inspect.getfile(f)  # type: ignore
+        # get module name for instances in the same file as the __main__ module
         mod_name, _ = InstanceTrackingMeta._find_instance_module()
         return f"{mod_name}.{name}", mod_name, name, os.path.abspath(inspect_file)
 
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -40,7 +40,7 @@ class InstanceTrackingMeta(type):
 
         file = Path(file)
         try:
-            file_relative = file.relative_to(curdir)
+            file_relative = file.resolve().relative_to(curdir)
         except ValueError:
             return None
 
@@ -328,9 +328,8 @@ def extract_task_module(f: Union[Callable, TrackedInstance]) -> Tuple[str, str, 
             f = f.task_function
         # If the module is __main__, we need to find the actual module name based on the file path
         inspect_file = inspect.getfile(f)  # type: ignore
-        file_name, _ = os.path.splitext(os.path.basename(inspect_file))
-        mod_name = get_full_module_path(f, file_name)  # type: ignore
-        return name, mod_name, name, os.path.abspath(inspect_file)
+        mod_name, _ = InstanceTrackingMeta._find_instance_module()
+        return f"{mod_name}.{name}", mod_name, name, os.path.abspath(inspect_file)
 
     mod_name = get_full_module_path(mod, mod_name)
     return f"{mod_name}.{name}", mod_name, name, os.path.abspath(inspect.getfile(mod))
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -842,6 +842,7 @@ class FlyteRemote(object):
         :param serialization_settings: The serialization settings to be used
         :param default_launch_plan: This should be true if a default launch plan should be created for the workflow
         :param options: Additional execution options that can be configured for the default launchplan
+        :param fast: fast register if true
         :return:
         """
         if fast:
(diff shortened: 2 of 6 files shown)


flytekit/remote/remote.py
```diff
@@ -828,6 +828,44 @@ def register_workflow(
         fwf._python_interface = entity.python_interface
         return fwf
 
+    def register_workflow_script_mode(
+        self,
+        entity: PythonFunctionWorkflow,
+        image_config: typing.Optional[ImageConfig] = None,
+        project: typing.Optional[str] = None,
```
pingsutw, line 835: I like the idea of using `register_script` for fast register here, but could we add a flag (fast) to `register_workflow` instead. like

```python
def register_workflow(
    self,
    entity: WorkflowBase,
    serialization_settings: typing.Optional[SerializationSettings] = None,
    version: typing.Optional[str] = None,
    fast: bool = False,
    default_launch_plan: typing.Optional[bool] = True,
    options: typing.Optional[Options] = None,
):
   if fast:
       ...
       return self.register_script(...)
   else:
       ...
       return ...
```

flytekit/remote/remote.py
```diff
@@ -827,8 +828,42 @@ def register_workflow(
         :param serialization_settings: The serialization settings to be used
         :param default_launch_plan: This should be true if a default launch plan should be created for the workflow
         :param options: Additional execution options that can be configured for the default launchplan
+        :param fast: fast register if true
         :return:
         """
+        if fast:
```
pingsutw, line 834: cc @samhita-alla @cosmicBboy 
Which one do you prefer?
1. remote.register_workflow(my_task, fast=True)
2. remote.fast_register_workflow(my_task)

pingsutw: another option might be `remote.register_workflow(my_task, source_root="./")`

thomasjpfan: I do not think like using `fast` here because we already use "fast" with "fast registration". In this case, it's just calling `self.register_script` by defining `source_path` and `module_name` from `entity`. I'll prefer another method all together.

thomasjpfan (commented):
@pingsutw Does this implementation work with imported workflows?

```python
from my_lib.workflows import my_wf

remote = FlyteRemote(...)
remote.register_workflow(my_wf)
```

novahow:
> @pingsutw Does this implementation work with imported workflows?
> 
> ```python
> from my_lib.workflows import my_wf
> from flytekit.remote import FlyteRemote
> 
> remote = FlyteRemote(...)
> remote.register_workflow(my_wf)
> ```

@thomasjpfan Hi, this supports imported workflows. As for the fast flag, I can use a separate function just as my earlier version does if that's the desired implementation.

flytekit/core/tracker.py
```diff
@@ -40,7 +40,9 @@ def _get_module_from_main(globals) -> Optional[str]:
 
         file = Path(file)
         try:
-            file_relative = file.relative_to(curdir)
+            root_dir = os.path.commonpath([file.resolve(), curdir.resolve()])
+            file_relative = Path(os.path.relpath(file.resolve(), root_dir))
+            curdir = Path(root_dir)
```
pingsutw, line 45: nit: I think we can remove this line and change line 55 to
```
# make sure `/root` is in the PYTHONPATH.
sys.path.insert(0, root_dir)
```

pingsutw (approved):

wild-endeavor:
Hey @novahow - any chance you can spend a bit more time on this and help us understand a few things please?  Thank you for working on this, I know it's tricky.

This change is leading to some weird behavior and I think part of it is working by coincidence.  We noticed this because of a test that was failing in Pycharm but passing in CI.  But you can reproduce a similar error just with pytest.

For example, this command passes
```
pytest -svv tests/flytekit/unit/core/test_type_hints.py::test_union_type
```
but this command fails
```
python -m pytest -svv tests/flytekit/unit/core/test_type_hints.py::test_union_type
```
The error it fails with is
```
E        Regex: "Error\\ encountered\\ while\\ converting\\ inputs\\ of\\ 'tests\\.flytekit\\.unit\\.core\\.test_type_hints\\.t2':\\\n\\ \\ Cannot\\ convert\\ from\\ Flyte\\ Serialized\\ object\\ \\(Literal\\):"
E        Input: "Error encountered while converting inputs of 'envs.flytekit.lib.python3.12.site-packages.pytest.__main__.tests.flytekit.unit.core.test_type_hints.t2':\n  Cannot convert from Flyte Serialized object (Literal):\n  scalar:\n    union:\n      value:\n        scalar:\n          primitive:\n            string_value: 2\n      type:\n        simple: 3\n        structure:\n          tag: str to typing.Union[float, dict] (using tag str)"
```

Both of these tests are run from the base of flytekit at `/Users/ytong/go/src/github.com/flyteorg/flytekit.`  The difference is in the error message, notice the name of the task is different, in the latter case, it is prepended with the actual `pytest` file.  (Note this is a separate discussion than the `{exec_prefix}` string that's in the test, that's not relevant here.)

If you uncomment [line 42](https://github.com/flyteorg/flytekit/blob/dbg/pr2303/flytekit/core/tracker.py#L42) in this branch though then that second command will also work.

The reason is that line 42 throws a ValueError, because there is no relative path from `/Users/ytong/envs/flytekit/lib/python3.12/site-packages/pytest/__main__.py` to the flytekit directory (which is cwd).

Why does running `pytest` directly work?  Because when you run it directly, `file` on line 58 is `/Users/ytong/envs/flytekit/bin/pytest` instead of the `__main__.py` in site packages, and `/Users/ytong/envs/flytekit/bin/pytest` is not loadable.  So it fails, raises a `ModuleNotFoundError`, resulting in `_get_module_from_main` returning `None`.  Returning `None` from this function basically causes a bunch of other logic to interpret the file holding the task as not a `if __name__ == '__main__'` type file.

Can you revisit this section a bit please?  Was the switch from `relative_to` to `commonpath` only to get the `../test_register.py` case working?  If so, can we think of another way around this?  The above behavior makes me feel like there's another way to do this.  Can you explain what the error was that made this change necessary?  Is it only in the run time case (in the task container at run time)?

Also separately, on line 56, does `root_dir` need to be at index 0?  Or can it be appended to the end?  If it needs to be at 0 could you explain why?

novahow:
> Hey @novahow - any chance you can spend a bit more time on this and help us understand a few things please? Thank you for working on this, I know it's tricky.
> 
> This change is leading to some weird behavior and I think part of it is working by coincidence. We noticed this because of a test that was failing in Pycharm but passing in CI. But you can reproduce a similar error just with pytest.
> 
> For example, this command passes
> 
> ```
> pytest -svv tests/flytekit/unit/core/test_type_hints.py::test_union_type
> ```
> 
> but this command fails
> 
> ```
> python -m pytest -svv tests/flytekit/unit/core/test_type_hints.py::test_union_type
> ```
> 
> The error it fails with is
> 
> ```
> E        Regex: "Error\\ encountered\\ while\\ converting\\ inputs\\ of\\ 'tests\\.flytekit\\.unit\\.core\\.test_type_hints\\.t2':\\\n\\ \\ Cannot\\ convert\\ from\\ Flyte\\ Serialized\\ object\\ \\(Literal\\):"
> E        Input: "Error encountered while converting inputs of 'envs.flytekit.lib.python3.12.site-packages.pytest.__main__.tests.flytekit.unit.core.test_type_hints.t2':\n  Cannot convert from Flyte Serialized object (Literal):\n  scalar:\n    union:\n      value:\n        scalar:\n          primitive:\n            string_value: 2\n      type:\n        simple: 3\n        structure:\n          tag: str to typing.Union[float, dict] (using tag str)"
> ```
> 
> Both of these tests are run from the base of flytekit at `/Users/ytong/go/src/github.com/flyteorg/flytekit.` The difference is in the error message, notice the name of the task is different, in the latter case, it is prepended with the actual `pytest` file. (Note this is a separate discussion than the `{exec_prefix}` string that's in the test, that's not relevant here.)
> 
> If you uncomment [line 42](https://github.com/flyteorg/flytekit/blob/dbg/pr2303/flytekit/core/tracker.py#L42) in this branch though then that second command will also work.
> 
> The reason is that line 42 throws a ValueError, because there is no relative path from `/Users/ytong/envs/flytekit/lib/python3.12/site-packages/pytest/__main__.py` to the flytekit directory (which is cwd).
> 
> Why does running `pytest` directly work? Because when you run it directly, `file` on line 58 is `/Users/ytong/envs/flytekit/bin/pytest` instead of the `__main__.py` in site packages, and `/Users/ytong/envs/flytekit/bin/pytest` is not loadable. So it fails, raises a `ModuleNotFoundError`, resulting in `_get_module_from_main` returning `None`. Returning `None` from this function basically causes a bunch of other logic to interpret the file holding the task as not a `if __name__ == '__main__'` type file.
> 
> Can you revisit this section a bit please? Was the switch from `relative_to` to `commonpath` only to get the `../test_register.py` case working? If so, can we think of another way around this? The above behavior makes me feel like there's another way to do this. Can you explain what the error was that made this change necessary? Is it only in the run time case (in the task container at run time)?
> 
> Also separately, on line 56, does `root_dir` need to be at index 0? Or can it be appended to the end? If it needs to be at 0 could you explain why?

sorry for the inconvenience this PR brought, I will take a look at it, thanks for the reminder.

novahow:
@wild-endeavor Yeah I believe the `commanpath` change was for executing from inside a module. Currently I haven't come up with a workaround if we want to support that without affecting the `python -m pytest` issue, as for now I can only disable that support to make the testing work

wild-endeavor:
> but can't support additional feature

What is the additional feature?

novahow:
> What is the additional feature?

executing something like `python ../entrypoint.py` would be disabled. while we may still be able to do `python entrypoint.py` without requiring to make it an module with some `__init__.py`

wild-endeavor:
could we do that then?  i think that will make it a lot clearer, and when sufficient people ask for the .. pattern we can try a different way to support it.

PR #2312, opened 2024-03-29
Add way to create FlyteFile/Directory from remote location

wild-endeavor:
Add a couple of class methods
```
def from_source(cls, source: str | os.PathLike) -> FlyteFile
def from_source(cls, source: str | os.PathLike) -> FlyteDirectory
```
meant to be used with remote paths.  This is because if you create a file with an s3/gs link as the path using the normal constructor, it doesn't come with a download function, forcing users to manually call into the data persistence layer manually.


ggydush (dismissed):
Thank you!!

eapolinario (approved):

PR #2314, opened 2024-04-01
Core/enable deck

novahow:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/5095_

closes #5095

~Make the ui display deck by default, while some users don't want to render large input/output such as dataframes~
let users customize displayed decks. 

1. add an parameter to task: `decks`, which takes a tuple of str and elements should be one of ["input", "output", "dependencies", "source_code", "Timeline"]
2. maintains compatibility with deprecated `disable_deck` params
~3. Currently by default, the flyteconsole will display `Timeline` and `Source Code` deck.~
4. utilize a `ExecutionParameters` builder pattern to inject expected decks in `dispatch_execute`
5. Added the `output_metadata_prefix ` param to builder of `ExecutionParameters`, not sure why it wasn't there initially. 

## How was this patch tested?
unit tests

### Screenshots
code snippet:
```python
from math import sqrt
from flytekit import workflow, task, LaunchPlan
from typing import List, Union
import random
import pdb

@task(enable_deck=False)
def standard_deviation(values: List[float], mu: float) -> float:
    variance = sum([(x - mu) ** 2 for x in values])
    return sqrt(variance)

@task(disable_deck=False)
def standard_scale(values: List[float], mu: float, sigma: float) -> List[float]:
    return [(x - mu) / sigma for x in values]

from flytekit import workflow
from flytekitplugins.deck.renderer import ImageRenderer


from flytekit.deck.deck import Deck, DeckFields
@task(decks=[DeckFields.INPUT.value, DeckFields.TIMELINE.value, DeckFields.DEPENDENCIES.value], enable_deck=True)
def image_renderer(image: FlyteFile) -> None:
    Deck("Image Renderer", ImageRenderer().to_html(image_src=image))


@task(decks=[DeckFields.OUTPUT])
def mean(values: List[float]) -> float:
    return sum(values) / len(values)

@task
def mean2(values: List[float]) -> float:
    return sum(values) / len(values)

@task
def generate_data(num_samples: int, seed: int) -> List[float]:
    random.seed(seed)
    return [random.random() for _ in range(num_samples)]


@workflow
def standard_scale_workflow(values: List[float]) -> List[float]:
    mu = mean(values=values)
    _ = image_renderer(image="https://bit.ly/3KZ95q4")
    sigma = standard_deviation(values=values, mu=mu)
    return standard_scale(values=values, mu=mu, sigma=sigma)


@workflow
def workflow_with_subworkflow(num_samples: int, seed: int) -> List[float]:
    data = generate_data(num_samples=num_samples, seed=seed)
    return standard_scale_workflow(values=data)


# pdb.set_trace()
print("Compiled")
# LaunchPlan.CACHE.pop("standard_scale_lp", None)
standard_scale_launch_plan = LaunchPlan.get_or_create(
    standard_scale_workflow,
    name="standard_scale_lp",
    default_inputs={"values": [3.0, 4.0, 5.0]}
)


if __name__ == "__main__":
    print(standard_scale_launch_plan())

```

**enable_deck is None, decks ignored**
image
**`enable_deck=True`, specified decks+custom image deck**
image
**enable_deck=False**
image
**`disable_deck=False`, displaying default decks(source_code, dependencies)**
image


kumare3:
I do not prefer default decks. There are security and performance implications. Can we enable these decks all using the same flag

If enable_deck is true then all, else you can pass a deck selection object to it

novahow:
> I do not prefer default decks. There are security and performance implications. Can we enable these decks all using the same flag
> 
> If enable_deck is true then all, else you can pass a deck selection object to it

Sure. Additionally, could you provide a brief example of a deck selection object? I'm unsure if it implies a list of DeckFields members or something else. Thanks.

pingsutw:
cc @thomasjpfan

flytekit/core/base_task.py
```diff
@@ -477,6 +478,8 @@ def __init__(
                 execution of the task. Supplied as a dictionary of key/value pairs
             disable_deck (bool): (deprecated) If true, this task will not output deck html file
             enable_deck (bool): If true, this task will output deck html file
+            additional_decks (Optional[List[str]]): List of additional decks besides [timeline and source code] to be
```
novahow, line 481: Should we define it as a List[DeckFields] instead?

thomasjpfan: I am +0 on using enums if we reimport it in `flytekit.deck`. Specifically:

```python
from flytekit.deck import PrebuiltDeck

@task(decks=[PrebuiltDeck.SourceCode, PrebuiltDeck.Inputs], enable_deck=True)
```

Enums is nice from a programming point of view, but they can make the feature harder to discover. For example, using strings means no additional imports:

```python
@task(decks=["source_code", "inputs"], enable_deck=True)
```

@novahow From a user point of view, do you have a preference?

thomasjpfan (commented):
> Sure. Additionally, could you provide a brief example of a deck selection object? I'm unsure if it implies a list of DeckFields members or something else. Thanks.

I think the intention is to have all the decks be explicit. For example, this will only show the "source code" deck and **not** the input and output:

```python
@task(decks=("source_code",), enable_deck=True)
```

If one wants to show the inputs and outputs, then they need to be explicit:

```python
@task(decks=("source_code", "inputs", "outputs"), enable_deck=True)
```

In the examples above the TimelineDeck is **not** shown because it was not explicitly shown.

The other question is what the default should be. Personally, I want the default `decks` to be `("source_code",)`, because I think seeing the source code is generally useful. (The default is a tuple, so it is not mutable.) With this proposal, if `enable_deck` is false, then `decks` is ignored.

flytekit/core/base_task.py
novahow: @thomasjpfan  Yeah I think using strings is fine.

novahow:
@thomasjpfan. Thanks for the comment. 
So when 
1. enable_deck=True, ignore decks and output all
2. enable_deck=None, use decks
3. enable_deck=False, ignore decks

Is that the desired behavior? Thanks

thomasjpfan:
> enable_deck=True, ignore decks and output all

I want the default `decks` to be `("source_code", "python_dependencies")`, so that when `enable_deck=True` and `decks` is not passed in, then the source code and python dependencies are shown. I have not found many use case for the "Inputs" and "Outputs" deck, and the `TimelineDeck` requires a plugin, so I rather not show it by default.

> enable_deck=None, use decks
> enable_deck=False, ignore decks

In these cases, I prefer not to not to show any decks. I like having an explicit `enable_deck=True` to mean "show decks" and anything else means there is no decks.

flytekit/core/context_manager.py
```diff
@@ -89,6 +89,8 @@ class Builder(object):
         execution_date: typing.Optional[datetime] = None
         logging: Optional[_logging.Logger] = None
         task_id: typing.Optional[_identifier.Identifier] = None
+        rendered_decks: Optional[List[str]] = None
```
thomasjpfan, line 92: At this point, I think having both `decks` and `rendered_decks` complicates this a bit too much. I prefer to have `decks` be the the same as "rendered decks" and not have two notions of decks.

If the user passes in `decks=["source_code"]`, then the "input", and "output" decks are never created. This should be simple to do in `PythonTask._write_decks`.

The tough one is the timeline deck, which is created by the context manager and not `PythonTask`. Here is a workaround I see:

- Override `TimeLineDeck.__init__` to accept a `auto_add_to_deck`. If this is `False`, then the 
- In `PythonTask._output_to_literal_map` and add the deck before `@timeit` is called and  set it to `auto_add_to_deck=True` if the user passes in "timeline" in `@task(decks=)`. Otherwise, `auto_add_to_deck` is False.
- Since `PythonTask` already added the `TimelineDeck`, the `ExecutionParameters.timeline` deck will not created another one itself.

flytekit/core/context_manager.py
```diff
@@ -89,6 +89,8 @@ class Builder(object):
...
+        output_metadata_prefix: Optional[str] = None
```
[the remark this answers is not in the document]
novahow: For the current implementation, in https://github.com/flyteorg/flytekit/blob/d504753cd74ff1b1ad5f399d5e5c3573dfbdad0d/flytekit/bin/entrypoint.py#L177, a new param is passed in, so `output_metadata_prefix ` has to be included otherwise the field will be empty and cause error.

flytekit/core/context_manager.py
novahow: Does that imply we need to create a Timelinedeck before any of the `timeit` calls or decorators? Or am I missing anything? Thanks.

pingsutw: But the timeline is created before we [load the task](https://github.com/flyteorg/flytekit/blob/c78142b5aa24a97694150d9a6714703c4b3d56fd/flytekit/core/python_auto_container.py#L250-L256)

Could we just keep the `Timeline` deck as default? we already show the [warning](https://github.com/flyteorg/flytekit/blob/b7de700a3a779866c895f37550b95173498d2758/flytekit/deck/deck.py#L95) in the deck if users don't install the required dependencies.

pingsutw: > At this point, I think having both decks and rendered_decks complicates this a bit too much. I prefer to have decks be the the same as "rendered decks" and not have two notions of decks.

agree, I don't think we need `rendered_decks`

flytekit/core/base_task.py
pingsutw: I prefer enum. how do users know what other decks we have? Using string is also error-prone. what if people use `sourceCode` or `Input`

thomasjpfan: We list the strings out in the docstring and remember to add new ones to the list. If someone uses `sourceCode` or `Input`, then we raise an error.

For me, I come from the PyData world, where most APIS are strings. For example, SciPy's [`minimize)` function](https://docs.scipy.org/doc/scipy/reference/optimize.html#local-multivariate-optimization) uses string for `method`. The newly release [torch.comple](https://pytorch.org/docs/stable/generated/torch.compile.html#torch-compile) uses strings for `backend` and `mode`.

Personally, I prefer enums too, but I think most users in the Python data space are more familiar with strings.

flytekit/core/context_manager.py
thomasjpfan: > Could we just keep the Timeline deck as default? we already show the warning in the deck if users don't install the required dependencies.

In it's current state, I think it's a poor user experience to have a timeline deck that shows an error message. Personally, I do not like installing `flytekitplugins-deck-standard` because it's dependencies make the image size much bigger. (Last time I checked it was +500MB).

Can we make `TimelineDeck` not be added to `decks` by default? Concretely, I mean to change the following:

https://github.com/flyteorg/flytekit/blob/c78142b5aa24a97694150d9a6714703c4b3d56fd/flytekit/core/context_manager.py#L284-L286

to `TimeLineDeck("Timeline", auto_add_to_deck=False)`. Then in `PythonTask`, it'll manually add the timeline deck if it is in `@task(decks=[...])`.

The alternative is to have simplified timeline deck that renders a normal HTML table without the graph. I see the TimelineDeck as a debugging tool that is not useful day to day, which means I would not recommend installing the `flytekitplugins-deck-standard` by default. If there is an issue with a task, then a simple HTML table that breaks down the time spend for each step is more than enough.

novahow: I guess simply adding auto_add_to_deck=False doesn't work because the first `timeit` call happens in `load_task`, which invokes the https://github.com/flyteorg/flytekit/blob/c78142b5aa24a97694150d9a6714703c4b3d56fd/flytekit/core/context_manager.py#L284-L286 before we can know if timeline deck is in decks, thus we may lose access to the originally created timeline_deck when we enter the execution part of the `PythonTask`. 
I guess it would be feasible as long as we make timelinedeck an attribute of `ExecutionParameters `. For instance, `ExecutionParameters._timelinedeck `, and we can save the timelinedeck upon `timeline_deck` property access, then we can later decide whether to add it to the decks in `PythonTask`.

thomasjpfan: > before we can know if timeline deck is in decks

From reading the code, when `TimelineDeck` gets created, it automatically adds itself to decks, when it calls `super().__init__` which ends up here:

https://github.com/flyteorg/flytekit/blob/f4d894af2ebb66f4492506ebb121509ef8cb8aa3/flytekit/deck/deck.py#L58

The `auto_add_to_deck=False` will turn this off:

```python
    def __init__(self, name: str, html: Optional[str] = "", auto_add_to_deck: bool = True):
        self._name = name
        self._html = html
        if auto_add_to_deck
            FlyteContextManager.current_context().user_space_params.decks.append(self)
```

novahow: I agree, but if I turn it off and not save it anywhere else, I think I can't retrieve the initially created timeline_deck when I want to append a new time info to the original deck. Since in the current version, the only place where the `Timelinedeck` is saved is in the `decks`. For example, when we are loading task in `load_task`, it automatically creates a timelinedeck due to a `timeit` decorator, and now we turn it off so it is not saved in the `ExecutionParams`. Later when we finally load the task module and get access to the `decks` of the task, where I finally get to know if timelinedeck should be in `ExecutionParams.decks`, I think I lost the access of the previously created timelinedeck since we turned the flag off and it was absent in my context, so I can't really append any other time info to it. In this case, when another `timeit` is invoked, it will create a new `TimelineDeck` instead of using the original one.

flytekit/core/context_manager.py
```diff
@@ -274,16 +278,20 @@ def default_deck(self) -> Deck:
...
         for deck in self.decks:
             if isinstance(deck, TimeLineDeck):
                 time_line_deck = deck
                 break
         if time_line_deck is None:
-            time_line_deck = TimeLineDeck("Timeline")
+            if self._timeline_deck is not None:
+                time_line_deck = self._timeline_deck
+            else:
+                time_line_deck = TimeLineDeck(DeckFields.TIMELINE.value)
```
thomasjpfan, line 292: I know it's the same, but to be consistent with the other decks:

```suggestion
                time_line_deck = TimeLineDeck("Timeline", auto_add_to_deck=False)
```

flytekit/core/python_function_task.py
```diff
@@ -351,18 +351,23 @@ def dynamic_execute(self, task_function: Callable, **kwargs) -> Any:
...
 
-                source_code_deck = Deck("Source Code")
+                source_code_deck = Deck(
+                    DeckFields.SOURCE_CODE.value, auto_add_to_deck=DeckFields.SOURCE_CODE in self.decks
+                )
                 renderer = SourceCodeRenderer()
                 source_code_deck.append(renderer.to_html(source_code))
 
-            python_dependencies_deck = Deck("Dependencies")
+            python_dependencies_deck = Deck(
```
thomasjpfan, line 368: Same here regarding not needing to creating the deck, if it is not added to the deck.

```python
if DeckFields.DEPENDENCIES in self.decks:
    python_dependencies_deck = Deck("Dependencies")
```

flytekit/deck/deck.py
```diff
@@ -10,6 +11,18 @@
...
 
+class DeckFields(str, enum.Enum):
+    """
+    DeckFields is used to specify the fields that will be rendered in the deck.
+    """
+
+    INPUT = "input"
+    OUTPUT = "output"
+    SOURCE_CODE = "source_code"
+    TIMELINE = "Timeline"
```
thomasjpfan, line 22:
```suggestion
    TIMELINE = "timeline"
```

flytekit/core/base_task.py
```diff
@@ -487,23 +491,35 @@ def __init__(
...
         self._task_config = task_config
+        self._decks = list(decks) if (decks is not None and (enable_deck is True or disable_deck is False)) else []
+
+        deck_members = set([_field.value for _field in DeckFields])
+        # enumerate additional decks, check if any of them are invalid
+        for deck in self._decks:
+            if deck not in deck_members:
+                raise ValueError(
+                    f"Element {deck} from decks param is not a valid deck field. Please use one of {deck_members}"
+                )
```
thomasjpfan, line 502: Can this validation happen after `self._disable_deck` is defined:

```python
if not self._disable_deck and decks is not None:
    self._decks = list(decks)
else:
    self._decks = []

# validate `self._decks` here
```

flytekit/core/base_task.py
```diff
@@ -462,6 +463,7 @@ def __init__(
         environment: Optional[Dict[str, str]] = None,
         disable_deck: Optional[bool] = None,
         enable_deck: Optional[bool] = None,
+        decks: Optional[Tuple[str, ...]] = None,
```
thomasjpfan, line 466: This is my preferred default, but I'm open to other opinions:

```suggestion
        decks: Optional[Tuple[str, ...]] = ("source_code", "dependencies"),
```

flytekit/core/python_function_task.py
```diff
@@ -351,18 +351,23 @@ def dynamic_execute(self, task_function: Callable, **kwargs) -> Any:
...
 
             # These errors are raised if the source code can not be retrieved
             with suppress(OSError, TypeError):
                 source_code = inspect.getsource(self._task_function)
                 from flytekit.deck.renderer import SourceCodeRenderer
 
-                source_code_deck = Deck("Source Code")
+                source_code_deck = Deck(
+                    DeckFields.SOURCE_CODE.value, auto_add_to_deck=DeckFields.SOURCE_CODE in self.decks
+                )
```
thomasjpfan, line 364: Same here regarding not creating the deck, if is not added to the deck.

flytekit/deck/deck.py
```diff
@@ -79,8 +93,8 @@ class TimeLineDeck(Deck):
     Instead, the complete data set is used to create a comprehensive visualization of the execution time of each part of the task.
     """
 
-    def __init__(self, name: str, html: Optional[str] = ""):
-        super().__init__(name, html)
+    def __init__(self, name: str, html: Optional[str] = "", auto_add_to_deck: bool = False):
```
thomasjpfan, line 96: I think it can be confusing to have a different default here. Can this be the same default as `Deck`, but have the caller set `auto_add_to_deck=False`?

```suggestion
    def __init__(self, name: str, html: Optional[str] = "", auto_add_to_deck: bool = True):
```

thomasjpfan (commented):
Thanks for the updates! Overall, I like the approach.

flytekit/deck/deck.py
novahow: Done, thanks for the advice!

flytekit/core/base_task.py
```diff
@@ -477,6 +479,8 @@ def __init__(
                 execution of the task. Supplied as a dictionary of key/value pairs
             disable_deck (bool): (deprecated) If true, this task will not output deck html file
             enable_deck (bool): If true, this task will output deck html file
+            decks (Tuple[str]): Tuple of decks to be
+                generated for this task. Valid values can be selected from fields of ``flytekit.deck.deck.DeckFields`` enum
```
thomasjpfan, line 483: Can you reimport `DeckFields` in `flytekit.deck`?
```suggestion
                generated for this task. Valid values can be selected from fields of ``flytekit.deck.DeckFields`` enum
```

flytekit/core/base_task.py
thomasjpfan: For me, it's more about consistency. If I wanted to debug `PythonTask`, then I'll have to go back to `def task` to figure out that the default is actually ("Source Code", "Dependencies").

thomasjpfan (commented):
The [CI errors](https://github.com/flyteorg/flytekit/actions/runs/8621740170/job/23688016772?pr=2314) look related.

flytekit/core/base_task.py
```diff
@@ -462,6 +463,7 @@ def __init__(
         environment: Optional[Dict[str, str]] = None,
         disable_deck: Optional[bool] = None,
         enable_deck: Optional[bool] = None,
+        decks: Optional[Tuple[str, ...]] = ("Source Code", "Dependencies"),
```
thomasjpfan, line 466: @novahow I know we've been moving back and forth about this, but using capitalized placeholders seem weird to me.

At this point, I'm okay with using the enums: `DeckField.SOURCE_CODE`, etc.

flytekit/deck/deck.py
```diff
@@ -10,6 +11,18 @@
 DECK_FILE_NAME = "deck.html"
 
 
+class DeckFields(str, enum.Enum):
```
thomasjpfan, line 14: Nit: I think enums in this context is singular:

```suggestion
class DeckField(str, enum.Enum):
```

(Many places needs to be adjusted with this change)

thomasjpfan (commented):
Thanks for the update!

flytekit/core/base_task.py
```diff
@@ -462,6 +463,7 @@ def __init__(
         environment: Optional[Dict[str, str]] = None,
         disable_deck: Optional[bool] = None,
         enable_deck: Optional[bool] = None,
+        decks: Optional[Tuple[str, ...]] = (DeckField.SOURCE_CODE.value, DeckField.DEPENDENCIES.value),
```
thomasjpfan, line 466: Can we go all in with the enums?

```suggestion
        decks: Optional[Tuple[DeckField, ...]] = (DeckField.SOURCE_CODE, DeckField.DEPENDENCIES),
```

thomasjpfan (commented):
Last bit about the enum API, otherwise this looks good to go!

flytekit/core/base_task.py
```diff
@@ -789,6 +810,13 @@ def disable_deck(self) -> bool:
         """
         return self._disable_deck
 
+    @property
+    def decks(self) -> List[DeckField]:
```
thomasjpfan, line 814: I think it's better to call this `deck_fields`

```suggestion
    def decks_fields(self) -> List[DeckField]:
```

A user can add custom decks that are not a preconfigured deck in `DeckField`.

flytekit/core/base_task.py
```diff
@@ -477,6 +479,8 @@ def __init__(
                 execution of the task. Supplied as a dictionary of key/value pairs
             disable_deck (bool): (deprecated) If true, this task will not output deck html file
             enable_deck (bool): If true, this task will output deck html file
+            decks (Tuple[DeckField]): Tuple of decks to be
```
thomasjpfan, line 482: I think naming this `deck_fields` is better to distinguish it from all the actual decks:

```suggestion
            deck_fields (Tuple[DeckField]): Tuple of decks to be
```

flytekit/core/base_task.py
```diff
@@ -462,6 +463,7 @@ def __init__(
         environment: Optional[Dict[str, str]] = None,
         disable_deck: Optional[bool] = None,
         enable_deck: Optional[bool] = None,
+        deck_fields: Optional[Tuple[DeckField, ...]] = (DeckField.SOURCE_CODE, DeckField.DEPENDENCIES),
```
thomasjpfan, line 466: After speaking with @pingsutw and @EngHabu, we still need to keep the current default decks to keep backward compatibility. Can you update this to include all of them? (source code + dependencies + timeline + inputs + outputs).

pingsutw:
I made some changes.
- Render time table without installing `flytekitplugins-deck-standard`
- Add timeline chart to the deck if `flytekitplugins-deck-standard` is installed

Screenshot 2024-07-01 at 2 05 33 PM
Screenshot 2024-07-01 at 2 08 47 PM

pingsutw (approved):

thomasjpfan (approved):

PR #2318, opened 2024-04-03
Ensure that annotations are dropped in the case of the dataclass and dict type transformers

eapolinario:
## Tracking issue
NA

The original exception in the `DictTransformer` was put in place to catch the case of `FlyteAnnotation` being set passed to annotated dictionary, e.g. `Annotated[dict, FlyteAnnotation("annotation-1")]` is invalid. A similar argument applies to the dataclass type transformer.

Keep the original exception, but only raise it if the annotation contains a `FlyteAnnotation` in both the dataclass and the dictionary transformers.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -356,6 +356,7 @@ class DataclassTransformer(TypeTransformer[object]):
         # However, FooSchema is created by flytekit and it's not equal to the user-defined dataclass (Foo).
         # Therefore, we should iterate all attributes in the dataclass and check the type of value in dataclass matches the expected_type.
 
+        expected_type = get_underlying_type(expected_type)
         expected_fields_dict = {}
         for f in dataclasses.fields(expected_type):
             expected_fields_dict[f.name] = f.type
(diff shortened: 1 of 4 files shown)


flytekit/core/type_engine.py
```diff
@@ -1642,12 +1642,16 @@ def get_dict_types(t: Optional[Type[dict]]) -> typing.Tuple[Optional[type], Opti
...
         _args = get_args(t)
         if _origin is not None:
-            if _origin is Annotated:
-                raise ValueError(
-                    f"Flytekit does not currently have support \
-                        for FlyteAnnotations applied to dicts. {t} cannot be \
-                        parsed."
-                )
+            if _origin is Annotated and _args:
+                # _args holds the type arguments to the dictionary, i.e. get_args(Dict[int, str]) == (int, str).
```
thomasjpfan, line 1646: Although this comment is informational, I find a little out of place and can be confusing to a reader. Concretely, inside this `if` block `_origin` is always `Annotated` and not `Dict`.

thomasjpfan (dismissed):
Nit regarding the comment, otherwise LGTM

wild-endeavor (dismissed):

pingsutw (approved):

wild-endeavor (approved):

PR #2322, opened 2024-04-03
Run ignore query

wild-endeavor:
When a query is detected, don't do anything, and just say that it'll be resolved.  If running locally though, it must be specified.

Make it so that pyflyte just doesn't pass along queries.

## How was this patch tested?
Tested locally, tested against union, and tested with local data

### Screenshots
Working
image

Help screen
image

Local error
image

### DIFF
--- a/flytekit/interaction/click_types.py
+++ b/flytekit/interaction/click_types.py
@@ -14,6 +14,7 @@ from dataclasses_json import DataClassJsonMixin
 from pytimeparse import parse
 
 from flytekit import BlobType, FlyteContext, FlyteContextManager, Literal, LiteralType, StructuredDataset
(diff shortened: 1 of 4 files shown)


flytekit/interaction/click_types.py
```diff
@@ -274,7 +293,7 @@ def modify_literal_uris(lit: Literal):
     SimpleType.STRING: click.STRING,
     SimpleType.BOOLEAN: click.BOOL,
     SimpleType.DURATION: DurationParamType(),
-    SimpleType.DATETIME: click.DateTime(),
+    SimpleType.DATETIME: DateTimeType(),
```
eapolinario, line 296: good catch.

eapolinario (approved):

PR #2332, opened 2024-04-05
Fix Fast registration removes datetime metadata from files

ChungYujoyce:
## Tracking issue
_https://github.com/flyteorg/flyte/issues/4711_

Remove the filter function and leave it with default filter=None 

## How was this patch tested?


pingsutw (commented):
Thanks! some tests are failing. mind taking a look

wild-endeavor:
Does this work?  it still produces the same version if you register twice?  when we implemented this we traced the duplicate versions to the timestamp of the files.

PR #2339, opened 2024-04-10
Added remote path helper function

pryce-turner:

Users will sometimes want to push blobs to custom buckets but maintain the naming convention that flyte creates by default

This adds a small helpers module with this function and associated tests

## How was this patch tested?

2 small unit tests, one for each new function

### DIFF
--- a/flytekit/extras/helpers.py
+++ b/flytekit/extras/helpers.py
@@ -1,0 +1,33 @@
+import os
+import random
+import string
+
+
+def rand_str(len: int) -> str:
+    """Generate a random string of length `len`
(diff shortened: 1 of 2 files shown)


pryce-turner:
I feel like this can _almost_ be accomplished with this [method](https://github.com/flyteorg/flytekit/blob/177571b0c85e3b7d95bb24e299bed70fbec992ae/flytekit/types/directory/types.py#L159) and then just replacing the bucket name with a string edit. Is there a cleaner way to set the bucket name before calling `new_remote`?

PR #2349, opened 2024-04-13
Adds custom string format parameter to customize ImageSpec's tag

thomasjpfan:

This PR allows users to generate any tag they want for their images. This is useful for teams that want to have custom names for managing their docker images.

This PR adds a `tag_format` to `ImageSpec` which accepts a format string to custom the image's tag.

## How was this patch tested?

I added a unit test to check the new behavior.


pingsutw (dismissed):

flytekit/image_spec/image_spec.py
```diff
@@ -85,6 +88,9 @@ def image_name(self) -> str:
     def _image_name(self) -> str:
         """Construct full image name with tag."""
         tag = calculate_hash_from_image_spec(self)
+        if self.tag_format:
+            tag = self.tag_format.format(spec_hash=spec_hash)
```
eapolinario, line 92:
```suggestion
            tag = self.tag_format.format(spec_hash=tag)
```

this should be ^, right?

pingsutw (approved):

PR #2356, opened 2024-04-16
Add default inputs to version hash

pingsutw:
## Tracking issue
NA

Failed to run a workflow again when using `datetime.datetime` as default input because flytekit will try to register the same version of LP with the different default input

image

Add default inputs to version hash

## How was this patch tested?
Integration test and pyflyte run 

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -23,6 +23,7 @@ from datetime import datetime, timedelta
 from typing import Dict
 
 import click
+import cloudpickle
 import fsspec
 import requests
 from flyteidl.admin.signal_pb2 import Signal, SignalListRequest, SignalSetRequest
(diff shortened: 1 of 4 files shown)


flytekit/remote/remote.py
```diff
@@ -939,6 +941,9 @@ def _version_from_hash(
         for s in additional_context:
             h.update(bytes(s, "utf-8"))
 
+        if default_inputs:
+            h.update(cloudpickle.dumps(default_inputs))
```
thomasjpfan, line 945: If the default inputs are big in-memory objects, then `clockpickle.dumps` will add more overhead. I do not think this is the case for Flyte.

Do you see use cases where the default inputs are big in-memory objects?

pingsutw: That's not the common use case IMO.

wild-endeavor (approved):

PR #2372, opened 2024-04-23
Fix use default inputs with remote LaunchPlan

andresgomezfrr:
## Tracking issue

https://github.com/flyteorg/flyte/issues/5270

These changes are needed to use remote LP without fixed inputs when input + default inputs are provided.

In line comment

## How was this patch tested?
Unit test and manual test

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -965,9 +965,8 @@ def create_and_link_node_from_remote(
     for k in sorted(typed_interface.inputs):
         var = typed_interface.inputs[k]
         if k not in kwargs:
-            if _inputs_not_allowed and _ignorable_inputs:
-                if k in _ignorable_inputs or k in _inputs_not_allowed:
-                    continue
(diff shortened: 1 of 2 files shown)


flytekit/core/promise.py
```diff
@@ -965,9 +965,8 @@ def create_and_link_node_from_remote(
     for k in sorted(typed_interface.inputs):
         var = typed_interface.inputs[k]
         if k not in kwargs:
-            if _inputs_not_allowed and _ignorable_inputs:
```
andresgomezfrr, line 968: if `_inputs_not_allowed` is None, which means no fixed inputs the exception is thrown, so is not possible to use a remote LP without fixed inputs even if all inputs are provided using `*kwargs` and default inputs.
https://github.com/flyteorg/flytekit/blob/b85a4fca36ccc47c5bd9ca1ef74043af1e5eae2d/flytekit/core/promise.py#L971

pingsutw (approved):
LGTM. Thank you!

andresgomezfrr:
> LGTM. Thank you!
thank you for the review! :) 

I think you need to merge this one, I can't do it @pingsutw

PR #2375, opened 2024-04-23
refactor(import): Remove meaningless underscore imports

MortalHappiness:
## Tracking issue
N/A

There are many `import pkg as _pkg` patterns in the codebase, and many of them are meaningless. For example, there are some files containing the following weird code snippet

```python
import os
import os as _os
```

https://github.com/flyteorg/flytekit/blob/f1689b7dbd3e872dbb48e92d968c87826ffc6bde/flytekit/clis/flyte_cli/main.py#L3-L4

which imports the `os` package twice.

Based on my own inference, perhaps the predecessors wanted to prevent exporting package names. However, renaming the builtin packages is meaningless.

This PR only deals with the `import pkg as _pkg` case. There are still many cases but refactoring them may introduce some unexpected behaviors. One can find these cases by searching `as _` string in the codebase.


wild-endeavor (approved):

eapolinario:
How did you choose which ones to remove? Asking because we still have a fair amount of imports like that: https://gist.github.com/eapolinario/c340dd6fae56c25729091f616b6b1cdd.

MortalHappiness:
@eapolinario I choose the imports with `import pkg as _pkg` pattern. I do not remove those imports like `import nested.pkg as _pkg` or `from pkg import obj as _obj`.

I do not remove those imports because I do not know the context why the predecessors rename the imports by adding the underscores as I mentioned in the PR description. Perhaps they want to keep them private? I am afraid that my changes will break some functionalities, so only those obviously meaningless imports are deleted.

PR #2377, opened 2024-04-24
Add unauthenticated Rust Flytekit Remote Client Task endpoints

austin362667:
## Tracking issue

https://github.com/flyteorg/flyte/issues/5344
https://github.com/flyteorg/flytekit/pull/2328

This PR aims to be a sub-task that validating feasibility of `Replace Python gRPC with Rust`.
As a minimum viable PR, we can end-to-end test no-auth `Task` APIs only, currently.

The communication between `flytekit.remote`(Python) `flyrs.FlyteRemote`(Rust) is throrugh serialization/deserialization as bytes string.

Ported 4 task-related APIs in `flytekit/clients/friendly_rs.py` and `flyrs/src/lib.py`, 
- `create_task()`
- `list_task_ids_paginated()`
- `list_tasks_paginated()`
- `get_task()`

So we can use Rust gRPC Client without authentication by setting `enable_rust` to `True` in `flytekit.remote.FlyteRemote`, default `False`, e.g.,
`FlyteRemote(Config.auto(), default_project="flytesnacks", default_domain="development", enable_rust=True)`.
This will make a no-auth client, `FlyteRemote._client = RustSynchronousFlyteClient()`.

## How was this patch tested?

`pytest ./flyrs/tests/test_rust_remote.py`

> You can even test some Rust-only functions (without Python `grpc`).
To do so, remove Python `grpc` dependencies by `pip uninstall grpcio grpcio-status` inside you Python environment.

### Setup process

In `/flyrs`, 
`maturin init`
then,
`maturin develop`
or 
`maturin develop --release` (pass `release` flag to `cargo build`)


tests/flytekit/unit/clients/test_raw.py
```diff
@@ -16,7 +15,6 @@ def test_update_project(mock_channel, mock_admin):
 
 
 @mock.patch("flytekit.clients.raw._admin_service")
-@mock.patch("flytekit.clients.raw.grpc.insecure_channel")
```
austin362667, line 19: same here,
Removed this line because we refactored `flytekit.clients.raw.py` in order to defer `import grpc` lazily.

tests/flytekit/unit/clients/test_raw.py
```diff
@@ -7,7 +7,6 @@
...
```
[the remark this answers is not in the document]
austin362667: So no attribute `grpc` for raw will cause unit test issues.

flytekit/clients/friendly_rs.py
```diff
@@ -0,0 +1,166 @@
...
+import flyrs
+from flytekit.clients.friendly import SynchronousFlyteClient as _SynchronousFlyteClient
+from flytekit.configuration import PlatformConfig
+from flytekit.models import common as _common
+from flytekit.models import filters as _filters
+from flytekit.models import task as _task
+from flytekit.models.core import identifier as _identifier
+
+
+class RustSynchronousFlyteClient(_SynchronousFlyteClient):
```
eapolinario, line 13: Can we a disclaimer that this currently only works with unauthenticated requests?

Also, we should discuss how Auth is going to be implemented (essentially reflect [this](https://github.com/flyteorg/flytekit/blob/e3f92cf748c7852164881bc566481dfe5b3d3e40/flytekit/clients/raw.py#L53-L58) logic in rust).

flytekit/clients/friendly_rs.py
```diff
@@ -0,0 +1,166 @@
+from flyteidl.admin import common_pb2 as _common_pb2
+from flyteidl.admin import task_pb2 as _task_pb2
+
+import flyrs
```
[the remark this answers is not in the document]
austin362667: Sure!
btw, I wanna discuss about the Github Action CI process that builds the whole underlying Python extension form Rust.

flytekit/clients/friendly_rs.py
austin362667: Of course, thanks for reminding.

Plan to work on `Auth` and `Rust gRPC channel interceptor` in subsequent PRs.

I'd love to discuss how `Auth` is going to be implemented and the ability to test all the various `Oauth` setups.
I'm totally a newbie in authentication mechanisms.

flyrs/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+    ) -> PyResult<PyObject> {
+        let bytes = bytes_obj.as_bytes();
+        let decoded: admin::NamedEntityIdentifierListRequest =
+            match Message::decode(&bytes.to_vec()[..]) {
+                Ok(de) => de,
+                Err(error) => panic!(
+                    "Failed at decoding requested object from bytes string: {:?}",
+                    error
+                ),
+            };
```
pingsutw, line 139: Should we add a helper function to decode the proto? something like
```rust
let (req, err) = decode_proto(...)
```

flyrs/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+        let res = (match self.runtime.block_on(self.admin_service.list_task_ids(req)) {
+            Ok(res) => res,
+            Err(error) => panic!(
+                "Failed at awaiting response from  gRPC service server: {:?}",
+                error
+            ),
+        })
+        .into_inner();
+        let mut buf = vec![];
+        match res.encode(&mut buf) {
```
pingsutw, line 151: same here

flyrs/tests/test_rust_remote.py
```diff
@@ -0,0 +1,57 @@
...
+def test_register_task():
+    @task()
+    def my_test_task(n: int) -> int:
+        return n
+
+    remote_rs = FlyteRemote(Config.auto(), default_project=PROJECT, default_domain=DOMAIN, enable_rust=True)
+    mock_image_config = MagicMock(default_image=MagicMock(full="fake-cr.io/image-name:tag"))
+    flyte_task = remote_rs.register_task(
+        entity=my_test_task,
+        serialization_settings=SerializationSettings(image_config=mock_image_config),
```
pingsutw, line 28:
```suggestion
        serialization_settings=SerializationSettings(image_config=ImageConfig(default_image="flyte-cr.io/image-name:tag")),
```

flyrs/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "flyrs"
+version = "0.1.0"
+edition = "2021"
```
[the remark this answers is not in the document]
austin362667: I think the [latest](https://doc.rust-lang.org/book/appendix-05-editions.html#:~:text=At%20the%20time%20of%20this,Rust%202018%2C%20and%20Rust%202021.) Rust / Cargo edition is `rust2021`?

flyrs/src/lib.rs
```diff
@@ -0,0 +1,200 @@
...
+#[pyclass(subclass)]
+pub struct FlyteClient {
+    admin_service: AdminServiceClient<Channel>,
+    runtime: Runtime,
+}
+
+#[pymethods]
+impl FlyteClient {
+    #[new] // Without this, you cannot construct the underlying class in Python.
+    pub fn new(endpoint: &str) -> PyResult<FlyteClient> {
```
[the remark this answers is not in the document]
austin362667: Certainly, thanks for reminding!

I modified a bit by adding bounds in [example](https://pyo3.rs/v0.21.1/function/signature#using-pyo3signature--).

flyrs/tests/test_rust_remote.py
```diff
@@ -0,0 +1,57 @@
+import os
```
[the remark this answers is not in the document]
austin362667: Sure, and I think it's better moving `test_rust_remote.py`to integration tests after CI can compile (`maturin build`) this extension and install it correctly.

flyrs/tests/test_rust_remote.py
```diff
@@ -0,0 +1,57 @@
...
+    task_py = remote_py.fetch_task(project=PROJECT, domain=DOMAIN, name=TASK_NAME, version=VERSION_ID)
+    assert isinstance(task_py, FlyteTask)
+    assert f"{task_py.id}" == f"TASK:{PROJECT}:{DOMAIN}:{TASK_NAME}:{VERSION_ID}"
+
+
+def test_fetch_task_and_compare():
+    remote_py = FlyteRemote(Config.auto(), default_project=PROJECT, default_domain=DOMAIN)
+    remote_rs = FlyteRemote(Config.auto(), default_project=PROJECT, default_domain=DOMAIN, enable_rust=True)
+
+    task_py = remote_py.fetch_task(project=PROJECT, domain=DOMAIN, name=TASK_NAME, version=VERSION_ID)
```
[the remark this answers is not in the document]
austin362667: removed, thanks!

flyrs/tests/test_rust_remote.py
austin362667: Encounter error `AttributeError: 'str' object has no attribute 'full'`,

Screenshot 2024-04-30 at 3 46 37 AM

flyrs/Cargo.toml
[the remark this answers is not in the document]
pingsutw: ah, I see.

flyrs/tests/test_rust_remote.py
pingsutw: try this
```
ImageConfig.auto(img_name="flyte-cr.io/image-name:tag")
```

austin362667: done, thx

flyrs/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+    #[pyo3(signature = (endpoint, **kwargs))]
+    pub fn new(endpoint: &str, kwargs: Option<&Bound<'_, PyDict>>) -> PyResult<FlyteClient> {
+        // Use Atomic Reference Counting abstractions as a cheap way to pass string reference into another thread that outlives the scope.
+        let s = Arc::new(endpoint);
+        // Check details for constructing Tokio asynchronous `runtime`: https://docs.rs/tokio/latest/tokio/runtime/struct.Builder.html#method.new_current_thread
+        let rt = match Builder::new_current_thread().enable_all().build() {
+            Ok(rt) => rt,
+            Err(error) => panic!("Failed to initiate Tokio multi-thread runtime: {:?}", error),
+        };
+        // Check details for constructing `channel`: https://docs.rs/tonic/latest/tonic/transport/struct.Channel.html#method.builder
```
thomasjpfan, line 96: Currently, a third party library can extend `FlyteRemote` by subclassing it and use `RawSynchronousFlyteClient._channel` directly as the authenticated channel for GRPC calls.

Assuming, we get intercepters + authentication to work in Rust, is there a way to reuse the authenticated channel in a third party library?

For example, can we wrap the channel object and expose it in Python? Then the channel can be passed to other clients that want to use the channel for other services. Concretely:

```python
channel = flyrs.get_authenticated_channel(endpoint)

flytekit_client = flyrs.FlyteClient(channel)
another_client = another_lib.Client(channel)
```

eapolinario: It might be worth considering that if the aim is to create a subclass of `FlyteRemote`, it's preferable for those classes to rely on the Flyte client configured in the parent class, rather than assuming that the flyte client is implemented via grpc and assume that there's an underlying `_channel`. Consider for a second the following, we could imagine a scenario where there's a different FlyteClient that doesn't use grpc at all but instead interacts with the HTTP endpoints directly. (Just to clarify, this isn't a reality yet; I'm simply using it as an example to emphasize the importance of exposing a channel as part of the Flyte client.) In that case it's not ok to assume that there's an underlying channel.

You might be talking about the case where someone is implementing a separate flyte client. If that's the case, I think it's fair to assume that they can (should?) bypass our implementation and re-implement auth+grpc.

wdyt?

thomasjpfan: > In that case it's not ok to assume that there's an underlying channel.

This makes sense in terms of the user level `FlyteRemote`. I'm considering the case of extending this PR's `RustFlyteRemote`, which I assume will be GRPC. In my above example, `another_lib.Client` is a external Rust client that uses the same channel as the `RustFlyteRemote`. 

> If that's the case, I think it's fair to assume that they can (should?) bypass our implementation and re-implement auth+grpc.

From a software design point of view, this is reasonable. It matches the current Python implementation which already hides the underlying `_channel`. An external library can create it's own channel, but that will show the authentication screen twice for PCKE auth. (One auth screen for `flytekit` and another one for the external library). To work around this, an external library can reuse the already authenticated `_channel` from `flytekit` and use that for GRPC calls.

---

For this PR, I see three possible paths for external libraries

1. Wrap Rust's GRPC Channel in Python so that external library can use it to make GRPC calls from Rust with the authenticated channel.
2. External library will build it's own `another_lib.RustFlyteClient` that extends `flyrs.RustFlyteClient`. The external library will use `another_lib.RustFlyteClient` instead of `flyters.RustFlyteClient`.
3. Have Auth happen in Python (outside of Rust) and pass in the credentials to `flyrs.RustFlyteClient`. External library will use the same credentials in it's client. This would be nice for HTTP clients as well, because I suspect they will need the same credentials. 

If we can get option 3 to work, I prefer it because it is the most flexible.

austin362667: I like option 3.
We reuse the credentials from the Python (`flytekit`) authenticator and pass them into `flyrs.RustFlyteClient` to wrap it up with Rust grpc channel via Interceptors.

flyrs/tests/test_rust_remote.py
```diff
@@ -0,0 +1,55 @@
...
+            image_config=ImageConfig.auto(img_name="flyte-cr.io/image-name:tag")
+        ),
+        version=VERSION_ID,
+    )
+    assert isinstance(flyte_task, FlyteTask)
+    assert f"{flyte_task.id}" == f"TASK:{PROJECT}:{DOMAIN}:{TASK_NAME}:{VERSION_ID}"
+
+
+def test_fetch_task_without_grpc():
+    subprocess.run(["pip", "uninstall", "grpcio", "grpcio-status"])
```
thomasjpfan, line 34: Having a test uninstall `grpcio` during a run feels like an anti-pattern. I think `flyrs` should have it's own CI environment without `grpcio` for testing.

flyrs/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+[lib]
+name = "flyrs"
+crate-type = ["cdylib"]
+
+[dependencies]
+pyo3 = "0.21.2"
```
thomasjpfan, line 12: I want to see if this works with `abi3-py38`. This will help a lot of building the packaging.

**Note**: This can be a follow up item.

flytekit/clients/friendly_rs.py
```diff
@@ -0,0 +1,168 @@
...
+            the database must match the existing definition exactly. Furthermore, as long as the request
+            remains identical, calling this method multiple times will result in success.
+
+        :param flytekit.models.core.identifier.Identifier task_identifer: The identifier for this task.
+        :param flytekit.models.task.TaskSpec task_spec: This is the actual definition of the task that
+            should be created.
+        :raises flytekit.common.exceptions.user.FlyteEntityAlreadyExistsException: If an identical version of the
+            task is found, this exception is raised.  The client might choose to ignore this exception because the
+            identical task is already registered.
+        :raises grpc.RpcError:
```
thomasjpfan, line 62: What error does this raise from Rust?

flyrs/Cargo.toml
austin362667: Sure, updated TODO list at https://github.com/flyteorg/flytekit/pull/2328#issue-2226044220

flytekit/clients/friendly_rs.py
austin362667: removed, thanks!

flyrs/tests/test_rust_remote.py
austin362667: Sure, updated TODO list at https://github.com/flyteorg/flytekit/pull/2328#issue-2226044220

flyrs/pyproject.toml
```diff
@@ -0,0 +1,15 @@
...
+requires = ["maturin>=1.5,<2.0"]
+build-backend = "maturin"
+
+[project]
+name = "flyrs"
+requires-python = ">=3.8"
+classifiers = [
+    "Programming Language :: Rust",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
```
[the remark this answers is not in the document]
austin362667: Removed, and it was generated by running command `maturin init`.

flyrs/src/lib.rs
thomasjpfan: I have not tried manually refreshing, but I feel like it will work. Although, if I manually refresh before a GRPC call, I may end up authenticating without needing to make a GRPC call. Currently, we authenticate in a GRPC intercepter to do auth at the very last moment.

For the follow up for this PR and option 3 in https://github.com/flyteorg/flytekit/pull/2377/files#r1589988041, I'll like to do the Auth when `FlyteRemote.client` is created. Concretely, I want to assume that `FlyteRemote.client` is created only if we need to communicate with GRPC.

flyrs/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+        // TODO: generally handle more protocols, e.g., `https://`
+        let endpoint_uri = match format!("http://{}", *s.clone()).parse::<Uri>() {
+            Ok(uri) => uri,
+            Err(error) => panic!(
```
troychiu, line 100: Are the panics intended?

austin362667: I don't think so; I'm on my way to removing it.

pingsutw:
Closed this since we have merged it into `flyrs` branch

PR #2381, opened 2024-04-25
Only check for secret groups during registration context

thomasjpfan:
## Tracking issue

Follow up to https://github.com/flyteorg/flytekit/pull/2355

This is required to make sure that the secrets groups check never happens during execution time.

With this PR, the secrets group check only happens during registration time.

## How was this patch tested?

There are current tests that checks `Secrets` construction when `Execution.mode` is `None`. I added more tests to make sure that the groups are not checked during execution time.

### DIFF
--- a/flytekit/models/security.py
+++ b/flytekit/models/security.py
@@ -44,7 +44,7 @@ class Secret(_common.FlyteIdlEntity):
         from flytekit.configuration.plugin import get_plugin
         from flytekit.core.context_manager import FlyteContextManager
 
-        # Only check for the groups during registration
+        # Only check for the groups during registration.
         execution = FlyteContextManager.current_context().execution_state
         in_registration_context = execution.mode is None
         if in_registration_context and get_plugin().secret_requires_group() and self.group is None:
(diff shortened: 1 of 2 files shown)


flytekit/models/security.py
```diff
@@ -42,8 +42,12 @@ class MountType(Enum):
 
     def __post_init__(self):
         from flytekit.configuration.plugin import get_plugin
+        from flytekit.core.context_manager import FlyteContextManager
 
-        if get_plugin().secret_requires_group() and self.group is None:
+        # Only check for the groups during registration.
+        execution = FlyteContextManager.current_context().execution_state
+        in_registration_context = execution.mode is None
```
pmahindrakar-oss, line 49: Should we have this to be until function which can be used at other places if needed to determine if we are in registration context

pmahindrakar-oss (dismissed):

flytekit/models/security.py
thomasjpfan: If this comes up more often, I would add a new `is_registration` method to `ExecutionState`:

https://github.com/flyteorg/flytekit/blob/afa1fc7f271e042b3f80d6349abc529a13898c84/flytekit/core/context_manager.py#L453

As for this PR, I rather not add public API for now. This is first time in the codebase, we check for `execution.mode is None`.

eapolinario (approved):

PR #2386, opened 2024-04-29
Add rust-built `target/` to GitIgnore

austin362667:
## Tracking issue

Register takes too much time to complete.
(longer than 3 minutes, usually the serialization completes in seconds)

Screenshot 2024-04-29 at 4 39 59 PM

Register takes too much time to complete in terms of the [compression](https://github.com/flyteorg/flytekit/blob/0c1d182bc83550ed0fd4375db1f7e7b2a41ad843/flytekit/tools/fast_registration.py#L55) of the tarball.

There are usually two factors result in different elapsed time in compression.
1. Total size of the source files waiting to be compressed.
2. The compress level of the underlying compression algorithm.

The `compresslevel` of gzip library remains unchanged, please check https://docs.python.org/3/library/gzip.html

> The `compresslevel` argument is an integer from `0` to `9` controlling the level of compression; `1` is fastest and produces the least compression, and `9` is slowest and produces the most compression. `0` is no compression. The default is `9`.

So we came to the conclusion that there were too many files being compressed.
The excessive compression of files happens to be the rust compiled files in `flyrs/target/`.
We have to ignore them when serialization in `fast_package()`.

By adding path `flyrs/target/` to `.gitignore`, the [`IgnoreGroup`](https://github.com/flyteorg/flytekit/blob/0c1d182bc83550ed0fd4375db1f7e7b2a41ad843/flytekit/tools/fast_registration.py#L33) from `fast_package()` will also ignore that path.

Will it cause any issue in `build`/`install` `flyrs` python extension in the future?

## How was this patch tested?

Screenshot 2024-04-29 at 4 40 35 PM

### Setup process

`pyflyte register wf.py` when you have built Rust dependencies inside `flytekit` repo via `cargo build` or `maturin develop` or something like that.

## Related PRs

https://github.com/flyteorg/flytekit/pull/2377


.gitignore
```diff
@@ -34,6 +34,7 @@ htmlcov
 docs/source/_tags/
 .hypothesis
 .npm
+/flyrs/target
```
pingsutw, line 37: nit
```suggestion
/**/target
```

pingsutw (approved):

eapolinario (approved):

PR #2390, opened 2024-05-06
pickle dictionary when it isn't JSON serializable

samhita-alla:

This is useful for serializing dictionaries that aren't JSON serializable. To enable pickling the data, the `dict` type needs to be annotated with `allow_pickle` set to `True`:

```python
@task
def t1() -> Annotated[dict, kwtypes(allow_pickle=True)]
    ...
```

Pickling only happens if the dict isn't JSON serializable. 

Modify `DictTransformer` to add support for pickling dictionaries and include relevant test cases. I've also updated the `BotoTask` output type to allow pickling when the dict isn't JSON serializable.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1716,7 +1716,7 @@ class DictTransformer(TypeTransformer[dict]):
         return False, base_type
 
     @staticmethod
-    def dict_types(python_type: Type) -> typing.Tuple[Type[typing.Any], Type[typing.Any]]:
+    def dict_types(python_type: Type) -> typing.Tuple[typing.Any, typing.Any]:
         if get_origin(python_type) is Annotated:
             base_type, *_ = DictTransformer.extract_types_or_metadata(python_type)
             tp = get_args(base_type)
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1130,7 +1130,7 @@ class TypeEngine(typing.Generic[T]):
             from flytekit.extras import sklearn  # noqa: F401
         if is_imported("pandas"):
             try:
-                pass
+                from flytekit.types.schema.types_pandas import PandasSchemaReader, PandasSchemaWriter  # noqa: F401
             except ValueError:
                 logger.debug("Transformer for pandas is already registered.")
             try:
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -55,13 +55,7 @@ from flytekit.models.literals import (
     Union,
     Void,
 )
-from flytekit.models.types import (
-    LiteralType,
-    SimpleType,
-    StructuredDatasetType,
-    TypeStructure,
-    UnionType,
-)
+from flytekit.models.types import LiteralType, SimpleType, StructuredDatasetType, TypeStructure, UnionType
 
 T = typing.TypeVar("T")
 DEFINITIONS = "definitions"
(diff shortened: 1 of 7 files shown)


flytekit/core/type_engine.py
```diff
@@ -54,7 +55,14 @@
...
 )
-from flytekit.models.types import LiteralType, SimpleType, StructuredDatasetType, TypeStructure, UnionType
+from flytekit.models.types import (
+    LiteralType,
+    SimpleType,
+    StructuredDatasetType,
+    TypeStructure,
+    UnionType,
+)
+from flytekit.types.pickle import FlytePickle
```
dansola, line 65: I think this is a circular import. We might need to import it inside `dict_to_generic_literal()`

plugins/flytekit-aws-sagemaker/flytekitplugins/awssagemaker_inference/boto3_agent.py
```diff
@@ -54,9 +57,20 @@ async def do(self, task_template: TaskTemplate, inputs: Optional[LiteralMap] = N
             inputs=inputs,
         )
 
-        outputs = None
+        outputs = {"result": {"result": None}}
```
pingsutw, line 60: Isn't None Json serializable? could we create a dummy class here

Screenshot 2024-05-08 at 7 12 33 PM

flytekit/core/type_engine.py
```diff
@@ -1679,22 +1677,53 @@ def get_dict_types(t: Optional[Type[dict]]) -> typing.Tuple[Optional[type], Opti
...
     def get_literal_type(self, t: Type[dict]) -> LiteralType:
         """
         Transforms a native python dictionary to a flyte-specific ``LiteralType``
         """
-        tp = self.get_dict_types(t)
+        if get_origin(t) is Annotated:
+            base_type, *_ = self.extract_types_or_metadata(t)
+            tp = get_args(base_type)
+        else:
+            tp = self.extract_types_or_metadata(t)
```
pingsutw, line 1725: nit: could we move the logic here into `self.extract_types_or_metadata`? something like

```python
def extract_types_or_metadata(self, python_type):
    if get_origin(python_type) is Annotated:
        python_type = get_args(python_type)[0]
    ... # original code here
```

so we can change the code here to
```python
tp = self.extract_types_or_metadata(t)
```

flytekit/core/type_engine.py
```diff
@@ -1710,21 +1739,37 @@ def to_literal(
...
         return Literal(map=LiteralMap(literals=lit_map))
 
     def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[dict]) -> dict:
         if lv and lv.map and lv.map.literals is not None:
-            tp = self.get_dict_types(expected_python_type)
+            if get_origin(expected_python_type) is Annotated:
+                base_type, *_ = self.extract_types_or_metadata(expected_python_type)
+                tp = get_args(base_type)
+            else:
+                tp = self.extract_types_or_metadata(expected_python_type)
```
pingsutw, line 1771: ditto

plugins/flytekit-aws-sagemaker/flytekitplugins/awssagemaker_inference/boto3_agent.py
samhita-alla: modified it to `None`

flytekit/core/type_engine.py
samhita-alla: i'm not sure if this is possible 'cause i don't want to always return the dict key and value types. i also need access to the annotation metadata to check if `allow_pickle` is set to `True`.

```python
def is_pickle(self, python_type: Type[dict]) -> typing.Tuple[bool, Type]:
        base_type, *metadata = self.extract_types_or_metadata(python_type)

        for each_metadata in metadata:
            if isinstance(each_metadata, OrderedDict):
                allow_pickle = each_metadata.get("allow_pickle", False)
                return allow_pickle, base_type

        return False, base_type
```

samhita-alla: so if `extract_types_or_metadata` returns the args, i can either fetch the metadata or the actual dict types depending on what i need. makes sense?

samhita-alla: i could probably encapsulate this logic in a separate function to avoid repetition. ok?

pingsutw (approved):

PR #2391, opened 2024-05-06
Fix `__annotations__` key not found

austin362667:
## Tracking issue

cc @pingsutw 
image
image

The error happened from some Python version (like `3.10.13`) but worked in some others.
`Failed with Unknown Exception <class 'KeyError'> Reason: '__annotations__'`

However, I can't reproduce the error on my python environment with version `3.10.13`.

There might be other package version issues, because it works in different python environment with different set of packages installed.

In spite of the unknown root cause of the error, we can still eliminate error by simply using `.` to select attribute path, `aa.__annotations__`.

Original approach returns the `__dict__` of a class as type Dictionary via `vars()` then get it by `obj['__annotations__']`, see https://docs.python.org/3/library/functions.html#vars .

As the usage of selecting attribute `__annotations__`, please check https://docs.python.org/3.12/howto/annotations.html .

image

### Setup process

`pyflyte register ...` and I can't assure 100% you can reproduce the error.

### Screenshots

1. _3.10.13_
image

3. _3.10.0_
image

4. _3.9.19_
image

6. _3.8.19_
image

## Related PRs

https://github.com/flyteorg/flytekit/commit/0cc8bbcd795d7f7e3f5898a077a7ef77ccfee951

### DIFF
--- a/flytekit/types/structured/structured_dataset.py
+++ b/flytekit/types/structured/structured_dataset.py
@@ -161,8 +161,7 @@ def extract_cols_and_format(
             if hasattr(aa, "__annotations__"):
                 # handle dataclass argument
                 d = collections.OrderedDict()
-                dm = vars(aa)
-                d.update(dm["__annotations__"])
+                d.update(aa.__annotations__)
                 ordered_dict_cols = d
             elif isinstance(aa, dict):
                 d = collections.OrderedDict()


eapolinario:
flytekit also supports python 3.8 and 3.9. Is this going to work in those cases as well?

austin362667:
>flytekit also supports python 3.8 and 3.9. Is this going to work in those cases as well?

Yes, It also works in python 3.8 and 3.9.

And be aware that `typing.Annotated` has been supported since python 3.9. 
https://docs.python.org/3/library/typing.html#typing.Annotated

Before python 3.9, use `typing_extensions.Annotated`.

pingsutw (approved):

PR #2399, opened 2024-05-08
[wip] Use Json type to represent naked dictionaries

eapolinario:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5318

As described in https://github.com/flyteorg/flyte/issues/5318, we can use a json string as the transport for a variety of types, including naked dictionaries.

Leverage the new Json type defined in https://github.com/flyteorg/flyte/pull/5337 to replace the usage of protobuf Struct as the transport for naked dictionaries with a json-encoded string. 

Keep in mind that this is WIP. TODO:
1. [ ]  define how custom encoders are passed in, if necessary,
2. [ ] make this backwards compatible with the old Struct 
3. [ ] fix+add tests

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -15,6 +15,7 @@ from abc import ABC, abstractmethod
 from functools import lru_cache
 from typing import Dict, List, NamedTuple, Optional, Type, cast
 
+import jsonpickle
 from dataclasses_json import DataClassJsonMixin, dataclass_json
 from flyteidl.core import literals_pb2
 from google.protobuf import json_format as _json_format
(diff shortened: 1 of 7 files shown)


flytekit/core/type_engine.py
```diff
@@ -1702,15 +1704,16 @@ def get_literal_type(self, t: Type[dict]) -> LiteralType:
...
+        return _type_models.LiteralType(simple=_type_models.SimpleType.JSON)
 
     def to_literal(
         self, ctx: FlyteContext, python_val: typing.Any, python_type: Type[dict], expected: LiteralType
     ) -> Literal:
         if type(python_val) != dict:
             raise TypeTransformerFailedError("Expected a dict")
 
-        if expected and expected.simple and expected.simple == SimpleType.STRUCT:
+        # TODO: back-compat with struct
```
wild-endeavor, line 1715: I don't think we should do any backwards compat right?

eapolinario:
msgpack-binary idl subsumes this.

PR #2401, opened 2024-05-09
feat(bindings): Task arguments default value binding

MortalHappiness:
## Tracking issue

Resolves: flyteorg/flyte#5321

See the issue description for details.

if the key is not in `kwargs` but in `interface.inputs_with_defaults`, add the value in `interface.inputs_with_defaults` to `kwargs`.

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -1086,16 +1086,7 @@ def create_and_link_node(
                     raise _user_exceptions.FlyteAssertion("Cannot use non-hashable object as default argument")
                 kwargs[k] = default_val
             else:
-                from flytekit.core.base_task import Task
-
                 error_msg = f"Input {k} of type {interface.inputs[k]} was not specified for function {entity.name}"
-
-                _, _default = interface.inputs_with_defaults[k]
-                if isinstance(entity, Task) and _default is not None:
-                    error_msg += (
-                        ". Flyte workflow syntax is a domain-specific language (DSL) for building execution graphs which "
-                        "supports a subset of Python’s semantics. When calling tasks, all kwargs have to be provided."
-                    )
 
                 raise _user_exceptions.FlyteAssertion(error_msg)
         v = kwargs[k]
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -690,6 +690,7 @@ def binding_data_from_python_std(
     t_value: Any,
     t_value_type: type,
     nodes: List[Node],
+    is_union_type_variants_expanded: bool = True,
 ) -> _literals_models.BindingData:
     # This handles the case where the given value is the output of another task
     if isinstance(t_value, Promise):
@@ -712,6 +713,7 @@ def binding_data_from_python_std(
         # If it is a container type, then we need to iterate over the variants in the Union type, try each one. This is
         # akin to what the Type Engine does when it finds a Union type (see the UnionTransformer), but we can't rely on
         # that in this case, because of the mix and match of realized values, and Promises.
+    elif t_value is not None and is_union_type_variants_expanded and expected_literal_type.union_type is not None:
         for i in range(len(expected_literal_type.union_type.variants)):
             try:
                 lt_type = expected_literal_type.union_type.variants[i]
@@ -778,9 +780,17 @@ def binding_from_python_std(
     expected_literal_type: _type_models.LiteralType,
     t_value: Any,
     t_value_type: type,
+    is_union_type_variants_expanded: bool = True,
 ) -> Tuple[_literals_models.Binding, List[Node]]:
     nodes: List[Node] = []
-    binding_data = binding_data_from_python_std(ctx, expected_literal_type, t_value, t_value_type, nodes)
+    binding_data = binding_data_from_python_std(
+        ctx,
+        expected_literal_type,
+        t_value,
+        t_value_type,
+        nodes,
+        is_union_type_variants_expanded=is_union_type_variants_expanded,
+    )
     return _literals_models.Binding(var=var_name, binding=binding_data), nodes
 
 
@@ -1072,6 +1082,7 @@ def create_and_link_node(
 
     for k in sorted(interface.inputs):
         var = typed_interface.inputs[k]
+        is_default_arg_used = False
         if var.type.simple == SimpleType.NONE:
             raise TypeError("Arguments do not have type annotation")
         if k not in kwargs:
@@ -1085,6 +1096,7 @@ def create_and_link_node(
                 if not isinstance(default_val, Hashable):
                     raise _user_exceptions.FlyteAssertion("Cannot use non-hashable object as default argument")
                 kwargs[k] = default_val
+                is_default_arg_used = True
             else:
                 error_msg = f"Input {k} of type {interface.inputs[k]} was not specified for function {entity.name}"
                 raise _user_exceptions.FlyteAssertion(error_msg)
@@ -1104,6 +1116,7 @@ def create_and_link_node(
                 expected_literal_type=var.type,
                 t_value=v,
                 t_value_type=interface.inputs[k],
+                is_union_type_variants_expanded=not is_default_arg_used,
             )
             bindings.append(b)
             nodes.extend(n)
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -2,13 +2,12 @@ from __future__ import annotations
 
 import collections
 import inspect
-import typing
 from copy import deepcopy
 from enum import Enum
-from typing import Any, Coroutine, Dict, List, Optional, Set, Tuple, Union, cast
+from typing import Any, Coroutine, Dict, Hashable, List, Optional, Set, Tuple, Union, cast, get_args
 
 from google.protobuf import struct_pb2 as _struct
-from typing_extensions import Protocol, get_args
+from typing_extensions import Protocol
 
 from flytekit.core import constants as _common_constants
 from flytekit.core import context_manager as _flyte_context
@@ -24,7 +23,13 @@ from flytekit.core.context_manager import (
 )
 from flytekit.core.interface import Interface
 from flytekit.core.node import Node
-from flytekit.core.type_engine import DictTransformer, ListTransformer, TypeEngine, TypeTransformerFailedError
+from flytekit.core.type_engine import (
+    DictTransformer,
+    ListTransformer,
+    TypeEngine,
+    TypeTransformerFailedError,
+    UnionTransformer,
+)
 from flytekit.exceptions import user as _user_exceptions
 from flytekit.exceptions.user import FlytePromiseAttributeResolveException
 from flytekit.loggers import logger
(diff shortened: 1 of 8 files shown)


pingsutw (commented):
Also, should we remove these [lines](https://github.com/flyteorg/flytekit/blob/cc1341574feaeca9cee98c34a101a71398c10859/flytekit/core/promise.py#L1076-L1081)? since if statement will always be false.

MortalHappiness:
> Also, should we remove these [lines](https://github.com/flyteorg/flytekit/blob/cc1341574feaeca9cee98c34a101a71398c10859/flytekit/core/promise.py#L1076-L1081)? since if statement will always be false.

Done

wild-endeavor:
could you add more tests please?  take a look at serialization tests, it should be pretty simple to test the compiled workflow.

do we do anything special for [] and {}?  You're not supposed to use those as function defaults anyways but some people might.

MortalHappiness:
> could you add more tests please? take a look at serialization tests, it should be pretty simple to test the compiled workflow.
> 
> do we do anything special for [] and {}? You're not supposed to use those as function defaults anyways but some people might.

Done

flytekit/core/promise.py
```diff
@@ -36,6 +37,33 @@
...
+
+    >>> _is_none_valid_instance_of(Optional[List[int]])
+    True
+
+    >>> _is_none_valid_instance_of(Union[str, Optional[int]])
+    True
+    """
+    if get_origin(t) is None:
+        return t is type(None)
+    return any(_is_none_valid_instance_of(arg) for arg in get_args(t))
```
wild-endeavor, line 64: `get_args` is [generic](https://docs.python.org/3/library/typing.html#typing.get_args) right?  It's used not just for Union, is there a way to scope this call to just look at Union types?  a bit worried that this check might cause a false positive somehow.

flytekit/core/promise.py
```diff
@@ -1057,32 +1085,20 @@ def create_and_link_node(
...
-                            )
-                        is_optional = True
-            if not is_optional:
-                from flytekit.core.base_task import Task
-
+            if k in interface.inputs_with_defaults and (
+                interface.inputs_with_defaults[k][1] is not None
+                or _is_none_valid_instance_of(interface.inputs_with_defaults[k][0])
+            ):
+                default_val = interface.inputs_with_defaults[k][1]
```
wild-endeavor, line 1095: `[1]` is the default arg right?  can you add a comment here reminding that?

tests/flytekit/unit/core/test_dynamic.py
```diff
@@ -95,6 +95,42 @@ def ranged_int_to_str(a: int) -> typing.List[str]:
...
+        with context_manager.FlyteContextManager.with_context(
+            ctx.with_execution_state(
+                ctx.execution_state.with_params(
+                    mode=ExecutionState.Mode.TASK_EXECUTION,
+                )
+            )
+        ) as ctx:
+            input_literal_map = TypeEngine.dict_to_literal_map(ctx, {"a": input_val})
+            dynamic_job_spec = dt.dispatch_execute(ctx, input_literal_map)
+            assert len(dynamic_job_spec.nodes) == 1
```
wild-endeavor, line 130: add
```
assert dynamic_job_spec.nodes[0].inputs[0].binding.scalar.primitive is not None
```

flytekit/core/promise.py
```diff
@@ -36,6 +37,33 @@
 from flytekit.models.types import SimpleType
 
 
+def _is_none_valid_instance_of(t: Type) -> bool:
```
wild-endeavor, line 40: can you copy these examples into a unit test also?

wild-endeavor (commented):
thanks!  just minor comments.  can you take a look at the failing unit test also?  for some performance reason, we split out tests that rely on pandas into a separate gh action.  Take a look at https://github.com/flyteorg/flytekit/blob/91cfb973eecca2ef35f0f5959f9754a1f51af451/tests/flytekit/unit/core/test_artifacts.py#L26

tests/flytekit/unit/core/test_serialization.py
```diff
@@ -84,6 +122,287 @@ def raw_container_wf(val1: int, val2: int) -> int:
     assert sum_spec.template.container.image == "alpine"
 
 
+@pytest.mark.parametrize(
```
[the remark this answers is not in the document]
MortalHappiness: Done

flytekit/core/promise.py
```diff
@@ -1057,32 +1085,20 @@ def create_and_link_node(
 
     for k in sorted(interface.inputs):
         var = typed_interface.inputs[k]
+        if var.type.simple == SimpleType.NONE:
```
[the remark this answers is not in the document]
MortalHappiness: I changed the error message to `Arguments do not have type annotation`. This case corresponds to this testcase https://github.com/flyteorg/flytekit/pull/2401/files#diff-fa72fab667c8393cca8d5afd3943db0edc55794423b7cfc3ab33d0f8c3146650R683-R699

flytekit/core/promise.py
MortalHappiness: Done

tests/flytekit/unit/core/test_dynamic.py
MortalHappiness: Done

flytekit/core/promise.py
MortalHappiness: This function is deleted. I use `UnionTransformer.is_optional_type` instead as mentioned above.

flytekit/core/promise.py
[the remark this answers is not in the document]
wild-endeavor: thanks

wild-endeavor (approved):
still need to look at monodocs.  i think we can do that later today.

wild-endeavor:
eduardo fixed the build separately https://github.com/flyteorg/flyte/pull/5425/files  can merge this after passing

tests/flytekit/unit/core/test_serialization.py
```diff
@@ -475,3 +486,302 @@ def z(a: int, b: str) -> typing.Tuple[int, str]:
...
+        return t1(a=input_val)
+
+    @workflow
+    def wf_with_sub_wf() -> tuple[typing.Optional[int], typing.Optional[int]]:
+        return (wf_no_input(), wf_with_input())
+
+    wf_no_input_spec = get_serializable(OrderedDict(), serialization_settings, wf_no_input)
+    wf_with_input_spec = get_serializable(OrderedDict(), serialization_settings, wf_with_input)
+
+    assert wf_no_input_spec.template.nodes[0].inputs[0].binding.value == Scalar(
```
wild-endeavor, line 654: actually sorry, i think we need to think about this case a bit more.

MortalHappiness: Do you mean I need to bind this to Union instead of Primitive?

wild-endeavor: exactly

MortalHappiness: Done. See [the latest commit](https://github.com/flyteorg/flytekit/pull/2401/commits/c9cb9df622b8ee0746ccb7d68490eb6d22412526) for details. Here I added a bool argument `is_union_type_variants_expanded`. And if the default argument is used, then `is_union_type_variants_expanded` will be set to `False`, which will bypass the for loop expansion for union types here
https://github.com/flyteorg/flytekit/blob/c9cb9df622b8ee0746ccb7d68490eb6d22412526/flytekit/core/promise.py#L706-L718

Also, I added two tests for testing `Union[List[int]]`.

flytekit/core/promise.py
```diff
@@ -684,6 +690,7 @@ def binding_data_from_python_std(
     t_value: Any,
     t_value_type: type,
     nodes: List[Node],
+    is_union_type_variants_expanded: bool = True,
```
wild-endeavor, line 693: can we remove this now with the other change?

MortalHappiness: Done

wild-endeavor (approved):

eapolinario:
This is pretty awesome! Thank you for all the hard work here, @MortalHappiness and @wild-endeavor !

flytekit/core/promise.py
```diff
@@ -1066,32 +1078,22 @@ def create_and_link_node(
...
-
+            # interface.inputs_with_defaults[k][0] is the type of the default argument
+            # interface.inputs_with_defaults[k][1] is the value of the default argument
+            if k in interface.inputs_with_defaults and (
+                interface.inputs_with_defaults[k][1] is not None
+                or UnionTransformer.is_optional_type(interface.inputs_with_defaults[k][0])
+            ):
+                default_val = interface.inputs_with_defaults[k][1]
+                if not isinstance(default_val, Hashable):
+                    raise _user_exceptions.FlyteAssertion("Cannot use non-hashable object as default argument")
```
eapolinario, line 1092: Why do we need this extra-check? Any value that can't be handled by the existing type transformers at registration time is going to show pretty prominently, right?

eapolinario: Relying on `Hashable` prevents dataclasses from being used as default arguments. This is too restrictive, since even sophisticated linters like ruff are not able to catch this example:
```
from mashumaro import DataClassDictMixin

@dataclass
class DC(DataClassDictMixin):
    a: int
    b: str

def f(dc: DC = DC(42, "43"), xs: list[int] = [1,2], d: dict[str, int] = {"abc": 42}, s: set(int) = set([1,2,3])):
    ...

if __name__ == "__main__":
    f()

```
Running ruff with https://docs.astral.sh/ruff/rules/mutable-argument-default/ turned on returns:
```
❯ ruff check example
example/dc.py:9:46: B006 Do not use mutable data structures for argument defaults
   |
 7 |     b: str
 8 |
 9 | def f(dc: DC = DC(42, "43"), xs: list[int] = [1,2], d: dict[str, int] = {"abc": 42}, s: set(int) = set([1,2,3])):
   |                                              ^^^^^ B006
10 |     ...
   |
   = help: Replace with `None`; initialize within function

example/dc.py:9:73: B006 Do not use mutable data structures for argument defaults
   |
 7 |     b: str
 8 |
 9 | def f(dc: DC = DC(42, "43"), xs: list[int] = [1,2], d: dict[str, int] = {"abc": 42}, s: set(int) = set([1,2,3])):
   |                                                                         ^^^^^^^^^^^ B006
10 |     ...
   |
   = help: Replace with `None`; initialize within function

example/dc.py:9:100: B006 Do not use mutable data structures for argument defaults
   |
 7 |     b: str
 8 |
 9 | def f(dc: DC = DC(42, "43"), xs: list[int] = [1,2], d: dict[str, int] = {"abc": 42}, s: set(int) = set([1,2,3])):
   |                                                                                                    ^^^^^^^^^^^^ B006
10 |     ...
   |
   = help: Replace with `None`; initialize within function

Found 3 errors.
No fixes available (3 hidden fixes can be enabled with the `--unsafe-fixes` option).
```

I propose we handle these 3 common errors explicitly instead. This is also going to simplify the error message. I'll get a PR out soon.

eapolinario: FYI: https://github.com/flyteorg/flytekit/pull/2651

MortalHappiness: @eapolinario I saw your PR. Quite make senses for me. Those 3 container types are indeed the most commonly used non-hashable types.

MortalHappiness: How about `collections.deque`?

PR #2403, opened 2024-05-09
Use uv in CI and to build images

eapolinario:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5346

As described in the issue, let's adopt [`uv`](https://github.com/astral-sh/uv) as the package installer for flytekit.

Use `uv` in a few places:
1. when setting up a dev environment via the new make target `setup-uv`. This assumes that a virtual environment (e.g. `uv venv`) is already setup prior to calling this. This keeps the tradition of not being too prescriptive about venvs.
  i. we can make this the default once we're satisfied with it. This will require a slight change to the docs.
2. the default flytekit image is now built using `uv`
  i. Notice the use of `--system` to ensure the packages are installed in the global environment.
4. the dev flyte image is also built using `uv`
5. We should probably also build the agent image with `uv`.

 

## How was this patch tested?
Tested CI locally.
I built and used the other two images in local sandbox.

### DIFF
--- a/.github/workflows/pythonbuild.yml
+++ b/.github/workflows/pythonbuild.yml
@@ -65,6 +65,7 @@ jobs:
           uv pip freeze
       - name: Test with coverage
         run: |
+          source .venv/bin/activate
           make unit_test_codecov
       - name: Codecov
         uses: codecov/codecov-action@v3.1.4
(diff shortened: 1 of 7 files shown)


eapolinario:
This will probably break on windows.

eapolinario:
dbt failures are unrelated. Being tracked in https://github.com/flyteorg/flyte/issues/5351.

pingsutw (approved):

PR #2404, opened 2024-05-09
Catch silent subproc_execute failures

pryce-turner:
Under some circumstances, subproc_execute will exit with 0 but have a non-null stderr.

Adds an assert statement to handle 0 exit with stderr.

## How was this patch tested?
Unit test to capture AssertionError and ensure the message contains the right content.

### Setup process
`make setup`
`make unit_test`

### DIFF
--- a/flytekit/extras/tasks/shell.py
+++ b/flytekit/extras/tasks/shell.py
@@ -79,8 +79,14 @@ def subproc_execute(command: typing.Union[List[str], str], **kwargs) -> ProcessR
     try:
         # Execute the command and capture stdout and stderr
         result = subprocess.run(command, **kwargs)
+        print(result.check_returncode())
 
-        assert result.stderr == ""
+        if "|" in command and kwargs.get("shell"):
+            logger.warning(
+                """Found a pipe in the command and shell=True.
+                This can lead to silent failures if subsequent commands
+                succeed despite previous failures."""
+            )
 
         # Access the stdout and stderr output
         return ProcessResult(result.returncode, result.stdout, result.stderr)
@@ -94,9 +100,6 @@ def subproc_execute(command: typing.Union[List[str], str], **kwargs) -> ProcessR
             Did you specify a container image in the task definition if using
             custom dependencies?\n{e}"""
         )
-
-    except AssertionError:
-        raise Exception(f"Command: {command}\nexperienced a silent failure, likely due to shell=True:\n{result.stderr}")
 
 
 def _dummy_task_func():
--- a/flytekit/extras/tasks/shell.py
+++ b/flytekit/extras/tasks/shell.py
@@ -80,6 +80,8 @@ def subproc_execute(command: typing.Union[List[str], str], **kwargs) -> ProcessR
         # Execute the command and capture stdout and stderr
         result = subprocess.run(command, **kwargs)
 
+        assert result.stderr == ""
+
         # Access the stdout and stderr output
         return ProcessResult(result.returncode, result.stdout, result.stderr)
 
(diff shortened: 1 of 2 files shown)


tests/flytekit/unit/extras/tasks/test_shell.py
```diff
@@ -370,3 +370,27 @@ def test_subproc_execute_with_shell():
...
+    with pytest.raises(Exception) as e:
+        subproc_execute(cmd)
+    assert "Failed with return code" in str(e.value)
+
+
+def test_subproc_execute_shell_error(tmp_path):
+    # This is a corner case I ran into that really shouldn't
+    # ever happen. The assert catches anything in stderr despite
+    # a 0 exit.
+    cmd = " ".join(["bcftools", "isec", "|", "gzip", "-c", ">", f"{tmp_path.joinpath('out')}"])
```
eapolinario, line 393: the exitcode `0` only happens because the second command (i.e. `gzip`) returns a `0`. Bash implements a feature called PIPESTATUS, which captures the exit codes of all commands in a pipe. It's just an array that you can verify after the pipe is run, for example:
```
bash-5.2$ inexistet command --flag | echo bla | gzip -c > out
bash: inexistet: command not found
bash-5.2$ echo "${PIPESTATUS[@]}"
127 0 0
```
Notice how the first command returns `127` and the other two return `0`.

flytekit/extras/tasks/shell.py
```diff
@@ -80,6 +80,8 @@ def subproc_execute(command: typing.Union[List[str], str], **kwargs) -> ProcessR
         # Execute the command and capture stdout and stderr
         result = subprocess.run(command, **kwargs)
 
+        assert result.stderr == ""
```
eapolinario, line 83: sometimes commands write debug information to stderr, so we should have a more resilient way of checking if a subprocess ran to completion.

tests/flytekit/unit/extras/tasks/test_shell.py
pryce-turner: TIL, thank you!

flytekit/extras/tasks/shell.py
pryce-turner: I wish they didn't but you're right.. unfortunately, subprocess.run doesn't capture PIPESTATUS it seems, only showing the return code for the last command. Do you have any thoughts? The only thing I can think of would be to check stderr for "command not found" or "error", but that's pretty awkward.

eapolinario: Unfortunately I don't have a great solution. 

pipestatus is a bash feature, so this complicates things a little since we use `/bin/sh` as the [default shell](https://github.com/flyteorg/flytekit/blob/master/flytekit/extras/tasks/shell.py#L207) in `ShellTask` and depending on the OS that's not bash (e.g. on MacOS `/bin/sh` is symlinked to bash:
```
❯ uname -a
Darwin gondor 23.3.0 Darwin Kernel Version 23.3.0: Wed Dec 20 21:31:00 PST 2023; root:xnu-10002.81.5~7/RELEASE_ARM64_T6020 arm64

❯ readlink /private/var/select/sh
/bin/bash
```
).

A low-lift solution is to write a warning to the flytekit logs that we detected the use of a pipe in the command and that means we might end up not capturing the output of the command. That's not a great solution either, but at least we let the users know that they are in pipeland, which means they will have to handle it (maybe mention pipestatus?). wdyt?

pryce-turner: Yeah, I think that's the best solution given the current constraints. It is a bit of a corner case and I don't think we should be expected to handle *every* failure state. How does the latest commit look?

eapolinario (approved):

PR #2410, opened 2024-05-11
fix(ImageSpec): Do not build image during executions

thomasjpfan:
## Tracking issue
Partly address https://github.com/flyteorg/flyte/issues/4374

Dynamic workflows will try to build the image spec during execution.

This PR stops the image builder from building during any execution mode.

## How was this patch tested?

I added unit test to check that the image builder does not launch. I also ran the following to check that it worked in the sandbox. 

```python
from flytekit import ImageSpec, task, dynamic

repro_img = ImageSpec(
    name="flytekit",
    base_image="ghcr.io/flyteorg/flytekit:py3.11-latest",
    builder="envd",
    packages=[
        "git+https://github.com/thomasjpfan/flytekit.git@3700c4710a606f5b1e13f3c815c233c39cc9f425"
    ],
    apt_packages=["git"],
    registry="localhost:30000",
)


@task(container_image=repro_img)
def foo():
    print("foo")


@dynamic(container_image=repro_img)
def bar():
    foo()

```

Note that this does not completely solve the issue. For example, if the `dynamic` workflow uses a different image, the execution will hang:

```python
new_image = repro_img.with_packages(["curl"])
another_image = repro_img.with_packages(["wget"])


@task(container_image=new_image)
def foo():
    print("foo")


@dynamic(container_image=another_image)
def bar():
    foo()
```

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -14,7 +14,6 @@ import click
 import requests
 from packaging.version import Version
 
-from flytekit.core.context_manager import FlyteContextManager
 from flytekit.exceptions.user import FlyteAssertion
 
 DOCKER_HUB = "docker.io"
@@ -235,6 +234,8 @@ class ImageBuildEngine:
     @classmethod
     @lru_cache
     def build(cls, image_spec: ImageSpec):
+        from flytekit.core.context_manager import FlyteContextManager
+
         execution_mode = FlyteContextManager.current_context().execution_state.mode
         # Do not build in executions
         if execution_mode is not None:
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -14,6 +14,7 @@ import click
 import requests
 from packaging.version import Version
 
+from flytekit.core.context_manager import FlyteContextManager
 from flytekit.exceptions.user import FlyteAssertion
 
 DOCKER_HUB = "docker.io"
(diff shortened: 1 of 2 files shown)


cosmicBboy (approved):
I'm okay with these changes as the default for flytekit: with these changes, can we document somewhere in flytekit/flytesnacks the fact that when dynamic workflows call tasks using ImageSpec, the image must already by built. The happy path case is that the dynamic workflow imagespec is the same as the task imagespecs @neverett .

### Approaches for a proper fix:
- The `unionai` image builder plugin to actually build the images at runtime. 
- Locally, calling `pyflyte run/register` somehow plucks out the tasks that are called inside a dynamic workflow and builds those images. pro of this approach is that it will work for flytekit + unionai serverless folks

@eapolinario @pingsutw @wild-endeavor there are probably other approaches here too.

neverett:
Created a docs task in Linear task to address this:

> I'm okay with these changes as the default for flytekit: with these changes, can we document somewhere in flytekit/flytesnacks the fact that when dynamic workflows call tasks using ImageSpec, the image must already by built. The happy path case is that the dynamic workflow imagespec is the same as the task imagespecs @neverett .
> ### Approaches for a proper fix:
> 
>     * The `unionai` image builder plugin to actually build the images at runtime.
> 
>     * Locally, calling `pyflyte run/register` somehow plucks out the tasks that are called inside a dynamic workflow and builds those images. pro of this approach is that it will work for flytekit + unionai serverless folks, but implementation details unclear
> 
> 
> @eapolinario @pingsutw @wild-endeavor there are probably other approaches here too.

PR #2411, opened 2024-05-11
Adds typing to @task and @workflow

thomasjpfan:
## Tracking issue

Closes https://github.com/flyteorg/flyte/issues/3681

Type checkers do not work with `flytekit` right now.

With this PR, the `@task` and `@workflow` pretends to return their original types for type checking purposes. The typing in this PR is not strictly correct, because the workflow transforms everything into Promises. I think this PR is an okay solution to get typing to work for user's workflows.

As for actually typing the Promises, I do not see a way to actually type them. Python typing allows us to change the output of the decorator, but we can not change the inputs:

```python
def task_example(f: Callable[P, R]) -> Callable[P, Promise[R]]:
    # We need to somehow express that the inputs are also Promises, and not the original types.
    def inner(*args: P.args, **kwargs: P.kwargs) -> Promise[R]:
        return f(*args, **kwargs)

    return inner
```

## How was this patch tested?

Run `mypy` on workflow code and see failures when types do not match:

```python


@task(cache=False)
def add(a: int, b: int) -> int:
    return a + b


# note how `b` is the incorrect type
@workflow
def wf(a: int, b: str) -> int:
    return add(a=a, b=b)
```

### DIFF
--- a/flytekit/core/task.py
+++ b/flytekit/core/task.py
@@ -120,7 +120,7 @@ def task(
     pod_template: Optional["PodTemplate"] = ...,
     pod_template_name: Optional[str] = ...,
     accelerator: Optional[BaseAccelerator] = ...,
-) -> Callable[[Callable[P, R]], Union[Callable[P, R], PythonFunctionTask[T]]]:
+) -> Callable[[Callable[P, R]], Callable[P, R]]:
     ...
 
 
@@ -158,7 +158,7 @@ def task(
     pod_template: Optional["PodTemplate"] = ...,
     pod_template_name: Optional[str] = ...,
     accelerator: Optional[BaseAccelerator] = ...,
-) -> Union[PythonFunctionTask[T], Callable[P, R]]:
+) -> Callable[P, R]:
     ...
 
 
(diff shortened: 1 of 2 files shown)


eapolinario:
@thomasjpfan , I like this as this fixes the problem for `@task` and `@workflow` decorators, but we can't simply enable `py.typed`, right? Otherwise all entities exported by flytekit will be type-checkable.

thomasjpfan:
>  I like this as this fixes the problem for @task and @workflow decorators, but we can't simply enable py.typed, right?

If we remove the `py.typed`, then `mypy` will continue to skip `@task` and @workflow`:

```bash
file.py:1: error: Cannot find implementation or library stub for module named "flytekit"  [import-not-found]
file.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```

Users will have to wait until we add `py.typed` before they can type check their `@task` and `@workflow`.

wild-endeavor:
I think users are okay waiting.  there's just too much in flytekit itself to clean up.  This will be an ongoing effort.  In any case this will make pyright less red at least.  Let's remove and merge?  Is the lint error caused by the py.typed file itself?

eapolinario:
The rules for distributing type information are described [here](https://typing.readthedocs.io/en/latest/source/libraries.html#providing-type-annotations). We might need to play with adding `*.pyi` files alongside the `py.typed` marker to find the right combination.

kumare3:
can we just use Any for now?

PR #1870, opened 2023-10-04
Kubeflow TensorFlow Training Operator Add Evaluator

Future-Outlier:
# TL;DR
Enable running a data service in kubeflow tensorflow training operator by utilizing the evaluator section in the TF_CONFIG.

## Describe your changes

Enable running a data service by utilizing the evaluator section in the TF_CONFIG to configure data service worker information, as discussed in this [Slack conversation](https://flyte-org.slack.com/archives/CP2HDHKE1/p1694005926592289).

The use case previously doesn't include the evaluator section, so we have to give it a default value so that we can take the case into account.

##  Setup Process

I test it in two ways, by specifying the Dockerfile or using ImageSpec.
### Dockerfile
```dockerfile
FROM python:3.9-slim-buster
USER root
WORKDIR /root
ENV PYTHONPATH /root
RUN apt-get update && apt-get install build-essential -y
RUN apt-get install git -y
# The following line is an example of how to install your modified plugins. In this case, it demonstrates how to install the 'deck' plugin.
# RUN pip install -U git+https://github.com/Yicheng-Lu-llll/flytekit.git@"demo#egg=flytekitplugins-deck-standard&subdirectory=plugins/flytekit-deck-standard" # replace with your own repo and branch
RUN pip install -U git+https://github.com/Future-Outlier/flytekit.git@98ddd542a02551a9a9eb122b98004d0d092abbe9#subdirectory=plugins/flytekit-kf-tensorflow

RUN pip install -U git+https://github.com/Future-Outlier/flyte.git@647b8f4eeeab1a65866d19fab13c416ed0e4a07f#subdirectory=flyteidl

RUN pip install -U git+https://github.com/Future-Outlier/flytekit.git@98ddd542a02551a9a9eb122b98004d0d092abbe9

```
Use the code below
```python
from flytekit import ImageSpec, Resources, task
from flytekit.configuration import Image, ImageConfig, SerializationSettings
from flytekitplugins.kftensorflow import (PS, Chief, CleanPodPolicy, Evaluator,
                                          RestartPolicy, RunPolicy, TfJob,
                                          Worker)

task_config = TfJob(
    worker=Worker(replicas=1),
    chief=Chief(replicas=1),
    ps=PS(replicas=1),
    evaluator=Evaluator(replicas=1),
)


@task(
    task_config=task_config,
    cache=True,
    requests=Resources(cpu="1"),
    cache_version="1",
)
def my_tensorflow_task(x: int, y: str) -> int:
    return x


[...]

PR #2419, opened 2024-05-15
[Core feature] Flytekit should support `unsafe` mode for types

Mecoli1219:
## Tracking issue
Related to flyteorg/flyte#5319

Use Case: When the user wants to apply Flytekit to their Python codebase, but annotating the type is too hard or time-consuming, we should support `unsafe` mode for easy usage.

1. I add `unsafe` kwarg to both `task` and `workflow`, and the default value of `unsafe` is `False`
2. If the `unsafe` kwarg is `True` and the types of inputs or outputs are not specified (which is None), replace the type from None to Optional[Any]. This will trigger `FlytePickleTransformer` to handle each unspecified type.

## How was this patch tested?
Two tests are created to examine the ability of `unsafe` kwarg.
1. `test_unsafe_input_wf_and_task`: This test mainly examines the unspecified input types. If the `unsafe` is False, the workflow or task should throw errors. Otherwise, it should succeed.
2.  `test_unsafe_wf_and_task`: This test mimics the use case mentioned above, where the user might not specify all the types in their codebase. The workflow should succeed if the `unsafe` kwarg of all tasks and workflow are set to `True`.

## Related PRs
https://github.com/flyteorg/flyte/issues/5319


tests/flytekit/unit/core/test_type_hints.py
```diff
@@ -2000,3 +2007,59 @@ def my_wf(a: int, retries: int) -> int:
...
+    assert wf1_with_unsafe(a=1) == 2
+    assert wf1_with_unsafe(a="1") == 0
+    assert wf1_with_unsafe(a=None) == 0
+
+    @workflow
+    def wf1_wo_unsafe(a) -> int:
+        return t1(a=a)
+
+    @workflow
+    def wf1_wo_unsafe2(a: int) -> int:
```
pingsutw, line 2036: Did you test this workflow in the sandbox cluster?

Mecoli1219: No, I only test it on my local environment. Can I ask how to try it?

Future-Outlier: This pr can help you, you can either use imageSpec or Dockerfile to create your own image.
https://github.com/flyteorg/flytekit/pull/1870

Mecoli1219:
```
from flytekit import task, Secret, workflow, ImageSpec

flytekit_dev_version = "https://github.com/Mecoli1219/flytekit.git@c7c9a2ad2bdf6799ec324955a023281cca030038"
image = ImageSpec(
    registry="localhost:30000",
    platform="linux/amd64",
    apt_packages=["git"],
    packages=[
        f"git+{flytekit_dev_version}",
    ],
)


@task(unsafe=True, container_image=image)
def t1(a) -> int:
    if type(a) == int:
        return a + 1
    return 0


@workflow(unsafe=True)
def wf1_with_unsafe(a) -> int:
    return t1(a=a)
```

Mecoli1219: The above code will fail. The reason is detailed in https://github.com/flyteorg/flyte/pull/5261.

Mecoli1219:
@pingsutw I created 2 new unit tests, and both of them passed.

pingsutw (approved):

flytekit/core/task.py
```diff
@@ -130,6 +130,7 @@ def task(
     pod_template: Optional["PodTemplate"] = ...,
     pod_template_name: Optional[str] = ...,
     accelerator: Optional[BaseAccelerator] = ...,
+    unsafe: bool = ...,
```
thomasjpfan, line 133: I prefer to have a more explicit name. Something like:

```suggestion
    pickle_untyped: bool = ...,
```

flytekit/core/python_function_task.py
```diff
@@ -112,6 +112,7 @@ def __init__(
         node_dependency_hints: Optional[
             Iterable[Union["PythonFunctionTask", "_annotated_launch_plan.LaunchPlan", WorkflowBase]]
         ] = None,
+        pickle_untyped: bool = False,
```
thomasjpfan, line 115: @kumare3 Since proposed the `safe` parameter name, are you okay with using `pickle_untyped`?

My concerned is that `safe=False` could mean so many unsafe behavior. `pickle_untyped` is explicit about the behavior.

flytekit/types/pickle/pickle.py
```diff
@@ -86,13 +86,16 @@ def assert_type(self, t: Type[T], v: T):
...
+    def to_python_value(self, ctx: FlyteContext, lv: Literal, expected_python_type: Type[T]) -> Optional[T]:
+        if lv.scalar.blob is None:
+            return None
         uri = lv.scalar.blob.uri
         return FlytePickle.from_pickle(uri)
 
     def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
         if python_val is None:
-            raise AssertionError("Cannot pickle None Value.")
+            # raise AssertionError("Cannot pickle None Value.")
```
thomasjpfan, line 97: Can this be removed?

Mecoli1219: yes sure. I'll remove that recently

Mecoli1219:
There is something wrong when running on the cluster. Don't merge now.

Mecoli1219:
NVM, it is still the same problem as mentioned in the comment above 6 months ago.

```
@task(pickle_untyped=True, container_image=image)
def t1(a) -> int:
    if type(a) == int:
        return a + 1
    return 0

@workflow
def wf1_with_unsafe() -> Tuple[int, int, int]:
    a1 = t1(a=1)
    a2 = t1(a="1")
    a3 = t1(a=None)    #! This will fail
    return a1, a2, a3
```
Currently, the input with a value `None` will fail

Mecoli1219:
```python
@task(pickle_untyped=True, container_image=image)
def t1(a) -> int:
    if type(a) == int:
        return a + 1
    return 0


@workflow
def wf1_with_unsafe() -> Tuple[int, int, int]:
    a1 = t1(a=1)
    a2 = t1(a="1")
    # a3 = t1(a=None)  #! This will fail
    return a1, a2, a2
```

image
image

Mecoli1219:
I revert the support for `None` type in `pickle`. I think we could discuss that later and solve that in another PR if finally decided to do that.

pingsutw (approved):

PR #2420, opened 2024-05-15
Fix return type in `PanderaTransformer::to_python_value()`

danieldanciu:
Return an object of the correct type in the `to_python_value` of the pandera type transformer. Currently, the returned type is `pandas.core.frame.DataFrame`, but the declared return type is `pandera.typing.DataFrame`. 
I suspect this is because instantiating `pandera.DataFrameSchema` returns the former rather than what one would normally expect, which is the latter (the pandera DataFrame).

Closes flyteorg/flyte##5369

Typeguard is unhappy about Flyte workflows that use Pandera, because the type transformer returns an object of the incorrect type.

Simply mark the return type as pandera.typing.DataFrame (which inherits from the currently returned type, so I don't expect any major side-effects).

## How was this patch tested?
Created a test workflow, checked with typeguard and everything is now running.

### Setup process
Not needed.

- [ ] All new and existing tests passed. I haven't run any tests, as there aren't any directly related tests AFAICT. I am hoping the Gitlab CI/CD will run them for me.


pingsutw:
cc @cosmicBboy

cosmicBboy:
Thanks for the contribution @danieldanciu !

I made some changes to the plugin recently, namely renamed the modules and updated some functionality: https://github.com/flyteorg/flytekit/tree/master/plugins/flytekit-pandera/flytekitplugins/pandera

Do you mind applying the change to the latest `master` branch

plugins/flytekit-pandera/flytekitplugins/pandera/schema.py
```diff
@@ -95,7 +95,7 @@ def downloader(x, y):
             downloader=downloader,
             supported_mode=SchemaOpenMode.READ,
         )
-        return self._pandera_schema(expected_python_type)(df.open().all())
+        return pandera.typing.DataFrame(self._pandera_schema(expected_python_type)(df.open().all()))
```
cosmicBboy, line 98: On the topic of the type change, I'm not really sure what the desired behavior would be.

1. On one hand, we want the output of this to be of the specific pandera dataframe type `pandera.typing.DataFrame[<SCHEMA>]`. If this is the desired behavior, this should be `expected_python_type(self._pandera_schema(...))`, since `expected_python_type` is the generic type with the `<SCHEMA>` argument.
2. On the other, I'm not sure what the implications would be if we change the return signature of `to_python_value` to `pandas.DataFrame`... I guess this would violate a user expectation that the data being passed into the task function body should be.

So I'd opt for option (1), assuming that `expected_python_type(self._pandera_schema(...))` works.

PR #2432, opened 2024-05-20
[TypeTransformer] Support non-Any Python types as Any input in workflows

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5366

We want to allow the Any type to accept all kinds of inputs and outputs, not just the types we currently can't handle.

For example, workflow like this.
```python
@task
def foo(a: Any) -> int:
    if type(a) == int:
        return a + 1
    return 0

@workflow
def wf(a: int) -> int:
    return foo(a=a)
```

Note: This PR only supports `Any` in python, if we want to support java, we will need to add more code.

Check if the literal's scalar has a primitive type.
If it does, convert it to a Python value using the corresponding type transformer.

## How was this patch tested?
unit test and remote cluster

```python
from typing import Any

@task
def foo(a: Any) -> int:
    if type(a) == int:
        return a + 1
    return 0

@workflow
def wf(a: int) -> int:
    return foo(a=a)
```
```Dockerfile
FROM python:3.9-slim-buster
USER root
WORKDIR /root
ENV PYTHONPATH /root
RUN apt-get install git -y

RUN pip install -U git+https://github.com/flyteorg/flytekit.git@3804a5155a523e2028ece8e7e581f794ebabd788
```

### Setup process
```python
pyflyte run --remote --image localhost:30000/any:0522 any_task.py wf --a 1
```
### Screenshots
image
image

## Related PRs
https://github.com/flyteorg/flyte/pull/5408

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1178,7 +1178,8 @@ class TypeEngine(typing.Generic[T]):
             lv.hash = hash
 
         metadata = lv.metadata or {}
-        metadata.update({"py_type": python_type})
+        print("type engine python type", python_type.__name__)
+        metadata.update({"py_type": python_type.__name__})
         lv.set_metadata(metadata=metadata)
         return lv
 
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1183,6 +1183,7 @@ class TypeEngine(typing.Generic[T]):
         """
         Converts a Literal value with an expected python type into a python value.
         """
+        print("Expected Python Type: ", expected_python_type)
         transformer = cls.get_transformer(expected_python_type)
         return transformer.to_python_value(ctx, lv, expected_python_type)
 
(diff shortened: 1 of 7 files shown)


Future-Outlier:
Update: I'm not sure if it's ok to support primitive types only.

Future-Outlier:
It can only run in local execution and `pyflyte run --remote`, since I haven't discovered a way to send metadata by literal.
image
```python=
@dataclass
class Datum:
    x: int
    y: str
    z: dict[int, str]
    w: list[int]
    
  @task
def foo_dataclass(a: Datum) -> int:
    if type(a) == Datum:
        return a.x
    return 0

@workflow
def foo_wf(a: Datum) -> int:
    return foo_dataclass(a=a)
```

Future-Outlier:
Note:
`python dotted path` can solve type for Dataclass, while `eval` can solve most of python type,
we should use both of these ways

PR #2439, opened 2024-05-23
Time sensor is erroring out.

Sally-Yang-Jing-Ou:
**Not to be checked in** 

Hey everyone, I have an urgent follow up on this.
So I implemented a flytekit-timesensor plugin, I pip install -e .the whole local flytekit as well as pip installed this new plugin I developed. The logic of the timeDeltaSensor is in this PR.

I first ran `flytectl demo start`, then I `pip install -e .` everything in the flytekit and all the plugins, I built and pushed the docker image, updated the deployment according to this guide: https://docs.flyte.org/en/latest/flyte_agents/deploying_agents_to_the_flyte_sandbox.html , submitted a workflow that uses this plugin, and the flow fails with `rpc error: code = DeadlineExceeded desc = context deadline exceeded .`

This is how i use it in a workflow in Flytesnacks, I added this workflow:
```
from flytekitplugins.timesensor import TimeDeltaSensor
import time


time_delta_sensor = TimeDeltaSensor(name="Time_Delta_mySensor")


@task()
def t1():
    print("SUCCEEDED")


@workflow()
def wf():
    time_delta_sensor(timeDelta=11) >> t1()


if __name__ == "__main__":
    wf()
```

But the workflow terminates with error:
`rpc error: code = DeadlineExceeded desc = context deadline exceeded .`

any clue of what went wrong?


Sally-Yang-Jing-Ou:
see this discussion. https://flyte-org.slack.com/archives/CP2HDHKE1/p1715055406449849

PR #2442, opened 2024-05-23
Remote fetch array node

wild-endeavor:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5359

The fundamental issue leading to the error was in the order of the args in FlyteArrayNode promote_from_model.  That's been fixed and support for array nodes added.

* The `NodeExecutionMetaData` model hadn't been updated in a while (lacking some fields) but it's not widely used so I just removed it entirely opting for the raw proto generated file.
* Rename an import

## How was this patch tested?
* Added a sandbox test that doesn't run for when we have a nice integration testing framework.

### DIFF
--- a/flytekit/models/node_execution.py
+++ b/flytekit/models/node_execution.py
@@ -227,41 +227,6 @@ class NodeExecutionClosure(_common_models.FlyteIdlEntity):
         )
 
 
-# class NodeExecutionMetaData(_common_models.FlyteIdlEntity):
-#     def __init__(self, retry_group: str, is_parent_node: bool, spec_node_id: str,
-#                  ):
-#         self._retry_group = retry_group
-#         self._is_parent_node = is_parent_node
-#         self._spec_node_id = spec_node_id
-#
-#     @property
-#     def retry_group(self) -> str:
-#         return self._retry_group
-#
-#     @property
-#     def is_parent_node(self) -> bool:
-#         return self._is_parent_node
-#
-#     @property
-#     def spec_node_id(self) -> str:
-#         return self._spec_node_id
-#
-#     def to_flyte_idl(self) -> admin_node_execution_pb2.NodeExecutionMetaData:
-#         return admin_node_execution_pb2.NodeExecutionMetaData(
-#             retry_group=self.retry_group,
-#             is_parent_node=self.is_parent_node,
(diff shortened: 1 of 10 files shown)


flytekit/models/node_execution.py
```diff
@@ -296,22 +262,23 @@ def closure(self):
...
 
     @classmethod
-    def from_flyte_idl(cls, p: _node_execution_pb2.NodeExecution) -> "NodeExecution":
+    def from_flyte_idl(cls, p: admin_node_execution_pb2.NodeExecution) -> "NodeExecution":
         return cls(
             id=_identifier.NodeExecutionIdentifier.from_flyte_idl(p.id),
             input_uri=p.input_uri,
             closure=NodeExecutionClosure.from_flyte_idl(p.closure),
-            metadata=NodeExecutionMetaData.from_flyte_idl(p.metadata),
+            # metadata=NodeExecutionMetaData.from_flyte_idl(p.metadata),
```
cosmicBboy, line 282: remove?

flytekit/remote/entities.py
```diff
@@ -472,6 +477,7 @@ def promote_from_model(
             flyte_gate_node = FlyteGateNode.promote_from_model(model.gate_node)
         elif model.array_node is not None:
             flyte_array_node = FlyteArrayNode.promote_from_model(model.array_node)
+            print(flyte_array_node)
```
cosmicBboy, line 480: remove print

flytekit/remote/remote.py
```diff
@@ -2101,7 +2101,7 @@ def sync_node_execution(
                 sub_node_mapping = {n.id: n for n in sub_flyte_workflow.flyte_nodes}
                 execution._underlying_node_executions = [
                     self.sync_node_execution(FlyteNodeExecution.promote_from_model(cne), sub_node_mapping)
-                    for cne in child_node_executions
+                    for cne in child_node_executiFsync_execons
```
cosmicBboy, line 2104: where is `child_node_executiFsync_execons` coming from? also kind of a strange-looking variable name...

wild-endeavor: uh, not sure how that happened, thanks

eapolinario (approved):

PR #2446, opened 2024-05-28
[MINOR] Support Passing Dataclass Values via Command Line

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/4486

Currently, we don't support `passing dataclass values via command line` after this [PR](https://github.com/flyteorg/flytekit/pull/2279).

For example, using this command line will raise an error.
`pyflyte run --remote --image localhost:30000/any:0528-1445 dataclass_task.py foo_wf  --a '{ "x": 1, "y": "2", "z": { "1": "one", "2": "two" }, "w": [1, 2, 3] }'`

image

If we use only `dataclass` decorator only for our dataclass value, convert it to `dataclass_json`.

## How was this patch tested?
local execution, remote execution and unit test.

### Setup process
```python

@dataclass
class Datum:
    x: int
    y: str
    z: dict[int, str]
    w: list[int]

@task
def foo_dataclass(a: Datum) -> int:
    if type(a) == Datum:
        return a.x
    return 0

@workflow
def foo_wf(a: Datum) -> int:
    return foo_dataclass(a=a)
```

### Screenshots
local execution
image
remote execution
image
image
unit test
image

### DIFF
--- a/flytekit/interaction/click_types.py
+++ b/flytekit/interaction/click_types.py
@@ -10,7 +10,7 @@ from typing import cast
 import cloudpickle
 import rich_click as click
 import yaml
-from dataclasses_json import DataClassJsonMixin
+from dataclasses_json import DataClassJsonMixin, dataclass_json
 from pytimeparse import parse
 
 from flytekit import BlobType, FlyteContext, FlyteContextManager, Literal, LiteralType, StructuredDataset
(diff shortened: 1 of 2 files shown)


pingsutw (approved):

wild-endeavor:
renaming... no need for `[MINOR]` in the pr desc please.  That typically means that the change is non-backwards compatible and will need a minor release instead of a patch release.  (since we're kinda using minor as major at this point)

Future-Outlier:
> renaming... no need for `[MINOR]` in the pr desc please.  That typically means that the change is non-backwards compatible and will need a minor release instead of a patch release.  (since we're kinda using minor as major at this point)

No problem, thank you

PR #2447, opened 2024-05-28
Add . to the PYTHONPATH in ImageSpec

pingsutw:
## Tracking issue
https://flyte-org.slack.com/archives/CP2HDHKE1/p1716911881652929

Module not found while running on a remote cluster since Python can not find the local module unless you explicitly add the local directory to the `PYTHONPATH`

Note: when you are using non-default `base_image`, the `WORKDIR` might be others (`/` or `/home`, etc). 

Add . to PYTHONPATH

## How was this patch tested?
```python

new_flytekit = "git+https://github.com/flyteorg/flytekit@554dd6dd4882d703839dc4b846285b7b2f9f66dc"

image_spec = ImageSpec(
    name="helloworld",
    base_image="python:3.11.5-slim-bullseye",
    packages=[new_flytekit, "mypy"],
    apt_packages=["git"],
    registry="localhost:30000",
)


def say_hello() -> str:
    return "Hello, World!"


@workflow
def wf() -> str:
    res = say_hello()
    return res


if __name__ == "__main__":
    print(f"Running hello_world_wf() {wf()}")
```

### Screenshots
Screenshot 2024-05-28 at 12 40 22 PM

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/plugins/flytekit-envd/flytekitplugins/envd/image_builder.py
+++ b/plugins/flytekit-envd/flytekitplugins/envd/image_builder.py
@@ -104,7 +104,7 @@ def build():
     run(commands=[{run_commands}])
     install.python_packages(name=[{python_packages}])
     install.apt_packages(name=[{apt_packages}])
-    runtime.environ(env={env}, extra_path=['/root'])
+    runtime.environ(env={env}, extra_path=['/root', "/"])
 """
     if image_spec.pip_extra_index_url is None:
         envd_config += f'    config.pip_index(url="{pip_index}")\n'
--- a/plugins/flytekit-envd/flytekitplugins/envd/image_builder.py
+++ b/plugins/flytekit-envd/flytekitplugins/envd/image_builder.py
@@ -91,7 +91,7 @@ def create_envd_config(image_spec: ImageSpec) -> str:
     run_commands = _create_str_from_package_list(image_spec.commands)
     conda_channels = _create_str_from_package_list(image_spec.conda_channels)
     apt_packages = _create_str_from_package_list(image_spec.apt_packages)
-    env = {"PYTHONPATH": "/root:.", _F_IMG_ID: image_spec.image_name()}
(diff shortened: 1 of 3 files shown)


plugins/flytekit-envd/flytekitplugins/envd/image_builder.py
```diff
@@ -91,7 +91,7 @@ def create_envd_config(image_spec: ImageSpec) -> str:
     run_commands = _create_str_from_package_list(image_spec.commands)
     conda_channels = _create_str_from_package_list(image_spec.conda_channels)
     apt_packages = _create_str_from_package_list(image_spec.apt_packages)
-    env = {"PYTHONPATH": "/root", _F_IMG_ID: image_spec.image_name()}
+    env = {"PYTHONPATH": "/root:.", _F_IMG_ID: image_spec.image_name()}
```
thomasjpfan, line 94: The "." is easy to miss. I think using `PWD` works too?

```suggestion
    env = {"PYTHONPATH": "/root:$PWD", _F_IMG_ID: image_spec.image_name()}
```

thomasjpfan (commented):
For custom base images, should `ImageSpec` change the `WORKDIR` to `/root`?

plugins/flytekit-envd/flytekitplugins/envd/image_builder.py
wild-endeavor: you don't need `.` you just need the `:`

eapolinario: how does this work? I was under the impression that the current working directory was already part of `sys.path` when the python interpreter starts.

pingsutw: > For custom base images, should ImageSpec change the WORKDIR to /root?

we don't need to. It should use the workdir in the base image.

eapolinario:
Can we add a description of the problem to this PR? I'd love to understand why this fixes the problem.

PR #2458, opened 2024-06-04
[wip] Add Image builder `should_build` method for more customizable build behavior

cosmicBboy:

Prior to this PR, the build behavior is determined by the `ImageBuildEngine`. This makes the messages confusing on e.g. with the serverless `UCImageBuilder`, which has different logic that determines whether an image should be built or not. The CLI messages say the images doesn't exist even when it does.

This change introduces a `should_build` method to the `ImageSpecBuilder` base class. This method should take an image spec and output True if the image should be built or False if the image already exists.

## How was this patch tested?

This PR was manually tested on a Union tenant, unit tests are pending

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -260,12 +260,12 @@ class ImageBuildEngine:
             image_spec.base_image = image_spec.base_image.image_name()
 
         if image_spec.builder is None and cls._REGISTRY:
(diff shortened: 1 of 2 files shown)


cosmicBboy (commented):
This looks good to me! @katrina mind a taking a look? context: https://unionai.slack.com/archives/C05EHTFEFB9/p1717455305121219

katrogan (approved):

pingsutw (approved):

PR #2460, opened 2024-06-04
add escape hatch to binding union handling

wild-endeavor:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/5321.  This was discovered while implementing [that PR](https://github.com/flyteorg/flytekit/pull/2401).

The main issue is highlighted in the test added to the test type engine file.  When you call `TypeEngine.to_literal` on a Union type, you should get back a Flyte literal with a `Union` [message](https://github.com/flyteorg/flyte/blob/9ffc54fa4d69f3058ce635a53ce2fa594a9cc5d9/flyteidl/protos/flyteidl/core/literals.proto#L56).  This was happening correctly in the Type Engine but for some reason this wasn't happening for Binding Data.

That is, if you had

```python
@task
def say_hello(a: int | str) -> str:
    print(f"Type is {type(a)}")
    if type(a) == str:
        return f"Hello, {a}!"
    return "Hello, World!"

@workflow
def hello_world_wf()-> str:
    res = say_hello(a="world")
    return res
```

the node for the `say_hello` invocation was being bound to a BindingData Literal Scalar that was just the raw string "world", not a Union that contained that string literal.

This adds an escape mechanism and some comments in the binding data from python std function.

## How was this patch tested?
Before: https://www.test.cluster/console/projects/flytesnacks/domains/development/executions/f6ba77e8684e94e1092a
After: https://www.test.cluster/console/projects/flytesnacks/domains/development/executions/f06eaea2298ad4884952
Both runs succeed.

Before: https://www.test.cluster/api/v1/workflows/flytesnacks/development/unions.hello_world_wf/4jItAJN7eckVByTXKxjsPg
After: https://www.test.cluster/api/v1/workflows/flytesnacks/development/unions.hello_world_wf/gmbOol0QMeiiHF_y99A97g
Note the after definition binds the node to a union literal, not a plain scalar.

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -697,6 +697,13 @@ def binding_data_from_python_std(
         )
 
     elif t_value is not None and expected_literal_type.union_type is not None:
+        # If the value is not a container type, then we can directly convert it to a scalar in the Union case.
+        # This pushes the handling of the Union types to the type engine.
+        if not isinstance(t_value, list) and not isinstance(t_value, dict):
+            scalar = TypeEngine.to_literal(ctx, t_value, t_value_type or type(t_value), expected_literal_type).scalar
+            return _literals_models.BindingData(scalar=scalar)
+
+        # If it is a container type, then we need to iterate over the variants in the Union type.
         for i in range(len(expected_literal_type.union_type.variants)):
             try:
                 lt_type = expected_literal_type.union_type.variants[i]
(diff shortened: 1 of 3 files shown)


wild-endeavor:
cc @MortalHappiness can you take a look at this?  i think with this we can revert some of the changes in your pr...

eapolinario (approved):

MortalHappiness:
LGTM

PR #2461, opened 2024-06-04
Fix CI by restricting hugging face dataset version

Future-Outlier:
I just saw that the datasets module had its latest release, so I guess it is the root cause of the CI failure, and it works now.

https://pypi.org/project/datasets/#history

### DIFF
--- a/plugins/flytekit-huggingface/setup.py
+++ b/plugins/flytekit-huggingface/setup.py
@@ -6,7 +6,7 @@ microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
 
 plugin_requires = [
     "flytekit>=1.3.0b2,<2.0.0",
-    "datasets>=2.4.0",
+    "datasets>=2.4.0,<2.19.2",
 ]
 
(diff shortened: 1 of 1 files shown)


plugins/flytekit-huggingface/setup.py
```diff
@@ -6,7 +6,7 @@
 
 plugin_requires = [
     "flytekit>=1.3.0b2,<2.0.0",
-    "datasets>=2.4.0",
+    "datasets>=2.4.0,<2.19.2",
```
pingsutw, line 9: nit: I hope they will fix the bug in the next release
```suggestion
    "datasets>=2.4.0,!2.19.2",

```

Future-Outlier: > nit: I hope they will fix the bug in the next release

No problem, thank you

eapolinario (approved):
Thank you!

I couldn't find an issue on their [issue tracker](https://github.com/huggingface/datasets/issues).

PR #2468, opened 2024-06-07
Enable string conversion for `Scalar` objects that contain a `schema`

redartera:

If attempting to create a string representation of a literal whose scalar contains a `schema` (e.g. pandera DataFrames), an "Unknown scalar type" `ValueError` is raised.

It is often nice to just print out the outputs of a workflow execution, but that's currently not possible if a pandera-typed DataFrame is one of them. 

We simply return the schema's URI, similarly to how `structured_dataset` is handled.

## How was this patch tested?

I expanded on the existing unit test

### DIFF
--- a/flytekit/interaction/string_literals.py
+++ b/flytekit/interaction/string_literals.py
@@ -37,6 +37,8 @@ def scalar_to_string(scalar: Scalar) -> typing.Any:
         return scalar.error.message
     if scalar.structured_dataset:
         return scalar.structured_dataset.uri
+    if scalar.schema:
+        return scalar.schema.uri
     if scalar.blob:
         return scalar.blob.uri
     if scalar.binary:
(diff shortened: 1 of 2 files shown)


pingsutw (approved):
@redartera Thanks. Could you use the structured dataset? We have deprecated it for a while.

redartera:
> @redartera Thanks. Could you use the structured dataset? We have deprecated it for a while.

@pingsutw I don't use `FlyteSchema` but rather just `pd.DataFrame` with `pandera`.

Looking at the `flyteplugins-pandera` code [here](https://github.com/flyteorg/flytekit/blob/63f190e691e150dd83b031f8561d698bdace21c2/plugins/flytekit-pandera/flytekitplugins/pandera/schema.py#L77) - it creates a `Schema` FlyteIDL object which is why I needed to make this change

PR #2475, opened 2024-06-12
add nim plugin

samhita-alla:
## Tracking issue

Fixes https://github.com/flyteorg/flyte/issues/5478

This PR adds NIM to flytekit plugins to enable serving optimized model containers, which can include NVIDIA CUDA software, NVIDIA Triton Inference Server, and NVIDIA TensorRT-LLM software. Since each Flyte task runs in a Kubernetes pod, a NIM container can be easily deployed as a sidecar service, allowing users to invoke the model's endpoint as if it is hosted locally, thus minimizing network overhead. Additionally, Flyte can manage data pre-processing and post-processing, facilitating the creation of end-to-end batch inference pipelines with optimized models. 

- A new `ModelInferenceTemplate` has been added, providing a base for creating other inference solutions like Ollama and Hugging Face.
- The `NIM` class, which subclasses `ModelInferenceTemplate`, has been introduced as well.

```python
from flytekit import ImageSpec, Secret, task, Resources
from flytekit.extras.accelerators import A10G
from flytekitplugins.inference import NIM, NIMSecrets
from openai import OpenAI

image = ImageSpec(
    name="nim",
    registry="...",
    packages=["flytekitplugins-inference"],
)

nim_instance = NIM(
    image="nvcr.io/nim/meta/llama3-8b-instruct:1.0.0",
    secrets=NIMSecrets(
        ngc_image_secret="nvcrio-cred",
        ngc_secret_key=NGC_KEY,
        secrets_prefix="_FSEC_",
    ),
)


@task(
    container_image=image,
    pod_template=nim_instance.pod_template,
    accelerator=A10G,
    secret_requests=[
        Secret(
            key="ngc_api_key", mount_requirement=Secret.MountType.ENV_VAR
        )  # must be mounted as an env var
    ],
    requests=Resources(gpu="0"),
)
def model_serving() -> str:
    client = OpenAI(
        base_url=f"{nim_instance.base_url}/v1", api_key="nim"
    )  # api key required but ignored

    completion = client.chat.completions.create(
        model="meta/llama3-8b-instruct",
        messages=[
            {
                "role": "user",
                "content": "Write a limerick about the wonders of GPU computing.",
            }
        ],
        temperature=0.5,
        top_p=1,
        max_tokens=1024,
    )

    return completion.choices[0].message.content
```


kumare3:
How about we move this pattern to core flytekit. I don't see any extra requirements?

flytekit/core/inference.py
```diff
@@ -0,0 +1,182 @@
...
+            image=image,
+            health_endpoint=health_endpoint,
+            port=port,
+            cpu=cpu,
+            gpu=gpu,
+            mem=mem,
+            env=env,
+        )
+
+        self.nim_pod_template()
```
eapolinario, line 81: can we rename this to `setup_nim_pod_template`?

flytekit/core/inference.py
```diff
@@ -0,0 +1,182 @@
...
+        image: str = "nvcr.io/nim/meta/llama3-8b-instruct:1.0.0",
+        health_endpoint: str = "v1/health/ready",
+        port: int = 8000,
+        cpu: int = 1,
+        gpu: int = 1,
+        mem: str = "20Gi",
+        shm_size: str = "16Gi",
+        # kubernetes secrets
+        ngc_image_secret: Optional[str] = None,
+        ngc_secret_key: Optional[str] = None,
```
eapolinario, line 35: Do these have to be k8s secrets? Why can't we use Flyte secrets?

flytekit/core/inference.py
```diff
@@ -0,0 +1,182 @@
...
+        ####################
+        env: Optional[dict[str, str]] = None,
+        hf_repo_ids: Optional[list[str]] = None,
+        hf_token_key: Optional[str] = None,
```
eapolinario, line 39: Is this supposed to be a secret also?

pingsutw (commented):
It looks awesome! I think having `ModelInferenceTemplate` is really great, so users don't need to deal with the K8s pod spec

flytekit/core/inference.py
```diff
@@ -0,0 +1,182 @@
...
+        ]
+        self.pod_template.pod_spec.image_pull_secrets = [V1LocalObjectReference(name=self._ngc_image_secret)]
+
+        model_server_container = self.pod_template.pod_spec.init_containers[0]
+
+        if model_server_container.env:
+            model_server_container.env.append(
+                V1EnvVar(
+                    name="NGC_API_KEY",
+                    value=f"$(_UNION_{self._ngc_secret_key.upper()})",
```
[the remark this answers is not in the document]
eapolinario: we just sync'd offline and the idea is to use the [FLYTE_SECRETS_ENV_PREFIX](https://github.com/flyteorg/flyte/blob/master/flytepropeller/pkg/webhook/secrets.go#L19) to figure out which environment variable to use.

flytekit/core/inference.py
samhita-alla: `ngc_secret_key` must be the name of the Flyte secret, whereas `ngc_image_secret` must be a Kubernetes secret.

flytekit/core/inference.py
samhita-alla: This is a Flyte secret. Will update the secrets interface to accept group as well.

eapolinario (commented):
This is looking good. I have only minor comments.

flytekit/core/inference.py
```diff
@@ -0,0 +1,196 @@
+"""
+=========
+Inference
+=========
+
+.. currentmodule:: flytekit.core.inference
```
thomasjpfan, line 6: Having a `flytekit.core.inference` feels very ML/AI leaning, should this be a plugin?

flytekit/core/inference.py
```diff
@@ -0,0 +1,196 @@
...
+            V1Volume(
+                name="dshm",
+                empty_dir=V1EmptyDirVolumeSource(medium="Memory", size_limit=self._shm_size),
+            )
+        ]
+        self.pod_template.pod_spec.image_pull_secrets = [V1LocalObjectReference(name=self._secrets.ngc_image_secret)]
+
+        model_server_container = self.pod_template.pod_spec.init_containers[0]
+
+        secret_prefix = get_plugin().secret_prefix()
```
thomasjpfan, line 117: The configuration plugin system is designed to be global, but it is currently used for this plugin.

Can we hard code the prefix in `NIMSecrets` and use it from there?

flytekit/core/inference.py
samhita-alla: hmm... what about `serve`? i don't think this should be a plugin as it's just a pod template wrapper.

flytekit/core/inference.py
samhita-alla: cc @eapolinario

flytekit/core/utils.py
```diff
@@ -387,3 +387,81 @@ def get_extra_config(self):
...
+                            "cpu": self._cpu,
+                            "nvidia.com/gpu": self._gpu,
+                            "memory": self._mem,
+                        },
+                    ),
+                    restart_policy="Always",  # treat this container as a sidecar
+                    env=([V1EnvVar(name=k, value=v) for k, v in self._env.items()] if self._env else None),
+                    startup_probe=V1Probe(
+                        http_get=V1HTTPGetAction(path=self._health_endpoint, port=self._port),
+                        failure_threshold=100,
```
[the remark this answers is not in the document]
samhita-alla: added a comment: `The model server initialization can take some time, so the failure threshold is increased to accommodate this delay.`

flytekit/core/inference.py
```diff
@@ -0,0 +1,196 @@
...
+    :param secrets_prefix: The secrets prefix that Flyte appends to all mounted secrets. Default value is _UNION_.
+    """
+
+    ngc_image_secret: str  # kubernetes secret
+    ngc_secret_key: str
+    ngc_secret_group: Optional[str] = None
+    hf_token_group: Optional[str] = None
+    hf_token_key: Optional[str] = None
+
+    secrets_prefix: str = "_UNION_"
```
thomasjpfan, line 41: Can `secrets_prefix` be a required parameter?

flytekit/core/inference.py
thomasjpfan: It's a pod template wrapper that uses the huggingface API to download a model, which is very AI focused. Although I think the feature is interesting, I do not know if it fits flytekit core.

@eapolinario What do you think?

eapolinario: > It's a pod template wrapper that uses the huggingface API to download a model, which is very AI focused. Although I think the feature is interesting, I do not know if it fits flytekit core.
> 
> @eapolinario What do you think?

Agreed. What was the reasoning to put this in core, @samhita-alla ? 



> what about `serve`? i don't think this should be a plugin as it's just a pod template wrapper.

Even if it's just a pod template wrapper, it's a *very* useful wrapper.

What was the reasoning behind putting this in core, @samhita-alla ?

samhita-alla: @eapolinario given that this implementation requires no additional libraries and functions as a wrapper rather than a plugin, wouldn't it be more appropriate to include it in flytekit core? also, `flytekit.core` was Ketan's suggestion.

eapolinario: I disagree with adding this code to Flytekit core simply because it doesn't introduce extra dependencies. We already have an extension mechanism through Flytekit plugins. As @thomasjpfan pointed out, flytekit doesn't include ML-specific code yet and bringing NIM to core creates a precedent and puts more burden on the maintainers, for example, what do we do when we receive external contributions?

So maybe, let's take a step back and list what are the drawbacks of creating a Nim plugin instead?

@kumare3, can you chime in here?

samhita-alla: i'll move nim to flytekit plugins

plugins/flytekit-inference/setup.py
```diff
@@ -0,0 +1,38 @@
+from setuptools import setup
+
+PLUGIN_NAME = "inference"
+
+microlib_name = f"flytekitplugins-{PLUGIN_NAME}"
+
+plugin_requires = ["flytekit>=1.13.0,<2.0.0", "kubernetes", "openai"]
```
eapolinario, line 7: the plugin doesn't really require `openai`, right?

eapolinario (commented):
one thing to consider is what sort of inference plugins we're going to add to `flytekitplugins-inference`. Right now there's no extra dependencies (since NIM is essentially a pod template wrapper), but that might not be the case in all plugins. That said, we'll cross that bridge when we get there.

plugins/flytekit-inference/setup.py
samhita-alla: > what sort of inference plugins we're going to add to flytekitplugins-inference
I've got ollama working with the sidecar pattern, so will be adding that next. in `flytekitplugins-inference`, i believe we need to stick to serving plugins that can be implemented with the sidecar pattern. if not, they should be standalone plugins. 

this parent plugin should essentially capture inference providers that allow self-hosting and serving models within a flyte task. the ideal goal is to make it work with actors to enable near real-time inference.

 > the plugin doesn't really require openai, right?
you're right. however, since most inference providers now support the openai api, it'd be beneficial to include it as part of the plugin.

samhita-alla: @eapolinario

eapolinario (approved):
Thank you. This looks great.

PR #2476, opened 2024-06-12
fix: Prevent local files from being moved when using FlyteFile on local executions

ggydush:
## Tracking issue
N/A

When running local executions, local files that exist should not need to be copied. When running large files, this can become a bottleneck.

Check if running a local task execution when converting FlyteFile to literal.

## How was this patch tested?

```python
from flytekit import task


@task
def example(ff: FlyteFile) -> str:
    downloaded_path = ff.download()
    print(downloaded_path)
    return str(ff.path)


example(ff=__file__)
```

Before the fix, this printed:
```console
/var/folders/d1/n1tj86t11pj_drgfg7_br5n00000gn/T/flyte-qd4bk3t0/raw/0fee47924014b3cb5f12295a48330c0f/tmp.py
```

After the fix, this printed:
```console
/Users/ggydush/dev/delve/flytekit/flytekit/types/file/tmp.py
```

### DIFF
--- a/flytekit/types/file/file.py
+++ b/flytekit/types/file/file.py
@@ -435,7 +435,7 @@ class FlyteFilePathTransformer(TypeTransformer[FlyteFile]):
                     f" specified. Since a simpler type was specified, we'll skip uploading!"
                 )
                 should_upload = False
-            if ctx.execution_state.is_local_execution:
+            if ctx.execution_state.is_local_execution and python_val.remote_path is None:
                 should_upload = False
 
             # Set the remote destination if one was given instead of triggering a random one below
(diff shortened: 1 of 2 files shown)


eapolinario (commented):
Thanks for the PR, @ggydush . This is a great idea IMO. Can you take a look at the test failures?

flytekit/types/file/file.py
```diff
@@ -435,6 +435,8 @@ def to_literal(
                     f" specified. Since a simpler type was specified, we'll skip uploading!"
                 )
                 should_upload = False
+            if ctx.execution_state.is_local_execution and python_val.remote_path is None:
+                should_upload = False
```
pingsutw, line 439: Should we set should_upload to `False` by default instead? That way, we won't need `should_upload=False` everywhere.

eapolinario (approved):

PR #2477, opened 2024-06-12
Fix test_image_spec

pingsutw:
## Tracking issue
https://github.com/flyteorg/flytekit/actions/runs/9475012693/job/26105636011

Unit test is failing

if image registry is `localhost:30000`, we should not use `request` to check if the image exists on docker hub

## How was this patch tested?
unit test

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -127,7 +127,10 @@ class ImageSpec:
         except Exception as e:
             tag = calculate_hash_from_image_spec(self)
             # if docker engine is not running locally, use requests to check if the image exists.
-            if DOCKER_HUB in self.registry:
+            container_registry = DOCKER_HUB
+            if self.registry and "/" in self.registry:
+                container_registry = self.registry.split("/")[0]
+            if container_registry == DOCKER_HUB and "localhost" not in self.registry:
                 url = f"https://hub.docker.com/v2/repositories/{self.registry}/{self.name}/tags/{tag}"
                 response = requests.get(url)
                 if response.status_code == 200:
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -120,17 +120,14 @@ class ImageSpec:
             return True
         except APIError as e:
             if e.response.status_code == 404:
-                return True
+                return False
             return True
         except ImageNotFound:
-            return True
+            return False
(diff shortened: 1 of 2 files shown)


flytekit/image_spec/image_spec.py
```diff
@@ -104,7 +104,6 @@ def is_container(self) -> bool:
             return os.environ.get(_F_IMG_ID) == self.image_name()
         return True
 
-    @lru_cache
```
pingsutw, line 107: Remove lru cache since I don't think it makes sense here. 

1. image.exist() => False
2. build image
3. image.exist() => it should return true instead of reading from the cache

flytekit/image_spec/image_spec.py
```diff
@@ -127,11 +126,11 @@ def exist(self) -> bool:
             return False
         except Exception as e:
             tag = calculate_hash_from_image_spec(self)
-            # if docker engine is not running locally
+            # if docker engine is not running locally, use requests to check if the image exists.
             container_registry = DOCKER_HUB
             if self.registry and "/" in self.registry:
                 container_registry = self.registry.split("/")[0]
-            if container_registry == DOCKER_HUB:
+            if container_registry == DOCKER_HUB and "localhost:" not in self.registry:
```
wild-endeavor, line 133: don't understand this... if self.registry is localhost something, why is container_registry still DOCKER_HUB?  isn't it more like the if condition above needs to change?

flytekit/image_spec/image_spec.py
wild-endeavor: do we look things up once per task?  or just once per image spec?  if it's looked up once per task for some reason it's okay to have cache (though i feel we should change that behaviro)

flytekit/image_spec/image_spec.py
pingsutw: updated it

flytekit/image_spec/image_spec.py
pingsutw: Once per image spec. Also, we already have a cache at [build](https://github.com/flyteorg/flytekit/blob/1953b73708a63725ab8f9ff5ecc2c455244758a6/flytekit/image_spec/image_spec.py#L260), so we don't need it here.

wild-endeavor (approved):

PR #2480, opened 2024-06-13
Support ArrayNode mapping over Launch Plans

pvditt:
## Tracking issue
fixes: https://linear.app/unionai/issue/COR-1385/add-flytekit-support-for-mapping-over-launch-plans

ArrayNode currently only supports mapping over a subset of flyte entities (PythonFunctionTasks). To map over other flyte entities such as LaunchPlans we need a new Flytekit node type as the ArrayNodeMapTask implementation can only support Python tasks.

Create new ArrayNode class to map over launch plans. Will have a follow up PR to have this new ArrayNode class support Python tasks (as per discussion with @eapolinario and @wild-endeavor).

## How was this patch tested?

Ran in sandbox with private fork branch: https://github.com/unionai/flyte/pull/255

@hamersaw scale tested as well

### Setup process
Need to create custom image to run. I updated Dockerfile.dev to pulldown the OSS idl changes:

```
# Use a future version of SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FLYTEIDL such that uv resolution works.
RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FLYTEKIT=$PSEUDO_VERSION \
    SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FLYTEIDL=3.0.0dev0 \
        uv pip install --system --no-cache-dir -U \
            "git+https://github.com/flyteorg/flyte.git@exec-version-array-node#subdirectory=flyteidl" \
            -e /flytekit \
            -e /flytekit/plugins/flytekit-k8s-pod \
            -e /flytekit/plugins/flytekit-deck-standard \
            -e /flytekit/plugins/flytekit-flyteinteractive \
            scikit-learn \
    && apt-get clean autoclean \
    && apt-get autoremove --yes \
    && rm -rf /var/lib/{apt,dpkg,cache,log}/ \
    && useradd -u 1000 flytekit \
    && chown flytekit: /root \
    && chown flytekit: /home \
    && :
```

```
@task
def multiply(val: int, val1: int) -> int:
    # if val == 1:
    #     raise Exception("This is a test exception")
    return val * val1


@workflow
def parent_wf(num: int, num1: int) -> int:
    return multiply(val=num, val1=num1)


ex_lp1 = LaunchPlan.get_default_launch_plan(current_context(), parent_wf)


@workflow
def mapped_lp1() -> list[int]:
    a = map_task(ex_lp1, min_success_ratio=0.6)(num=[1, 3, 5], num1=[2, 4, 6])
    return a
```

## Related PRs

https://github.com/unionai/flyte/pull/255


flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+            self.target.python_interface, self._bound_inputs, output_as_list_of_optionals
+        )
+        self._collection_interface = collection_interface
+
+        self.metadata = None
+        if isinstance(target, LaunchPlan):
+            if self._execution_version is None:
+                self._execution_version = 1
+            if self._execution_version != 1:
+                raise ValueError("Only execution version 1 is supported for LaunchPlans.")
```
eapolinario, line 68: can you clarify what's this used for?

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+        # Part of SupportsNodeCreation interface
+        return self._collection_interface
+
+    @property
+    def bindings(self) -> List[_literal_models.Binding]:
+        return []
+
+    @property
+    def upstream_nodes(self) -> List[Node]:
+        return []
```
eapolinario, line 98: do we have to implement this?

eapolinario (commented):
There's a fair amount of code duplication between array_node.py and array_node_map_task.py. Have we thought about how to unify some of that code?

Also, I see mentions to `execution_version` but no correspondent field in the `ArrayNode` proto, maybe I'm missing a flyteidl change?

I'm slightly confused about the use of `mapped_entity`. I'll do another pass on the PR later today.

flytekit/models/core/workflow.py
```diff
@@ -381,7 +381,9 @@ def from_flyte_idl(cls, pb2_object: _core_workflow.GateNode) -> "GateNode":
 
 
 class ArrayNode(_common.FlyteIdlEntity):
-    def __init__(self, node: "Node", parallelism=None, min_successes=None, min_success_ratio=None) -> None:
+    def __init__(
+        self, node: "Node", parallelism=None, min_successes=None, min_success_ratio=None, execution_version=None
```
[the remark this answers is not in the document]
pvditt: this will be utilized for determining how propeller will handle the execution (mainly eventing and storing state) of an ArrayNode. Have the IDL change made on the branch in our private fork. Will merge the IDL change into OSS

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+class ArrayNode(object):
+    def __init__(
+        self,
+        target: LaunchPlan,
+        concurrency: Optional[int] = None,
+        min_successes: Optional[int] = None,
+        min_success_ratio: Optional[float] = None,
+        bound_inputs: Optional[Set[str]] = None,
+        execution_version: Optional[int] = None,
+        metadata: Optional[Union[_workflow_model.NodeMetadata, TaskMetadata]] = None,
```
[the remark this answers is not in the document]
pvditt: I forgot to clean this up when removing code. When I initially implemented this, this new ArrayNode class supported PythonTasks as well. Since we still plan to add this support as a follow up, would it be fine to leave this as is?

flytekit/core/array_node.py
pvditt: These are needed to be set due to the [get_serializable_node function](https://github.com/flyteorg/flytekit/blob/d29e30eedc48f0f3b8d72dd5ea42f9dbdb1f2f13/flytekit/tools/translator.py#L434).

pvditt:
@eapolinario 

> There's a fair amount of code duplication between array_node.py and array_node_map_task.py. Have we thought about how to unify some of that code?

I believe the goal is to have this eventually replace array node map task. We can unify some code, but unsure if it's worthwhile. I don't feel too strongly either way - can move some logic over to util functions if that's preferred.

> Also, I see mentions to execution_version but no correspondent field in the ArrayNode proto, maybe I'm missing a flyteidl change?

have this set in the private fork PR. Will merge in OSS - https://github.com/flyteorg/flyte/pull/5512

> I'm slightly confused about the use of mapped_entity. I'll do another pass on the PR later today.

Poor naming convention on my end. Just updated to make it clearer.

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+    translate_inputs_to_literals,
+)
+from flytekit.core.task import TaskMetadata
+from flytekit.loggers import logger
+from flytekit.models import literals as _literal_models
+from flytekit.models.core import workflow as _workflow_model
+from flytekit.models.literals import Literal, LiteralCollection, Scalar
+
+
+class ArrayNode(object):
```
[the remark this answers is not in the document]
pvditt: thanks for calling this out. Cleaned it up.

flytekit/core/array_node.py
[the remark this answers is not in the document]
eapolinario: fair. Leave it as is.

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+    @property
+    def execution_version(self) -> Optional[int]:
+        return self._execution_version
+
+
+def array_node(
+    target: Union[LaunchPlan],
+    concurrency: Optional[int] = None,
+    min_success_ratio: Optional[float] = None,
+    min_successes: Optional[int] = None,
```
eapolinario, line 189: since you're adding support for `min_successes`, what's the error if someone specifies `min_successes` and `min_success_ratio` at the same time? The proto has them [defined as a `oneof`](https://github.com/flyteorg/flyte/blob/1e54d21c4d4ee74245f799a57b4bb8a5534e8368/flyteidl/protos/flyteidl/core/workflow.proto#L126-L136).

Can you also add a few unit-tests around valid calls to `array_node` that pass `min_successes`?

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+                flyte_interface_types={} if typed_interface is None else typed_interface.inputs,
+                native_types=self.target.python_interface.inputs,
+            )
+            kwargs_literals = {k1: Promise(var=k1, val=v1) for k1, v1 in literal_map.items()}
+
+            try:
+                output = self.target.__call__(**kwargs_literals)
+                if outputs_expected:
+                    literals.append(output.val)
+
```
eapolinario, line 151: nit: remove extra line.

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+            except Exception as exc:
+                if outputs_expected:
+                    literal_with_none = Literal(scalar=Scalar(none_type=_literal_models.Void()))
+                    literals.append(literal_with_none)
+                    failed_count += 1
+                    if mapped_entity_count - failed_count < min_successes:
+                        logger.error("The number of successful tasks is lower than the minimum ratio")
```
eapolinario, line 158: Drop the word ratio. 
"lower than the minimum"

flytekit/core/array_node.py
```diff
@@ -0,0 +1,214 @@
...
+
+
+class ArrayNode:
+    def __init__(
+        self,
+        target: LaunchPlan,
+        concurrency: Optional[int] = None,
+        min_successes: Optional[int] = None,
+        min_success_ratio: Optional[float] = None,
+        bound_inputs: Optional[Set[str]] = None,
```
eapolinario, line 29: all the code related to `bound_inputs` should be removed since we can't partially eval a launchplan.

flytekit/core/array_node.py
eapolinario: Also, can we make `execution_version` a required parameter?

flytekit/core/array_node.py
eapolinario: Can you leave a comment saying this?

eapolinario (commented):
This is shaping up nicely.

flytekit/core/array_node.py
pvditt: I think the enum is preferable. I'll update that soon.

Yup, will make execution_version required.

flytekit/core/array_node.py
pvditt: good call out - the initial array node map task only enables the setting of min_success_ratio so this wouldn't be a possibility. Let me update this to handle both getting set.

flytekit/core/array_node.py
pvditt: PythonFunctionTasks will require them and we're planning on adding that as a follow up. Since this is a little more ingrained in the logic, I lean towards keeping this in as well. 

Also to note, there is a follow up task to handle bound_input support for mapping over launch plans. I don't know how soon that will become a priority add though.

eapolinario: How is this going to work though? For reference, the support for partial inputs in the case of tasks works today by having the bound inputs as part of the task template.

pvditt: I'll have to look into this further. I initially was looking to update ArrayNode idl to have bounded inputs list and update the validating interface functions in the admin side compiler. I'm unsure if that's the best route. Was looking to shift this mapping to the ArrayNode level and look into keeping the underlying entity's (LaunchPlan, Task, etc) interface unchanged.

eapolinario: We'll need to modify our approach to bound inputs. Right now we only store bound inputs in the task template, we'll have to shift that to a "higher-level" in IDL.

tests/flytekit/unit/core/test_array_node.py
```diff
@@ -0,0 +1,105 @@
...
+@pytest.mark.parametrize(
+    "min_success_ratio, should_raise_error",
+    [
+        (None, True),
+        (1, True),
+        (0.75, False),
+        (0.5, False),
+    ],
+)
+def test_local_exec_lp_min_success_ratio(min_success_ratio, should_raise_error):
```
eapolinario, line 83: Can you add a test where you use `map_task` and a launchplan too?

flytekit/core/array_node.py
pvditt: I think it makes more sense to have it take min_successes over min_success_ratio is both are set instead of favoring.

Also noticed why it wasn't working/implemented in array_node_map_task. Will get that fix up in a separate PR.

pvditt:
@eapolinario once this [PR](https://github.com/flyteorg/flyte/pull/5512) is merged, then these unit tests should pass. I believe the IDL change not being merged to master is causing some issues since I'm now importing the new array node execution mode enum

eapolinario:
> @eapolinario how/where do we bump flyteidl version for these PR builds?

We run tests using [master of flyteidl](https://github.com/flyteorg/flytekit/pull/2480#issuecomment-2251789771).

flytekit/core/array_node_map_task.py
```diff
@@ -347,6 +349,43 @@ def _raw_execute(self, **kwargs) -> Any:
...
+    Args:
+        target: The argument is a Flyte entity of which will be mapped over
+        concurrency: If specified, this limits the number of mapped tasks than can run in parallel to the given batch
+        size. If the size of the input exceeds the concurrency value, then multiple batches will be run serially until
+        all inputs are processed. If set to 0, this means unbounded concurrency. If left unspecified, this means the
+        array node will inherit parallelism from the workflow
+        min_successes: The minimum number of successful executions
+        min_success_ratio: The minimum ratio of successful executions
+        successfully before terminating this task and marking it successful.
+    """
```
eapolinario, line 370: Reformat this docstring to use the [reST format](http://daouzli.com/blog/docstring.html#id26) (e.g. [[1]](https://github.com/flyteorg/flytekit/blob/aa0b7c70450a5ff65956e4964f48f33649a25233/flytekit/configuration/file.py#L298-L304) or [[2]](https://github.com/flyteorg/flytekit/blob/aa0b7c70450a5ff65956e4964f48f33649a25233/flytekit/core/array_node_map_task.py#L44-L51)).

flytekit/tools/translator.py
```diff
@@ -617,6 +628,28 @@ def get_serializable_node(
...
+    options: Optional[Options] = None,
+) -> ArrayNodeModel:
+    array_node = node.flyte_entity
+    if array_node.min_successes is not None:
+        return ArrayNodeModel(
+            node=get_serializable_node(entity_mapping, settings, array_node, options=options),
+            parallelism=array_node.concurrency,
+            min_successes=array_node.min_successes,
+            execution_mode=array_node.execution_mode,
+        )
```
eapolinario, line 643: why do we need to special case `min_successes`?

flytekit/core/array_node.py
```diff
@@ -0,0 +1,221 @@
...
+    Map tasks that maps over tasks and other Flyte entities
+    Args:
+    :param target: The target Flyte entity to map over
+    :param concurrency: If specified, this limits the number of mapped tasks than can run in parallel to the given batch
+        size. If the size of the input exceeds the concurrency value, then multiple batches will be run serially until
+        all inputs are processed. If set to 0, this means unbounded concurrency. If left unspecified, this means the
+        array node will inherit parallelism from the workflow
+    :param min_successes: The minimum number of successful executions. If set, this takes precedence over
+        min_success_ratio
+    :param min_success_ratio: The minimum ratio of successful executions
```
eapolinario, line 205: Reformat docstring. Make sure to add `:return:`.

eapolinario:
> > @eapolinario how/where do we bump flyteidl version for these PR builds?
> 
> We run tests using [master of flyteidl](https://github.com/flyteorg/flytekit/pull/2480#issuecomment-2251789771).

Actually, plugin tests require a small change to make this a reality. PR incoming.

flytekit/tools/translator.py
pvditt: Since we default min_success_ratio to 1.0, it's always set. When passing in both into the proto the one_of will take the last assigned value - which would be min_success_ratio if both are set.

Could also handle this by nulling out min_success_ratio if min_successes is set

flytekit/core/array_node_map_task.py
```diff
@@ -347,6 +349,41 @@ def _raw_execute(self, **kwargs) -> Any:
...
 def map_task(
+    target: Union[LaunchPlan, PythonFunctionTask],
+    concurrency: Optional[int] = None,
+    min_successes: Optional[int] = None,
+    min_success_ratio: float = 1.0,
+    **kwargs,
+):
+    """
+    Wrapper that creates a map task utilizing either the existing ArrayNodeMapTask
+    or the drop in replacement ArrayNode implementation
```
eapolinario, line 360: Can you add a newline after the description? This is causing the monodocs [failure](https://github.com/flyteorg/flytekit/actions/runs/10155606942/job/28084523310?pr=2480) we're seeing in CI. The error is not intuitive, linking the relevant portion of it:
```
2024-07-30T04:55:03.9312901Z sphinx.errors.SphinxWarning: /home/runner/work/flytekit/flytekit/flyte/docs/_src/flytekit/flytekit/core/array_node_map_task.py:docstring of flytekit.core.array_node_map_task.map_task:8:Block quote ends without a blank line; unexpected unindent.
```

flytekit/tools/translator.py
eapolinario: what do you think of the idea of changing the type of `min_success_ratio` in `map_task`? From `float` to `Optional[float]` and set its default value to `None` but at the same time check if both `min_successes` and `min_success_ratio` are set to `None` we "favor" `min_success_ratio` and set that to `1.0`?

This way we centralize this weird interaction between two different parameters in a single function and we don't really validate anything in `ArrayNodeModel` (so that instantiating objects of that type will always take the arguments as-is) and no more special-casing.

eapolinario (commented):
@pvditt , this is getting very close, I have only one small comment.

Also, you mentioned that there was an issue running the sandbox with the launchplan watch api enabled. Is this being tracked somewhere else?

pvditt:
@eapolinario just made a [ticket](https://linear.app/unionai/issue/COR-1648/arraynode-mapping-over-launch-plans-breaks-w-launch-plan-watch-api) for that issue. This isn't an issue when running in any of our tenants. Work around locally is to just disable the [config](https://github.com/unionai/flyte/blob/9c0a0477dc5d3db31442389a01e550263dc0a017/flytepropeller/pkg/controller/nodes/subworkflow/launchplan/adminconfig.go#L22).

pvditt:
@eapolinario yup - I can implement that real quick

eapolinario (approved):
Amazing. Thank you for your persistence, this looks great!

PR #2482, opened 2024-06-14
Skip broken symlinks during calculation of digest (during fast registration)

eapolinario:
We had the erroneous assumption that all files are accessible during fast registration.

Verify that the file exists before trying to open it.

### DIFF
--- a/flytekit/tools/fast_registration.py
+++ b/flytekit/tools/fast_registration.py
@@ -15,6 +15,7 @@ import click
 
 from flytekit.core.context_manager import FlyteContextManager
 from flytekit.core.utils import timeit
+from flytekit.loggers import logger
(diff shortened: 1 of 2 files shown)


flytekit/tools/fast_registration.py
```diff
@@ -95,6 +95,9 @@ def compute_digest(source: os.PathLike, filter: Optional[callable] = None) -> st
 
         for fname in files:
             abspath = os.path.join(root, fname)
+            # Only consider files that exist (e.g. disregard symlinks that point to non-existent files)
+            if not os.path.exists(abspath):
```
wild-endeavor, line 99: should we add an info line?

wild-endeavor (approved):

flytekit/tools/fast_registration.py
eapolinario: good call.

wild-endeavor (approved):

PR #2485, opened 2024-06-17
Pin numpy version in dev requirements, only register type transformer if installed

cosmicBboy:
This is a follow up PR on https://github.com/flyteorg/flytekit/pull/2483.

This PR updates flytekit's imports of numpy:
- Only register the numpy ndarray type transformer if numpy is installed
- Only define the numpy supported types in the Schema and StructuredDataset types if numpy is installed
- Pin numpy < 2 in ci builds involving numpy

### DIFF
--- a/.github/workflows/pythonbuild.yml
+++ b/.github/workflows/pythonbuild.yml
@@ -428,7 +428,7 @@ jobs:
           uv pip install --system -U $GITHUB_WORKSPACE "protobuf<5"
           # TODO: remove this when numpy v2 in onnx has been resolved
           if [[ ${{ matrix.plugin-names }} == *"onnx"* ]]; then
-            uv pip install --system numpy<2.0.0
+            uv pip install --system "numpy<2.0.0"
           fi
           uv pip freeze
       - name: Test with coverage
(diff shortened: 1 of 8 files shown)


.github/workflows/pythonbuild.yml
```diff
@@ -141,7 +147,7 @@ jobs:
         run: |
           pip install uv
           make setup-global-uv
-          uv pip install --system --force-reinstall "${{ matrix.pandas }}"
+          uv pip install --system --force-reinstall "${{ matrix.pandas }}" "${{ matrix.numpy }}"
```
eapolinario, line 150: `numpy` should be a member of `matrix` (similar to [pandas](https://github.com/flyteorg/flytekit/blob/master/.github/workflows/pythonbuild.yml#L124)).

cosmicBboy: we don't want the cartesian product of all four combinations tho right?
```
- pandas<2.0.0, numpy<2.0.0
- pandas<2.0.0, numpy>=2.0.0
- pandas>=2.0.0, numpy<2.0.0
- pandas>=2.0.0, numpy>=2.0.0
```

eapolinario (approved):

PR #2511, opened 2024-06-19
remove upper bound of plugin dependencies for flytekit-onnx-pytorch

Mecoli1219:


kumare3 (commented):
Why are we removing flytekit upper bound of 2.0.0? This should not cause a problem. We should ensure lower bounds are right and upper bounds are indeed needed especially patch/minor upper bounds

PR #2522, opened 2024-06-20
[Feature] Support positional arguments

MortalHappiness:
## Tracking issue
Resolves: flyteorg/flyte#5320

The the issue description for details.

- Change the `inputs` and `outputs` attributes in the `Interface` class to `OrderedDict` to preserve the order.
- Write values in positional arguments to `kwargs`.

### DIFF
--- a/tests/flytekit/unit/core/test_serialization.py
+++ b/tests/flytekit/unit/core/test_serialization.py
@@ -943,3 +943,126 @@ def test_default_args_task_optional_list_type_default_list():
     )
 
     assert wf_with_input() == input_val
+
+def test_positional_args_task():
+    arg1 = 5
+    arg2 = 6
+    ret = 17
+
+    @task
+    def t1(x: int, y: int) -> int:
+        return x + y * 2
+
+    @workflow
+    def wf_pure_positional_args() -> int:
+        return t1(arg1, arg2)
+
+    @workflow
+    def wf_mixed_positional_and_keyword_args() -> int:
+        return t1(arg1, y=arg2)
+
+    wf_pure_positional_args_spec = get_serializable(OrderedDict(), serialization_settings, wf_pure_positional_args)
+    wf_mixed_positional_and_keyword_args_spec = get_serializable(OrderedDict(), serialization_settings, wf_mixed_positional_and_keyword_args)
+
+    arg1_binding = Scalar(primitive=Primitive(integer=arg1))
+    arg2_binding = Scalar(primitive=Primitive(integer=arg2))
+    output_type = LiteralType(simple=SimpleType.INTEGER)
+
+    assert wf_pure_positional_args_spec.template.nodes[0].inputs[0].binding.value == arg1_binding
+    assert wf_pure_positional_args_spec.template.nodes[0].inputs[1].binding.value == arg2_binding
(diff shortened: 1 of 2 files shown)


flytekit/core/base_sql_task.py
```diff
@@ -24,9 +25,9 @@ def __init__(
         query_template: str,
         task_config: Optional[T] = None,
         task_type="sql_task",
-        inputs: Optional[Dict[str, Tuple[Type, Any]]] = None,
```
pingsutw, line 27: > Changed in version 3.7: Dictionary order is guaranteed to be insertion order. This behavior was an implementation detail of CPython from 3.6.

https://docs.python.org/3.7/library/stdtypes.html#dict

flytekit only supports 3.8+, do we still need to use OrderedDict here?

pingsutw: https://mail.python.org/pipermail/python-dev/2017-December/151283.html

pingsutw:
Could you also test it on serverless? Thanks!

wild-endeavor:
@MortalHappiness for changes that only touch flytekit, and to give all reviewers peace of mind, could you please sign up for serverless, use this commit to build the image, and run some tests?  (incidentally, this also helps us test serverless 😄).  Could you link it in the pr description as well please?

tests/flytekit/unit/core/test_serialization.py
```diff
@@ -943,3 +943,102 @@ def wf_with_input() -> typing.Optional[typing.List[int]]:
...
+    arg2 = 6
+    ret = 11
+
+    @task
+    def t1(x: int, y: int) -> int:
+        return x + y
+
+    @workflow
+    def wf_pure_positional_args() -> int:
+        return t1(arg1, arg2)
```
wild-endeavor, line 958: could you also write a test (and run on serverless) where a downstream task takes both positional and named arguments from both 1) workflow inputs and 2) outputs of upstream tasks?

wild-endeavor (commented):
as always, thanks!  this has been bugging us for a while.

flytekit/core/base_sql_task.py
MortalHappiness: I know that the dictionary after Python 3.7 is ordered. I use `OrderedDict` because I found somewhere in the codebase still uses `OrderedDict`. After some discussions in meetings, it seems that those are legacy codes. Therefore, I've rolled back the changes of `OrderedDict`.

tests/flytekit/unit/core/test_serialization.py
MortalHappiness: Done. Please see if these tests are what you want. Thanks. Also, I'll run them on the serverless later.

https://github.com/flyteorg/flytekit/pull/2522/files#diff-fa72fab667c8393cca8d5afd3943db0edc55794423b7cfc3ab33d0f8c3146650R1022-R1044

MortalHappiness:
image

```python
from flytekit import ImageSpec, task, workflow

@task
def t1(x: int, y: int) -> int:
    return x + y * 2

@workflow
def wf() -> int:
    return t1(5, y=6)

```

wild-endeavor (approved):

PR #2527, opened 2024-06-24
Ensure that PythonAutoContainerTask has a container_image attribute

pingsutw:

Fix unit test. https://github.com/flyteorg/flytekit/actions/runs/9620004661/job/26621714719

Call `super().__init__` after setting container_image because PythonAutoContainerTask
is added to the FlyteEntities in `super().__init__`. The translator will iterate over
FlyteEntities and call container_image. Therefore, we should ensure the container_image attribute is set
before appending the task to FlyteEntities.

https://github.com/flyteorg/flytekit/blob/876877abd8d6f4f54dec2738a0ca07a12e9115b1/flytekit/tools/translator.py#L181

## How was this patch tested?
unit test

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/.github/workflows/monodocs_build.yml
+++ b/.github/workflows/monodocs_build.yml
@@ -1,7 +1,7 @@
 name: Monodocs Build
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
   cancel-in-progress: true
 
 on:
--- a/flytekit/core/python_auto_container.py
+++ b/flytekit/core/python_auto_container.py
@@ -95,7 +95,7 @@ class PythonAutoContainerTask(PythonTask[T], ABC, metaclass=FlyteTrackedABC):
 
         # Call super().__init__ after setting _container_image because PythonAutoContainerTask
         # is added to the FlyteEntities in super().__init__. The translator will iterate over
-        # FlyteEntities and call container_image.
+        # FlyteEntities and call entity.container_image().
         # Therefore, we need to ensure the _container_image attribute is set
         # before appending the task to FlyteEntities.
         # https://github.com/flyteorg/flytekit/blob/876877abd8d6f4f54dec2738a0ca07a12e9115b1/flytekit/tools/translator.py#L181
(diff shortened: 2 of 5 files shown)


.github/workflows/monodocs_build.yml
```diff
@@ -1,7 +1,7 @@
 name: Monodocs Build
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}
```
eapolinario, line 4: why are we removing this?

flytekit/core/python_auto_container.py
```diff
@@ -87,19 +87,28 @@ def __init__(
...
+        # Therefore, we need to ensure the _container_image attribute is set
+        # before appending the task to FlyteEntities.
+        # https://github.com/flyteorg/flytekit/blob/876877abd8d6f4f54dec2738a0ca07a12e9115b1/flytekit/tools/translator.py#L181
+
         super().__init__(
             task_type=task_type,
             name=name,
             task_config=task_config,
             security_ctx=sec_ctx,
+            environment=environment,
```
eapolinario, line 109: do we have to worry about `environment` being `None` after this call? (looking at the initialization to `{}` in the existing code).

eapolinario (commented):
I agree in principle with this PR, but I'd like to understand a bit more how this started failing all of a sudden (esp. after https://github.com/flyteorg/flytekit/pull/2485).

.github/workflows/monodocs_build.yml
pingsutw: It allows the GA workflow to run against all pushed commits without being canceled when a new PR is merged

flytekit/core/python_auto_container.py
pingsutw: PythonTask (parent class) will set a default value (`{}`) for it.
https://github.com/flyteorg/flytekit/blob/9d94c83b622b53f65055915efba081470f62c1bc/flytekit/core/base_task.py#L490

.github/workflows/monodocs_build.yml
eapolinario: Makes sense. We use this pattern in a [bunch of our repos](https://github.com/search?q=org%3Aflyteorg+github.event.pull_request.number+%7C%7C+github.ref+org&type=code), can you also make sure these are changed?

PR #2538, opened 2024-06-27
[fix] Validate interface variable names are python-valid

ddl-rliu:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5511

Rationale: flyte inputs names are "like" python variable names, which should not use special characters like . (see my.value in the repro code

Expected behavior: The workflow should actually fail at compile time at step 1. (e.g. perform extra validation)

Workflow should fail at compile time at step 1, when python-invalid variable name is used like "my.value".

### Setup process

```
import logging

from flytekit import ContainerTask, kwtypes, task, workflow

logger = logging.getLogger(__file__)

calculate_ellipse_area_shell = ContainerTask(
    name="ellipse-area-metadata-shell",
    input_data_dir="/var/inputs",
    output_data_dir="/var/outputs",
    inputs=kwtypes(**{"my.value": float, "c": float}),
    outputs=kwtypes(area=float, metadata=str),
    image="ghcr.io/flyteorg/rawcontainers-shell:v2",
    command=[
        "./calculate-ellipse-area.sh"
    ],
)

@task
def report_all_calculated_areas(
    area_shell: float,
    metadata_shell: str,
):
    logger.info(f"shell: area={area_shell}, metadata={metadata_shell}")

@workflow
def wf(a: float, b: float):
    area_shell, metadata_shell = calculate_ellipse_area_shell(**{"my.value": a, "c": b})

    # Report on all results in a single task to simplify comparison
    report_all_calculated_areas(
        area_shell=area_shell,
        metadata_shell=metadata_shell,
    )
```

### DIFF
--- a/flytekit/core/interface.py
+++ b/flytekit/core/interface.py
@@ -3,7 +3,6 @@ from __future__ import annotations
 import collections
 import copy
 import inspect
-import re
 import typing
 from collections import OrderedDict
 from typing import Any, Dict, Generator, List, Optional, Tuple, Type, TypeVar, Union, cast
--- a/flytekit/core/interface.py
+++ b/flytekit/core/interface.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections
 import copy
 import inspect
+import re
 import typing
 from collections import OrderedDict
 from typing import Any, Dict, Generator, List, Optional, Tuple, Type, TypeVar, Union, cast
@@ -70,6 +71,8 @@ class Interface(object):
         self._inputs: Union[Dict[str, Tuple[Type, Any]], Dict[str, Type]] = {}  # type: ignore
         if inputs:
             for k, v in inputs.items():
+                if not k.isidentifier():
+                    raise ValueError(f'Input name must be valid Python identifier: {k!r}')
                 if type(v) is tuple and len(cast(Tuple, v)) > 1:
                     self._inputs[k] = v  # type: ignore
                 else:
--- a/flytekit/core/interface.py
+++ b/flytekit/core/interface.py
@@ -71,7 +71,7 @@ class Interface(object):
         if inputs:
             for k, v in inputs.items():
                 if not k.isidentifier():
-                    raise ValueError(f'Input name must be valid Python identifier: {k!r}')
+                    raise ValueError(f"Input name must be valid Python identifier: {k!r}")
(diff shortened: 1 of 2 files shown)


flytekit/core/interface.py
```diff
@@ -70,6 +71,8 @@ def __init__(
         self._inputs: Union[Dict[str, Tuple[Type, Any]], Dict[str, Type]] = {}  # type: ignore
         if inputs:
             for k, v in inputs.items():
+                if not k.isidentifier():
```
eapolinario, line 74: Nice. I was about to mention `isidentifier`. 😄

flytekit/core/interface.py
```diff
@@ -70,6 +71,8 @@ def __init__(
...
+                    raise ValueError(f'Input must be valid Python identifier: {k!r}')
```
ddl-rliu, line 75: Adds validation on input naming in `Interface`. Note that this is already done on the outputs:
```
variables = [k for k in outputs.keys()]
...
collections.namedtuple(..., variables) # collections.namedtuple checks the variables are valid python identifiers
```
So it is sensible to add similar validation to the inputs, which must also by valid python types.

eapolinario (approved):

tests/flytekit/unit/core/test_interface.py
```diff
@@ -320,6 +320,16 @@ def z(a: int, b: str) -> typing.NamedTuple("NT", x_str=str, y_int=int):
     assert typed_interface.outputs.get("y_int").description == "description for y_int"
 
 
+def test_init_interface_with_invalid_parameters():
+    from flytekit.core.interface import Interface
+
+    with pytest.raises(ValueError, match=r"Input must be valid Python identifier:"):
+        _ = Interface({"my.input": int}, {})
+
+    with pytest.raises(ValueError, match=r"Type names and field names must be valid identifiers:"):
```
ddl-rliu, line 329: The error message is odd here, but it is the standard error message from `collections.namedtuple` see: https://github.com/flyteorg/flytekit/pull/2538/files#r1657915913  https://github.com/python/cpython/blob/main/Lib/collections/__init__.py#L401-L402

eapolinario:
@ddl-rliu , can you fix the lint error in CI? You can run `make lint` from the root to repro it locally.

ddl-rliu:
@eapolinario done!

eapolinario (approved):

PR #2543, opened 2024-07-01
Feat: Improve UX of pytorch-elastic plugin by configuring reasonable defaults

fg91:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/5339

As outlined in the tracking issue, when using the pytorch elastic plugin, users almost always have to configure the following on their own:

1. A pod template with a volume/volume mount which increases the shared memory as the default shared memory segment size of the container is very often too small when torch multiprocessing is used (e.g. for multithreaded data loaders).
2. A higher join timeout because it easily happens that some worker pods start quicker than others, causing the rendezvous of the workers to fail due to timeouts.

Configure reasonable defaults to improve the UX for users:

* Add a flag to `task_config=Elastic()` and `PyTorch()` which allows adding such a shared memory volume to the pod template. The flag defaults to true.
* Configure reasonable join timeouts of 15 minutes.
   
    15 minutes was chosen as an estimate for the time difference between the startup of a pod which is immediately assigned to a running node which has the image pulled (a few seconds) and a pod which requires a node to be scaled up and the image to be pulled.

    If users require a larger timeout, they can of course increase the values but should likely rather use a gang scheduler as described [here](https://docs.flyte.org/en/latest/deployment/plugins/k8s/index.html).

## How was this patch tested?

* Added unit tests
* Ran pytorch and elastic pytorch tasks in a cluster and ensured that the volume/volume mount is added

# Sentence for the release notes:

@wild-endeavor 

> The distributed pytorch and distributed elastic-pytorch tasks in `flytekitplugins-kfpytorch` by default increase the shared memory limit by mounting an `emptyDir` volume with medium `Memory` to to `/dev/shm` as this is almost always required when working with torch multiprocessing (e.g. multi-processed data loader workers or local worker group in distributed training). To disable this, pass `increase_shared_mem=False` to `task_config=PyTorch/Elastic`. `Elastic` tasks now also set a default join timeout of 15 minutes to prevent timeouts when some worker pods require a node scale-up. This setting can be modified via `task_config=Elastic(rdzv_configs{...})`.


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -103,13 +106,18 @@ class PyTorch(object):
...
+            and and one might have to increase the shared memory size. This option configures the task's pod template to mount
+            an `emptyDir` volume with medium `Memory` to to `/dev/shm`.
     """
 
     master: Master = field(default_factory=lambda: Master())
     worker: Worker = field(default_factory=lambda: Worker())
     run_policy: Optional[RunPolicy] = field(default_factory=lambda: None)
     # Support v0 config for backwards compatibility
     num_workers: Optional[int] = None
+    increase_shared_mem: bool = True
```
fg91, line 120: While the tracking issue only discussed the elastic pytorch plugin, I'd make the case that the non-elastic task type benefits from increasing the shared memory by default as well.
As soon as you use data loaders with multiple worker processes, you very likely need this. In the past, I have added the shared memory volume to the flyte pod template configured in the flyte namespace which is applied to all task pods. This has never caused a problem.

thomasjpfan: Can we have another parameter to configure the size limit of `V1EmptyDirVolumeSource`?

fg91: @thomasjpfan have you been in a situation where this was required?
I've run hundreds, if not thousands, of these jobs and have never had a problem not setting the size limit. I'm not opposed to setting the size limit if it is required but if not, I'd prefer to avoid an unnecessary flag here.

thomasjpfan: I always explicitly set it. I am okay with not having another flag if we document what the behavior is.

fg91: https://github.com/flyteorg/flytekit/pull/2543/commits/927e9012af7aa9a2a9f6d878a59ed2ac708c4ad2

wild-endeavor:
+1 from me, but will let thomas give the final go.

wild-endeavor:
and thanks for the blurb!

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
```diff
@@ -0,0 +1,36 @@
...
+        pod_template.pod_spec = V1PodSpec()
+
+    if pod_template.pod_spec.containers is None:
+        pod_template.pod_spec.containers = []
+
+    if pod_template.pod_spec.volumes is None:
+        pod_template.pod_spec.volumes = []
+    pod_template.pod_spec.volumes.append(shm_volume)
+
+    num_containers = len(pod_template.pod_spec.containers)
```
thomasjpfan, line 21: Nit: Can we do the error checking here? 

```python
if num_container >= 2:
    raise ValueError(...)

if num_containers != 1:
    pod_template.pod_spec.containers.append(V1Container(name="primary"))
```

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
```diff
@@ -0,0 +1,36 @@
+from kubernetes.client import V1Container, V1EmptyDirVolumeSource, V1PodSpec, V1Volume, V1VolumeMount
```
thomasjpfan, line 1: We need to add the `kubernetes` dependency to:

https://github.com/flyteorg/flytekit/blob/a7285067bfd90d53c373163c7db961fe09e3231d/plugins/flytekit-kf-pytorch/setup.py#L7

thomasjpfan (commented):
This looks good. There seems to be merge conflicts.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
fg91: https://github.com/flyteorg/flytekit/pull/2543/commits/d03b6af5f2a7d195ab4191e6ed07a0e47b1d1aa9

plugins/flytekit-kf-pytorch/tests/test_shared.py
```diff
@@ -0,0 +1,109 @@
...
+            assert test_task.pod_template.pod_spec.containers is not None
+            assert test_task.pod_template.pod_spec.containers[0].volume_mounts is not None
+
+            assert any([v == expected_volume for v in test_task.pod_template.pod_spec.volumes])
+            assert any(
+                [v == expected_volume_mount for v in test_task.pod_template.pod_spec.containers[0].volume_mounts]
+            )
+
+        else:
+            no_pod_template = test_task.pod_template is None
```
thomasjpfan, line 93: Can these all be asserts similar to the previous branch?

thomasjpfan (approved):
Minor nit, otherwise LGTM

plugins/flytekit-kf-pytorch/tests/test_shared.py
fg91: In the previous branch we simply always assert `is not None` at every nesting level while here in this branch we ultimately want to assert that there are no shared memory volumes+mounts but in theory we could have other volumes+mounts. Or we don't have a container, or no pod spec, or no pod template so we can't just have `assert is None` at every nesting level 🤔 

If I'm overlooking a nicer way to do this, I'm super open though!

thomasjpfan: In the parameterization, there are two entries with `needs_shm_volume=False`:

```python
        (PyTorch(num_workers=3, increase_shared_mem=False), None, False, False),
        (Elastic(nnodes=2, increase_shared_mem=False), None, False, False),
```

I would be okay with this complexity, if we add a new text case with `increase_shared_mem=False` and a `PodTemplate`.

Also can you add a comment to this branch to state that is checking that there are no shm volume mounted.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
```diff
@@ -0,0 +1,36 @@
...
+            "container yourself."
+        )
+
+    if num_containers != 1:
+        pod_template.pod_spec.containers.append(V1Container(name="primary"))
+
+    if pod_template.pod_spec.containers[0].volume_mounts is None:
+        pod_template.pod_spec.containers[0].volume_mounts = []
+
+    pod_template.pod_spec.containers[0].volume_mounts.append(shm_volume_mount)
```
thomasjpfan, line 36: What do we do if there is already a existing shared volume mounted?

plugins/flytekit-kf-pytorch/tests/test_shared.py
fg91: https://github.com/flyteorg/flytekit/pull/2543/commits/d6941d5655452538f2ee79ac18f2027960c06f3d

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
fg91: Good catch, we didn't handle this yet and would have added another mount.

https://github.com/flyteorg/flytekit/pull/2543/commits/adc3eaafb537e7615c6ac8788fe0f4583c69cf0c

thomasjpfan (approved):
LGTM

kumare3:
This is awesome - cc @thomasjpfan @samhita-alla can we improve all our examples to drop the volume config etc

PR #2553, opened 2024-07-03
Added entrypoint to imagespec and default builder

pryce-turner:

Base image entrypoints will sometimes conflict with `pyflyte-execute` this option allows the user to either nullify or replace the entrypoint of the base image.

Adds `entrypoint` attr to ImageSpec

## How was this patch tested?

unit tests

- [ x] I updated the documentation accordingly.
- [ x] All new and existing tests passed.
- [ x] All commits are signed-off.


requirements.txt
```diff
@@ -0,0 +1,325 @@
+#
+# This file is autogenerated by pip-compile with Python 3.10
```
thomasjpfan, line 2: Can this be reverted?

tests/flytekit/unit/core/image_spec/test_default_builder.py
```diff
@@ -28,6 +28,7 @@ def test_create_docker_context(tmp_path):
         requirements=os.fspath(other_requirements_path),
         source_root=os.fspath(source_root),
         commands=["mkdir my_dir"],
+        entrypoint=["/bin/bash"],
```
thomasjpfan, line 31: Can you add another test case to check the behavior of `entrypoint=[]`? 

(You can use `test_create_docker_context_with_git_subfolder` for a quick way to setup `create_docker_context`.)

pryce-turner: Ohhhh because empty lists are falsey, good catch! I found out some other niche behavior in Docker but this should be better now..

pingsutw (approved):

pingsutw:
I'll update the `envd` builder in the follow-up PR.

PR #2563, opened 2024-07-05
Allow for flytekit version to be specified in default image builder

thomasjpfan:

With `1.13.0`, the default image builder does not allow `flytekit` to be specified to a specific version. For example, the following will fail:

```

image1 = ImageSpec(
    packages=["numpy==1.26.4", "flytekit==1.12.3"],
    builder="default",
)


@task(container_image=image1)
def work():
    pass
```

```shell
 pyflyte build build_flyte.py work
```

with:

```
0.220   × No solution found when resolving dependencies:
0.221   ╰─▶ Because you require flytekit==1.13.0 and flytekit==1.12.3, we can
0.221       conclude that the requirements are unsatisfiable.
```

With this PR, we check for `flytekit` first. If it is not detected in `requirements`, then we inject `flytekit` into the requirements with the version used during registration.

## How was this patch tested?

Unit test was added to this PR.

### DIFF
--- a/flytekit/image_spec/default_builder.py
+++ b/flytekit/image_spec/default_builder.py
@@ -1,5 +1,6 @@
 import json
 import os
+import re
 import shutil
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


flytekit/image_spec/default_builder.py
```diff
@@ -124,6 +137,10 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
     if image_spec.packages:
         requirements.extend(image_spec.packages)
 
+    # Adds flytekit if if it is not specified
```
pingsutw, line 140:
```suggestion
    # Adds flytekit if it is not specified
```

pingsutw (approved):

PR #2565, opened 2024-07-05
Fix CSS with Flyte decks

thomasjpfan:

The Flyte Deck's width extends past the width of the window:

Screenshot 2024-07-05 at 5 06 33 PM

With this PR, the bounds are properly set:

Screenshot 2024-07-05 at 4 57 43 PM

## How was this patch tested?

I build an image with the new changes and ran a simple workflow.


pingsutw (approved):
Thank you so much! You’re a life saver

PR #2569, opened 2024-07-07
Fix: Set OMP_NUM_THREADS by default in Elastic

fellhorn:

`torchrun` sets the environment variable `OMP_NUM_THREADS` automatically if not specified ([code](https://github.com/pytorch/pytorch/blob/main/torch/distributed/run.py#L791)). If it is not set, we saw some special cases in which our experiments were stuck for several minutes with high CPU load.

Also see https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html#utilize-openmp for more details.

To make `Elastic` tasks behave the same as executions started with `torchrun`, I would propose copying this behavior over to flytekit.

See above

## How was this patch tested?

1. Added unit tests
2. Tested with internal workflows to check the behavior of `OMP_NUM_THREADS`.

### DIFF
--- a/plugins/flytekit-kf-pytorch/tests/test_pytorch_task.py
+++ b/plugins/flytekit-kf-pytorch/tests/test_pytorch_task.py
@@ -1,3 +1,4 @@
+import os
 import pytest
 from flytekitplugins.kfpytorch.task import CleanPodPolicy, Master, PyTorch, RestartPolicy, RunPolicy, Worker
 
@@ -16,7 +17,6 @@ def serialization_settings() -> SerializationSettings:
         image_config=ImageConfig(default_image=default_img, images=[default_img]),
     )
     return settings
-
 
 def test_pytorch_task(serialization_settings: SerializationSettings):
     @task(
--- a/plugins/flytekit-kf-pytorch/tests/test_pytorch_task.py
+++ b/plugins/flytekit-kf-pytorch/tests/test_pytorch_task.py
@@ -1,4 +1,3 @@
-import os
 import pytest
 from flytekitplugins.kfpytorch.task import CleanPodPolicy, Master, PyTorch, RestartPolicy, RunPolicy, Worker
 
@@ -17,6 +16,7 @@ def serialization_settings() -> SerializationSettings:
         image_config=ImageConfig(default_image=default_img, images=[default_img]),
     )
     return settings
+
 
 def test_pytorch_task(serialization_settings: SerializationSettings):
     @task(
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -332,6 +332,22 @@ class PytorchElasticFunctionTask(PythonFunctionTask[Elastic]):
                 )
             )
 
+        # If OMP_NUM_THREADS is not set, set it to 1 to avoid overloading the system.
+        # Doing so to copy the default behavior of torchrun.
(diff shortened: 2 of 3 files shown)


thomasjpfan (commented):
Although this is changing behavior of current code, I think it's a better default.

For my curiosity, how are coming across this issue with PyTorch? (PyTorch's DataLoader also sets the number of threads to one [here](https://github.com/pytorch/pytorch/blob/940e4477ab0b81eea25051447cf5f599080c903f/torch/utils/data/_utils/worker.py#L259) so it does not oversubscribe with `num_workers`.)

Can we add to `Elastic`'s docstring, how a user should set `OMP_NUM_THREADS`? Do you see users doing this:

```python
# my_workflow.py
import os

os.environ["OMP_NUM_THREADS"] = "2"

@task(task_config=Elastic(...))
def my_task():
    ....
```

kumare3 (dismissed):

kumare3:
@fellhorn i like @thomasjpfan's question. @thomasjpfan it should be set using 
'Pyflyte run --env'

fellhorn:
> @fellhorn i like @thomasjpfan's question. @thomasjpfan it should be set using 'Pyflyte run --env'

Good point, I added some information now in [95a10f9](https://github.com/flyteorg/flytekit/pull/2569/commits/95a10f9eeb5f40b5010becd46c8e47d6303041f9):

> Like `torchrun`, this plugin sets the environment variable `OMP_NUM_THREADS` to 1 if it is not set.
    Please see https://pytorch.org/tutorials/recipes/recipes/tuning_guide.html for potential performance improvements.
    To change `OMP_NUM_THREADS`, specify it in the environment dict of the flytekit task decorator or via `pyflyte run --env`.

thomasjpfan (dismissed):

pingsutw (approved):

plugins/flytekit-kf-pytorch/tests/test_pytorch_task.py
```diff
@@ -1,3 +1,4 @@
+import os
```
pingsutw, line 1: should we remove this?

fellhorn: ec68536f1

Thanks, surprised this was not automatically caught and fixed by the linter

fellhorn:
> Although this is changing behavior of current code, I think it's a better default.
> 
> For my curiosity, how are coming across this issue with PyTorch? (PyTorch's DataLoader also sets the number of threads to one [here](https://github.com/pytorch/pytorch/blob/940e4477ab0b81eea25051447cf5f599080c903f/torch/utils/data/_utils/worker.py#L259) so it does not oversubscribe with `num_workers`.)
> 
> Can we add to `Elastic`'s docstring, how a user should set `OMP_NUM_THREADS`? Do you see users doing this:
> 
> ```python
> # my_workflow.py
> import os
> 
> os.environ["OMP_NUM_THREADS"] = "2"
> 
> @task(task_config=Elastic(...))
> def my_task():
>     ....
> ```

@thomasjpfan regarding your other question: OpenMP is e.g. also used for `torch.jit.script` [inference](https://pytorch.org/docs/stable/notes/cpu_threading_torchscript_inference.html), where no default value for the number of threads is set.

plugins/flytekit-kf-pytorch/tests/test_pytorch_task.py
fellhorn: I saw all test folders are excluded from `ruff`:
```toml
extend-exclude = [
  "tests/",
  "**/tests/**",
]
```

I personally think it's better to ignore certain rules for test folders.

pingsutw (approved):

PR #2574, opened 2024-07-09
Rename databricks task type

pingsutw:
## Tracking issue
NA

For now, Spark tasks can only run on either k8s or Databricks, but they cannot run on both at the same time.

The databricks task's type (`spark`) should be renamed to support running both Spark and Databricks in the same workflow.

- Task: To keep backward compatible, we add a new task config (`DatabricksV2`), and the new task type for this config is `databricks`.

```python
from flytekitplugins.spark import Databricks # Old user

from flytekitplugins.spark import DatabricksV2 as Databricks # New user
```

- Agent: Always register two agetns, these agent supports `spark` and `databricks` respectively.

```python
AgentRegistry.register(DatabricksAgent())
AgentRegistry.register(DatabricksAgentV2())
```

## How was this patch tested?
local/remote

### Screenshots
Screenshot 2024-07-10 at 9 47 26 PM

## Related PRs
NA

## Docs link
NA


flytekit/extend/backend/base_agent.py
```diff
@@ -319,8 +320,11 @@ async def _create(
         self: PythonTask, task_template: TaskTemplate, output_prefix: str, inputs: Dict[str, Any] = None
     ) -> ResourceMeta:
         ctx = FlyteContext.current_context()
+        es = ctx.new_execution_state().with_params(mode=ExecutionState.Mode.TASK_EXECUTION)
```
pingsutw, line 323: Change the execution mode to `TASK_EXECUTION`, so that flytekit will updalod the flytefile to remote during the serialization.

pingsutw: since this [PR](https://github.com/flyteorg/flytekit/commit/bd4c7085e8cc6068900144bb55548e7c6c56f7c3), flytefile will not be uploaded in the `LOCAL_TASK_EXECUTION` mode

Future-Outlier (approved):

PR #2455, opened 2024-05-31
Remove pod-plugin and scikit-learn from official docker image

eapolinario:
## Why are the changes needed?
The flytekit pod plugin has been deprecated for a while and scikit used to be part of the getting started guide (emphasis on used to be).

## What changes were proposed in this pull request?
Remove pod plugin and scikit-learn packages from the flytekit image.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #2589, opened 2024-07-19
add kubernetes package to official docker image again

flixr:
as it is needed in order to use pod templates.
This was indirectly removed with #2455

needed to use pod templates with the official image

install kubernetes package in image


thomasjpfan (approved):
I think this is okay.

`kubernetes` dependencies overlaps with many of our other indirect dependencies, so it does not include the image size too much.

pingsutw (approved):

PR #2593, opened 2024-07-21
unify the image spec hash function

pingsutw:
## Tracking issue
NA

### Problem 1

There are three hash functions in the image spec
1. `__hash__` - used by the lru cache
2. `_calculate_deduped_hash_from_image_spec` - used as an identifier in the image config in the serialization context
3. `calculate_hash_from_image_spec` - used as an image tag 

we should be able to unify 1 and 2

### Probelm 2

Local and remote environments are generating different hash values from the same ImageSpec, which causes dynamic tasks to create tasks with the wrong image. Additionally, this discrepancy indirectly causes is_container always to return false, because the hash value is different from `_F_IMG_ID`

For example:
In flytekit 1.3.14, we added [entrypoint](https://github.com/flyteorg/flytekit/commit/3cd587683fbfc12592622d596391ee6e07548acb) to the ImageSpec. If a user registers a workflow with Flytekit 1.3.14 but uses Flytekit 1.12.0 in the ImageSpec, the `_F_IMG_ID` will always be different from the hash of the ImageSpec. The reason is that we calculate the hash from [asdict(image_spec)](https://github.com/flyteorg/flytekit/blob/a50eb4bd92f294cbe92fc5db4318071cc636b8b4/flytekit/image_spec/image_spec.py#L383)

The solution is to calculate the hash only from the non-None values in the ImageSpec to ensure the hash is consistent across different flytekit versions.
```python
asdict(spec, dict_factory=lambda x: {k: v for (k, v) in x if v is not None})
```

Add a few methods to ImageSpec
- `id` -> It returns a unique hash as identifier of the ImageSpec. it will be used to identify the imageSpec in the ImageConfig or check if the current container image in the pod is built from this image spec in `is_container()`.
- `__hash__` -> it will return the hash of ImageSpec ID
- `tag` -> the tag of the image. 

Remove
- _calculate_deduped_hash_from_image_spec
- calculate_hash_from_image_spec

## How was this patch tested?
```
pyflyte run --remote flyte-example/image_spec_dynamic.py wf
```

### Setup process
```python
from flytekit import ImageSpec, task, dynamic, workflow


new_flytekit = "git+https://github.com/flyteorg/flytekit.git@c81fb9226d76de006bb167a2ed6e1cc8b4c48143"

repro_img = ImageSpec(
    builder="envd",
    packages=["mypy", new_flytekit],
    apt_packages=["git"],
    registry="pingsutw",
)

image_foo = repro_img.with_apt_packages(["curl"])
image_d1 = repro_img.with_apt_packages(["wget"])
image_bar = repro_img.with_apt_packages(["vim"])


if image_foo.is_container():
    print("foo in container.")


# hello
@task(container_image=image_foo)
def foo():
    print("foo")


@task(container_image=image_bar)
def bar():
    print("bar")


@dynamic(container_image=image_d1)
def d1():
    bar()


@workflow
def wf():
    foo()
    d1()
```

## Related PRs
NA

## Docs link
NA


flytekit/image_spec/image_spec.py
```diff
@@ -79,6 +79,22 @@ def __post_init__(self):
         if self.registry:
             self.registry = self.registry.lower()
 
+    @property
```
thomasjpfan, line 82: We can use `functools.cached_property` to cache this: https://docs.python.org/3/library/functools.html#functools.cached_property.

flytekit/image_spec/image_spec.py
```diff
@@ -163,7 +181,7 @@ def exist(self) -> Optional[bool]:
             return None
 
     def __hash__(self):
-        return hash(asdict(self).__str__())
+        return hash(self.id)
```
thomasjpfan, line 184: Can this just return `self.id`?

pingsutw: it should return a number instead of string

flytekit/image_spec/image_spec.py
pingsutw: this is cool. I'll update it

pingsutw: Tests fail after I change to `functools.cached_property`. Maybe it's because __hash__ also calls id(), which could cause a race condition

pingsutw: added it

pingsutw:
@thomasjpfan mind taking a look again, thanks!

flytekit/image_spec/image_spec.py
```diff
@@ -91,37 +92,90 @@ def __post_init__(self):
...
-            parameter_is_None = attr is None
+            parameter_is_none = attr is None
             parameter_is_list_string = isinstance(attr, list) and all(isinstance(v, str) for v in attr)
-            if not (parameter_is_None or parameter_is_list_string):
+            if not (parameter_is_none or parameter_is_list_string):
                 error_msg = f"{parameter} must be a list of strings or None"
                 raise ValueError(error_msg)
 
+    @cached_property
+    def id(self) -> str:
```
thomasjpfan, line 102: Let's say there are two workflows in two different directories with this image spec:

```python
image = ImageSpec(name="my_name", requirements="requirements-prod.txt")
```

where `requirements-prod.txt` is not the same. They will end up with the same `image_spec.id`, but different `image_spec.tag`.

Is it okay for an image spec to have the same id, but a different tag?

thomasjpfan: Thinking about it more, having the same `id` and different `tag` is harder to come across because the local working directory is the context for `requirements` and `source_root`.

pingsutw: yup, exactly

flytekit/image_spec/image_spec.py
```diff
@@ -91,37 +92,90 @@ def __post_init__(self):
...
+        # won't rebuild the image if we change the registry_config path
+        spec.registry_config = None
+        image_spec_dict = asdict(spec, dict_factory=lambda x: {k: v for (k, v) in x if v is not None})
+        image_spec_bytes = image_spec_dict.__str__().encode("utf-8")
+        tag = (
+            base64.urlsafe_b64encode(hashlib.md5(image_spec_bytes).digest())
+            .decode("ascii")
+            .rstrip("=")
+            .replace("-", "_")
+        )
```
thomasjpfan, line 154: I think this works: 

```suggestion
        tag = spec.id.replace("-", "_")
```

At this point, can we have `.replace("-", "_")` in `.id` ?

thomasjpfan (approved):

PR #2594, opened 2024-07-21
Feat: Make exception raised by external command authenticator more actionable

fg91:

Flytekit's admin client can be configured to refresh credentials using an external command. When this command fails, the user is presented with this error:

```console
Problems refreshing token with command: Command '['some', 'failing', 'command']' returned non-zero exit status 1.
```

* As a platform engineer managing Flyte for ML engineering teams (who don't have to know the details of authentication), I've been asked more than once what to do in this situation - it seems to not always be clear that one needs to execute the command to debug why it fails.
* Even if one knows what to do in this situation, one still has to remove the `[`, `,`, `'`, ... in order to be able to execute the command which can be annoying if the command has multiple args (see e.g. [here](https://github.com/flyteorg/flytekit/tree/master/plugins/flytekit-identity-aware-proxy#configuring-the-token-generation-cli-provided-by-this-plugin)).

I propose to change the error message to ...

```console
Failed to refresh token with command `some failing command`. Please execute this command in your terminal to debug.
```

... so that it's very clear how to proceed and so that one can just copy paste the command.


flytekit/clients/auth/authenticator.py
```diff
@@ -179,12 +179,15 @@ def refresh_credentials(self):
...
         try:
             output = subprocess.run(self._cmd, capture_output=True, text=True, check=True)
-        except subprocess.CalledProcessError as e:
-            logging.error("Failed to generate token from command {}".format(self._cmd))
-            raise AuthenticationError("Problems refreshing token with command: " + str(e))
+        except subprocess.CalledProcessError:
+            logging.error("Failed to generate token from command `{}`".format(cmd_joined))
+            raise AuthenticationError(
+                f"Failed to refresh token with command `{cmd_joined}`. Please execute this command in your terminal to debug."
+            )
```
fg91, line 190: I briefly considered logging `.stderr` but decided against this as this might accidentally log and thus leak tokens.

PR #2599, opened 2024-07-22
Global default flytefile downloader

pingsutw:
## Tracking issue
NA

downloader is only picklable if they are defined at the top-level of a module.

### DIFF
--- a/flytekit/types/file/file.py
+++ b/flytekit/types/file/file.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 import mimetypes
 import os
 import pathlib
@@ -29,6 +30,10 @@ def noop(): ...
 
 
 T = typing.TypeVar("T")
+
+
+def _default_downloader(ctx: FlyteContext, remote_path: str, local_path: str) -> None:
+    return ctx.file_access.get_data(remote_path, local_path, is_multipart=False)
 
 
 @dataclass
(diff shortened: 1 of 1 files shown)


thomasjpfan (commented):
In what context did you have to pickle the output of `FlyteFile.to_python_value`?

Also, can you add a unit test that pickles the object?

tests/flytekit/unit/core/test_flyte_file.py
```diff
@@ -699,3 +701,26 @@ def test_new_remote_file():
     nf = FlyteFile.new_remote_file(name="foo.txt")
     assert isinstance(nf, FlyteFile)
     assert nf.path.endswith('foo.txt')
+
+
+def test_pickle_flyte_file():
```
thomasjpfan, line 706: This test is passing on `master`. I think it is going to this branch:

https://github.com/flyteorg/flytekit/blob/d29e30eedc48f0f3b8d72dd5ea42f9dbdb1f2f13/flytekit/types/file/file.py#L561-L562

and not getting all the way to the `_downloader` yet.

pingsutw: Good catch, thanks. I updated the test. We should not only move the downloader to a global scope but also need to ensure it is pickleable somehow. I haven't figured out how to do that yet, so I will revisit it later

PR #2601, opened 2024-07-22
Added alt prefix head to FlyteFile.new_remote

pryce-turner:
## Tracking issue
SE-75

Sometimes it would be nice to write to a non-default bucket from within a task. Being able to create a `new_remote` with an alt bucket provides that.

New arg to `new_remote`

## How was this patch tested?

unit tests

### DIFF
--- a/flytekit/types/file/file.py
+++ b/flytekit/types/file/file.py
@@ -179,13 +179,21 @@ class FlyteFile(SerializableType, os.PathLike, typing.Generic[T], DataClassJSONM
         return ""
 
     @classmethod
-    def new_remote_file(cls, name: typing.Optional[str] = None) -> FlyteFile:
+    def new_remote_file(cls, alt: typing.Optional[str] = None, name: typing.Optional[str] = None) -> FlyteFile:
         """
         Create a new FlyteFile object with a remote path.
+
+        :param alt: If you want to specify a different prefix head than the default one, you can specify it here.
+        :param name: If you want to specify a different name for the file, you can specify it here.
         """
         ctx = FlyteContextManager.current_context()
         r = name or ctx.file_access.get_random_string()
-        remote_path = ctx.file_access.join(ctx.file_access.raw_output_prefix, r)
+        pref = ctx.file_access.raw_output_prefix
+        if alt:
+            s_pref = pref.split("/")
+            s_pref[2] = alt
+            pref = "/".join(s_pref)
+        remote_path = ctx.file_access.join(pref, r)
         return cls(path=remote_path)
 
     @classmethod
--- a/flytekit/types/file/file.py
+++ b/flytekit/types/file/file.py
@@ -179,21 +179,15 @@ class FlyteFile(SerializableType, os.PathLike, typing.Generic[T], DataClassJSONM
         return ""
 
     @classmethod
-    def new_remote_file(cls, alt: typing.Optional[str] = None, name: typing.Optional[str] = None) -> FlyteFile:
+    def new_remote_file(cls, name: typing.Optional[str] = None, alt: typing.Optional[str] = None) -> FlyteFile:
         """
         Create a new FlyteFile object with a remote path.
 
-        :param alt: If you want to specify a different prefix head than the default one, you can specify it here.
         :param name: If you want to specify a different name for the file, you can specify it here.
+        :param alt: If you want to specify a different prefix head than the default one, you can specify it here.
         """
         ctx = FlyteContextManager.current_context()
-        r = name or ctx.file_access.get_random_string()
-        pref = ctx.file_access.raw_output_prefix
-        if alt:
-            s_pref = pref.split("/")
-            s_pref[2] = alt
-            pref = "/".join(s_pref)
-        remote_path = ctx.file_access.join(pref, r)
+        remote_path = ctx.file_access.get_new_path(alt=alt, stem=name)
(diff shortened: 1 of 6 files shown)


flytekit/types/file/file.py
```diff
@@ -179,13 +179,21 @@ def extension(cls) -> str:
         return ""
 
     @classmethod
-    def new_remote_file(cls, name: typing.Optional[str] = None) -> FlyteFile:
+    def new_remote_file(cls, alt: typing.Optional[str] = None, name: typing.Optional[str] = None) -> FlyteFile:
```
eapolinario, line 182: Can you add `alt` as the last parameter?

flytekit/types/file/file.py
```diff
@@ -179,13 +179,21 @@ def extension(cls) -> str:
...
+        :param name: If you want to specify a different name for the file, you can specify it here.
         """
         ctx = FlyteContextManager.current_context()
         r = name or ctx.file_access.get_random_string()
-        remote_path = ctx.file_access.join(ctx.file_access.raw_output_prefix, r)
+        pref = ctx.file_access.raw_output_prefix
+        if alt:
+            s_pref = pref.split("/")
+            s_pref[2] = alt
+            pref = "/".join(s_pref)
```
eapolinario, line 195: using `/` explicitly is not going to work on windows. Use `os.sep` instead.

flytekit/types/file/file.py
```diff
@@ -179,13 +179,21 @@ def extension(cls) -> str:
...
+        :param alt: If you want to specify a different prefix head than the default one, you can specify it here.
+        :param name: If you want to specify a different name for the file, you can specify it here.
         """
         ctx = FlyteContextManager.current_context()
         r = name or ctx.file_access.get_random_string()
-        remote_path = ctx.file_access.join(ctx.file_access.raw_output_prefix, r)
+        pref = ctx.file_access.raw_output_prefix
+        if alt:
+            s_pref = pref.split("/")
+            s_pref[2] = alt
```
eapolinario, line 194: let's make sure that `len(s_pref)` is at least 2 before executing this. I'm not sure if you intend to run this only in the case of blob stores (if that's the case we have to make this code more resilient and clarify its intent).

flytekit/types/file/file.py
pryce-turner: Why is that, out of curiosity? I put it ahead to match it's location in the final path.

flytekit/types/file/file.py
pryce-turner: Flawless only wants it for blob store, but I figured it could be used for any path that starts with `protocol://` where they'd want to change the first part. Behavior is slightly different in a local context, but still useful. This is mostly my open-question around how many fsspec protocols we should support..

flytekit/types/file/file.py
eapolinario: This is just to maintain backwards compatibility.

flytekit/types/file/file.py
eapolinario: actually, scratch all that, the method [join in data persistence](https://github.com/flyteorg/flytekit/blob/d29e30eedc48f0f3b8d72dd5ea42f9dbdb1f2f13/flytekit/core/data_persistence.py#L435-L455) already does the correct thing. You should be able to replace all this code with this just:
```
remote_path = ctx.file_access.join(ctx.file_access.raw_output_prefix, alt, r)
```

flytekit/types/file/file.py
pryce-turner: DUH. I'm so deep in named arguments now that I forgot about positional ones 🤦

flytekit/types/file/file.py
pryce-turner: But it's part of the raw output prefix that we're trying to replace, specifically the bucket name. Unless I'm misunderstanding.

pryce-turner:
@eapolinario please TAL at the `core/data_persistrence.py` change and let me know what you think about the approach. A couple considerations:
1. There are a good few cases where `file_access.join` is used throughout, but almost all with the same pattern and could be replaced with the new method except in one case.
2. We could validate at the protocol level since we went one layer deeper, but it would break tests for local and I don't think the local case is a problem as is.
3. How do we test for the protocols we _do_ care about e.g. `gcs`, and `s3`?

eapolinario:
This is shaping up to be a great addition to the arsenal of utility functions in flytekit. A few inline comments.

> @eapolinario please TAL at the `core/data_persistrence.py` change and let me know what you think about the approach. A couple considerations:
> 
> 1. There are a good few cases where `file_access.join` is used throughout, but almost all with the same pattern and could be replaced with the new method except in one case.
Cool, these can be tackled in a separate PR.

> 2. We could validate at the protocol level since we went one layer deeper, but it would break tests for local and I don't think the local case is a problem as is.

> 3. How do we test for the protocols we _do_ care about e.g. `gcs`, and `s3`?
Setting the raw output prefix in calls to `FileAccessProvider` is not enough or you're looking for integration-type tests?

pryce-turner:
Thanks for the comments @eapolinario, please lmk what you think of the FlyteFile and FlyteDirectory new_remote implementations.

pryce-turner:
@eapolinario LGTM?

eapolinario (approved):
Thank you, this looks great!

PR #2602, opened 2024-07-23
eliminate redundant literal conversion for `Iterator[JSON]` type

samhita-alla:

This change is necessary to enable sending an iterator as a default workflow input. Without this change, the iterator is iterated over twice, resulting in an `iterator is empty` error during literal conversion.

Added a conditional to skip selecting the default value when the type is `Iterator[JSON]` and the entity is a workflow. This bug in iterator/openai batch agent, introduced in flytekit 1.13.0, was caused by https://github.com/flyteorg/flytekit/pull/2525.

## How was this patch tested?

- Verified with OpenAI batch agent.
- Tested task and workflow with iterator as default input.
- Confirmed task and workflow with list of integers work correctly.

### DIFF
--- a/tests/flytekit/unit/cli/pyflyte/test_run.py
+++ b/tests/flytekit/unit/cli/pyflyte/test_run.py
@@ -12,10 +12,18 @@ from click.testing import CliRunner
 from flytekit.loggers import logging, logger
 
 from flytekit.clis.sdk_in_container import pyflyte
-from flytekit.clis.sdk_in_container.run import RunLevelParams, get_entities_in_file, run_command
+from flytekit.clis.sdk_in_container.run import (
+    RunLevelParams,
+    get_entities_in_file,
+    run_command,
+)
 from flytekit.configuration import Config, Image, ImageConfig
 from flytekit.core.task import task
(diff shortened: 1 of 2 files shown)


tests/flytekit/unit/cli/pyflyte/test_run.py
```diff
@@ -429,6 +484,125 @@ def check_image(*args, **kwargs):
...
+                ],
+            },
+        },
+    ]:
+        yield x
+
+
+@mock.patch("flytekit.configuration.default_images.DefaultImages.default_image")
+def test_pyflyte_run_with_iterator_json_type(
+    mock_image, mock_image_spec_builder, caplog
```
eapolinario, line 507: `caplog` ❤️ 

A lot of the tests in this test suite should use `caplog` instead. We can tackle this in a separate PR.

eapolinario (commented):
Can you take a look at the `build-with-pandas` unit test failures?

samhita-alla:
@eapolinario the tests pass now.

eapolinario (approved):

PR #2606, opened 2024-07-25
Enable Ray Fast Register

fiedlerNr9:
Today, when launching a Ray job in Flyte only the Head Node is retrieving fast register data - no applying for Ray Worker Nodes. This results in the need for the user of copy the source code into the containers working dir & rebuilding the container image for every change on the ray user code.

This PR utilizes rays functionality of adding files to the ray job on runtime passing the `runtime_env {"working_dir":"some_path"}`.

⚠️ There is an ray gcs storage limit of 500MiB. Thats why ray supports adding an `excludes` variable along with `working_dir` which functions as a .ignore file. This PR introduces a `excludes_working_dir` to the RayJobConfig, which allows control over what files to upload to the ray cluster.

- Adding `working_dir` & `excludes` init parameter when running remotely
- Docs on what this [parameter](https://docs.ray.io/en/latest/ray-core/handling-dependencies.html#api-reference) is doing
- TLDR: after head node downloads fast register data -> `ray.init("working_dir"="/root", excludes=["data"])` zips all fast register data, takes `excludes` into account and uploads to ray gcs -> ray worker nodes downloads zips from ray gcs and unzips

## How was this patch tested?
- Running the following workflow with Imagespec works:
``` python
from flytekit import ImageSpec, Resources, task, workflow, PodTemplate
from flytekitplugins.ray import HeadNodeConfig, RayJobConfig, WorkerNodeConfig
from ray.util.multiprocessing import Pool

import time
import math
import random

container_image = ImageSpec(
    name="rayworkerlove",
    apt_packages=["git", "wget"],
    packages=[
        "git+https://github.com/flyteorg/flytekit.git@d5549cee2a9bef8338bd9e94aa23397631386c10#subdirectory=plugins/flytekit-ray"
    ],
    registry="ghcr.io/fiedlerNr9",
)


def sample(num_samples):
    num_inside = 0
    for _ in range(num_samples):
        x, y = random.uniform(-1, 1), random.uniform(-1, 1)
        if math.hypot(x, y) <= 1:
            num_inside += 1
    return num_inside


ray_config = RayJobConfig(
    head_node_config=HeadNodeConfig(
        ray_start_params={"log-color": "True", "num-cpus": "1"}
    ),
    worker_node_config=[
        WorkerNodeConfig(
            group_name="ray-group", replicas=3, min_replicas=1, max_replicas=4
        )
    ],
    shutdown_after_job_finishes=True,
    ttl_seconds_after_finished=120,
    enable_autoscaling=False,
    excludes_working_dir=["micromamba"],
)


@task(
    task_config=ray_config,
    requests=Resources(mem="2Gi", cpu="1"),
    container_image=container_image,
)
def approximate_pi_distributed(num_samples: int):

    pool = Pool()

    start = time.time()
    num_inside = 0
    sample_batch_size = 100000
    for result in pool.map(
        sample, [sample_batch_size for _ in range(num_samples // sample_batch_size)]
    ):
        num_inside += result

    print("pi ~= {}".format((4 * num_inside) / num_samples))
    print("Finished in: {:.2f}s".format(time.time() - start))


@workflow
def wf(n: int = 60000000):
    approximate_pi_distributed(num_samples=n)

```

### Screenshots
image

### DIFF
--- a/plugins/flytekit-ray/flytekitplugins/ray/task.py
+++ b/plugins/flytekit-ray/flytekitplugins/ray/task.py
@@ -1,16 +1,22 @@
 import base64
 import json
+import os
 import typing
 from dataclasses import dataclass
 from typing import Any, Callable, Dict, Optional
-
+import os
 import yaml
-from flytekitplugins.ray.models import HeadGroupSpec, RayCluster, RayJob, WorkerGroupSpec
+from flytekitplugins.ray.models import (
+    HeadGroupSpec,
+    RayCluster,
+    RayJob,
+    WorkerGroupSpec,
+)
 from google.protobuf.json_format import MessageToDict
 
 from flytekit import lazy_module
 from flytekit.configuration import SerializationSettings
-from flytekit.core.context_manager import ExecutionParameters
+from flytekit.core.context_manager import ExecutionParameters, FlyteContextManager
 from flytekit.core.python_function_task import PythonFunctionTask
 from flytekit.extend import TaskPlugins
 
@@ -40,6 +46,7 @@ class RayJobConfig:
     address: typing.Optional[str] = None
     shutdown_after_job_finishes: bool = False
     ttl_seconds_after_finished: typing.Optional[int] = None
+    excludes_working_dir: typing.Optional[typing.List[str]] = None
 
 
 class RayFunctionTask(PythonFunctionTask):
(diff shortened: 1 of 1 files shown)


plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -50,11 +57,31 @@ class RayFunctionTask(PythonFunctionTask):
...
+            working_dir = os.getcwd()
+            init_params["runtime_env"] = {"working_dir": working_dir}
+
+            cfg = self._task_config
+            if cfg.excludes_working_dir:
+                init_params["runtime_env"]["excludes"] = cfg.excludes_working_dir
+
+            # fast register data with timestamp mtime=0 will be zipped and uploaded to ray gcs
+            # zip does not support timestamps before 1980 -> hacky workaround of touching all the files
+            os.system(f"touch `find {working_dir} -type f`")
```
thomasjpfan, line 82: This depends on the system having `touch` + `find`, which is usually the case.

To be safer, can this use [os.utime](https://docs.python.org/3/library/os.html#os.utime) + [os.walk](https://docs.python.org/3.9/library/os.html#os.walk) to update the mtimes?

plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -50,11 +57,31 @@ class RayFunctionTask(PythonFunctionTask):
...
+        if not ctx.execution_state.is_local_execution():
+            working_dir = os.getcwd()
+            init_params["runtime_env"] = {"working_dir": working_dir}
+
+            cfg = self._task_config
+            if cfg.excludes_working_dir:
+                init_params["runtime_env"]["excludes"] = cfg.excludes_working_dir
+
+            # fast register data with timestamp mtime=0 will be zipped and uploaded to ray gcs
+            # zip does not support timestamps before 1980 -> hacky workaround of touching all the files
```
thomasjpfan, line 81: Can this comment say explicitly how the data is being moved around? My understanding is:

1. Flyte's fast register and sends it to the head node
2. Ray then zips the working directory and sends it to the ray gcs
3. Ray worker pulls the files from ray's gcs.

Is this correct?

plugins/flytekit-ray/flytekitplugins/ray/task.py
fiedlerNr9: Good catch! I can try this out & adjust. Also this should only stay short term. @pingsutw mentioned there is work in progress that allows us to not set `mtime=0` for the fast register `tar.gz` cc @pingsutw

plugins/flytekit-ray/flytekitplugins/ray/task.py
fiedlerNr9: Its exactly this only with the addition that ray respects the given `excludes` variable (`list[str]`), which functions as a ignore file

plugins/flytekit-ray/flytekitplugins/ray/task.py
kumare3: i do not think we will do this right now @fiedlerNr9, because the mtime=0 makes it possible to get consistent hashes for the similar tar files, otherwise we get multiple uploads to admin.

I think you should do what @thomasjpfan recommends use `os.walk`. the os.system will break and will cause random bugs.

plugins/flytekit-ray/flytekitplugins/ray/task.py
pingsutw: qq: could we exclude the file during the fast register? (.gitignore)
Do we have a case where we want to upload a file to the head node but don't want it on the worker nodes?"

fiedlerNr9: > Do we have a case where we want to upload a file to the head node but don't want it on the worker nodes?
No i dont think so.

What file are you trying to exclude? The fast register `.tar.gz`? Actually I am thinking, its only the .tar.gz file coming with the `mtime=0` right? If we exclude that we dont need to modify any timestamp on files?

plugins/flytekit-ray/flytekitplugins/ray/task.py
fiedlerNr9: No need to manually adjust the timestamp for this now. I just catch the fast register data and dont zip it at all

fiedlerNr9:
Okay no need to manually update timestamps here. I am just always put the in the exclude variable, since there is no need to upload the .tar.gz to the ray gcs.

`"excludes": ["script_mode.tar.gz", "fast*.tar.gz"]`!

Works and tested with pyflyte run and register. Would appreciate another review

pingsutw (approved):

eapolinario (commented):
Jus to reiterate, this looks great. Thanks, @fiedlerNr9 !

PR #2607, opened 2024-07-25
Customize PyTorch task error files

bgedik:
## Tracking issue
TODO

TODO

TODO
## How was this patch tested?
TODO

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -559,6 +559,15 @@ class PythonTask(TrackedInstance, Task, Generic[T]):
         """
         return self._task_config
 
+    def get_error_file_name(self) -> Optional[str]:
+        """
+        Returns the error file name for the task. Typically used for customizing the default behavior
+        of using a fixed file name; such as for elastic tasks where multiple sub-tasks can generate 
+        errors that need to be mapped back to an error for the top level task. Such scenarios require 
+        customization of the error file name, so that the earliest error can be picked as the final result.
+        """
+        return None
+
     def get_type_for_input_var(self, k: str, v: Any) -> Type[Any]:
         """
         Returns the python type for an input variable by name.
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -562,9 +562,9 @@ class PythonTask(TrackedInstance, Task, Generic[T]):
     def get_error_file_name_suffix(self) -> Optional[str]:
         """
         Returns the error file name suffix for the task. Typically used for customizing the default behavior
-        of using a fixed file name; such as for elastic tasks where multiple sub-tasks can generate
-        errors that need to be mapped back to an error for the top level task. Such scenarios require
-        customization of the error file name, so that the earliest error can be picked as the final result.
+        of using a fixed file name; such as for tasks where multiple pods working on behalf of the task can
+        generate errors that need to be mapped back to an error for the top level task. Such scenarios require
+        customization of the error file name, so that the earliest error can be picked as the final error.
         """
         return None
 
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -292,7 +292,7 @@ class Task(object):
         except TypeTransformerFailedError as exc:
             msg = f"Failed to convert inputs of task '{self.name}':\n  {exc}"
             logger.error(msg)
-            raise TypeError(msg) from exc
+            raise TypeError(msg) from None
         input_literal_map = _literal_models.LiteralMap(literals=literals)
 
         # if metadata.cache is set, check memoized version
@@ -358,7 +358,7 @@ class Task(object):
         return flyte_entity_call_handler(self, *args, **kwargs)  # type: ignore
 
     def compile(self, ctx: FlyteContext, *args, **kwargs):
-        raise Exception("not implemented")
+        raise NotImplementedError
 
     def get_container(self, settings: SerializationSettings) -> Optional[_task_model.Container]:
         """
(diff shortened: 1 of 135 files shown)


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -334,6 +334,25 @@ def __init__(self, task_config: Elastic, task_function: Callable, **kwargs):
                 self.pod_template = PodTemplate()
             add_shared_mem_volume_to_pod_template(self.pod_template)
 
+    def get_error_file_name_suffix(self) -> Optional[str]:
+        """
+        Returns error file name suffix that represents the replica index.
+        """
+        # This is fragile, as it assumes hostname == podname, which is not the case if
```
bgedik, line 341: What's needed is to customize the pod spec to set an environment variable that captures the value of metadata.labels."training.kubeflow.org/replica-index" :

```
env:
 - name: FLYTE_PYTORCH_TASK_REPLICA_INDEX
   valueFrom:
     fieldRef:
       fieldPath: metadata.labels['training.kubeflow.org/replica-index']
```

flytekit/core/base_task.py
```diff
@@ -559,6 +559,15 @@ def task_config(self) -> Optional[T]:
         """
         return self._task_config
 
+    def get_error_file_name_suffix(self) -> Optional[str]:
```
fg91, line 562: Nit (can be ignored right now until we are sure of how to implement this feature):
We shouldn't mention specific plugins like elastic here but rather speak of plugins using multiple workers which might produce independent error files.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -455,6 +474,14 @@ def fn_partial():
         except SignalException as e:
             logger.exception(f"Elastic launch agent process terminating: {e}")
             raise IgnoreOutputs()
+        except Exception as e:
```
fg91, line 477: Can we do this in another PR please?
Is my understanding correct that any actual user error would have been a `ChildFailedError`? Do you have an example of an exception that isn't yet handled?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: Setting this env var via the downward api in the pod spec in flyteplugins would in theory work I'd say 👍
I agree that we shouldn't use the hostname.

I have two questions though:

* Does the error file name actually need to contain the group rank? If the error message itself would contain the info which group rank failed, could we get away with a random uuid suffix which just prevents overwriting?
* Do you know how torch distributed elastic determines the group rank for the worker pods/processes? I just checked on a running elastic task and (in line with what you said) only see the `PET_NNODES` and `PET_NPROC_PER_NODE` env vars which specify the number of nodes and processes but not the rank. Could it be that `worker-0` ends up being group rank 2 or how is this decided at rendezvous?

bgedik: The ranks are decided by the rendezvous and worker-0 is not necessarily the one that has rank 0 or group rank 0.

I agree that the suffix does not need to be the group rank. We can do uuid, but internal debugging might get a bit harder, as by looking at the error files, no one would know what pod actually generated it.

What about using the pod name? Is that available in Flyte via some mechanism or does that also require downward api usage?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
bgedik: I've seen `RendezvousError` happening. Typically it happens when there is a much earlier ChildError from another group. Pod A has a ChildError, Pod B has a RendezvousError, and if Pod B's error file takes precedence, then we end up preventing a task restart :( 

I have this change in our local environment (via monkey patching `elastic_launch`), because with the current behavior of task outputs overwriting each other, these exceptions are preventing task restarts. 

I agree that this can belong to a separate PR. Btw, if we fix the task error file issue and pick the earliest, then the impact of this is smaller (but I still think treating these as non-recoverable user errors is a surprising behavior).

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: > The ranks are decided by the rendezvous and worker-0 is not necessarily the one that has rank 0 or group rank 0.

Ok then it makes sense 👍 

> What about using the pod name? Is that available in Flyte via some mechanism or does that also require downward api usage?

I think it's not available unfortunately. Could be added though I'd say.

> I agree that the suffix does not need to be the group rank. We can do uuid, but internal debugging might get a bit harder, as by looking at the error files, no one would know what pod actually generated it

The [`ChiledFailedError`](https://github.com/pytorch/pytorch/blob/f3df7deab8953af76ff1723ed49094208057a834/torch/distributed/elastic/multiprocessing/errors/__init__.py#L234) has a ref to the global rank but that doesn't help us because it's not linked to the pod and we are interested in the pod name in order to look at the respective stackdriver logs for more info - correct?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
fg91: Can you please open another PR for this? I think we can merge this much faster than proper treatment of multiple error files.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
bgedik: `ChiledFailedError` is not the only error that comes out of `torch_elastic`, unfortunately. 

We can start with uuid and improve afterwards. Made the change.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
pingsutw: @fg91 Thanks for reviewing the PR 😄

kumare3:
@bgedik did you give up on this?

bgedik:
> @bgedik did you give up on this?

@kumare3 We can close this one. Currently working on backend changes on multi error file reader (there is an open PR). This PR will be reworked based on the latest RFC.

bgedik:
> > @bgedik did you give up on this?
> 
> @kumare3 We can close this one. Currently working on backend changes on multi error file reader (there is an open PR). This PR will be reworked based on the latest RFC.

https://github.com/flyteorg/flytekit/pull/2797 is the new PR

PR #2608, opened 2024-07-25
Add nvidia-l4 gpu accelerator

eapolinario:
The suffix `-vws` in the definition of the L4 gpu accelerator causes some confusion. According to google cloud [docs](https://cloud.google.com/compute/docs/gpus#l4-gpus), we can specify either `nvidia-l4` or `nvidia-l4-vws` to use L4 GPUs.

Rename the existing `L4` constant to `L4_VWS` just to be safe and define a new `L4` gpu accelerator with the value set to `nvidia-l4`.

### DIFF
--- a/flytekit/extras/accelerators.py
+++ b/flytekit/extras/accelerators.py
@@ -133,7 +133,11 @@ A10G = GPUAccelerator("nvidia-a10g")
 
 #: use this constant to specify that the task should run on an
 #: `NVIDIA L4 Tensor Core GPU <https://www.nvidia.com/en-us/data-center/l4/>`_
-L4 = GPUAccelerator("nvidia-l4-vws")
+L4 = GPUAccelerator("nvidia-l4")
+
+#: use this constant to specify that the task should run on an
+#: `NVIDIA L4 Tensor Core GPU <https://www.nvidia.com/en-us/data-center/l4/>`_
+L4_VWS = GPUAccelerator("nvidia-l4-vws")
 
 #: use this constant to specify that the task should run on an
 #: `NVIDIA Tesla K80 GPU <https://www.nvidia.com/en-gb/data-center/tesla-k80/>`_


mhotan (approved):

flytekit/extras/accelerators.py
```diff
@@ -133,7 +133,11 @@ def to_flyte_idl(self) -> tasks_pb2.GPUAccelerator:
 
 #: use this constant to specify that the task should run on an
 #: `NVIDIA L4 Tensor Core GPU <https://www.nvidia.com/en-us/data-center/l4/>`_
-L4 = GPUAccelerator("nvidia-l4-vws")
+L4 = GPUAccelerator("nvidia-l4")
```
thomasjpfan, line 136: @mhotan Do you think changing the id can be BC breaking?

GCP is okay with both strings, but is this the same for AWS?

eapolinario: AFAIK, we've never tested this gpu accelerator code in aws (we couldn't get those gpus at the time). Looking at the code this should work, but it requires verification.

eapolinario:
@thomasjpfan , I also noticed that this change is going to break this `unionai` test: https://github.com/unionai/unionai/blob/main/tests/unit/actor/test_actor.py#L119

flytekit/extras/accelerators.py
samhita-alla: @eapolinario i've used `node_selector={"k8s.amazonaws.com/accelerator": "nvidia-tesla-l4"}` to get l4 to work on aws. we've l4 enabled on the demo hosted tenant.

katrogan (approved):

thomasjpfan (approved):

thomasjpfan:
>  I also noticed that this change is going to break this unionai test:

@eapolinario I'll adjust the unit tests after this is merged.

PR #2609, opened 2024-07-25
Fix source code link

pingsutw:
## Tracking issue
NA

We only show the repo link in the UI, which is not helpful. It would be great if we could display the link to the file.

- Updated the source link (`<github_repo_link>` => `<github_repo_link>/blob/<git_sha>/<file_path>`)
- Only show the link if the file has been pushed

## How was this patch tested?
```
pyflyte run ...
```

```python
from datetime import datetime

def t1(x: int) -> datetime:
    print(x)
    return datetime.now()


@workflow
def wf():
    """
    Dummy workflow for functional test.
    """
    t1(x=10)
```

### Screenshots
Screenshot 2024-08-30 at 12 24 15 PM

## Related PRs
NA

## Docs link
NA


flytekit/remote/remote.py
```diff
@@ -165,6 +166,12 @@ def _get_git_repo_url(source_path: str):
             .rstrip()
             .decode("utf-8")
         )
+        git_branch = (
```
thomasjpfan, line 169: If you are using a detached head, this is empty.

pingsutw:
@thomasjpfan @cosmicBboy, I'd like to get your thoughts here. Showing the source code link on the UI is quite useful, but I found a UX issue.

currently, the link is <github_repo_link>/blob/<git_sha>/<file_path>

### Problems:
The link doesn't work if users haven't pushed the code yet. (I think this is a common use case. People always register and run the task first, and then push their code.)

And the link is immutable, so they have to register a new version of the task/workflow again to see the link after they push their code.

### Proposed solution

If users haven't pushed the code yet

1. Should we just not show the link?
2. or show the link of the current or master branch. (`<github_repo_link>/blob/master/<file_path>`)
3. not show the link, but show a warning message while registering
4. only show the link of the git repo. like https://github.com/unionai/unionai-examples
5. Instead of showing the link, display the entire code on the UI. (maybe too much) 

Screenshot 2024-08-30 at 12 24 15 PM

thomasjpfan:
> The link doesn't work if users haven't pushed the code yet. (I think this is a common use case. People always register and run the task first, and then push their code.)

- During registration, check if the URL 404s. If it exists, then include the link.
- If it does not exists, show the remote URL and state that code was not committed. If we want to be clever, show the URL to the branch but still state that the code as not commited.

I rather not warn since this is a common use case and warning is too noisy.

I would love for a way to link to the code in the UI. Maybe another VSCode integration. In this case it's "explore your files".

---

As for the UI, I would show a link and not a full URL. Also each version of the workflow should have it's own link, because the code could be different.

flytekit/core/interface.py
```diff
@@ -573,3 +574,30 @@ def remap_shared_output_descriptions(output_descriptions: Dict[str, str], output
...
+        return doc
+
+    if docstring.short_description:
+        doc.short_description = docstring.short_description
+    if docstring.long_description:
+        if sys.getsizeof(docstring.long_description) > 16 * 1024 * 1024:
+            # Check if the size of long description exceeds 16KB
+            raise ValueError(
+                "Long Description of the flyte entity exceeds the 16KB size limit. Please specify the uri in the long description instead."
+            )
```
thomasjpfan, line 601: Erroring on a long docstring does not feel great. I'll rather slice the docstring with `textwrap.shorten` and add some context saying that "the docstring was too long"

Since this behavior is already in `master`, I am okay with leaving this as is in this PR.

flytekit/tools/translator.py
```diff
@@ -820,21 +819,73 @@ def get_serializable(
...
+    :param settings: the serialization settings
+    :return: the git source code link
+    """
+    if not settings.git_repo:
+        return None
+
+    from flytekit.remote.remote import _get_git_root
+
+    if _is_file_pushed(settings.source_root, module):
+        return settings.git_repo + module.removeprefix(_get_git_root(settings.source_root))
```
thomasjpfan, line 849: `removeprefix` is Python >= 3.9. (Our minimum support is 3.8)

flytekit/tools/translator.py
```diff
@@ -820,21 +819,73 @@ def get_serializable(
...
+        and entity.module_file
+    ):
+        # Extract the repo URL from the git config, and assign it to the link of the source code of the description entity
+        entity.docs.source_code = SourceCode(link=_get_git_link(str(entity.module_file), settings))
+    # This needs to be at the bottom, not the top - i.e., dependent tasks get added before the workflow containing it
     entity_mapping[entity] = cp_entity
     return cp_entity
 
 
+def _get_git_link(module: str, settings: SerializationSettings) -> Optional[str]:
```
thomasjpfan, line 834: Should this be filename?

```suggestion
def _get_git_link(filename: str, settings: SerializationSettings) -> Optional[str]:
```

flytekit/tools/translator.py
```diff
@@ -820,21 +819,73 @@ def get_serializable(
...
-                    )
-            entity.docs.source_code = SourceCode(link=settings.git_repo)
-    # This needs to be at the bottom not the top - i.e. dependent tasks get added before the workflow containing it
+    if (
+        isinstance(entity, (PythonTask, WorkflowBase))
+        and not isinstance(entity, ReferenceEntity)
+        and entity.module_file
+    ):
+        # Extract the repo URL from the git config, and assign it to the link of the source code of the description entity
+        entity.docs.source_code = SourceCode(link=_get_git_link(str(entity.module_file), settings))
```
thomasjpfan, line 828: `module_file` is already a string

```suggestion
        entity.docs.source_code = SourceCode(link=_get_git_link(entity.module_file, settings))
```

flytekit/tools/translator.py
```diff
@@ -820,21 +819,73 @@ def get_serializable(
...
-                if sys.getsizeof(entity.docs.long_description.value) > 16 * 1024 * 1024:
-                    raise ValueError(
-                        "Long Description of the flyte entity exceeds the 16KB size limit. Please specify the uri in the long description instead."
-                    )
-            entity.docs.source_code = SourceCode(link=settings.git_repo)
-    # This needs to be at the bottom not the top - i.e. dependent tasks get added before the workflow containing it
+    if (
+        isinstance(entity, (PythonTask, WorkflowBase))
+        and not isinstance(entity, ReferenceEntity)
+        and entity.module_file
```
thomasjpfan, line 825: For the edge case where `module_file=""` (I do not think it happens, but to be safe)

```suggestion
        and entity.module_file is not None
```

flytekit/tools/translator.py
eapolinario: we just dropped support for python 3.8, so this is no longer an issue.

eapolinario:

PR #2600, opened 2024-07-22
JSON IDL

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5318

## What changes were proposed in this pull request?
- [x] type transformer tries to convert to `JSON` first, then `Struct`
- [x] backward compatibility
- [x] attribute access for local execution

## How was this patch tested?
local execution, remote execution and unit tests.
Note: pydantic Transformer will be supported in the follow-up PR.

<details>

<summary>Tips for collapsed sections</summary>

### dataclass transformer
```python
from dataclasses import dataclass, fields
from typing import Dict
from flytekit.types.file import FlyteFile

from flytekit import task, workflow, ImageSpec

flytekit_hash = "a964d7d5461fd79f3f8442c433825bcba830f32d"
flyteidl_hash = "0b32845e92cfe354c197ddf5ae3c8605a17f079b"

json_idl = f"git+https://github.com/flyteorg/flyte.git@{flyteidl_hash}#subdirectory=flyteidl"
flytekit = f"git+https://github.com/flyteorg/flytekit.git@{flytekit_hash}"

image = ImageSpec(
    packages=[json_idl, flytekit],
    apt_packages=["git"],
    registry="futureoutlier",
)

@dataclass
class DC_inside:
    a: int
    b: float
    # ff: FlyteFile

@dataclass
class DC:
    a: int
    b: float
    c: str
    d: Dict[str, int]
    e: DC_inside


@task(container_image=image)
def dataclass_task(input: DC) -> DC:
    def print_dataclass_info(instance, indent=0):
        for field in fields(instance):
            value = getattr(instance, field.name)
            print(f"{' ' * indent}Field: {field.name}, Type: {field.type}, Value: {value}")
            if isinstance(value, dict):
                for k, v in value.items():
                    print(f"{' ' * (indent + 4)} Type: {type(k)}, Key: {k}")
                    print(f"{' ' * (indent + 4)} Type: {type(v)}, Value: {v}")

    print("DC fields:")
    print_dataclass_info(input)

    return input

@task(container_image=image)
def dataclass_attr_dict_task(a: float, d: Dict[str, int]):
    print("DC Attr:")
    print("a:", type(a), a)
    print("d:", type(d), d)
    for k, v in d.items():
        print(type(k), k)
        print(type(v), v)

@task(container_image=image)
def dataclass_attr_dc_task(dc: DC_inside):
    print("DC Attr:")
    print("DC Attr:")
[...]

PR #2614, opened 2024-07-26
Allowing dataclasses which require list protobuf deserialization

JackUrb:
## Tracking issue
closes flyteorg/flyte/issues/5593
closes flyteorg/flyte/issues/4581

As described in flyteorg/flyte/issues/5593, when handling the protobuf serialization/deserialization for a dataclass, attributes that are lists end up being deserialized to lists rather than dicts (despite the function being `_MessageToDict`).

The primary change is to the `ProtobufTransformer`'s `to_literal` method. If the `_MessageToDict` method returns a `list` instead of a `dict`, rather than try to construct a protobuf `Struct` (which can't be done for a list), we use a separate `_handle_list_literal` method, which ultimately treats the processing more similarly to the `ListTransformer`.

An additional change is made to the `assert_type` method to avoid an issue where the `isinstance` clause was unnecessarily evaluated (due to the logic ordering) in cases where the type `t` could be a generic (for which the assertion error `isinstance() argument 2 cannot be a parameterized generic` would be thrown). Swapping the logic order means that if `get_origin` is not None (which occurs in parameterized generics), the `isinstance` will not be called.

## How was this patch tested?

Adds a new test to the `TypeEngine` testing, `test_DataclassTransformer_for_list_attributes`.

We've also been running this change [locally](https://github.com/datologyai/flytekit/pull/1/files) for some time, and it has been stable.


flytekit/core/type_engine.py
```diff
@@ -706,10 +706,24 @@ def tag(expected_python_type: Type[T]) -> str:
...
     def to_literal(self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType) -> Literal:
         struct = Struct()
         try:
-            struct.update(_MessageToDict(cast(Message, python_val)))
+            message_dict = _MessageToDict(cast(Message, python_val))
+            try:
+                struct.update(message_dict)
+            except Exception:
+                if isinstance(message_dict, list):
+                    return self._handle_list_literal(ctx, message_dict)
```
pingsutw, line 725: nit: could we do this?
```python
if isinstance(message_dict, list):
    return self._handle_list_literal(ctx, message_dict)
else:
    struct.update(message_dict)
```

pingsutw (commented):
thanks. overall lgtm. Some tests are failing in python 3.9. mind taking a look

flytekit/core/type_engine.py
JackUrb: Seems like yes, will update

JackUrb:
@pingsutw I'm having a hard time telling quite why the test is failing in the narrow case of py3.9 tensorflow file conversion. Is there any retry logic somewhere else in flyte that effectively checks for successful conversion, and then does something else on failure? It's hard to navigate why this conversion is this specific case breaks, and the `FlyteFile` `_SpecificFormatClass` portion is pretty hard to understand.

JackUrb:
Note, it's possible this is entirely superseded by #2600 - happy to assist on that effort if extra hands would help.

Future-Outlier:
> Note, it's possible this is entirely superseded by #2600 - @Future-Outlier happy to assist on that effort if extra hands would help.

@JackUrb 
I just back from the military service!
I'm in charge of JSON IDL, do you want to collaborate with me? or help review my PR?
I can hop on a call with you.

JackUrb:
Indeed, this entire issue is resolved by #5742. Very excited for v1.14.0 to come out! Thanks @Future-Outlier

PR #2579, opened 2024-07-12
[wip] Jupyter notebooks support - run tasks on remote cluster!

kumare3:
## Why are the changes needed?
I would love to develop tasks and workflows in notebooks and run it from the notebook

## What changes were proposed in this pull request?
Support for task execution from notebook to remote cluster for
- [x] Regular tasks
- [x] Workflows
- [x] Map tasks (needs backend change)
- [ ] Dynamic workflows (cannot be supported - reason pickled objects create a problem in the backend)
- [ ] Eager workflows (local entry point should already work)

## How was this patch tested?
https://www.loom.com/share/1369e80571e8427fbfc8e2c6ad127d67?sid=658c01ce-da5c-4bff-b50d-57c61107fba6

##TODO##
- [ ] unit tests
- [ ] integration tests
- [ ]test with various plugins (help needed)
- [ ]Support for dynamic, eager and map
- [ ]Able to use python files and import from these and intermingle with new tasks authored in jupyter
```python
from mycompany.team_a.module_b import taskA

@workflow
def foo():
   taskA()
```

Problem when enabling dynamic task
```
      File "/usr/local/lib/python3.9/site-packages/flytekit/exceptions/scopes.py", line 183, in system_entry_point
        return wrapped(*args, **kwargs)
      File "/usr/local/lib/python3.9/site-packages/flytekit/core/base_task.py", line 733, in dispatch_execute
        native_outputs = self.execute(**native_inputs)
      File "/usr/local/lib/python3.9/site-packages/flytekit/core/python_function_task.py", line 206, in execute
        return self.dynamic_execute(self._task_function, **kwargs)
      File "/usr/local/lib/python3.9/site-packages/flytekit/core/python_function_task.py", line 358, in dynamic_execute
        return self.compile_into_workflow(ctx, task_function, **kwargs)
      File "/usr/local/lib/python3.9/site-packages/flytekit/core/python_function_task.py", line 240, in compile_into_workflow
        self._create_and_cache_dynamic_workflow()
      File "/usr/local/lib/python3.9/site-packages/flytekit/core/python_function_task.py", line 214, in _create_and_cache_dynamic_workflow
        self._wf = PythonFunctionWorkflow(self._task_function, metadata=workflow_meta, default_metadata=defaults)
      File "/usr/local/lib/python3.9/site-packages/flytekit/core/tracker.py", line 81, in __call__
        o = super(InstanceTrackingMeta, cls).__call__(*args, **kwargs)
[...]

PR #2618, opened 2024-07-27
Interface for remote execution in Jupyter Notebook

Mecoli1219:
## Tracking issue or PR
https://github.com/flyteorg/flytekit/pull/2579

Enable this usage:
```
init_remote(Config.auto(), default_project="flytesnacks", default_domain="development")

@task
def hello(name: str) -> str:
    return f"Hello {name}"

@task
def world(pre: str) -> str:
    return f"{pre} Welcome to the world!"

@workflow
def wf(name: str) -> str:
    return world(pre=hello(name=name))

future = wf.remote(name="Flyte says...")
out = future.get()
```

- [x]  a `remote()` function for workflow and task.
- [x] a `init_remote()` function to store the remote configuration.
- [x] `FlyteFuture` class to store the remote execution data.
- [x] a new integration unit-test testing the Jupyter Notebook environment.
- [x] Default option, currently set as another variable, should merge into FlyteRemote in the future
- [x] Doc
- [x] get_deck
- [x] Add translator unit-test if `interactive_mode_enabled` => check pickle & no dest-dir
- [x] update unit-test to remove version
- [x] Remove fast_register_file_uploader

## How was this patch tested?

All the tests were tested locally.
1. I created a `ipynb` file similar to the one provided in https://github.com/flyteorg/flytekit/pull/2579 (https://www.loom.com/share/1369e80571e8427fbfc8e2c6ad127d67?sid=658c01ce-da5c-4bff-b50d-57c61107fba6) with the new interface.
    - [My Demo](https://www.loom.com/share/d49d7498e5f946e2878159a0b15d40bd?sid=ced19859-615f-4aa6-9dda-8f2536695565)
2. I tested the PyTorch example in [Flytesnack](https://github.com/flyteorg/flytesnacks/blob/master/examples/kfpytorch_plugin/kfpytorch_plugin/pytorch_mnist.py).
3. I tested the new integration unit test with the docker image built from `Dockerfile.dev`.

## Screenshots
1. `get_deck()`
image

## Related PRs
https://github.com/flyteorg/flytekit/pull/2579


flytekit/core/base_task.py
```diff
@@ -791,6 +792,15 @@ def execute(self, **kwargs) -> Any:
         """
         pass
 
+    def remote(self, version=None, **kwargs) -> FlyteFuture:
+        """
+        This method will be invoked to execute the task remotely. This will return a FlyteFuture object that can be
+        used to track the progress of the task execution.
+
+        This method should be executed after specifying the remote configuration via `flytekit.remote.init_remote()`.
```
kumare3, line 800: Let's write docs for all arguments and outputs.
Especially because we are mirroring Flytekit.remote, but folks will just start using this interface.

flytekit/core/future.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import TYPE_CHECKING
+
+import cloudpickle
+
+if TYPE_CHECKING:
+    from flytekit.core.base_task import PythonTask
+    from flytekit.core.workflow import WorkflowBase
+
+
+class FlyteFuture:
```
kumare3, line 19: This is really cool, I am thinking maybe we can also use the `init_remote` to set some parts of the Context that allows us to not rely on `ipython_check` of when to pickle. For example, we could simply set a new variable like `interactive_mode_enabled=True` and then this will allow the pickling?

flytekit/remote/init_remote.py
```diff
@@ -0,0 +1,32 @@
...
+    """
+    Initializes the global remote client. This is required before executing any tasks or workflows remotely via `.remote()` method.
+
+    :param default_project: default project to use when fetching or executing flyte entities.
+    :param default_domain: default domain to use when fetching or executing flyte entities.
+    :param data_upload_location: this is where all the default data will be uploaded when providing inputs.
+        The default location - `s3://my-s3-bucket/data` works for sandbox/demo environment. Please override this for non-sandbox cases.
+    :return:
+    """
+    global REMOTE_ENTRY
```
kumare3, line 25: Let put this in a mutex and make it thread safe

flytekit/remote/init_remote.py
```diff
@@ -0,0 +1,32 @@
...
+
+REMOTE_ENTRY = None
+
+
+def init_remote(
+    config: Config,
+    default_project: typing.Optional[str] = None,
+    default_domain: typing.Optional[str] = None,
+    data_upload_location: str = "flyte://my-s3-bucket/",
+    **kwargs,
```
kumare3, line 14: I think we should also allow folks to set Options (which have service account etc)

flytekit/tools/translator.py
```diff
@@ -163,6 +191,64 @@ def fn(settings: SerializationSettings) -> List[str]:
...
+) -> SerializationSettings:
+    # If the entity is not a PythonAutoContainerTask, we don't need to do anything, as only Tasks with container |
+    # user code in container need to be serialized as pickled objects.
+    if not isinstance(entity, (PythonAutoContainerTask, ArrayNodeMapTask)):
+        return serialization_settings
+
+    from flytekit.tools.interactive import ipython_check
+
+    # Let's check if we are in an interactive environment like Jupyter notebook
+    if ipython_check():
```
kumare3, line 207: This is what I was referring to, if instead of this we can check the settings from a Context.
the context could be explicitly set or we could set it based on ipython availability

kumare3:
OHH WOW this is amazing stuff. I am really excited about this. The only problem I see is if we simply convert this into a script it will fail, but we can solve that problem as a separate PR.

flytekit/remote/init_remote.py
Mecoli1219: I think that `Options` is used for each execution, while `init_function()` is aimed at setting up the `FlyteRemote` object. Should we enable customizing it in the .remote() function instead?

Another option is to set a default Options for `.remote()`. However, I’m unsure if we should also modify the constructor of FlyteRemote for this functionality since it might also be useful in the normal usage of FlyteRemote.

flytekit/remote/init_remote.py
Mecoli1219: Updated!

flytekit/remote/init_remote.py
kumare3: i think its ok to modify and set default options. otherwise it will be hard to use with service account. We can do this as a separate PR

flytekit/remote/init_remote.py
```diff
@@ -0,0 +1,37 @@
...
+    Initializes the global remote client. This is required before executing any tasks or workflows remotely via `.remote()` method.
+
+    :param default_project: default project to use when fetching or executing flyte entities.
+    :param default_domain: default domain to use when fetching or executing flyte entities.
+    :param data_upload_location: this is where all the default data will be uploaded when providing inputs.
+        The default location - `s3://my-s3-bucket/data` works for sandbox/demo environment. Please override this for non-sandbox cases.
+    :return:
+    """
+    global REMOTE_ENTRY
+    with REMOTE_ENTRY_LOCK:
```
kumare3, line 30: If it is initialized already we should raise an exception sayin init should be called only once?

kumare3:
@Mecoli1219 this is great, how about few more unit tests for translator, tracker and entrypoint.
I would love to have unit tests for entrypoint specifically. Otherwise it will be subtly break.

Also if you could test if this thing works with some large scale examples, like maybe when we have some state in jupyter notebook etc. Basically a few more real world tests. Also if ray/spark work?

flytekit/remote/init_remote.py
Mecoli1219: Roger that

flytekit/remote/init_remote.py
Mecoli1219: I'll restrict it to be called once only but I don't think multiple initializations are a problem here. 
Instead, if we restrict this function to be called once and the user wants to rerun the Jupyter Notebook, the cell with `init_remote()` should be skipped or it will throw an error. Nevertheless, I don't think it is a big problem to skip it either.

flytekit/core/base_task.py
pingsutw: Will it raise an exception if the remote is not initialized?

flytekit/bin/entrypoint.py
```diff
@@ -198,6 +199,17 @@ def get_one_of(*args) -> str:
...
 
+def _fast_register_file_uploader(file_access: FileAccessProvider, dest: str, src: str):
+    """
+    This function can be used during dynamic task execution, where the downstream task needs to be pickled and uploaded
+
+    :param file_access: The file access provider
+    :param src: The source file to upload
+    """
+    # return md5_bytes and path
+    return "", file_access.put_data(src, dest)
```
pingsutw, line 210: if it always returns the `""`, why do we need it here?

Mecoli1219:
@kumare3 Default options and `interactive_mode_enabled` kwargs are added, and all the existing and new unit tests pass. Could you take a look at it? Thanks!
BTW, we can have a 1-on-1 chat if needed

pingsutw (commented):
Could we move the log (Jupyter notebook..) to `init_remote`?
Also, I think we can remove `Uploading Pickled representation of Task to remote storage...
`

Screenshot 2024-08-05 at 12 44 53 PM

flytekit/bin/entrypoint.py
```diff
@@ -206,6 +218,7 @@ def setup_execution(
     prev_checkpoint: Optional[str] = None,
     dynamic_addl_distro: Optional[str] = None,
     dynamic_dest_dir: Optional[str] = None,
+    pickled: bool = False,
```
pingsutw, line 221: update the docstring below.
`:param pickled ...`

flytekit/bin/entrypoint.py
```diff
@@ -400,6 +428,8 @@ def _execute_map_task(
...
+    pkl_file: Optional[str] = None,
```
pingsutw, line 432: ditto, docstring

flytekit/core/future.py
```diff
@@ -0,0 +1,78 @@
...
+        self._version = version
+        self._exe = self._remote_entry.execute(entity, version=version, inputs=kwargs, options=options)
+
+    def wait(
+        self,
+        timeout: typing.Optional[timedelta] = None,
+        poll_interval: typing.Optional[timedelta] = None,
+        sync_nodes: bool = True,
+    ) -> FlyteWorkflowExecution:
+        return self._remote_entry.wait(
```
pingsutw, line 65: Add some comments here

flytekit/core/future.py
```diff
@@ -0,0 +1,78 @@
...
+            self._exe,
+            timeout=timeout,
+            poll_interval=poll_interval,
+            sync_nodes=sync_nodes,
+        )
+
+    @property
+    def version(self) -> str:
+        return self._version
```
pingsutw, line 74: comments

flytekit/core/future.py
```diff
@@ -0,0 +1,78 @@
...
+
+    @property
+    def exe(self) -> FlyteWorkflowExecution:
```
pingsutw, line 77: comments

flytekit/core/tracker.py
```diff
@@ -337,6 +337,10 @@ def extract_task_module(f: Union[Callable, TrackedInstance]) -> Tuple[str, str,
 
 
 def get_full_module_path(mod: ModuleType, mod_name: str) -> str:
+    from flytekit.tools.interactive import ipython_check
+
+    if ipython_check():
+        return ""
```
pingsutw, line 343: Do we really need it here? I feel like we can remove it. We already pickle the task entity, so `get_full_module_path` should not be called.

flytekit/remote/remote.py
```diff
@@ -755,6 +763,12 @@ async def _serialize_and_register(
             )
         if serialization_settings.version is None:
             serialization_settings.version = version
+        serialization_settings.interactive_mode_enabled = self.interactive_mode_enabled
+
+        if options is None:
+            options = Options()
```
pingsutw, line 769: nit 
```suggestion
        options = options or Options()
```

flytekit/remote/remote.py
```diff
@@ -755,6 +763,12 @@ async def _serialize_and_register(
...
+        if options.file_uploader is None:
+            options.file_uploader = self.upload_file
```
pingsutw, line 771: nit
```suggestion
        options.file_uploader = options.file_uploader or self.upload_file
```

flytekit/remote/remote.py
```diff
@@ -803,9 +817,18 @@ def register_task(
                 domain=self.default_domain,
             )
 
-        ident = asyncio.run(
-            self._serialize_and_register(entity=entity, settings=serialization_settings, version=version)
-        )
+        try:
+            import nest_asyncio
```
pingsutw, line 821: This code is the same as line 869~879; could we add a common function for it?

flytekit/remote/remote.py
```diff
@@ -1734,19 +1765,23 @@ def execute_local_task(
...
-                image_config=image_config,
+                image_config=image_config or ImageConfig.auto_default_image(),
                 project=project or self.default_project,
                 domain=domain or self._default_domain,
                 version=version,
             )
-            flyte_task: FlyteTask = self.register_task(entity, ss)
+            try:
+                flyte_task: FlyteTask = self.register_task(entity, ss)
+            except Exception as e:
```
pingsutw, line 1783: hmm. Why do we need try except here?

flytekit/tools/translator.py
```diff
@@ -165,6 +193,57 @@ def fn(settings: SerializationSettings) -> List[str]:
...
+        from flytekit.configuration import FastSerializationSettings
+
+        if options is None or options.file_uploader is None:
+            raise FlyteAssertion("To work interactively with Flyte, a code transporter/uploader should be configured.")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dest = pathlib.Path(tmp_dir, "pkl.gz")
+            with gzip.GzipFile(filename=dest, mode="wb", mtime=0) as gzipped:
+                cloudpickle.dump(entity, gzipped)
+            rich.get_console().print("[yellow]Uploading Pickled representation of Task to remote storage...[/ yellow]")
```
pingsutw, line 239: could we use `logger.debug` for it? I don't think it's necessary to show it to users

thomasjpfan:
Done in https://github.com/flyteorg/flytekit/pull/2733. We can open a new PR if we want to revisit the interface.

PR #2620, opened 2024-07-29
[Artifacts/Elastic] Skip partitions

wild-endeavor:
This relates to an issue we found when using the artifact class with Elastic while also specifying partitions.

As far as I can tell, what's happening is that the kfpytorch plugin [serializes](https://github.com/flyteorg/flytekit/blob/5bc5d5c0db193fdeea1bcaa2af6ee1af2316499c/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L408), using cloud pickle, and then [deserializes](https://github.com/flyteorg/flytekit/blob/5bc5d5c0db193fdeea1bcaa2af6ee1af2316499c/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L276) within the worker.  Because of how cloudpickle works, this was causing an infinite recursion in the `__getattr__` call of the partitions object.  Basically the getattr function is called only if the field is missing from the object, but cloudpickle doesn't run the init function for the object, which leads to an infinite recursion.  (The added unit test fails with infinite recursion if the code change to getattr is removed.)  (See more background [here.](https://nedbatchelder.com/blog/201010/surprising_getattr_recursion.html))

```
  File "/Users/ytong/go/src/github.com/flyteorg/flytekit/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py", line 276, in spawn_helper
    fn = cloudpickle.loads(fn)
         ^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ytong/go/src/github.com/flyteorg/flytekit/flytekit/core/artifact.py", line 321, in __getattr__
    if self.partitions and item in self.partitions:
       ^^^^^^^^^^^^^^^
  File "/Users/ytong/go/src/github.com/flyteorg/flytekit/flytekit/core/artifact.py", line 312, in partitions
    return self._partitions
           ^^^^^^^^^^^^^^^^
  File "/Users/ytong/go/src/github.com/flyteorg/flytekit/flytekit/core/artifact.py", line 321, in __getattr__
    if self.partitions and item in self.partitions:
       ^^^^^^^^^^^^^^^
```

This was hard to find because the exception was getting swallowed by Elastic somewhere.  This stack trace was taken from a file which looks something like `/var/folders/jx/54tww2ls58n8qtlp9k31nbd80000gp/T/torchelastic_wg0jjz6j/local_r5pr8f9n/attempt_0/0/error.json`.  The behavior of the pytorch plugin is such that this just hangs (at least until some timeout), rather than failing even though there is an error.json.

## How was this patch tested?
Tested locally.

### DIFF
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -447,12 +447,10 @@ class PytorchElasticFunctionTask(PythonFunctionTask[Elastic]):
         from torch.distributed.elastic.multiprocessing.errors import ChildFailedError
 
         try:
-            print("Here+++++++++++++++++++++++++ 1")
             out = elastic_launch(
                 config=config,
                 entrypoint=launcher_target_func,
             )(*launcher_args)
-            print("Here+++++++++++++++++++++++++ 2")
         except ChildFailedError as e:
             _, first_failure = e.get_first_failure()
             if is_recoverable_worker_error(first_failure):
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -241,7 +241,7 @@ class ElasticWorkerResult(NamedTuple):
 
     return_value: Any
     decks: List[flytekit.Deck]
-    om: Optional[OutputMetadata]
+    om: Optional[OutputMetadata] = None
 
 
 def spawn_helper(
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
@@ -241,7 +241,7 @@ class ElasticWorkerResult(NamedTuple):
 
     return_value: Any
     decks: List[flytekit.Deck]
-    om: OutputMetadata
+    om: Optional[OutputMetadata]
 
 
 def spawn_helper(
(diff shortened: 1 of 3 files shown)


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -241,7 +241,7 @@ class ElasticWorkerResult(NamedTuple):
 
     return_value: Any
     decks: List[flytekit.Deck]
-    om: OutputMetadata
+    om: Optional[OutputMetadata]
```
pingsutw, line 244: nit
```suggestion
    om: Optional[OutputMetadata] = None
```

pingsutw (approved):

samhita-alla:
@fg91 do you know why the plugin doesn't fail immediately when there's an error.json file?

> This was hard to find because the exception was getting swallowed by Elastic somewhere. This stack trace was taken from a file which looks something like /var/folders/jx/54tww2ls58n8qtlp9k31nbd80000gp/T/torchelastic_wg0jjz6j/local_r5pr8f9n/attempt_0/0/error.json. The behavior of the pytorch plugin is such that this just hangs (at least until some timeout), rather than failing even though there is an error.json.

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -241,7 +241,7 @@ class ElasticWorkerResult(NamedTuple):
 
     return_value: Any
     decks: List[flytekit.Deck]
-    om: OutputMetadata
+    om: Optional[OutputMetadata] = None
```
fg91, line 245: Nit: Could we please add `om` to `ElasticWorkerResult`'s doc string? 🙇

fg91:
> @fg91 do you know why the plugin doesn't fail immediately when there's an error.json file?

I don't know unfortunately, sorry. But I can confirm that this is not the normal behavior, normally it fails immediately:

> The behavior of the pytorch plugin is such that this just hangs (at least until some timeout), rather than failing even though there is an error.json.

PR #2622, opened 2024-07-30
ci: Add Python 3.9 to build matrix

pingsutw:
## Tracking issue
NA

Some tests are failing in Python 3.8/3.9/3.10

- list[int]
Subscripts in native collections (e.g. `list[int]`) were only enabled in 3.9 via https://peps.python.org/pep-0585. We had some tests in `test_type_engine.py` that were using those, so [skipping](https://github.com/flyteorg/flytekit/pull/2622/commits/096bbaaeaba8238eba4d89c3c114ec774c00747e) them if running in python 3.8.
- is_relative_to
The ability to list Flyte entities via `pyflyte run <path/to/file/containing/flyte/entities.py>` does not work as [it uses](https://github.com/flyteorg/flytekit/blob/e2971f0413acd10f9775407970fbc3c37a49d2c0/flytekit/clis/sdk_in_container/run.py#L344) of `is_relative_to` which is only available on python 3.9+.
- integration
   sandbox has disk pressure, so I removed the github action cache to get more disk space.
- ray plugin
   ray.is_initialized always returns true after we execute the task. we haven't run this test in CI before.
- mlflow plugin
   python dependencies and source code have been added to the default deck, so update the number of decks in the test.
- pytorch plugin
   replace `FlyteContextManager.current_context()` with `flytekit.current_context()` since the former doesn't have `decks` attribute
- test_print_agents_metadata_output
  we have removed `secho` and use rich to print the agent metadata
   

## How was this patch tested?
unit test

## Related PRs
NA

## Docs link
NA


tests/flytekit/unit/extend/test_agent.py
```diff
@@ -399,16 +399,3 @@ def sample_agents():
...
-@patch("flytekit.extend.backend.base_agent.AgentRegistry.list_agents")
-def test_print_agents_metadata_output(list_agents_mock, mock_secho, sample_agents):
-    list_agents_mock.return_value = sample_agents
-    print_agents_metadata()
-    expected_calls = [
-        (("Starting Sensor that supports task categories ['sensor']",), {"fg": "blue"}),
-        (("Starting ChatGPT Agent that supports task categories ['chatgpt']",), {"fg": "blue"}),
-    ]
-    mock_secho.assert_has_calls(expected_calls, any_order=True)
-    assert mock_secho.call_count == len(expected_calls)
```
eapolinario, line 414: why are we removing this test?

.github/workflows/pythonbuild.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           if [[ ${{ github.event_name }} == "schedule" ]]; then
             echo "python_versions=[\"3.8\",\"3.9\",\"3.10\",\"3.11\",\"3.12\"]" >> $GITHUB_ENV
           else
-            echo "python_versions=[\"3.12\"]" >> $GITHUB_ENV
+            echo "python_versions=[\"3.8\", \"3.12\"]" >> $GITHUB_ENV
```
eapolinario, line 31: python 3.8 is hitting EOL in November according to https://devguide.python.org/versions/#supported-versions. What do you think of using 3.9 instead (in preparation for the removal of python 3.8)?

tests/flytekit/unit/extend/test_agent.py
pingsutw: the test is failing because we use rich to print the metadata instead of secho. https://github.com/flyteorg/flytekit/blob/2b49bb3343c8480c388d1c81f340e2f42626e9f2/flytekit/clis/sdk_in_container/serve.py#L105

.github/workflows/pythonbuild.yml
pingsutw: I'll update it later

tests/flytekit/unit/extras/tasks/test_shell.py
```diff
@@ -43,7 +43,7 @@ def test_shell_task_access_to_result():
     t()
 
     assert t.result.returncode == 0
-    assert t.result.output == "Hello World!"  # ShellTask strips carriage returns
+    assert "Hello World!" in t.result.output  # ShellTask strips carriage returns
```
pingsutw, line 46: it returns `'"Hello World!"'` on Windows

.github/workflows/pythonbuild.yml
```diff
@@ -248,6 +250,8 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
+      - name: 'Clear action cache'
+        uses: ./.github/actions/clear-action-cache
```
wild-endeavor, line 254: why did we need this?  can you add a comment?

eapolinario:
@pingsutw , we've been skipping these failing plugins tests in 3.12 for a long time: https://github.com/flyteorg/flytekit/blob/master/.github/workflows/pythonbuild.yml#L390-L403.

plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -76,10 +75,6 @@ def pre_execute(self, user_params: ExecutionParameters) -> ExecutionParameters:
                 "excludes": ["script_mode.tar.gz", "fast*.tar.gz"],
             }
 
-            cfg = self._task_config
```
pingsutw, line 79: cc @fiedlerNr9 mind taking a look at this?

.github/workflows/pythonbuild.yml
pingsutw: updated it, sandbox has disk pressure

.github/workflows/pythonbuild.yml
pingsutw: updated

Future-Outlier (approved):
Lgtm!!!

eapolinario (approved):

plugins/flytekit-ray/flytekitplugins/ray/task.py
fiedlerNr9: LGTM

pingsutw:
Thank you all!

PR #2623, opened 2024-07-30
Feature gate for FlyteMissingReturnValueException

pingsutw:
## Tracking issue
closed https://github.com/flyteorg/flyte/issues/5608

CPython versions <3.10.10 https://github.com/python/cpython/issues/102647 in the inspect module, which causes missing source lines for certain decorated functions.

Check if function has a return value only for python >=3.10 since `inspect` doesn't work correctly in python 3.8/3.9

## How was this patch tested?
unit test

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flytekit/core/interface.py
+++ b/flytekit/core/interface.py
@@ -386,8 +386,8 @@ def transform_function_to_interface(fn: typing.Callable, docstring: Optional[Doc
         ctx.execution_state
         # Only check if the task/workflow has a return statement at compile time locally.
         and ctx.execution_state.mode is None
-        # inspec module does not work correctly with Python <3.10. https://github.com/flyteorg/flyte/issues/5608
-        and sys.version_info >= (3, 10)
+        # inspect module does not work correctly with Python <3.10.10. https://github.com/flyteorg/flyte/issues/5608
+        and sys.version_info >= (3, 10, 10)
         and return_annotation
         and type(None) not in get_args(return_annotation)
         and return_annotation is not type(None)
--- a/flytekit/core/interface.py
+++ b/flytekit/core/interface.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import collections
 import copy
 import inspect
+import sys
 import typing
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


pingsutw:
cc @fellhorn

flytekit/core/interface.py
```diff
@@ -381,10 +382,12 @@ def transform_function_to_interface(fn: typing.Callable, docstring: Optional[Doc
     return_annotation = type_hints.get("return", None)
 
     ctx = FlyteContextManager.current_context()
-    # Only check if the task/workflow has a return statement at compile time locally.
     if (
         ctx.execution_state
+        # Only check if the task/workflow has a return statement at compile time locally.
         and ctx.execution_state.mode is None
+        # inspec module does not work correctly with Python <3.10. https://github.com/flyteorg/flyte/issues/5608
```
fellhorn, line 389:
```suggestion
        # inspect module does not work correctly with Python <3.10. https://github.com/flyteorg/flyte/issues/5608
```

fellhorn (approved):
Thank you!

flytekit/core/interface.py
```diff
@@ -381,10 +382,12 @@ def transform_function_to_interface(fn: typing.Callable, docstring: Optional[Doc
...
+        and sys.version_info >= (3, 10)
```
fg91, line 390: Should we compare the micro level here too?

fg91: Thank you for the quick reaction @pingsutw btw 🙇

fellhorn, line 390:
```suggestion
        and sys.version_info >= (3, 10, 10)
```

flytekit/core/interface.py
pingsutw: good catch, thanks. updated it.

flytekit/core/interface.py
pingsutw: nice, good call. updated it.

eapolinario (approved):
LGTM.

PR #2624, opened 2024-07-31
Richer printing for some artifact objects

wild-endeavor:
Just improving the experience.

Adding  `__rich_repr__` and `_repr_html_` to TimePartition, Partition, and Partitions objects.
Also adding `__rich_repr__` to LiteralType.

## How was this patch tested?
Tested in jupyter locally, along with the corresponding change, it now looks like
image

### DIFF
--- a/flytekit/core/artifact.py
+++ b/flytekit/core/artifact.py
@@ -273,6 +273,24 @@ class TimePartition(object):
         self.reference_artifact: Optional[Artifact] = None
         self.granularity = granularity
 
+    def __rich_repr__(self):
+        if self.value:
+            if isinstance(self.value, art_id.LabelValue):
+                if self.value.HasField("time_value"):
+                    yield "Time Partition", str(self.value.time_value.ToDatetime())
+                elif self.value.HasField("input_binding"):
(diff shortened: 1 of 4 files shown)


flytekit/core/artifact.py
```diff
@@ -273,6 +273,24 @@ def __init__(
         self.reference_artifact: Optional[Artifact] = None
         self.granularity = granularity
 
+    def __rich_repr__(self):
```
thomasjpfan, line 276: I see that this is used by `_repr_html_`, but is there a use case where `__rich_repr__` is used directly by `rich`?

eapolinario: Yeah, in fact this is a protocol that rich [proposes](https://rich.readthedocs.io/en/stable/pretty.html#rich-repr-protocol) and expects that will become more available in the future. From the rich docs:
```
Note that you can add __rich_repr__ methods to third-party libraries without including Rich as a dependency. If Rich is not installed, then nothing will break. Hopefully more third-party libraries will adopt Rich repr methods in the future.
```

eapolinario (approved):

thomasjpfan (approved):

PR #2626, opened 2024-07-31
Remove use of multiprocessing from the OAuth client

rdeaton-freenome:

Multiprocessing in Python is a bit fraught with issues, as it requires that every dependency in the transitive dependency tree of your project supports multiprocessing. We've noticed issues with the way that this code interacts with concurrent task registration that results in trying to open multiple instances of the server, which fails due to listening on the same port.

The multiprocessing usage here didn't actually bring any real benefit, as the intended use was to simply block the requestor's control flow anyways with the use of a queue. This retains the same behavior without the use of any multiprocessing.

## How was this patch tested?

### DIFF
--- a/flytekit/clients/auth/auth_client.py
+++ b/flytekit/clients/auth/auth_client.py
@@ -14,6 +14,7 @@ from dataclasses import dataclass
 from http import HTTPStatus as _StatusCodes
 from multiprocessing import get_context
 from urllib.parse import urlencode as _urlencode
+from queue import Queue
 
 import click
 import requests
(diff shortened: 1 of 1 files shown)


thomasjpfan (approved):
I tested this PR on an OAuth client and it works. LGTM

thomasjpfan:
@rdeaton-freenome There is a linting issue: https://github.com/flyteorg/flytekit/actions/runs/10185219976/job/28174935429?pr=2626

rdeaton-freenome:
Sorry about that linting issue, fixed in the last commit. Thanks for the quick review!

pingsutw (approved):

rdeaton-freenome:
I'll just quickly note I expect we'll have a second PR in adjacent areas shortly. We're finding that there's still concurrency issues with registration with auth enabled, but we're certain multiprocessing was both exacerbating the issues and are aware of a long history of multiprocessing issues with grpcio.

eapolinario (approved):
This is awesome, thank you! This PR solves an outstanding issue we've had on Windows for a long time (https://github.com/flyteorg/flyte/issues/5079 and https://github.com/flyteorg/flyte/issues/4982). I just tested on a windows machine and it worked beautifully.

PR #2635, opened 2024-08-01
Fix race conditions in the Authentication client

rdeaton-freenome:
Related: https://github.com/flyteorg/flytekit/pull/2626

The AuthorizationClient may need to pop up a webbrowser for human interaction to auth, and may do so from multiple threads and multiple grpcio channels. This manifested for us in race conditions during `pyflyte register` because it uses concurrent registration https://github.com/flyteorg/flytekit/blob/1cc2b110806f7cf16cd2cf404e04538b97ac1a7f/flytekit/tools/repo.py#L310-L311

We block all threads and allow a single request through for authorization and cache that result for a short window to make sure that we don't need a separate request per-thread, and avoid the race condition on setting up the local HTTP server for handling the redirect.

## How was this patch tested?

Lots of calls to pyflyte register in a repository with a large number of workflows.

- [x] I updated the documentation accordingly. (docstrings, at least)

### DIFF
--- a/flytekit/clients/auth/auth_client.py
+++ b/flytekit/clients/auth/auth_client.py
@@ -6,6 +6,8 @@ import http.server as _BaseHTTPServer
 import logging
 import os
 import re
+import threading
+import time
 import typing
 import urllib.parse as _urlparse
 import webbrowser
@@ -236,6 +238,9 @@ class AuthorizationClient(metaclass=_SingletonPerEndpoint):
         self._verify = verify
         self._headers = {"content-type": "application/x-www-form-urlencoded"}
         self._session = session or requests.Session()
+        self._lock = threading.Lock()
+        self._cached_credentials = None
+        self._cached_credentials_ts = None
 
         self._request_auth_code_params = {
             "client_id": client_id,  # This must match the Client ID of the OAuth application.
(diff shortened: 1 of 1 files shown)


flytekit/clients/auth/auth_client.py
```diff
@@ -339,25 +344,36 @@ def _request_access_token(self, auth_code) -> Credentials:
...
+        needs to open a port on localhost and may be called from a
+        multithreaded context (e.g. pyflyte register), this call may block
+        multiple threads and return a cached result for up to 60 seconds.
         """
         # In the absence of globally-set token values, initiate the token request flow
-        q = Queue()
+        with self._lock:
+            cache_ttl_s = 60
+            if self._cached_credentials_ts is not None and self._cached_credentials_ts + cache_ttl_s < time.monotonic():
+                self._cached_credentials = None
```
thomasjpfan, line 357: Minor nit:
```suggestion
            # Clear cache if it's been more than 60 seconds since the last check
            cache_ttl_s = 60
            if self._cached_credentials_ts is not None and self._cached_credentials_ts + cache_ttl_s < time.monotonic():
                self._cached_credentials = None

```

thomasjpfan (dismissed):
Otherwise LGTM

eapolinario (approved):

PR #2638, opened 2024-08-01
update markdown renderer to create tables

dansola:

It would be nice to render pandas dataframes as tables in decks using markdown (`df.to_markdown()`). To do this we need to the `table` plugin of `markdown-it-py` enabled.

Enable `table` plugin for `markdown-it-py`.

## How was this patch tested?

```python
data = {
    'id': np.arange(1, 11),
    'name': [f'Item {i}' for i in range(1, 11)],
    'value': np.random.rand(10) * 100,
    'category': np.random.choice(['A', 'B', 'C'], size=10),
    'timestamp': pd.date_range('2023-01-01', periods=10, freq='D')
}
df = pd.DataFrame(data)
md_df = df.to_markdown()
flytekit.current_context().default_deck.append(
        MarkdownRenderer().to_html(md_df)
    )
```

### html before change:
image

### html after change:
image

## Docs link

Link to the markdown-it-py docs: https://markdown-it-py.readthedocs.io/en/latest/using.html#the-parser

### DIFF
--- a/flytekit/deck/renderer.py
+++ b/flytekit/deck/renderer.py
@@ -55,7 +55,8 @@ class MarkdownRenderer:
     """Convert a markdown string to HTML and return HTML as a unicode string."""
 
     def to_html(self, text: str) -> str:
-        return MarkdownIt().render(text)
+        md = MarkdownIt("gfm-like").enable('table')  # enable table plugin
+        return md.render(text)
 
 
 class SourceCodeRenderer:


thomasjpfan (commented):
Can you add a test that renders a small markdown table and asserts that a HTML table tag is in the output? (Something as simple as `<tr>` is rendered.)

PR #2639, opened 2024-08-01
Make configurable execution console url

ddl-rliu:

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -2298,7 +2298,9 @@ class FlyteRemote(object):
         This will automatically determine if this is an execution or an entity and change the type automatically
         """
         if isinstance(entity, (FlyteWorkflowExecution, FlyteNodeExecution, FlyteTaskExecution)):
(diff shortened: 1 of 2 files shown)


flytekit/remote/remote.py
```diff
@@ -2298,7 +2298,9 @@ def generate_console_url(
         This will automatically determine if this is an execution or an entity and change the type automatically
         """
         if isinstance(entity, (FlyteWorkflowExecution, FlyteNodeExecution, FlyteTaskExecution)):
-            return f"{self.generate_console_http_domain()}/console/projects/{entity.id.project}/domains/{entity.id.domain}/executions/{entity.id.name}"  # noqa
+            from flytekit.configuration.plugin import get_plugin
```
ddl-rliu, line 2301: the import is declared here rather than at the file start, to avoid circular import e.g. https://github.com/flyteorg/flytekit/blob/ea6fa0dffcdacc4f6e43c968fa24ba1811d7183f/flytekit/models/security.py#L44

PR #2641, opened 2024-08-01
In `FlyteRemote.upload_file`, pass the file object directly rather than the entire bytes buffer

redartera:

In the `upload_file` function, we read the entire file's content and hold onto it in memory - this is problematic when dealing with large files. Moreover, there is an [issue](https://github.com/psf/requests/issues/2717) where `requests.put` can raise an `OverflowError` exception when dealing with large byte buffers. 

For a demo replicating the issue and demonstrating the fix, please refer to this [comment](https://github.com/flyteorg/flytekit/pull/2634#issuecomment-2264191444). 

Setup: Python 3.9, flytekit 1.12.3 and Ubuntu 22.04

The proposed changes follow the [recommendation](https://docs.python-requests.org/en/latest/user/advanced/#streaming-uploads) from the official requests documentation

Additionally, we add an integration test for `FlyteRemote.upload_file` that is run separately from the rest of our integration tests, using a new `lftransfers` pytest marker. Changes were made to our `Makefile`, our `pythonbuild.yml` as well as the markers section of `pyproject.toml` accordingly.

## Related PRs

[
](https://github.com/flyteorg/flytekit/pull/2634)

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -901,12 +901,9 @@ class FlyteRemote(object):
         extra_headers = self.get_extra_headers_for_protocol(upload_location.native_url)
         extra_headers.update(upload_location.headers)
         encoded_md5 = b64encode(md5_bytes)
-        with open(str(to_upload), "+rb") as local_file:
-            # Move to the end of the file object to get its length
-            local_file.seek(0, os.SEEK_END)
-            content_length = local_file.tell()
-            # Move back to the beginning of the file object to make it ready for reading
-            local_file.seek(0)
+        local_file_path = str(to_upload)
+        content_length = os.stat(local_file_path).st_size
+        with open(local_file_path, "+rb") as local_file:
             headers = {"Content-Length": str(content_length), "Content-MD5": encoded_md5}
             headers.update(extra_headers)
             rsp = requests.put(
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -904,7 +904,7 @@ class FlyteRemote(object):
         with open(str(to_upload), "+rb") as local_file:
             # Move to the end of the file object to get its length
             local_file.seek(0, os.SEEK_END)
-            content_length = len(local_file.tell())
+            content_length = local_file.tell()
             # Move back to the beginning of the file object to make it ready for reading
             local_file.seek(0)
             headers = {"Content-Length": str(content_length), "Content-MD5": encoded_md5}
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -902,13 +902,16 @@ class FlyteRemote(object):
         extra_headers.update(upload_location.headers)
         encoded_md5 = b64encode(md5_bytes)
         with open(str(to_upload), "+rb") as local_file:
-            content = local_file.read()
-            content_length = len(content)
+            # Move to the end of the file object to get its length
+            local_file.seek(0, os.SEEK_END) 
+            content_length = len(local_file.tell()) 
+            # Move back to the beginning of the file object to make it ready for reading
+            local_file.seek(0)
             headers = {"Content-Length": str(content_length), "Content-MD5": encoded_md5}
             headers.update(extra_headers)
             rsp = requests.put(
                 upload_location.signed_url,
-                data=content,
+                data=local_file, # NOTE: We pass the file object directly to stream our upload.
                 headers=headers,
                 verify=False
                 if self._config.platform.insecure_skip_verify is True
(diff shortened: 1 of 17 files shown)


redartera:
any recommendations on how this change might be tested in a more systematic way?

pingsutw:
@redartera Thank you so much for fixing it. we can register a task with large file in the integration test. https://github.com/flyteorg/flytekit/blob/ea6fa0dffcdacc4f6e43c968fa24ba1811d7183f/tests/flytekit/integration/remote/test_remote.py#L31-L49

redartera:
I added a simple integration test which 

> @redartera Thank you so much for fixing it.
> 
> > any recommendations on how this change might be tested in a more systematic way?
> 
> We can register a task with a large file in the integration test.
> 
> https://github.com/flyteorg/flytekit/blob/ea6fa0dffcdacc4f6e43c968fa24ba1811d7183f/tests/flytekit/integration/remote/test_remote.py#L31-L49
> 
> btw, this is awesome. I didn't know `request` supports streaming uploads. It will reduce lots of memory overhead at compile time.

sounds good - I added a test which writes a large file to a module then runs `pyflyte register ...` on that module.

redartera:
> > @redartera Thank you so much for fixing it.
> > > any recommendations on how this change might be tested in a more systematic way?
> > 
> > 
> > We can register a task with a large file in the integration test.
> > https://github.com/flyteorg/flytekit/blob/ea6fa0dffcdacc4f6e43c968fa24ba1811d7183f/tests/flytekit/integration/remote/test_remote.py#L31-L49
> > 
> > btw, this is awesome. I didn't know `request` supports streaming uploads. It will reduce lots of memory overhead at compile time.
> 
> sounds good - I added a test which writes a large file to a module then runs `pyflyte register ...` on that module.

seems like the GH runners are running out of disk space, I reduced the file size to 2 and 3gb while adding the `@pytest.mark.serial` marker to avoid running these tests in parallel and overloading the runner's disk.

pingsutw (dismissed):

tests/flytekit/integration/remote/test_remote.py
```diff
@@ -483,3 +484,34 @@ def test_execute_workflow_with_maptask(register):
         wait=True,
     )
     assert execution.outputs["o0"] == [4, 5, 6]
+
+@pytest.mark.serial
+@pytest.mark.parametrize("gigabytes", [2, 3])
+def test_can_register_with_large_file(gigabytes):
+    with tempfile.TemporaryDirectory(dir=MODULE_PATH) as tempdir:
```
redartera, line 491: @pingsutw I see failures in the CI - maybe we can just test with a single 3Gb file?

```suggestion
def test_can_register_with_large_file():
    gigabytes = 3
    with tempfile.TemporaryDirectory(dir=MODULE_PATH) as tempdir:
```

flytekit/remote/remote.py
```diff
@@ -902,13 +902,16 @@ def upload_file(
         extra_headers.update(upload_location.headers)
         encoded_md5 = b64encode(md5_bytes)
         with open(str(to_upload), "+rb") as local_file:
-            content = local_file.read()
-            content_length = len(content)
+            # Move to the end of the file object to get its length
+            local_file.seek(0, os.SEEK_END)
+            content_length = local_file.tell()
+            # Move back to the beginning of the file object to make it ready for reading
+            local_file.seek(0)
```
eapolinario, line 909: Using [stat](https://en.wikipedia.org/wiki/Stat_(system_call)) is at least one order of magnitude faster than seeking. In python, the size of a file is exposed via https://docs.python.org/3/library/stat.html#stat.ST_SIZE.

eapolinario:
> seems like the GH runners are running out of disk space, I reduced the file size to 2 and 3gb while adding the `@pytest.mark.serial` marker to avoid running these tests in parallel and overloading the runner's disk.

What if we added a separate [pytest marker](https://github.com/flyteorg/flytekit/blob/master/pyproject.toml#L97-L99) to run just this test?

redartera:
> > seems like the GH runners are running out of disk space, I reduced the file size to 2 and 3gb while adding the `@pytest.mark.serial` marker to avoid running these tests in parallel and overloading the runner's disk.
> 
> What if we added a separate [pytest marker](https://github.com/flyteorg/flytekit/blob/master/pyproject.toml#L97-L99) to run just this test?

I like this approach - will find a cleaner way to adopt it

redartera:
@eapolinario @pingsutw I made a few major changes to both the implementation and the integration test:
- I now use `stat` following the recommendation of @eapolinario - works as expected.
- I now use a separate `lftransfers` pytest marker also following the recommendation of @eapolinario. Changes were to the `Makefile`, `pythonbuild.yml` and `pyproject.toml` to allow those tests to be ran separately from the rest of the integration tests
- The test itself has changed, where I now simply test the `upload_file` method, thus making it easier to clean all objects which were created by the test. If i were to register a module and upload 2/3gb files each time, it would be difficult to cleanly delete all artifacts which were generated by each test, and the sandbox volume of users or GH runners would quickly become overloaded if our large files are persisted.

eapolinario (approved):
Thank you!

PR #2644, opened 2024-08-02
Fix Get Agent Secret Using Key

Future-Outlier:
## Tracking issue

If we install union sdk, when using `get_agent_secret`, we will ignore the group, and right now we use `group` to retrieve the agent's secret.

### DIFF
--- a/flytekit/extend/backend/utils.py
+++ b/flytekit/extend/backend/utils.py
@@ -39,7 +39,7 @@ def is_terminal_phase(phase: TaskExecution.Phase) -> bool:
 
 
 def get_agent_secret(secret_key: str) -> str:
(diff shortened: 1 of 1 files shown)


flytekit/extend/backend/utils.py
```diff
@@ -39,7 +39,7 @@ def is_terminal_phase(phase: TaskExecution.Phase) -> bool:
 
 
 def get_agent_secret(secret_key: str) -> str:
-    return flytekit.current_context().secrets.get(secret_key)
+    return flytekit.current_context().secrets.get(key=secret_key)
```
pingsutw, line 42: thanks, good catch. Could we add a small unit test for it?

pingsutw (approved):

PR #2647, opened 2024-08-03
Improve jupyter repr and `__repr__` for Flyte models

thomasjpfan:

Improves jupyter repr for `flytekit.models`.

This PR adds a `_repr_html_` that generically renders any `flytekit.model`. Internally it uses `_repr_idl_yaml_like` that renders any flyteidl as a yaml like repr.

## How was this patch tested?

I added a simple unit test.

#### This PR

Screenshot 2024-08-02 at 10 25 00 PM

#### Current

Screenshot 2024-08-02 at 10 26 28 PM

### DIFF
--- a/flytekit/models/common.py
+++ b/flytekit/models/common.py
@@ -47,8 +47,7 @@ class FlyteType(FlyteABCMeta):
 def _repr_idl_yaml_like(idl, indent=0) -> str:
     """Formats an IDL into a YAML-like representation."""
     if not hasattr(idl, "ListFields"):
-        str_repr = shorten(str(idl).strip(), width=80)
-        return str_repr
+        return str(idl)
 
     with closing(StringIO()) as out:
         for descriptor, field in idl.ListFields():
@@ -60,6 +59,7 @@ def _repr_idl_yaml_like(idl, indent=0) -> str:
                     out.write(" " * indent + descriptor.name + ":" + os.linesep)
                     out.write(_repr_idl_yaml_like(field, indent + 2))
             except AttributeError:
+                # No ListFields -> Must be a scalar
                 str_repr = shorten(str(field).strip(), width=80)
                 if str_repr:
                     out.write(" " * indent + descriptor.name + ": " + str_repr + os.linesep)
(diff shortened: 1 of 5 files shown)


pingsutw (approved):

flytekit/models/common.py
```diff
@@ -40,6 +44,29 @@ def from_flyte_idl(cls, idl_object):
         pass
 
 
+def _repr_idl_yaml_like(idl, indent=0) -> str:
```
kumare3, line 47: Why not make this the default repr

thomasjpfan: I wanted to make it the default, but I saw the current repr strips away the newlines:

https://github.com/flyteorg/flytekit/blob/e9f349910e660f25f5e33e360ea16813099b27d6/flytekit/models/common.py#L63

Was there a reason for making it all one line?

wild-endeavor: i don't think so, not a good one anyways.  feel free to overwrite.  the reason likely was logging platforms (cloudwatch logs, splunk, grafana, etc) don't render logs with new lines well.  but i don't know how much this could matter for python.  that's more of a backend issue.

wild-endeavor (approved):

flytekit/models/common.py
```diff
@@ -60,9 +86,9 @@ def short_string(self):
         """
         :rtype: Text
         """
-        literal_str = re.sub(r"\s+", " ", str(self.to_flyte_idl())).strip()
+        str_repr = _repr_idl_yaml_like(self.to_flyte_idl(), indent=2).rstrip(os.linesep)
         type_str = type(self).__name__
-        return f"[Flyte Serialized object: Type: <{type_str}> Value: <{literal_str}>]"
+        return f"Flyte Serialized object ({type_str}):" + os.linesep + str_repr
```
eapolinario, line 91: Can we add a few more tests to cover the different types of literals (collections, maps, etc)? These are useful especially in flyteremote when debugging.

thomasjpfan: I added more tests in https://github.com/flyteorg/flytekit/pull/2647/commits/21372e4748801a2baf68b8b46e31f01a0875ad56

eapolinario (approved):

PR #2653, opened 2024-08-05
Added warning for command list and shell true

pryce-turner:
Stubbed my toe on some rather arcane behavior of `subprocess.run` so I decided to log a warning under this circumstance. More details here: https://medium.com/python-pandemonium/a-trap-of-shell-true-in-the-subprocess-module-6db7fc66cdfd


pingsutw (approved):

eapolinario (approved):
Wow, it's insane that someone thought that this was a good api. 🤯 

Thanks for adding the warning!

PR #2656, opened 2024-08-06
[fix] Validate workflow input name is lowercase

ddl-rliu:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5640

Click does not support uppercase in option names: https://github.com/pallets/click/issues/837

This causes an unintuitive error in flytekit when uppercase is used in task input names:
```
$ pyflyte run --remote wf.py wf --A 1
Failed with Unknown Exception <class 'flytekit.core.type_engine.TypeTransformerFailedError'>
Reason: Python value cannot be None, expected <class 'int'>/<FlyteLiteral(LiteralType) simple: INTEGER>
```

This PR aims to improve the error message. A long-term fix might be to support uppercase in workflow input names, but this seems to be incompatible with the click.Option implementation. Lowercase+snake_case is a convention that click aims to follow.

Add validation against the workflow input names to ensure it is lowercase.

## How was this patch tested?

New error is improved:
```
$ pyflyte run --remote wf.py wf --A 1
Failed with Unknown Exception <class 'ValueError'> Reason: Workflow input name must be lowercase: 'A'
```

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -379,6 +379,10 @@ def to_click_option(
     This handles converting workflow input types to supported click parameters with callbacks to initialize
     the input values to their expected types.
     """
+    if input_name != input_name.lower():
+        # Click does not support uppercase option names: https://github.com/pallets/click/issues/837
+        raise ValueError(f"Workflow input name must be lowercase: {input_name!r}")
+
     run_level_params: RunLevelParams = ctx.obj
 
     literal_converter = FlyteLiteralConverter(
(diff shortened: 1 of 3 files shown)


pingsutw:
@ddl-rliu mind taking a look the test errors
```bash
FAILED tests/test_onnx_scikitlearn.py::test_onnx_scikitlearn_simple - ValueError: Input name must be lowercase: 'X_test'
```

ddl-rliu:
Just pushed a fix, I'll review the CI checks when they complete.

(The original fix had a mistake where the validation is done on task input names, but the uppercase/lowercase issue is specific to workflow input names, so I've updated it by moving the validation to the `run.py` level)

pingsutw (approved):

PR #1859, opened 2023-09-28
Run notebook task on aws batch

pingsutw:
# TL;DR
Override `loader_args`  in the default task_resolver, and extract task module from the notebook task instead of `config_task_instance`.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
Before:
image

After:
image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3809

## Follow-up issue
_NA_

PR #2660, opened 2024-08-07
Fix overriding of loader_args task resolver in papermill plugin

eapolinario:
In https://github.com/flyteorg/flytekit/pull/1859/ (which was released in flytekitplugins-papermill 1.10.0) we ended up overriding the task resolver's `loader_args` method in order to generate the correct `NotebookTask` container arguments. However we never restored that function to its original value, which has the side effect that in the case of shared task resolvers (e.g. the `default_task_resolver` used for regular tasks) we end up reusing the last `NotebookTask` registered as the task name.

This PR restores the loader_args in papermill plugin's `get_container` and `get_k8s_pod`.


pingsutw (approved):

pingsutw (commented):
lgtm, thank you!

PR #2661, opened 2024-08-07
Make additional execution info configurable via plugin

ddl-rliu:


pingsutw (approved):

flytekit/configuration/plugin.py
```diff
@@ -52,6 +53,12 @@ def get_default_image() -> Optional[str]:
     def get_auth_success_html(endpoint: str) -> Optional[str]:
         """Get default success html for auth. Return None to use flytekit's default success html."""
 
+    @staticmethod
+    def get_additional_info_for_execution(
```
pingsutw, line 57: cc @thomasjpfan maybe we can override this as well?

thomasjpfan: I can not think of something right away.

flytekit/configuration/plugin.py
```diff
@@ -52,6 +53,12 @@ def get_default_image() -> Optional[str]:
...
+        console_http_domain: str, entity: Union[FlyteWorkflowExecution, FlyteNodeExecution, FlyteTaskExecution]
```
thomasjpfan, line 58: @ddl-rliu  What do you think of passing in `FlyteRemote` instead of `console_http_domain`? This will give the CLI plugin more context if they need it.

ddl-rliu: Great, that makes sense to me! Done: https://github.com/flyteorg/flytekit/pull/2661/commits/c4231c2de6ca184ba67082fe87acbe3dfb8c9963

ddl-rliu: @pingsutw (sorry for the ping) I think my commit here caused your approval to expire, could you take another look when you have the time? Thank you!

pingsutw (approved):

thomasjpfan:
I’m concerned with the precedence this PR sets for extending console printing. There are many places in flytekit that prints to the console. Where do we draw the line for console printing extensions?

Overall, I prefer not to have the configuration plugin be used to customize printing.

ddl-rliu:
@thomasjpfan Thanks for the feedback, it's a valid concern. As mentioned in the [issue](https://github.com/flyteorg/flyte/issues/5641), an alternative we considered is to maintain a custom CLI "custom-pyflyte-wrapper-cli" which wraps pyflyte, additionally including the various execution info/console output that we desire.

We have not started this effort, but it could be the next alternative to making the execution info console output pluggable.

PR #2663, opened 2024-08-08
Copy user files that were imported by workflow in pyflyte run

thomasjpfan:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/5343

Currently, `pyflyte run` requires `--copy-all` when the workflow imports files from other directories. This is a common gotcha because `pyflyte run workflow.py` works, but `pyflyte run --remote workflow.py` does not.

With this PR:
- We look at all the modules in `sys.modules` and copy over all files that are necessary for the workflow to run.
- This way `pyflyte run --remote workflow.py` will work out of the box most of the time.
- Less copying than `--copy-all` because it only copies over the required files. If you have a project with tests or docs, it will not be copied.

I think the new coping mechanism does what `copy_module_to_destination` does, but I did not want to adjust it just in case I break anything. I added a new `add_imported_modules_from_source` to additionally copy over files. 

## How was this patch tested?

I added unit tests to make sure the copying works. I also had a project with this structure:

```
.
└── workflows
    ├── __init__.py
    ├── another_module
    │   └── wow.py
    ├── main.py
    └── util.py
```

and ran `pyflyte run --remote workflows/main.py` and everything copies correctly. Simple workflow in the root directory also works:

```
.
├── main.py
└── util.py
```

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -44,7 +44,7 @@ from flytekit.models.types import SimpleType
 from flytekit.remote import FlyteLaunchPlan, FlyteRemote, FlyteTask, FlyteWorkflow, remote_fs
 from flytekit.remote.executions import FlyteWorkflowExecution
 from flytekit.tools import module_loader
-from flytekit.tools.script_mode import _find_project_root, compress_scripts
+from flytekit.tools.script_mode import _find_project_root, compress_scripts, get_all_modules
 from flytekit.tools.translator import Options
 
 
@@ -493,7 +493,8 @@ def _update_flyte_context(params: RunLevelParams) -> FlyteContext.Builder:
     if output_prefix and ctx.file_access.is_remote(output_prefix):
         with tempfile.TemporaryDirectory() as tmp_dir:
             archive_fname = pathlib.Path(os.path.join(tmp_dir, "script_mode.tar.gz"))
-            compress_scripts(params.computed_params.project_root, str(archive_fname), list(sys.modules.values()))
+            modules = get_all_modules(params.computed_params.project_root, params.computed_params.module)
+            compress_scripts(params.computed_params.project_root, str(archive_fname), modules)
             remote_dir = file_access.get_random_remote_directory()
             remote_archive_fname = f"{remote_dir}/script_mode.tar.gz"
             file_access.put_data(str(archive_fname), remote_archive_fname)
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -493,7 +493,7 @@ def _update_flyte_context(params: RunLevelParams) -> FlyteContext.Builder:
     if output_prefix and ctx.file_access.is_remote(output_prefix):
         with tempfile.TemporaryDirectory() as tmp_dir:
             archive_fname = pathlib.Path(os.path.join(tmp_dir, "script_mode.tar.gz"))
-            compress_scripts(params.computed_params.project_root, str(archive_fname), params.computed_params.module)
+            compress_scripts(params.computed_params.project_root, str(archive_fname), list(sys.modules.values()))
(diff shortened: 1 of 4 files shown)


flytekit/tools/script_mode.py
```diff
@@ -46,6 +50,8 @@ def compress_scripts(source_path: str, destination: str, module_name: str):
 
         visited: typing.List[str] = []
         copy_module_to_destination(source_path, destination_path, module_name, visited)
+        add_imported_modules_from_source(source_path, destination_path, list(sys.modules.values()))
```
pingsutw, line 53: If we remove `copy_module_to_destination`, does it work as well? If so, I think we can remove that.

For more context:
`copy_module_to_destination` will only copy the module to remote if and only if that module has a flyte task/workflow.


in this case, pyflyte run --remote uploads `foo_wf` and `main`
```python
└── workflows
    ├── __init__.py
    ├── another_module
    │   └── foo_wf.py
    ├── main.py  # it import foo_wf
```

I did this because I didn't know how to do `add_imported_modules_from_source` at that time. 

Now you have added it, I think we should be able to remove `copy_module_to_destination`

flytekit/tools/script_mode.py
```diff
@@ -127,6 +133,65 @@ def tar_strip_file_attributes(tar_info: tarfile.TarInfo) -> tarfile.TarInfo:
     return tar_info
 
 
+def add_imported_modules_from_source(source_path: str, destination: str, modules: List[ModuleType]):
```
pingsutw, line 136: If `utils` imports another local module, will we copy that to the remote storage as well?
```
.
└── workflows
    ├── __init__.py
    ├── another_module
    │   └── wow.py
    ├── main.py  # only import util.py
    └── util.py # import util2.py
     └── util2.py     
```

thomasjpfan: Any files that is a part of the import chain from the original workflow will be copied over.

Details: If `main.py` imports from `util.py` and `util.py` imports from `util2.py` then `util2.py` is part of `sys.modules` when `main.py` is loaded. With this PR, if the module is in `sys.modules` and in the same working directory as `main.py`, then it will be uploaded.

pingsutw: nice, perfect

flytekit/tools/script_mode.py
wild-endeavor: +1, remove this one line and merge.  tested this against dev but feel free to again.  the only difference i saw is that with the copy_module... call, the main script itself is already in the destination folder, and because it exists it's skipped when add_imported... sees it.  Without it, the main file itself is copied by add_imported...

wild-endeavor (dismissed):

pingsutw (approved):

flytekit/clis/sdk_in_container/run.py
```diff
@@ -493,7 +493,7 @@ def _update_flyte_context(params: RunLevelParams) -> FlyteContext.Builder:
     if output_prefix and ctx.file_access.is_remote(output_prefix):
         with tempfile.TemporaryDirectory() as tmp_dir:
             archive_fname = pathlib.Path(os.path.join(tmp_dir, "script_mode.tar.gz"))
-            compress_scripts(params.computed_params.project_root, str(archive_fname), params.computed_params.module)
+            compress_scripts(params.computed_params.project_root, str(archive_fname), list(sys.modules.values()))
```
thomasjpfan, line 496: @pingsutw I'm concerned with this change. At this point, is the main workflow imported by `flytekit`?

thomasjpfan: I added a new `get_all_modules` function that returns all modules. It adds the `params.computed_params.module` if it is not already imported.

pingsutw: sgtm, thank you!

pingsutw (approved):

PR #2673, opened 2024-08-09
Add pythonpath "." before loading modules

arbaobao:

Sometimes User defined Dockfile is missing "ENV PYTHONPATH /root", it will cause execute errors.

This change will add pythonpath "." before  loading modules

## How was this patch tested?

build a docker image with Dockfile missing "ENV PYTHONPATH /root"

### Screenshots
截圖 2024-08-10 清晨7 38 41


pingsutw (commented):
Could you also remove the PYTHONPATH from this [dockerfile](https://github.com/flyteorg/flytekit/blob/c2b5c454e922f5dff1795f2a320868691429de3b/Dockerfile.dev#L54)? Just want to make sure it works in the integration tests after we remove the PYTHONPATH.

flytekit/bin/entrypoint.py
```diff
@@ -376,6 +376,9 @@ def _execute_task(
         dynamic_addl_distro,
         dynamic_dest_dir,
     ) as ctx:
+        import sys
+
+        sys.path.append(".")
```
pingsutw, line 381: nit
```suggestion
        if "." not in sys.path:
          sys.path.append(".")
```

arbaobao:
@pingsutw I has removed the pythonpath in [Dockerfile.dev](https://github.com/flyteorg/flytekit/blob/c2b5c454e922f5dff1795f2a320868691429de3b/Dockerfile.dev#L54).
And It works.

wild-endeavor (commented):
do you know what happens if both the current working directory, as an absolute path, and this new '.' are both in the sys path?  does that affect python module loading at all?

arbaobao:
@wild-endeavor I think append "." into sys.path is not going to affect python module  loading. Just to make sure that user defined Dockerfile works without pythonpath=/root

flytekit/bin/entrypoint.py
```diff
@@ -376,6 +376,10 @@ def _execute_task(
         dynamic_addl_distro,
         dynamic_dest_dir,
     ) as ctx:
+        import sys
+
+        if "." not in sys.path:
+            sys.path.append(".")
```
thomasjpfan, line 382: I think it's best to be explicit here and add the fully path for the current working directory:

```suggestion
        sys.path.append(os.getcwd())
```

Although, I do not think there is harm to include the working directory twice. If you want to make sure to not add the working directory twice:

```python
working_dir = os.getcwd()
if all(os.path.realpath(path) != working_dir for path in sys.path):
    sys.path.append(working_dir)
```

pingsutw (commented):
Could we add [dest_dir](https://github.com/flyteorg/flytekit/blob/4309c5ea92014f67204d245108e536ed2b82d83e/flytekit/bin/entrypoint.py#L532-L533) to PYTHONPATH as well?

flytekit/bin/entrypoint.py
wild-endeavor: Yes agreed.  This seems safer than adding `.`.  I'm worried about users who use both `.relative` imports as well as absolute imports.  We've seen weird cases of repeat-importing of modules before.

thomasjpfan:
> Could we add [dest_dir](https://github.com/flyteorg/flytekit/blob/4309c5ea92014f67204d245108e536ed2b82d83e/flytekit/bin/entrypoint.py#L532-L533) to PYTHONPATH as well?

Yea, that would give a better experience. In that case, I would inject it into `PYTHONPATH` in `Popen`:

https://github.com/flyteorg/flytekit/blob/4309c5ea92014f67204d245108e536ed2b82d83e/flytekit/bin/entrypoint.py#L545

I'll say do it in another PR, so we can discuss `dest_dir` independent of this PR.

flytekit/bin/entrypoint.py
arbaobao: Got it. Thank you for your advice.

arbaobao:
> I'll say do it in another PR, so we can discuss `dest_dir` independent of this PR.

I think i will create another PR to discuss `dest_dir`. 
Some of the test errors are due to this error.

wild-endeavor:
yeah let's discuss dest dir in another pr... if we ever move to a different model, like one where another worker maybe written in a different language is first in charge of downloading and extracting the package, it may be weird for that worker to be setting the pythonpath.

wild-endeavor (approved):

arbaobao:
@thomasjpfan @wild-endeavor @pingsutw 
I think we can add `dest_dir` instead of adding "."  because if we don't give `--destination-dir=/xxx` when we executes, it uses the default value "." .

thomasjpfan:
> I think we can add dest_dir instead of adding "." because if we don't give --destination-dir=/xxx when we executes, it uses the default value "." .

I think it's possible to have `dest_dir` be different from the working directory.

For example, one can set `dest_dir=/home/flytekit/my_dest` and working directory is `/root`. There could be python modules in `working_dir=/root` baked into the image and then fast registration copies files to `dest_dir=/home/flytekit/my_dest`. In this case, adding `/home/flytekit/my_dest` to `sys.path` is not enough. Either the image or `flytekit` needs to still needs to add `/root` into the `PYTHONPATH`.

arbaobao:
Got it. Thank you. i will create another PR to discuss dest_dir

wild-endeavor (approved):

Dockerfile.dev
```diff
@@ -50,8 +50,5 @@ RUN SETUPTOOLS_SCM_PRETEND_VERSION_FOR_FLYTEKIT=$PSEUDO_VERSION \
     && chown flytekit: /home \
     && :
 
-
-ENV PYTHONPATH="/flytekit:/flytekit/plugins/flytekit-k8s-pod:/flytekit/plugins/flytekit-deck-standard:"
```
eapolinario, line 54: removing this broke [flyteremote integration tests](https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/integration/remote/test_remote.py) due to the way tasks+workflows are [registered](https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/integration/remote/test_remote.py#L38-L55). The tests end up present in the image under the `tests` module, which lives under the `flytekit` directory and since we had `/flytekit` as part of `PYTHONPATH` the task was able to be run once `pyflyte-execute` ran.

eapolinario: fix in https://github.com/flyteorg/flytekit/pull/2701.

PR #2676, opened 2024-08-11
Add docker commands

mao3267:
We want to support adding other docker commands while creating ImageSpec.

1. Add a new "docker_commands" member in ImageSpec, which is a list of docker commands.
2. Insert these commands into the Dockerfile template.

## How was this patch tested?
It can not use the new ImageSpec while creating a container remotely, therefore, we didn't add test at this time.
We provide an example of building an image.
```
from flytekit.image_spec import ImageSpec
import flytekit

image_spec = ImageSpec(
    name="my_image",
    registry="localhost:30000",
    builder="default",
    apt_packages=["wget", "git", "build-essential"],
    docker_commands=[
        "RUN git config --global user.email 'chenvincent610@gmail.com'",
        "RUN git config --global user.name 'mao3267'",
        "RUN git clone https://github.com/mao3267/flytekit.git",
        "RUN wget https://raw.githubusercontent.com/opensource4you/gossip/main/README.md",
        "COPY . /root",
    ],
)


def my_task() -> str:
    try:
        with open("/root/README.md", "r") as f:
            return f.read().split("\n")[0]
    except Exception as e:
        return str(e) + f" {flytekit.__version__}"


@workflow
def my_wf() -> str:
    return my_task()

```
As the screenshots show, the image is successfully built. However, the task will fail since the flytekit version of the remote container is still `1.13`.

### Setup process
```
git clone https://github.com/mao3267/flytekit.git
git checkout add-docker-commands
make setup && pip install -e .
```
### Screenshots
image
image

## Related PRs
None

## Docs link
None


flytekit/image_spec/default_builder.py
```diff
@@ -232,6 +233,11 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
     else:
         run_commands = ""
 
+    if image_spec.docker_commands:
```
pingsutw, line 236: Could we also add some tests for it?

pingsutw: We should raise an exception in `envd` builder when users specify `docker_command` since `envd` doesn't support it.

mao3267: Unit tests of creating docker content and building image have been added, and also the exception for `envd` builder. Do they look good?

flytekit/image_spec/image_spec.py
```diff
@@ -72,6 +73,7 @@ class ImageSpec:
     entrypoint: Optional[List[str]] = None
     commands: Optional[List[str]] = None
     tag_format: Optional[str] = None
+    docker_commands: Optional[List[str]] = None
```
thomasjpfan, line 76: In `__post_init__` can you add `docker_commands` to the `parameters_str_list` list?

mao3267: @thomasjpfan Sure! Thanks for your quick review. We added `docker_commands` to the `parameters_str_list` now.

thomasjpfan (commented):
Currently the docker context for building is a temporary directory with the files copied over.

@mao3267 Do you think adding `DOCKER_COMMANDS` will be confusing in terms of Docker's context? A user may assume that the context is the current working directory.

PR #2678, opened 2024-08-12
Default execution name should be generated in flyteadmin

wayner0628:
- When user didn't specify execution prefix or name, `flytekit` should not generate execution name, but send an empty string and let `flyteadmin` generate

- When nothing is specified, set `execution_name` to an empty string, see the implementation [here](https://github.com/flyteorg/flytekit/blob/41d655629a75bc1c167c3fa4b75f93677846bb11/flytekit/remote/remote.py#L1150)

## How was this patch tested?
unit testing

### Setup process
```bash
pytest ./tests/flytekit/unit/remote/test_remote.py
```

## Related PRs
https://github.com/flyteorg/flyte/pull/5637

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -16,7 +16,6 @@ import pathlib
 import tempfile
 import time
 import typing
-import uuid
 from base64 import b64encode
 from collections import OrderedDict
 from dataclasses import asdict, dataclass
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -1144,10 +1144,9 @@ class FlyteRemote(object):
         """
         if execution_name is not None and execution_name_prefix is not None:
             raise ValueError("Only one of execution_name and execution_name_prefix can be set, but got both set")
+        # todo: The prefix should be passed to the backend
         if execution_name_prefix is not None:
             execution_name = execution_name_prefix + "-" + uuid.uuid4().hex[:19]
-        if execution_name is None and execution_name_prefix is None:
-            execution_name = ""
         if not options:
             options = Options()
         if options.disable_notifications is not None:
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -16,6 +16,7 @@ import pathlib
 import tempfile
 import time
 import typing
+import uuid
 from base64 import b64encode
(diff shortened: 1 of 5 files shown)


flytekit/remote/remote.py
```diff
@@ -1782,8 +1721,6 @@ def execute_local_workflow(
         domain: str = None,
         name: str = None,
         version: str = None,
-        execution_name: typing.Optional[str] = None,
```
pingsutw, line 1785: We should still allow users to specify the name. If they don't specify it, FlyteAdmin should generate a random name instead of Flytekit.

flytekit/remote/remote.py
```diff
@@ -1144,8 +1144,10 @@ def _execute(
         """
         if execution_name is not None and execution_name_prefix is not None:
             raise ValueError("Only one of execution_name and execution_name_prefix can be set, but got both set")
-        execution_name_prefix = execution_name_prefix + "-" if execution_name_prefix is not None else None
-        execution_name = execution_name or (execution_name_prefix or "f") + uuid.uuid4().hex[:19]
+        if execution_name_prefix is not None:
+            execution_name = execution_name_prefix + "-" + uuid.uuid4().hex[:19]
+        if execution_name is None and execution_name_prefix is None:
+            execution_name = ""
```
pingsutw, line 1150: nit: we can remove these two lines, right? we should be able to pass none to the backend

flytekit/remote/remote.py
```diff
@@ -1144,8 +1144,10 @@ def _execute(
         """
         if execution_name is not None and execution_name_prefix is not None:
             raise ValueError("Only one of execution_name and execution_name_prefix can be set, but got both set")
-        execution_name_prefix = execution_name_prefix + "-" if execution_name_prefix is not None else None
-        execution_name = execution_name or (execution_name_prefix or "f") + uuid.uuid4().hex[:19]
+        if execution_name_prefix is not None:
+            execution_name = execution_name_prefix + "-" + uuid.uuid4().hex[:19]
```
pingsutw, line 1148: The prefix should be passed to the backend, and the backend should use this prefix to generate the new execution name. We can address this in a follow-up PR.

pingsutw (approved):

PR #2680, opened 2024-08-13
add motherduck support for duckdb plugin

dansola:

We would like to integrate motherduck support with the current duckdb plugin such that a `DuckDBQuery` can query the motherduck data warehouse and/or in memory files at the same time.

Simply add a connection to motherduck and authenticate via a secret. `duckdb` handles the rest.

## How was this patch tested?

Tested on a motherduck free trial with a combinarion query like:
```
"""
WITH db1 AS (
    SELECT
        SUM(a) AS sum_a,
    FROM mydf
),
db2 AS (
    SELECT
        MEAN(trip_time) AS mean_tt
    FROM sample_data.nyc.rideshare
)
SELECT
    db1.sum_a + db2.mean_tt AS total_sum,
FROM db1, db2;
"""
```

### DIFF
--- a/plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
+++ b/plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
@@ -12,6 +12,10 @@ pd = lazy_module("pandas")
 pa = lazy_module("pyarrow")
 
 
+class MissingSecretError(ValueError):
+    pass
+
+
 def connect_local(token: Optional[str]):
     """Connect to local DuckDB."""
     return duckdb.connect(":memory:")
@@ -89,6 +93,8 @@ class DuckDBQuery(PythonInstanceTask):
                 group_version=self._connect_secret.group_version,
             )
         if isinstance(self._provider, DuckDBProvider):
+            if not connect_token and self._provider != DuckDBProvider.LOCAL:
+                raise MissingSecretError(f"A secret_requests must be provided for the {self._provider.name} provider.")
             return self._provider.value(connect_token)
         else:  # callable
             return self._provider(connect_token)
--- a/plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
+++ b/plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
@@ -46,7 +46,6 @@ class DuckDBQuery(PythonInstanceTask):
         query: Union[str, List[str]],
         inputs: Optional[Dict[str, Union[StructuredDataset, list]]] = None,
         provider: DuckDBProvider = DuckDBProvider.LOCAL,
-        hosted_secret: Optional[Secret] = None,
         **kwargs,
     ):
         """
@@ -57,11 +56,11 @@ class DuckDBQuery(PythonInstanceTask):
             query: DuckDB query to execute
             inputs: The query parameters to be used while executing the query
             provider: DuckDB provider (e.g., LOCAL, MOTHERDUCK, ANOTHERPRODUCT)
-            hosted_secret: Secret containing authentication token for the hosted provider
         """
         self._query = query
         self._provider = provider
-        self._hosted_secret = hosted_secret
+        secret_requests: Optional[list[Secret]] = kwargs.get("secret_requests", None)
+        self._hosted_secret = secret_requests[0] if secret_requests else None
 
         outputs = {"result": StructuredDataset}
 
@@ -73,25 +72,21 @@ class DuckDBQuery(PythonInstanceTask):
             **kwargs,
         )
 
-    def _connect_to_duckdb(self, provider: DuckDBProvider, hosted_secret: Optional[Secret]):
+    def _connect_to_duckdb(self):
         """
         Handles the connection to DuckDB based on the provider.
-
-        Args:
-            provider: DuckDB provider
-            hosted_secret: Secret containing authentication token for the hosted provider
 
         Returns:
             A DuckDB connection object.
         """
 
-        if provider not in DuckDBProvider:
-            raise ValueError(f"Unknown DuckDB provider: {provider}")
+        if self._provider not in DuckDBProvider:
+            raise ValueError(f"Unknown DuckDB provider: {self._provider}")
 
-        if provider != DuckDBProvider.LOCAL and hosted_secret is None:
-            raise ValueError(f"A secret is required for the {provider} provider.")
+        if self._provider != DuckDBProvider.LOCAL and self._hosted_secret is None:
+            raise ValueError(f"A secret is required for the {self._provider} provider.")
 
-        return provider.value(hosted_secret)
+        return self._provider.value(self._hosted_secret)
 
     def _execute_query(
         self, con: duckdb.DuckDBPyConnection, params: list, query: str, counter: int, multiple_params: bool
@@ -125,7 +120,7 @@ class DuckDBQuery(PythonInstanceTask):
 
     def execute(self, **kwargs) -> StructuredDataset:
         # TODO: Enable iterative download after adding the functionality to structured dataset code.
-        con = self._connect_to_duckdb(self._provider, self._hosted_secret)
+        con = self._connect_to_duckdb()
 
         params = None
         for key in self.python_interface.inputs.keys():
--- a/plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
+++ b/plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
@@ -1,7 +1,7 @@
 import json
 from enum import Enum
 from functools import partial
-from typing import Dict, List, NamedTuple, Optional, Union
+from typing import Callable, Dict, List, NamedTuple, Optional, Union
 
 from flytekit import PythonInstanceTask, Secret, current_context, lazy_module
 from flytekit.extend import Interface
@@ -12,19 +12,14 @@ pd = lazy_module("pandas")
 pa = lazy_module("pyarrow")
 
 
-def connect_local(hosted_secret: Optional[Secret]):
+def connect_local(token: Optional[str]):
     """Connect to local DuckDB."""
     return duckdb.connect(":memory:")
 
 
-def connect_motherduck(hosted_secret: Secret):
+def connect_motherduck(token: str):
     """Connect to MotherDuck."""
-    motherduck_token = current_context().secrets.get(
-        group=hosted_secret.group,
-        key=hosted_secret.key,
-        group_version=hosted_secret.group_version,
-    )
-    return duckdb.connect("md:", config={"motherduck_token": motherduck_token})
+    return duckdb.connect("md:", config={"motherduck_token": token})
 
 
 class DuckDBProvider(Enum):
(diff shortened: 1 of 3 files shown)


plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -32,10 +33,17 @@ def __init__(
...
             query: DuckDB query to execute
             inputs: The query parameters to be used while executing the query
+            md_token_name: Name of the secret containing a MotherDuck authentication token
         """
+        if md_token_name is None:
+            # create an in-memory database that's non-persistent
+            self._con = duckdb.connect(":memory:")
+        else:
+            # connect to motherduck
+            md_token = current_context().secrets.get(md_token_name)
```
samhita-alla, line 43: i don't think this will work for k8s secrets. we should add support for group, key, and group version.

dansola: ah thanks! Let me know if the latest commit is what you're looking for

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -32,10 +33,21 @@ def __init__(
...
+            # create an in-memory database that's non-persistent
+            self._con = duckdb.connect(":memory:")
+        else:
+            # connect to motherduck
+            md_token = current_context().secrets.get(
+                group=md_secret.group,
+                key=md_secret.key,
+                group_version=md_secret.group_version,
+            )
+            self._con = duckdb.connect("md:", config={'motherduck_token': md_token})
```
thomasjpfan, line 48: What is the recommended way to connect to motherduck? In their docs, it uses `md:?motherduck_token=<token>`.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -23,6 +23,7 @@ def __init__(
         name: str,
         query: Union[str, List[str]],
         inputs: Optional[Dict[str, Union[StructuredDataset, list]]] = None,
+        md_secret: Optional[Secret] = None,
```
thomasjpfan, line 26: If we want to have a motherduck specific parameter, I think it's better to be explicit with the name:

```suggestion
        motherduck_secret: Optional[Secret] = None,
```

`md_secret` looks like "markdown secret"

thomasjpfan (commented):
There are linting issues: https://github.com/flyteorg/flytekit/actions/runs/10374544320/job/28722185574?pr=2680

Can you run `make fmt`? (Also you can configure pre-commit to automatically lint  your files.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
dansola: They recommend both ways. I thought this was was cleaner and more readable.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -32,10 +33,21 @@ def __init__(
...
+            # create an in-memory database that's non-persistent
+            self._con = duckdb.connect(":memory:")
+        else:
+            # connect to motherduck
+            motherduck_token = current_context().secrets.get(
+                group=motherduck_secret.group,
+                key=motherduck_secret.key,
+                group_version=motherduck_secret.group_version,
+            )
+            self._con = duckdb.connect("md:", config={"motherduck_token": motherduck_token})
```
thomasjpfan, line 48: My concern is that this is hardcoded for motherduck. If there was a non-motherduck provider for duckdb, then we'll need to add another if statement & `another_provider_sercret` to configure them.

The only way I see doing this is to provide a callable:

```python
def motherduck_connect(token):
    return duckdb.connect("md:", config={"motherduck_token": token})

def memory_connect(token):
    return duckdb.connect(":memory:")

class DuckDBQuery:
    def __init__(self, ..., construct_duckdb_connect=motherduck_connect):
        if secret is None:
            self._con = construct_duckdb_connect(None)
        else:
            token = current_conect()...
            self._con = construct_duckdb_connect(token)
```

The above API also makes the rest of `duckdb.connect` configurable to the user.

dansola: good point. let me make this a bit more general

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -23,6 +29,8 @@ def __init__(
         name: str,
         query: Union[str, List[str]],
         inputs: Optional[Dict[str, Union[StructuredDataset, list]]] = None,
+        provider: Optional[DuckDBProvider] = DuckDBProvider.LOCAL,
```
samhita-alla, line 32: this need not be optional.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -32,10 +40,11 @@ def __init__(
             name: Name of the task
             query: DuckDB query to execute
             inputs: The query parameters to be used while executing the query
+            provider: DuckDB provider (e.g., LOCAL, MOTHERDUCK, ANOTHERPRODUCT)
+            hosted_secret: Secret containing authentication token for the hosted provider
         """
+        self._con = self._connect_to_duckdb(provider, hosted_secret)
```
samhita-alla, line 46: check this PR: https://github.com/flyteorg/flytekit/pull/2684

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -47,6 +56,43 @@ def __init__(
...
+        if provider != DuckDBProvider.LOCAL and hosted_secret is None:
+            raise ValueError(f"A secret is required for the {provider.value} provider.")
+
+        return connection_map[provider](hosted_secret)
+
+    def _connect_local(self, hosted_secret: Optional[Secret]):
+        """Connect to local DuckDB."""
+        return duckdb.connect(":memory:")
+
+    def _connect_motherduck(self, hosted_secret: Secret):
```
samhita-alla, line 87: could you add some test cases?

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -47,6 +56,43 @@ def __init__(
...
+        Handles the connection to DuckDB based on the provider.
+
+        Args:
+            provider: DuckDB provider
+            hosted_secret: Secret containing authentication token for the hosted provider
+
+        Returns:
+            A DuckDB connection object.
+        """
+        connection_map = {
```
samhita-alla, line 70: instead of creating a map, you could use `functools.partial`: https://www.linkedin.com/pulse/how-enumerate-functions-python-fan-wu/

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -21,8 +43,9 @@ class DuckDBQuery(PythonInstanceTask):
     def __init__(
         self,
         name: str,
-        query: Union[str, List[str]],
+        query: Optional[Union[str, List[str]]] = None,
         inputs: Optional[Dict[str, Union[StructuredDataset, list]]] = None,
+        provider: DuckDBProvider = DuckDBProvider.LOCAL,
```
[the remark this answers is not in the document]
dansola: Will do!! Thank you.

I will also use `secret_requests` from `PythonAutoContainerTask` and check for it in the constructor rather than adding a new secret argument.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -44,6 +81,27 @@ def __init__(
...
+            A DuckDB connection object.
+        """
+        connect_token = None
+        if self._connect_secret:
+            connect_token = current_context().secrets.get(
+                group=self._connect_secret.group,
+                key=self._connect_secret.key,
+                group_version=self._connect_secret.group_version,
+            )
+        if isinstance(self._provider, DuckDBProvider):
```
thomasjpfan, line 98: Can we raise this error in `__init__` so it's caught during registration time? 

Specifically, if `self._connect_secret is None and isinstance(self._provider, DuckDBProvider)`, then raise an error.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -21,19 +42,35 @@ class DuckDBQuery(PythonInstanceTask):
     def __init__(
         self,
         name: str,
-        query: Union[str, List[str]],
+        query: Optional[Union[str, List[str]]] = None,
```
thomasjpfan, line 45: Can you update the docstring to example that `query=None` is useful for runtime queries?

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
dansola: Sure. I think `self._connect_secret is None and isinstance(self._provider, DuckDBProvider)` will cause an error if we are using DuckDBProvider.LOCAL. I think I need `if self._connect_secret is None and isinstance(self._provider, DuckDBProvider) and self._provider != DuckDBProvider.LOCAL`

thomasjpfan (approved):

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -21,19 +42,53 @@ class DuckDBQuery(PythonInstanceTask):
...
+        Note that the provider can be one of the default providers listed in DuckDBProvider or a custom callable like the following:
+
+        def custom_connect_motherduck(token: str):
+            return duckdb.connect("md:", config={"motherduck_token": token, "another_config": "hello"})
+
+        DuckDBQuery(..., provider=custom_connect_motherduck)
+
+        Also note that a query can be provided at runtime if query=None is provided.
+
+        duckdb_query = DuckDBQuery(
```
samhita-alla, line 62: since the inputs must be provided at registration time, does it make sense to provide the query at runtime?

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -10,6 +12,25 @@
 pa = lazy_module("pyarrow")
 
 
+class MissingSecretError(ValueError):
```
samhita-alla, line 15: can't we move this to the test module since this class isn't being used anywhere?

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
```diff
@@ -44,6 +99,25 @@ def __init__(
...
+        if self._connect_secret:
+            connect_token = current_context().secrets.get(
+                group=self._connect_secret.group,
+                key=self._connect_secret.key,
+                group_version=self._connect_secret.group_version,
+            )
+        if isinstance(self._provider, DuckDBProvider):
+            return self._provider.value(connect_token)
+        else:  # callable
+            return self._provider(connect_token)
```
samhita-alla, line 119: how about we not send the token when there isn't one provided?

samhita-alla: can you also remove the token param from the connect_local function?

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
dansola: The actual contents of the query could differ at runtime based on some logic in say a dynamic task. If the input type is a dataframe, you can still query it in different ways.

plugins/flytekit-duckdb/flytekitplugins/duckdb/task.py
dansola: sure, I removed the token and gave the option for callables that don't have tokens

samhita-alla (approved):

PR #2688, opened 2024-08-16
proof of concept refactor of FlyteRemote._version_from_hash (todo: needs tests, linting)

ddl-rliu:

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -942,8 +942,7 @@ class FlyteRemote(object):
     def _version_from_hash(
         md5_bytes: bytes,
         serialization_settings: SerializationSettings,
-        default_inputs: typing.Optional[Dict[str, typing.Any]] = None,
-        *additional_context: str,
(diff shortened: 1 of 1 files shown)


flytekit/remote/remote.py
```diff
@@ -942,8 +942,7 @@ def upload_file(
     def _version_from_hash(
         md5_bytes: bytes,
         serialization_settings: SerializationSettings,
-        default_inputs: typing.Optional[Dict[str, typing.Any]] = None,
-        *additional_context: str,
+        entity: typing.Union[PythonAutoContainerTask, WorkflowBase],
     ) -> str:
```
ddl-rliu, line 946: A plugin can implement a subclass of FlyteRemote which overrides _version_from_hash to control the version hash calculation (idea from [here](https://github.com/flyteorg/flytekit/pull/2428#pullrequestreview-2243627484))

A concern with this idea is that if future PRs modify the behavior of `FlyteRemote._version_from_hash`, it may become out-of-sync with the custom subclass of `FlyteRemote`. On the other hand, such an occurrence may be very rare in practice.

PR #2690, opened 2024-08-17
Streamline fast register options with new flag

wild-endeavor:
## Background
Flytekit has a notion of a "fast register" ([additional docs](https://docs.flyte.org/en/latest/flyte_fundamentals/registering_workflows.html#fast-registration)) which copies a set of files from the local host, compresses and uploads them, and instructs tasks registered to download and extract those files before execution.

### Current UX
The behavior today and how the user invokes the behavior is as follows

#### pyflyte run
`pyflyte run` was initially written as a way for users to quickly run _one_ script.  The fast registration aspect of this command copied this one script into the zip file and into the final container.  Over time, support for additional Python files was added via the `--copy-all` flag.  Setting this flag copies _all_ unignored files from the detected root folder into the archive.

More recently, default behavior for `run` was [updated](https://github.com/flyteorg/flytekit/pull/2663) to automatically detect and include all local Python files that were loaded in the course of loading the target script.

#### pyflyte register
`pyflyte register` by default runs in 'fast' mode as well.  By default, it behaves like the `--copy-all` flag in `run`, copying all unignored files from the detected root folder.

To turn off this behavior, the `--non-fast` flag was added.

#### pyflyte package
`pyflyte package` by default does not run in 'fast' mode.  That is by default, there are no files copied/extracted.

To enable fast packaging, `--fast` must be added.

The differences in behavior, the differences in switches, and the fact that a new way to more intelligently detect which files should be copied means we have the opportunity to streamline the ux here.

### Aligned UX
This PR introduces a `--copy` flag (listed currently as a beta) that will deprecate the current set of flags.  For now, both sets of flags can be used.

In short, use

* `--copy all` for the current fast register behavior (copying all non-ignored files)
* `--copy auto` for the new behavior of copying only loaded Python modules.
* `--copy none` to not copy any files.  This is basically turning off fast registration entirely.

#### Examples
Example use of old and new flag.  Again, all old flags will continue to work.  In the examples, the `--copy` switch is explicitly written to a) just highlight what the default is and b) if specified, the logic is _slightly_ different than if not specifying it (see the discussion on changes below).

##### Run
To smart-copy just loaded python modules (the default)
Old: `pyflyte run a/c/d.py`
New: `pyflyte run [--copy auto] a/c/d.py`

To copy all files
Old: `pyflyte run --copy-all a/c/d.py`
New: `pyflyte run --copy all a/c/d.py`

##### Register
To copy all unignored files (the default)
Old: `pyflyte register a/b,a/c/d.py`
New: `pyflyte register [--copy all] a/b,a/c/d.py`

To smart copy only relevant Python modules
`pyflyte register --copy auto a/b,a/c/d.py`

To turn off fast register completely:
Old: `pyflyte register --non-fast a/b,a/c/d.py`
New: `pyflyte register --copy none a/b,a/c/d.py`

#### Package
To package normally with no files copied (the default)
Old: `pyflyte --pkgs a.b,c.d package`
New: `pyflyte --pkgs a.b,c.d package [--copy none]

To package with all files:
Old: `pyflyte --pkgs a.b,c.d package --fast`
New: `pyflyte --pkgs a.b,c.d package --copy all`

To package with only loaded Python modules:
`pyflyte --pkgs a.b,c.d package --copy auto`

#### Debugging
To help users understand what files are slated to be copied, each command also takes a `--ls-files` flag.

```
Files to be copied for fast registration...
📂 /Users/you/dev/project1/src (detected source root)
┣━━ my.img
┣━━ resources
┃   ┗━━ file.txt
┗━━ work
    ┣━━ __init__.py
    ┣━━ tt.py
    ┗━━ wfs.py
```

### Serialize
`pyflyte serialize` is missing from the above discussions because it is for the most part a duplicate of the `package` command.  We are planning to deprecate `serialize` actually.

### Changes
To make the new flag possible, especially in the context of the `auto` behavior of detecting which modules are loaded, this pr had to
* Load the user requested modules before uploading.  This changes the behavior for `register` which previously had uploaded files first.
* The manner in which files are added to the tar file are different.  Previously, folders were added into the tar file wholesale, with the tar library responsible for adding files recursively.  With the new logic, files are listed and added one at a time.  For folders with a large number of files, this may take longer.

> [!NOTE]  
> If you use the default/use the old flags, the old tar logic will be triggered.  Only if you use the new `--copy` flag will the new logic for tar be used.

> [!NOTE]  
> This means that your code can no longer depend on the result of the upload.  We think this is an extreme edge case, but let us know if not.

### Code updates
* The `serialize` function was split up into two functions to allow pyflyte to first load python modules, then upload files, then actually compile the Flyte entities.
* `load_packages_and_modules` renamed to `list_packages_and_modules` because it no longer loads anything.
* Two fields were added to the `FastPackageOptions` class to capture the new behavior.
* Two helper functions were moved from fast registration to script mode.
* Added a simple error handler to the walk_packages call just to not swallow errors.


flytekit/clis/sdk_in_container/package.py
```diff
@@ -50,8 +53,26 @@
...
+    "--copy",
+    required=False,
+    type=click.Choice(["all", "auto", "none"], case_sensitive=False),
+    default=None,  # this will be changed to "none" after removing fast option
+    callback=parse_copy,
+    help="[Beta] Specify how and whether to use fast register"
+    " 'all' will behave as the current fast flag copying all files, 'auto' copies only loaded Python modules",
+)
+@click.option(
+    "--ls-files",
```
thomasjpfan, line 70: From our call, we wanted to remove `--ls-files` and print all the files when `--verbose` is passed in.

flytekit/clis/sdk_in_container/package.py
```diff
@@ -50,8 +53,26 @@
...
+    " deploys of flyte workflows and tasks. You can specify --copy all/auto instead"
+    " Note this needs additional configuration, refer to the docs.",
+)
+@click.option(
+    "--copy",
+    required=False,
+    type=click.Choice(["all", "auto", "none"], case_sensitive=False),
+    default=None,  # this will be changed to "none" after removing fast option
+    callback=parse_copy,
+    help="[Beta] Specify how and whether to use fast register"
```
thomasjpfan, line 66: Can this help be rephrased to remove the concept of "fast register"? In the end it's different copying behavior.

flytekit/tools/fast_registration.py
```diff
@@ -3,27 +3,42 @@
...
 
 
+class CopyFileDetection(Enum):
+    LOADED_MODULES = 1
+    ALL = 2
+    # This is a temporary option to be removed in the future. In the future this value of the enum should simply
+    # be Python None. Here now to distinguish between users explicitly setting --copy none and not setting the flag.
+    # This is only used for register, not for package or run because run doesn't have a no-fast-register option and
+    # package is by default non-fast.
+    TEMP_NO_COPY = 3
```
thomasjpfan, line 39: I think it's better to keep this enum and not use Python `None`. This way `CopyFileDetection` fully describes all the different copy modes.

flytekit/tools/fast_registration.py
```diff
@@ -32,6 +47,34 @@ class FastPackageOptions:
 
     ignores: list[Ignore]
     keep_default_ignores: bool = True
+    copy_style: Optional[CopyFileDetection] = None
+    ls_files: bool = False
+
+
+def print_ls_tree(source: os.PathLike, ls: typing.List[str]):
+    click.secho("Files to be copied for fast registration...", fg="bright_blue")
+    fff = []
```
thomasjpfan, line 56: I do not see what `fff` means. Can this be renamed?

flytekit/clis/sdk_in_container/package.py
```diff
@@ -113,6 +136,12 @@ def package(
     object contains the WorkflowTemplate, along with the relevant tasks for that workflow.
     This serialization step will set the name of the tasks to the fully qualified name of the task function.
     """
+    if copy == CopyFileDetection.TEMP_NO_COPY:
```
thomasjpfan, line 139: I say we let users specify `--copy none`. It's technically a valid option, but it is not the default.

flytekit/clis/sdk_in_container/package.py
```diff
@@ -113,6 +136,12 @@ def package(
...
+        raise ValueError("--copy none doesn't need to be specified, package by default does not copy files")
+    elif copy == CopyFileDetection.ALL or copy == CopyFileDetection.LOADED_MODULES:
+        # for those migrating, who only set --copy all/auto but don't have --fast set.
+        fast = True
+
```
thomasjpfan, line 144: Can we disallow `--copy` and `--fast` at the same time?

Also if `--fast` is passed in, I'll want to raise a warning message that points people to docs that explain "how to use --copy".

flytekit/clis/sdk_in_container/register.py
```diff
@@ -148,14 +172,27 @@ def register(
     """
     see help
     """
+
```
thomasjpfan, line 175: Same here can we prevent users from passing in `--non-fast` and `--copy` at the same time? (We do not check if the values are consistent)

flytekit/tools/script_mode.py
```diff
@@ -79,17 +84,114 @@ def tar_strip_file_attributes(tar_info: tarfile.TarInfo) -> tarfile.TarInfo:
...
+def list_imported_modules_as_files(source_path: str, modules: List[ModuleType]) -> List[str]:
     """Copies modules into destination that are in modules. The module files are copied only if:
 
     1. Not a site-packages. These are installed packages and not user files.
     2. Not in the bin. These are also installed and not user files.
     3. Does not share a common path with the source_path.
     """
+    # source path is the folder holding the main script.
+    # but in register/package case, there are multiple folders.
+    # identify a common root amongst the packages listed?
```
thomasjpfan, line 189: In this PR, how are you handling the case of multiple folders?

thomasjpfan (commented):
This is a first past review.

flytekit/tools/fast_registration.py
wild-endeavor: debugging... removed.

flytekit/tools/fast_registration.py
wild-endeavor: sounds good.  I will update the comment and rename the value.

flytekit/clis/sdk_in_container/package.py
wild-endeavor: reworded.

flytekit/clis/sdk_in_container/package.py
wild-endeavor: removing the flag, but it's helpful to have the option in the fastpackageoptions object so leaving it there unless there are objections.

flytekit/clis/sdk_in_container/package.py
wild-endeavor: will remove the exception and add a check for conflicts with the fast switch.  But what do you mean by default?  not copying any files is and will continue to be the default for package.

flytekit/tools/script_mode.py
wild-endeavor: not changing that behavior here at all, but all the commands resolve down to one folder for now.

flytekit/clis/sdk_in_container/package.py
wild-endeavor: adding an exception, but no docs to point to as yet.  can update after those have been published.

flytekit/clis/sdk_in_container/register.py
wild-endeavor: added.

flytekit/clis/sdk_in_container/register.py
```diff
@@ -93,7 +98,16 @@
...
+    help="[Will be deprecated, see --copy] Skip zipping and uploading the package. You can specify --copy none instead",
+)
+@click.option(
+    "--copy",
+    required=False,
+    type=click.Choice(["all", "auto", "none"], case_sensitive=False),
+    default=None,  # this will be changed to "all" after removing non-fast option
+    callback=parse_copy,
+    help="[Beta] Specify how and whether to use fast register"
+    " 'all' is the current behavior copying all files from root, 'auto' copies only loaded Python modules",
```
pingsutw, line 110: should we add a description for "none" here

pingsutw (commented):
This is so cool!
Screenshot 2024-08-22 at 1 28 43 PM

flytekit/tools/fast_registration.py
```diff
@@ -58,28 +100,74 @@ def fast_package(
...
+        else:
+            # This triggers listing of all files, mimicking the old way of creating the tar file.
+            ls, ls_digest = ls_files(str(source), [], deref_symlinks, ignore)
+
+        logger.debug(f"Hash digest: {ls_digest}", fg="green")
+
+        if options.ls_files:
+            print_ls_tree(source, ls)
+
+        # print(f"Digest check: old {digest} ==? new {ls_digest} -- {digest == ls_digest}")
```
pingsutw, line 124: remove this?

pingsutw (dismissed):

flytekit/clis/sdk_in_container/register.py
```diff
@@ -148,14 +164,30 @@ def register(
     """
     see help
     """
+    if copy is not None and non_fast:
+        raise ValueError("--non-fast and --copy cannot be used together. Use --copy none instead.")
+
+    # Error handling for non-fast/copy conflicts
```
thomasjpfan, line 170: Since `copy`  and `non_fast` can not be used together, there should be no conflicts?

flytekit/tools/fast_registration.py
```diff
@@ -58,28 +100,74 @@ def fast_package(
         ignores = default_ignores
     ignore = IgnoreGroup(source, ignores)
 
+    # Remove this after original tar command is removed.
```
thomasjpfan, line 103: Why are we keeping the original tar command? Is it for backward compatibility?

flytekit/clis/sdk_in_container/register.py
```diff
@@ -148,14 +164,30 @@ def register(
...
+        non_fast = True
+        # Set this to None because downstream logic currently detects None to mean old logic.
+        copy = None
+    elif copy == CopyFileDetection.ALL:
+        if non_fast:
+            raise ValueError("Conflicting options: cannot specify both --non-fast and --copy all")
+    elif copy == CopyFileDetection.LOADED_MODULES:
+        if non_fast:
+            raise ValueError("Conflicting options: cannot specify both --non-fast and --copy auto")
+    ls_files = ctx.obj[constants.CTX_VERBOSE] > 0
```
thomasjpfan, line 181: Can we be explicit and rename `ls_files` to "show_files" everywhere?

```suggestion
    show_files = ctx.obj[constants.CTX_VERBOSE] > 0
```

flytekit/tools/fast_registration.py
wild-endeavor: yup.

flytekit/clis/sdk_in_container/register.py
wild-endeavor: i want to keep the function name, but these variables yeah sure.

pingsutw (approved):

PR #2695, opened 2024-08-20
Polars lazyframe

cosmicBboy:
## Tracking issue

Fixes https://github.com/flyteorg/flyte/issues/5678

The flytekit plugin for polars only supports `polars.DataFrame`. Many polars users rely on the `LazyFrame` API to make data handling more memory efficient.

This PR:
- Adds and registers a `LazyFrame` encoder/decoder to the `StructuredDataset` type.

## How was this patch tested?

Unit tests run on CI.

### DIFF
--- a/plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
+++ b/plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
@@ -33,10 +33,11 @@ class PolarsDataFrameRenderer:
 
     def to_html(self, df: typing.Union[pl.DataFrame, pl.LazyFrame]) -> str:
         assert isinstance(df, (pl.DataFrame, pl.LazyFrame))
-        if isinstance(df, pl.LazyFrame):
-            describe_df = df.collect().describe()
-        else:
+        try:
             describe_df = df.describe()
+        except AttributeError:
+            # LazyFrames in polars <= 0.19 does not support `describe`
+            describe_df = df.collect().describe()
 
         columns = describe_df["describe"]
         html_repr = describe_df.drop("describe").transpose(column_names=columns)._repr_html_()
@@ -111,23 +112,25 @@ class PolarsLazyFrameToParquetEncodingHandler(StructuredDatasetEncoder):
     ) -> literals.StructuredDataset:
         lf = typing.cast(pl.LazyFrame, structured_dataset.dataframe)
 
-        output_bytes = io.BytesIO()
         # The pl.LazyFrame.sink_parquet method uses streaming mode, which is currently considered unstable. Until it is
         # stable, we collect the dataframe and write it to a BytesIO buffer.
         df = lf.collect()
-        # Polars 0.13.12 deprecated to_parquet in favor of write_parquet
+
         if hasattr(df, "write_parquet"):
-            df.write_parquet(output_bytes)
+            # Polars 0.13.12 deprecated to_parquet in favor of write_parquet
+            _write_method = df.write_parquet
         else:
-            df.to_parquet(output_bytes)
+            _write_method = df.to_parquet
 
         if structured_dataset.uri is not None:
             fs = ctx.file_access.get_filesystem_for_path(path=structured_dataset.uri)
             with fs.open(structured_dataset.uri, "wb") as s:
-                s.write(output_bytes)
+                _write_method(s)
             output_uri = structured_dataset.uri
         else:
+            output_bytes = io.BytesIO()
             remote_fn = "00000"  # 00000 is our default unnamed parquet filename
+            _write_method(output_bytes)
             output_uri = ctx.file_access.put_raw_data(output_bytes, file_name=remote_fn)
         return literals.StructuredDataset(uri=output_uri, metadata=StructuredDatasetMetadata(structured_dataset_type))
 
--- a/plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
+++ b/plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
@@ -153,19 +153,16 @@ class ParquetToPolarsLazyFrameDecodingHandler(StructuredDatasetDecoder):
     ) -> pl.LazyFrame:
         uri = flyte_value.uri
 
-        protocol = fsspec_utils.get_protocol(uri)
-        if protocol == "file":
-            kwargs = None
-        else:
-            kwargs = get_fsspec_storage_options(
-                protocol=protocol,
-                data_config=ctx.file_access.data_config,
-            )
-
+        kwargs = get_fsspec_storage_options(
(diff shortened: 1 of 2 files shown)


plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
```diff
@@ -14,20 +14,33 @@
...
     """
 
-    def to_html(self, df: pl.DataFrame) -> str:
-        assert isinstance(df, pl.DataFrame)
-        describe_df = df.describe()
-        return pd.DataFrame(describe_df.transpose(), columns=describe_df.columns).to_html(index=False)
+    def to_html(self, df: typing.Union[pl.DataFrame, pl.LazyFrame]) -> str:
+        assert isinstance(df, (pl.DataFrame, pl.LazyFrame))
+        if isinstance(df, pl.LazyFrame):
+            describe_df = df.collect().describe()
```
thomasjpfan, line 37: `describe` does not need to be collected first with a lazy dataframe:

```python
import polars as pl

df = pl.DataFrame({"A": [1, 2, 3], "b": [4, 5, 6]}).lazy()
assert isinstance(df, pl.LazyFrame)

df.describe()
```

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
```diff
@@ -82,6 +95,76 @@ def decode(
...
+        # Polars 0.13.12 deprecated to_parquet in favor of write_parquet
+        if hasattr(df, "write_parquet"):
+            df.write_parquet(output_bytes)
+        else:
+            df.to_parquet(output_bytes)
+
+        if structured_dataset.uri is not None:
+            fs = ctx.file_access.get_filesystem_for_path(path=structured_dataset.uri)
+            with fs.open(structured_dataset.uri, "wb") as s:
+                s.write(output_bytes)
```
thomasjpfan, line 127: Can we write directly into the cloud storage and not use `io.BytesIO()`?

```python
with fs.open(structured_dataset.uri, "wb") as s:
    df.write_parquet(s)
```


REF: https://docs.pola.rs/user-guide/io/cloud-storage/

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
cosmicBboy: it does for polars <= 0.19. Unless we place a lower bound on the polars version supported by this plugin.

thomasjpfan: Can we try `describe` first and fallback to `collect` if it does not work?

```python
try:
    describe_df = df.describe()
except AttributeError:
    # LazyFrames in polars <= 0.19 does not support `describe` 
    describe_df = df.collect().describe()
```

eapolinario (dismissed):

thomasjpfan (commented):
On the flyte sandbox, the following stalls for me:

```python
import polars as pl

polars_plugin = "git+https://github.com/flyteorg/flytekit@55db243b51bfe5137942b68aa212ecb8edc27bcd#subdirectory=plugins/flytekit-polars"

image = ImageSpec(
    name="polars",
    packages=[polars_plugin, "flytekit"],
    registry="localhost:30000",
    apt_packages=["git"],
)


@task(container_image=image)
def create_lazy_frame() -> pl.LazyFrame:
    return pl.LazyFrame({"col1": [1, 3, 2], "col2": list("abc")})


@task(container_image=image)
def read_pl(df: pl.LazyFrame) -> int:
    return df.select(pl.col("col1").sum()).collect().item()


@workflow
def main() -> int:
    df = create_lazy_frame()
    return read_pl(df=df)

```

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
```diff
@@ -82,6 +98,79 @@ def decode(
...
+
+# Register the Polars DataFrame handlers
 StructuredDatasetTransformerEngine.register(PolarsDataFrameToParquetEncodingHandler())
 StructuredDatasetTransformerEngine.register(ParquetToPolarsDataFrameDecodingHandler())
 StructuredDatasetTransformerEngine.register_renderer(pl.DataFrame, PolarsDataFrameRenderer())
+
+# Register the Polars LazyFrame handlers
+StructuredDatasetTransformerEngine.register(PolarsLazyFrameToParquetEncodingHandler())
+StructuredDatasetTransformerEngine.register(ParquetToPolarsLazyFrameDecodingHandler())
+StructuredDatasetTransformerEngine.register_renderer(pl.LazyFrame, PolarsDataFrameRenderer())
```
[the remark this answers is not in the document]
cosmicBboy: should we make the renderer for this explicit? (either using an `Annotated` type or in the task function body?)

cosmicBboy:
[failing greatexpectations plugin](https://github.com/flyteorg/flytekit/actions/runs/10530626765/job/29181131760?pr=2695) should be fixed in a separate PR

plugins/flytekit-polars/flytekitplugins/polars/sd_transformers.py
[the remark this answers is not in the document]
cosmicBboy: @thomasjpfan I removed `register_renderer` for the LazyFrame... I think it's reasonable to think that any user who uses this doesn't want to perform potentially computationally intensive operation at the end of the task without doing it explicitly

eapolinario:
@cosmicBboy , @thomasjpfan , can we merge this?

cosmicBboy:
one sec @eapolinario let's not merge yet, still testing on flyte sandbox and serverless

pingsutw (approved):

eapolinario (approved):

HansBambel:
Cool feature!
Is the documentation then just not up to date? There is no mention of polars at all in the type-mapping:

https://docs.flyte.org/en/latest/user_guide/data_types_and_io/index.html#python-to-flyte-type-mapping

kumare3:
Yes @HansBambel - would you mind adding an update

HansBambel:
@kumare3 Yes, opened a PR for this.

HansBambel:
For better understanding: Even though `LazyFrame` is supported, using it across tasks does not yield any benefits such as low memory usage and lazy evaluation as `StructuredDataset` calls `.collect()` to serialize it.

This results in using `LazyFrame` over `DataFrame` when dealing with huge files that do not fit in memory useless, right?

cosmicBboy:
@HansBambel correct. IIRC streaming a LazyFrame to blob store was not supported at the time of this PR (I believe this is still the case). Tried using `sink_parquet`: https://docs.pola.rs/api/python/stable/reference/api/polars.LazyFrame.sink_parquet.html but it's not considered stable and didn't work with fsspec (neither did `scan_parquet`).

PR #2696, opened 2024-08-21
fix: cover 500 errors for new Harbor repos

dylanspag-lmco:

ImageSpec builders currently do not support pushing images to repositories on Harbor. For new repositories (i.e. ones that do not exist yet on the registry), the `APIError` except block in `ImageSpec.exist` only checks for 404 errors before returning `False`. However, unlike other registries such as GitLab's, Harbor raises a 500 error not a 404. Therefore, the existence check fails and returns `None` which leads `ImageSpec` to conclude that the image should not be built.

In `ImageSpec.exist` I added an additional check after the 404 status code check that compares the error's explanation against a simple regex pattern aligned with what Harbor produces in this case. If there is a match, then the method returns `False`. 

## How was this patch tested?

Utilizing an on-prem Harbor instance, I first attempted to run `pyflyte package` without the patch and got this (sanitized) output:

```
Failed to check if the image exists with error:
 500 Server Error for http+docker://localhost/v1.46/distribution/<image-name>/json: Internal Server Error ("unknown: repository <image-name-without-fqdn-or-tag> not found")
Flytekit assumes the image <image-name> already exists.
```

After introducing the patch, I ran the exact same command and got the following output:

```
Image <image-name> not found. building...
<docker-build-and-push-logs>
```

I then let it finish pushing the image and ran the exact same command again (`ImageSpec` has `tag_format` set to a fixed value) which produced this output:

```
Image <image-name> found. Skip building.
```

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -145,6 +145,7 @@ class ImageSpec:
                 return False
 
             if re.match(f"unknown: repository .*{self.name} not found", e.explanation):
+                click.secho(f"Received 500 error with explanation: {e.explanation}", fg="yellow")
                 return False
 
             click.secho(f"Failed to check if the image exists with error:\n {e}", fg="red")
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -3,6 +3,7 @@ import copy
 import hashlib
 import os
 import pathlib
+import re
 import typing
 from abc import abstractmethod
 from dataclasses import asdict, dataclass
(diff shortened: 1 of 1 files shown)


flytekit/image_spec/image_spec.py
```diff
@@ -143,6 +144,9 @@ def exist(self) -> Optional[bool]:
             if e.response.status_code == 404:
                 return False
 
+            if re.match(f"unknown: repository .*{self.name} not found", e.explanation):
```
wild-endeavor, line 147: don't we want to show the error message as well?

wild-endeavor (commented):
thank you for the contribution!

In the long run though, the harbor api will change right?  5xx is not really the right return code.

flytekit/image_spec/image_spec.py
dylanspag-lmco: I just added a print out with the explanation. Let me know if looks okay.

wild-endeavor (approved):

PR #2711, opened 2024-08-26
Make task loading a part of dispatch execute

thomasjpfan:

Currently, if `load_task` raises an error, it is not captured by `_dispatch_execute`. For example, `load_task` can raise import errors if the container environment is not consistent with the registration environment.

With this PR, the error will be captured by flytekit and written to `error.pb`.

## How was this patch tested?

I added a unit test to check this change. The other unit tests were updated because `_dispatch_execute`'s signature changed.

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -10,7 +10,7 @@ import sys
 import tempfile
 import traceback
 from sys import exit
-from typing import List, Optional
+from typing import Callable, List, Optional
 
 import click
 from flyteidl.core import literals_pb2 as _literals_pb2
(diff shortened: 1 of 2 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -318,18 +331,6 @@ def setup_execution(
         yield ctx
 
 
-def _handle_annotated_task(
```
pingsutw, line 321: I also want to remove this; thanks!

wild-endeavor (approved):

flytekit/bin/entrypoint.py
wild-endeavor: same

PR #2720, opened 2024-08-29
Update eager: make sure client secret can be specified as env var

cosmicBboy:

Currently, the `@eager` workflows are too strict with its handling of the client secret values. Both secret group and key need to be specified for the secret request to be passed down to the underlying task. There is also an issue with the async event loop not being available for the downstream data persistence step in the task execution.

This PR loosens this requirement to be compatible with Flyte backends that do not require secret groups. This PR also:

- Allows the user to bind the client secret to an environment variable in cases where the `FlyteRemote` object can authenticate with just the client secret
- Makes sure that an async loop is available for downstream data persistence step in the task execution

## How was this patch tested?

The following workflow was tested on flyte sandbox
```python

flytekit = "git+https://github.com/flyteorg/flytekit@54c4471a56311ff8201ba76d9bda726bc5b4689b"

image = ImageSpec(
    name="polars",
    packages=["flytekit", flytekit],
    registry="localhost:30000",
    apt_packages=["git"],
)


@task
def add_one(x: int) -> int:
    return x + 1


@task
def double(x: int) -> int:
    return x * 2


@eager(
    container_image=image,
    remote=FlyteRemote(
        config=Config.for_sandbox(),
        default_domain="development",
    ),
)
    out = await add_one(x=x)
    if out < 0:
        return -1
    return await double(x=out)

```

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -68,6 +68,15 @@ def _compute_array_job_index():
     return offset
 
 
+def _reset_event_loop_if_needed():
+    """Create new event loop if it doesn't exist."""
+    try:
+        asyncio.get_event_loop()
+    except RuntimeError:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+
+
 def _dispatch_execute(
     ctx: FlyteContext,
     load_task: Callable[[], PythonTask],
@@ -107,13 +116,14 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
+
+            # make sure event loop exists before running the coroutine
+            _reset_event_loop_if_needed()
+
             outputs = asyncio.run(outputs)
+
             # make sure an event loop exists for data persistence step
-            try:
-                asyncio.get_event_loop()
-            except RuntimeError:
-                loop = asyncio.new_event_loop()
-                asyncio.set_event_loop(loop)
+            _reset_event_loop_if_needed()
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -177,13 +177,6 @@ def _dispatch_execute(
     for k, v in output_file_dict.items():
         utils.write_proto_to_file(v.to_flyte_idl(), os.path.join(ctx.execution_state.engine_dir, k))
 
-    # make sure an event loop exists for data persistence step
-    try:
-        asyncio.get_event_loop()
-    except RuntimeError:
-        loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(loop)
-
     ctx.file_access.put_data(ctx.execution_state.engine_dir, output_prefix, is_multipart=True)
     logger.info(f"Engine folder written successfully to the output prefix {output_prefix}")
 
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -108,6 +108,9 @@ def _dispatch_execute(
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
             outputs = asyncio.run(outputs)
+            # make sure an event loop exists for data persistence step
+            loop = asyncio.new_event_loop()
+            asyncio.set_event_loop(loop)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -180,7 +180,7 @@ def _dispatch_execute(
     # make sure an event loop exists for data persistence step
     try:
         asyncio.get_event_loop()
-    except Exception:
+    except RuntimeError:
         loop = asyncio.new_event_loop()
         asyncio.set_event_loop(loop)
 
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -109,8 +109,11 @@ def _dispatch_execute(
             logger.info("Output is a coroutine")
             outputs = asyncio.run(outputs)
             # make sure an event loop exists for data persistence step
-            loop = asyncio.new_event_loop()
-            asyncio.set_event_loop(loop)
+            try:
+                asyncio.get_event_loop()
+            except RuntimeError:
+                loop = asyncio.new_event_loop()
+                asyncio.set_event_loop(loop)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -108,12 +108,6 @@ def _dispatch_execute(
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
             outputs = asyncio.run(outputs)
-            # make sure an event loop exists for data persistence step
-            try:
-                asyncio.get_event_loop()
-            except RuntimeError:
-                loop = asyncio.new_event_loop()
-                asyncio.set_event_loop(loop)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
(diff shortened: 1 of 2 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -177,6 +177,13 @@ def _dispatch_execute(
     for k, v in output_file_dict.items():
         utils.write_proto_to_file(v.to_flyte_idl(), os.path.join(ctx.execution_state.engine_dir, k))
 
+    # make sure an event loop exists for data persistence step
```
cosmicBboy, line 180: this issue comes up not just with eager. Frameworks like Langchain that have async functionality will cause the event loop to end before the flyte task can run the async data persistence operation.

flytekit/bin/entrypoint.py
```diff
@@ -177,6 +177,13 @@ def _dispatch_execute(
...
+    try:
+        asyncio.get_event_loop()
+    except Exception:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
```
Future-Outlier, line 185: Should we specify which exception we might have?

Future-Outlier (commented):
thank you for this!
can we have
1. register the task to remote cluster (sandbox or dogfood gcp) to prove it works?
2. add workflow in unit tests.

flytekit/bin/entrypoint.py
cosmicBboy: ah yes this should be `RuntimeError`

cosmicBboy:
> thank you for this! can we have
> 
> 1. register the task to remote cluster (sandbox or dogfood gcp) to prove it works?
> 2. add workflow in unit tests.

Unit test for this exist [here](https://github.com/flyteorg/flytekit/tree/master/tests/flytekit/integration/experimental) and [here](https://github.com/flyteorg/flytekit/tree/master/tests/flytekit/unit/experimental). The changes made to the PR (adding the new `client_secret_env_var` argument) and loosening the strictness of secrets only really applies to Union serverless.

The code snippet in the PR description should work on flyte sandbox
image
image

Future-Outlier (dismissed):
LGTM, thank you

flytekit/experimental/eager_function.py
```diff
@@ -605,16 +617,25 @@ def _internal_demo_remote(remote: FlyteRemote) -> FlyteRemote:
...
     client_secret = secrets_manager.get(client_secret_group, client_secret_key)
     # get the raw output prefix from the context that's set from the pyflyte-execute entrypoint
     # (see flytekit/bin/entrypoint.py)
+
+    if client_secret_env_var is not None:
+        # this creates a remote client where the env var client secret is sufficient for authentication
+        os.environ[client_secret_env_var] = client_secret
+        try:
+            remote_cls = type(remote)
+            return remote_cls()
```
thomasjpfan, line 635: This assumes that the other `FlyteRemote` object can be created without a `Config` and uses an environment variable. Is this okay for OSS?

Also, should this carry over the `default_domain` and `default_project` from the original remote obj? Or should the other remote object figure out the domain and project from the execution environment?

cosmicBboy: ooo yeah it should probably carry over those args.

admittedly this code path is basically: "tell me you only run on serverless without saying you run on serverless" 🙃

cosmicBboy: Basically `client_secret_env_var` only really applies to serverless

flytekit/experimental/eager_function.py
```diff
@@ -533,8 +542,10 @@ async def wrapper(*args, **kws):
                 await cleanup_fn()
 
     secret_requests = kwargs.pop("secret_requests", None) or []
-    if client_secret_group is not None and client_secret_key is not None:
+    try:
         secret_requests.append(Secret(group=client_secret_group, key=client_secret_key))
+    except ValueError:
+        pass
```
wild-endeavor, line 548: what if there's nothing in the client_secret_env_var environment variable?

flytekit/experimental/eager_function.py
```diff
@@ -396,6 +398,8 @@ def eager(
     :param local_entrypoint: If True, the eager workflow will can be executed locally but use the provided
         :py:func:`~flytekit.remote.FlyteRemote` object to create task/workflow executions. This is useful for local
         testing against a remote Flyte cluster.
+    :param client_secret_env_var: if specified, binds the client secret to the specified environment variable for
```
wild-endeavor, line 401: okay with this for now, but I think we should see if we can't move this to some higher level in the future.  the secret here we're talking about is the secret for hitting admin itself right?

flytekit/experimental/eager_function.py
```diff
@@ -605,16 +617,28 @@ def _internal_demo_remote(remote: FlyteRemote) -> FlyteRemote:
...
-    assert client_secret_group is not None, "secret_group must be defined when using a remote cluster"
-    assert client_secret_key is not None, "secret_key must be defined a remote cluster"
     secrets_manager = current_context().secrets
     client_secret = secrets_manager.get(client_secret_group, client_secret_key)
     # get the raw output prefix from the context that's set from the pyflyte-execute entrypoint
     # (see flytekit/bin/entrypoint.py)
+
+    if client_secret_env_var is not None:
+        # this creates a remote client where the env var client secret is sufficient for authentication
+        os.environ[client_secret_env_var] = client_secret
```
wild-endeavor, line 632: confused what's happening here.  if we're removing the asserts, then that means the group/key might be None.  If they're none, won't the secrets_manager.get fail?  So then what is the client_secret?  Or is there some default in the secrets manager that somehow gets returned?
I don't get why we're setting the environment variable here to the value of client secret.

wild-endeavor (commented):
some questions, but this event loop thing... can you explain why it's needed in entrypoint.py but there doesn't seem to be another one for local execution?  Won't both need it?

flytekit/experimental/eager_function.py
cosmicBboy: `client_secret_env_var` doesn't figure into this try except block, or am I missing something?

flytekit/experimental/eager_function.py
cosmicBboy: this is to make it easy to authenticate with `UnionRemote` on serverless (via `UNION_SERVERLESS_API_KEY ` env var). Secrets on serverless only needs a secret key (no group).

flytekit/experimental/eager_function.py
wild-endeavor: no yeah i know, i mean, what happens if all three (`client_secret_env_var`, `client_secret_group` and `client_secret_key`) are None?

flytekit/experimental/eager_function.py
wild-endeavor: just wondering, in the following scenario:
* `client_secret_key` is  `None`
* `client_secret_group` is `None`
(and this scenario is now valid because the asserts are now removed)
after line 626 `client_secret = secrets_manager.get(client_secret_group, client_secret_key)`
what is the value of `client_secret`?

wild-endeavor: going through the secretsmanager, it looks like it raises an IndexError if you call `get(group=None, key=None)`, so I don't think it ever gets past line 626.  Could we move the assertion that at least one of those has be be non-None into the `get` function?

wild-endeavor (commented):
Had a few more comments and went over this with @eapolinario.  Could we
* Add an assert to secretsmanager.get that at least one of group/key should be non-None
* Could we move the async loop stuff to another PR?  I think it'd be helpful to discuss it separately.  Also do you remember how you triggered the issue?  I wanna repro locally to understand more what langchain is doing.  Ideally we handle in such a way that it's completely independent of any other library.

cosmicBboy:
> Had a few more comments and went over this with @eapolinario. Could we
> 
> * Add an assert to secretsmanager.get that at least one of group/key should be non-None

Updated PR with client secret key/group assertion.

> * Could we move the async loop stuff to another PR?  I think it'd be helpful to discuss it separately.  Also do you remember how you triggered the issue?  I wanna repro locally to understand more what langchain is doing.  Ideally we handle in such a way that it's completely independent of any other library.

Deleted this. For some reason I'm unable to repro the async issue I was seeing earlier on eager or with langchain async document loaders https://github.com/unionai-oss/union-rag/blob/main/union_rag/langchain.py#L138-L141

wild-endeavor (approved):

PR #2724, opened 2024-08-30
Better error for min_success_ratio<1

pingsutw:
## Tracking issue
NA

when min_success_ratio < 1, flytekit automatically makes the output interface optional (str => Optional[str])
If the input of the downstream task is not optional, Flytekit will fail to deserialize the output and display a confusing error message

check the type at compile time

### Setup process
```
pyflyte run flyte-example/map_optional_list.py wf
```

```python
import os
import tempfile
from typing import List

from flytekit import map_task, workflow, task


@task
def generate_directory(word: str) -> FlyteDirectory:
    temp_dir1 = tempfile.TemporaryDirectory()
    with open(os.path.join(temp_dir1.name, "file.txt"), "w") as tmp:
        tmp.write(f"Hello world {word}!\n")
    return FlyteDirectory(path=temp_dir1.name)


@task
def consume_directories(dirs: List[FlyteDirectory]):
    for d in dirs:
        print(f"Directory: {d.path} {d._remote_source}")
        for path_info, other_info in d.crawl():
            print(path_info)


mt = map_task(generate_directory, min_success_ratio=0.1)


@workflow
def wf():
    dirs = mt(word=["one", "two", "three"])
    consume_directories(dirs=dirs
```

### Screenshots
Before:
Screenshot 2024-09-03 at 11 47 01 AM

After:
Screenshot 2024-09-03 at 11 46 00 AM

## Related PRs
NA

## Docs link
NA


tests/flytekit/unit/core/test_array_node_map_task.py
```diff
@@ -435,3 +437,28 @@ def test_wf():
...
+    def generate_directory(word: str) -> FlyteDirectory:
+        temp_dir1 = tempfile.TemporaryDirectory(delete=False)
+        with open(os.path.join(temp_dir1.name, "file.txt"), "w") as tmp:
+            tmp.write(f"Hello world {word}!\n")
+        return FlyteDirectory(path=temp_dir1.name)
+
+    @task
+    def consume_directories(dirs: List[FlyteDirectory]):
+        for d in dirs:
+            print(f"Directory: {d.path} {d._remote_source}")
```
wild-endeavor, line 453: nix this

wild-endeavor: oh this won't run nm, keep

wild-endeavor (dismissed):

wild-endeavor:
i think it's good to have this...  i don't want flytekit to fall into the trap of duplicating the compiler, but this is a common  and tricky enough case that I think it's okay to explicitly catch this.

wild-endeavor (approved):

PR #2726, opened 2024-09-02
Fix: Prevent UnionTransformer type ambiguity in combination with PyTorchTypeTransformer

fg91:

```py
@task
def foo() -> Union[list[torch.Tensor], torch.Tensor]:
    return [torch.tensor([1, 2, 3])]


@workflow
def wf():
    foo()
```

This example fails with:

```console
TypeError: Failed to convert outputs of task 'wf.foo' at position 0.
Failed to convert type <class 'list'> to type typing.Union[list[torch.Tensor], torch.Tensor].
Error Message: Ambiguous choice of variant for union type.
```

The type transformer choice should, however, not be ambiguous. 

The problem arises as follows:

[Here](https://github.com/flyteorg/flytekit/blob/839f555f497c29bac6f12897697a5a93a182fc8d/flytekit/core/type_engine.py#L1526) in the `UnionTransformer` we test all union variants and raise the observed error if more than one of the union variants work.

In this case, the two union working variants are:

* Call the `ListTransformer` for a list of tensors and then the `PyTorchTensorTransformer` for each element in the list. This is the expected behaviour and does work.
* Call the `PyTorchTensorTransformer` for the list of tensors. This, however, works as well as `torch.save([some_tensor], ...)` works. This is not intended and hence causes the ambiguity.

We need to prevent the pytorch type transformers from serializing lists of tensors or modules since the type engine (i.e. the `UnionTransformer` expects these cases to be handled by the `ListTransformer`).

## How was this patch tested?

Tested with the snippet above and added a unit test that would have caught this.

### DIFF
--- a/flytekit/extras/pytorch/native.py
+++ b/flytekit/extras/pytorch/native.py
@@ -28,6 +28,9 @@ class PyTorchTypeTransformer(TypeTransformer[T]):
         python_type: Type[T],
         expected: LiteralType,
     ) -> Literal:
+        if not isinstance(python_val, torch.Tensor) and not isinstance(python_val, torch.nn.Module):
+            raise TypeTransformerFailedError("Expected a torch.Tensor or nn.Module")
+
         meta = BlobMetadata(
             type=_core_types.BlobType(
                 format=self.PYTORCH_FORMAT,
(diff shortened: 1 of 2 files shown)


thomasjpfan (commented):
Can we include one more unit test based on your example in https://github.com/flyteorg/flytekit/pull/2726#issue-2501201315 to make sure this continues to work?

fg91:
> ```python
> 
> @workflow
> def wf():
>     foo()
> ```

Absolutely https://github.com/flyteorg/flytekit/pull/2726/commits/99df4fe20dd8b9042c876f444eba4f3f83eec6e5

thomasjpfan (approved):

PR #2729, opened 2024-09-04
Feat: Optionally use pigz to speed up tarball compression

fg91:

In an (admittedly large) repository, fast registration has become quite slow for us. The culprit appears to be the compression of the code tarball which I have observed to take up to a minute. (I'm already ignoring non-relevant parts of the code base using a `.flyteignore`.)

In the fast package helper, where we currently compress the code tarball using cpython's `gzip`, I propose to check whether the user has [`pigz`](https://zlib.net/pigz/) installed and if yes, leverage it to speed up the compression.

In our case, I'm seeing a speedup from ~60s to ~1-2s.

I expect that the majority of users will not run into these performance issues but for those that might, I propose to add a log message with a hint to install `pigz` in case the default compression mechanism takes to long.

I see that you recently refactored the fast registration logic @wild-endeavor, what do you think about this proposal?

### DIFF
--- a/flytekit/tools/fast_registration.py
+++ b/flytekit/tools/fast_registration.py
@@ -152,8 +152,9 @@ def fast_package(
                 end_time = time.time()
                 warning_time = 30
                 if end_time - start_time > warning_time:
-                    logger.info(
-                        f"Code tarball compression took {end_time - start_time:.0f} seconds. Consider installing `pigz` for faster compression."
+                    click.secho(
+                        f"Code tarball compression took {end_time - start_time:.0f} seconds. Consider installing `pigz` for faster compression.",
+                        fg="yellow",
                     )
 
     # Original tar command - This condition to be removed in the future.
@@ -188,8 +189,9 @@ def fast_package(
                 end_time = time.time()
                 warning_time = 30
                 if end_time - start_time > warning_time:
-                    logger.info(
-                        f"Code tarball compression took {end_time - start_time:.0f} seconds. Consider installing `pigz` for faster compression."
+                    click.secho(
+                        f"Code tarball compression took {end_time - start_time:.0f} seconds. Consider installing `pigz` for faster compression.",
+                        fg="yellow",
                     )
 
     return archive_fname
--- a/flytekit/tools/fast_registration.py
+++ b/flytekit/tools/fast_registration.py
@@ -77,6 +77,26 @@ def print_ls_tree(source: os.PathLike, ls: typing.List[str]):
     rich_print(tree_root)
 
 
+def compress_tarball(source: os.PathLike, output: os.PathLike) -> None:
+    """Compress code tarball using pigz if available, otherwise gzip"""
+    if pigz := shutil.which("pigz"):
+        with open(output, "wb") as gzipped:
+            subprocess.run([pigz, "-c", source], stdout=gzipped, check=True)
+    else:
+        start_time = time.time()
+        with gzip.GzipFile(filename=output, mode="wb", mtime=0) as gzipped:
+            with open(source, "rb") as source_file:
+                gzipped.write(source_file.read())
+
+        end_time = time.time()
+        warning_time = 10
+        if end_time - start_time > warning_time:
+            click.secho(
+                f"Code tarball compression took {end_time - start_time:.0f} seconds. Consider installing `pigz` for faster compression.",
+                fg="yellow",
+            )
+
+
 def fast_package(
     source: os.PathLike,
     output_dir: os.PathLike,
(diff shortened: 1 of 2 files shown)


flytekit/tools/fast_registration.py
```diff
@@ -139,9 +141,21 @@ def fast_package(
...
+                        gzipped.write(tar_file.read())
+
+                end_time = time.time()
+                warning_time = 30
+                if end_time - start_time > warning_time:
+                    logger.info(
+                        f"Code tarball compression took {end_time - start_time:.0f} seconds. Consider installing `pigz` for faster compression."
+                    )
 
     # Original tar command - This condition to be removed in the future.
```
fg91, line 160: Not removing code duplication because the else branch appears to be removed soon anyways.

flytekit/tools/fast_registration.py
```diff
@@ -139,9 +141,21 @@ def fast_package(
...
+            else:
+                start_time = time.time()
+                with gzip.GzipFile(filename=archive_fname, mode="wb", mtime=0) as gzipped:
+                    with open(tar_path, "rb") as tar_file:
+                        gzipped.write(tar_file.read())
+
+                end_time = time.time()
+                warning_time = 30
+                if end_time - start_time > warning_time:
+                    logger.info(
```
pingsutw, line 156: Should we use `click.secho` instead? we don't print `logger.info` by default

pingsutw (approved):
omg, this is amazing!

flytekit/tools/fast_registration.py
fg91: https://github.com/flyteorg/flytekit/pull/2729/commits/1f8018134616fb97d47e5436665131eb581c8d43

flytekit/tools/fast_registration.py
```diff
@@ -139,9 +141,21 @@ def fast_package(
...
+            if shutil.which("pigz"):
+                subprocess.run([f"pigz -c {tar_path} > {archive_fname}"], shell=True, check=True)
+            else:
+                start_time = time.time()
+                with gzip.GzipFile(filename=archive_fname, mode="wb", mtime=0) as gzipped:
+                    with open(tar_path, "rb") as tar_file:
+                        gzipped.write(tar_file.read())
+
+                end_time = time.time()
+                warning_time = 30
```
pingsutw, line 153:
```suggestion
                warning_time = 10
```

fg91: Already set to 10s in this commit: https://github.com/flyteorg/flytekit/pull/2729/commits/908e1a6161bc61aec76d40386ca410c82f785b0c

flytekit/tools/fast_registration.py
```diff
@@ -139,9 +141,21 @@ def fast_package(
                         filter=lambda x: tar_strip_file_attributes(x),
                     )
 
-            with gzip.GzipFile(filename=archive_fname, mode="wb", mtime=0) as gzipped:
-                with open(tar_path, "rb") as tar_file:
-                    gzipped.write(tar_file.read())
+            if shutil.which("pigz"):
+                subprocess.run([f"pigz -c {tar_path} > {archive_fname}"], shell=True, check=True)
```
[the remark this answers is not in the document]
fg91: You mean like before https://github.com/flyteorg/flytekit/pull/2729/commits/75c534ec83dcf0f7c10af181910bffa4e60542cf?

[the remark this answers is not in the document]
fg91: https://github.com/flyteorg/flytekit/pull/2729/commits/121607b373b2657f88755d8bc5263823f99fa412

flytekit/tools/fast_registration.py
```diff
@@ -139,9 +141,22 @@ def fast_package(
                         filter=lambda x: tar_strip_file_attributes(x),
                     )
 
-            with gzip.GzipFile(filename=archive_fname, mode="wb", mtime=0) as gzipped:
-                with open(tar_path, "rb") as tar_file:
-                    gzipped.write(tar_file.read())
+            if shutil.which("pigz"):
+                with open(archive_fname, "wb") as gzipped:
+                    subprocess.run(["pigz", "-c", tar_path], stdout=gzipped, check=True)
```
thomasjpfan, line 146: Can we store the output of `shutil.which` and use it here?

thomasjpfan (commented):
@fg91 Do you think it's worth refactoring the taring code path into a helper function?

flytekit/tools/fast_registration.py
pingsutw: Hmm, you change that at line 192. Don't we need to change it here as well?

flytekit/tools/fast_registration.py
fg91: https://github.com/flyteorg/flytekit/pull/2729/commits/d32822314c401205ee93b7801dcbd53af9301775

fg91:
> @fg91 Do you think it's worth refactoring the taring code path into a helper function?

Done https://github.com/flyteorg/flytekit/pull/2729/commits/d32822314c401205ee93b7801dcbd53af9301775

flytekit/tools/fast_registration.py
fg91: You are right, sorry. Now obsolete as the logic exists only a single time in a helper function https://github.com/flyteorg/flytekit/pull/2729/commits/d32822314c401205ee93b7801dcbd53af9301775

wild-endeavor (commented):
If it's identical, can you add a note to `clis.rst` as well in the fast register section suggesting that people download this tool?  I'm not sure how we never knew about this, sad.  Thanks for enlightening us!

wild-endeavor:
let's get this in and we can all test it.  we're shooting for a beta release soon with some larger changes, possibly some that is a bit backwards incompatible in some edge cases, so we will do a longer testing period.

fg91:
> If it's identical, can you add a note to `clis.rst` as well in the fast register section suggesting that people download this tool? I'm not sure how we never knew about this, sad. Thanks for enlightening us!

https://github.com/flyteorg/flytekit/pull/2729/commits/a1f6418cdd3436e5b61d9856404913cbeec099a3

wild-endeavor (approved):

PR #2715, opened 2024-08-27
[Flytekit] Support extra copy commands in ImageSpec

mao3267:
## Tracking issue
Initially, we aimed to support adding custom Docker commands when creating images with ImageSpec. However, after discussions with @thomasjpfan and @pingsutw in #2676 , we prioritized supporting additional COPY commands first, as they were deemed more important.

## Why are the changes needed?
When creating images with ImageSpec on a remote cluster, we may need to copy additional files or directories required during the container build process. To address this, we aim to support extra copy commands in ImageSpec for copying files and directories into the container’s `/root` directory.

## What changes were proposed in this pull request?
1. Add the `ImageSpec.with_copy(src)` function to the ImageSpec class, allowing the inclusion of additional copy commands. The src parameter can be a single path string or a list of path-like strings, specifying files or directories to be copied.

- Example Usage:

`input.json`
```json
{
    "b": "Hello"
}
```

`build_image.py`
```py
from flytekit.image_spec import ImageSpec
from flytekit import task, workflow
import json

image_spec = ImageSpec(
    name="test_image",
    registry="localhost:30000",
    builder="default",
    apt_packages=["git", "gh"],
    # For remote testing before merged
    env={"GH_TOKEN": "<YOUR_GITHUB_TOKEN>"},
    commands=[
        "git clone https://github.com/flyteorg/flytekit.git", 
        "cd flytekit", 
        "gh pr checkout 2715", 
        "pip install -e ."
    ]
)

image_spec.with_copy("local_test/input.json")
# image_spec.with_copy(["local_test/input.json", "local_test/input.yaml"]

@task(container_image=image_spec)
def my_task() -> str:
    with open("/root/local_test/input.json", "r") as f:
        data = json.load(f)
        return data["b"]


@workflow
def my_wf() -> str:
    return my_task()
```

2. Support copying files or directories into `/root` in the container by setting `ImageSpec.copy`, a list of path strings.

- Example usage:

`build_image.py`
```py
from flytekit.image_spec import ImageSpec
from flytekit import task, workflow
import json

image_spec = ImageSpec(
    name="test_image",
    registry="localhost:30000",
    builder="default",
    apt_packages=["git", "gh"],
    copy=["local_test/input.json"],
    # For remote testing before merged
    env={"GH_TOKEN": "<YOUR_GITHUB_TOKEN>"},
    commands=[
        "git clone https://github.com/flyteorg/flytekit.git", 
        "cd flytekit", 
        "gh pr checkout 2715", 
[...]

PR #2731, opened 2024-09-05
Imgspec/copy auto

wild-endeavor:
    ImageSpec, added a copy option. Note that this is the first time ImageSpec has a non-primitive option. Think it's
    okay but wanted to be aware of it.
    
    NB the tag property changed based on when the new update function is called in relation to it. Also true for the
      id property. Is there a way around this?  We can make update_image_spec_copy_handling not mutate, but that also
      doesn't work since tag/id need correct source_root/copy in order to correctly compute the tag.
      Basically the question is how to make sure serialization settings has been taken into account before these
      two properties are called. 
    
    guidelines for how to raise errors while developing flytekit
    
    moved the logic that modifies ImageSpec common to ContainerTask and PythonAutoContainerTask to a helper function
    
    deref symlink behavior
    
    when you do non-fast register, package, serialize, with ImageSpec
        * by default, the entire source root is copied into the image
        * source root of IS, is set to SS.local source root

    (add copy auto functionality)
    non-fast register but copy auto for image spec
    
    (make sure this is still true)
    when you do --copy none register, package, serialize, with ImageSpec
        * by default, the entire source root is copied into the image

    
    when you do fast register, package, serialize, run
        * no files copied into the image
    

### DIFF
--- a/flytekit/core/python_auto_container.py
+++ b/flytekit/core/python_auto_container.py
@@ -8,6 +8,7 @@ from typing import Callable, Dict, List, Optional, TypeVar, Union
 from flyteidl.core import tasks_pb2
 
 from flytekit.configuration import ImageConfig, SerializationSettings
+from flytekit.constants import CopyFileDetection
 from flytekit.core.base_task import PythonTask, TaskMetadata, TaskResolverMixin
 from flytekit.core.context_manager import FlyteContextManager
 from flytekit.core.pod_template import PodTemplate
@@ -185,10 +186,10 @@ class PythonAutoContainerTask(PythonTask[T], ABC, metaclass=FlyteTrackedABC):
         return self._get_command_fn(settings)
 
     def get_image(self, settings: SerializationSettings) -> str:
-        if settings.fast_serialization_settings is None or not settings.fast_serialization_settings.enabled:
-            if isinstance(self.container_image, ImageSpec):
-                # Set the source root for the image spec if it's non-fast registration
-                self.container_image.source_root = settings.source_root
+        """Update image spec based on fast registration usage, and return string representing the image"""
+        if isinstance(self.container_image, ImageSpec):
+            update_image_spec_copy_handling(self.container_image, settings)
+
         return get_registerable_container_image(self.container_image, settings.image_config)
 
     def get_container(self, settings: SerializationSettings) -> _task_model.Container:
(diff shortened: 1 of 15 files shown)


flytekit/core/python_auto_container.py
```diff
@@ -273,6 +274,37 @@ def get_all_tasks(self) -> List[PythonAutoContainerTask]:  # type: ignore
...
+
+    # Handle the default behavior of setting the behavior based on the inverse of fast register usage
+    # The default behavior additionally requires that serializa
+    elif settings.fast_serialization_settings is None or not settings.fast_serialization_settings.enabled:
+        # Set the source root for the image spec if it's non-fast registration
+        # Unfortunately whether the source_root/copy instructions should be set is implicitly dependent also on the
+        # existence of the source root in settings.
+        if settings.source_root is not None or image_spec.source_root is not None:
+            if image_spec.source_root is None:
+                image_spec.source_root = settings.source_root
```
pingsutw, line 303:
```suggestion
              image_spec.source_root = settings.source_root
```

flytekit/image_spec/default_builder.py
```diff
@@ -165,16 +167,32 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
...
-            dirs_exist_ok=True,
-        )
+
+        if image_spec.copy == CopyFileDetection.LOADED_MODULES:
+            # This is the 'auto' semantic by default used for pyflyte run, it only copies loaded .py files.
+            sys_modules = list(sys.modules.values())
+            ls, _ = ls_files(str(image_spec.source_root), sys_modules, deref_symlinks=False, ignore_group=ignore)
+        else:
+            # This triggers listing of all files
+            ls, _ = ls_files(str(image_spec.source_root), [], deref_symlinks=False, ignore_group=ignore)
```
pingsutw, line 186: could we update ls_files to 
```python
def ls_files(
    source_path: str,
    copy_file_detection: CopyFileDetection,
    deref_symlinks: bool = False,
    ignore_group: Optional[IgnoreGroup] = None,
) -> Tuple[List[str], str]:
   ...
```
then, we can change the code here to
```suggestion
        ls, _ = ls_files(str(image_spec.source_root), image_spec.copy, deref_symlinks=False, ignore_group=ignore)
```

flytekit/image_spec/image_spec.py
```diff
@@ -74,13 +84,19 @@ class ImageSpec:
     entrypoint: Optional[List[str]] = None
     commands: Optional[List[str]] = None
     tag_format: Optional[str] = None
+    copy: Optional[CopyFileDetection] = None
```
pingsutw, line 87: This [PR](https://github.com/flyteorg/flytekit/pull/2715/files) already uses `copy`. should we use the other name?

flytekit/core/python_auto_container.py
wild-endeavor: why?  what if it's set manually by the user in the constructor of ImageSpec?

flytekit/image_spec/image_spec.py
wild-endeavor: can you suggest a good name?

flytekit/image_spec/default_builder.py
wild-endeavor: yeah let me do this tomorrow.

flytekit/core/python_auto_container.py
wild-endeavor: and if users are not supposed to set that then we should remove it from the constructor.  If users are allowed to set it, we should add a comment that says this means it'll rebuild every time those files change.

pingsutw: oops, sorry. my bad. I thought line 301 was
```
if settings.source_root is not None and image_spec.source_root is not None:
```

flytekit/image_spec/image_spec.py
wild-endeavor: yeah copy is bad because the `--copy` switch on the command line has the opposite effect.  --copy none on the command line means it should copy files into the image.

wild-endeavor: how about `source_copy_mode` and `addl_copies` for the other pr?

pingsutw: These names sound good to me. 👍

flytekit/image_spec/image_spec.py
```diff
@@ -51,12 +52,20 @@ class ImageSpec:
...
+                  Python files into the image.
+
+            If the option is set by the user, then that option is of course used.
     """
 
     name: str = "flytekit"
     python_version: str = None  # Use default python in the base image if None.
     builder: Optional[str] = None
-    source_root: Optional[str] = None
+    source_root: Optional[str] = None  # a.txt:auto
```
pingsutw, line 68: `a.txt:auto`. is there a plan to support this?
it looks nicer honestly

wild-endeavor: separate pr?

pingsutw (approved):

PR #2276, opened 2024-03-18
pyflyte `run`& `register` asynchronously

austin362667:
## Tracking issue

NA

## Why are the changes needed?

To make register  `task`, `workflow`, and `launchplan` **x2.5 ~ x4 faster**  in `pyflyte run` and  `pyflyte register` by leveraging `asyncio`.

## What changes were proposed in this pull request?

- https://github.com/flyteorg/flytekit/pull/2267: A neat workaround by @pingsutw for sync grpc clients.
- Entities serialization is not enough in async approach.
    - It's necessary to separate registered entities into "task" and "non-task" categories in orders. This is because we can ensure flyteadmin returns a task registration response before registering workflows that depend on it.

### Setup process

After introducing localhost network latency to `1000ms` per request, we can simulate network IO-intensive operations to cloud services (e.g., object storage, Kubernetes cluster, FlyteAdmin) on your own desktop. This way, we can finally evaluate performance improvements through `asyncio` without setup a real cluster outside. Otherwise, the Flyte sandbox cluster residing in macOS is just too fast to measure.

Steps to reproduce network condition:

1. `sudo pfctl -f -`: packfilter out proto tcp from any to any pipe `1`.
2. `sudo dnctl pipe 1 config delay 1000`: Adding 1000ms latency to every dummynet requests.
3. `sudo pfctl -E`: Enabling the latency setup in packfilter.

> Don't forget set latency back to `0` to get your original network speed.

> Screenshot 2024-03-18 at 3 15 17 PM
  
### Screenshots

- The registration elapsed time reduced from `34.1s` to `16.1s` after introducing asynchronous registration.
- Register, 9 `task`, 1 `wf`, 1 `lp`.

#### 1. Register in the `sync` manner: **`21.1s`**

  1. `python3 -m cProfile -o sync_requests_500ms.prof ./flytekit/clis/sdk_in_container/pyflyte.py register ./workflow.py`
  2. `snakeviz ./sync_requests_500ms.prof`

  - Screenshot 2024-03-23 at 7 43 15 AM

#### 2. Register in the `async` manner: **`6.05s`**
  1. `python3 -m cProfile -o async_requests_500ms.prof ./flytekit/clis/sdk_in_container/pyflyte.py register ./workflow.py`
  2. `snakeviz ./async_requests_500ms.prof`
  
  - Screenshot 2024-03-23 at 7 44 57 AM

#### 3. Run in the `async` manner: **`5.95s`**
[...]

PR #2733, opened 2024-09-06
Pickle remote task for Jupyter Notebook Environment

Mecoli1219:
## Tracking issue
https://github.com/flyteorg/flytekit/pull/2579
Original PR: https://github.com/flyteorg/flytekit/pull/2618

Some people may love to develop tasks and workflows in notebooks and run them from the notebook.
Feel free to check out previous Pull Requests. This PR's goal is just to split two functions into separate PRs.

Jupyter Notebooks store code in a different format, as a JSON structure, rather than a standard Python script. It is also difficult to locate the code during runtime because Jupyter does not provide direct access to the file path or the exact code location in a standard `__file__` variable. This makes it challenging for Flytekit to extract the task code or module when registering a task, as it relies on standard Python modules or scripts for task registration.

#### Main Idea: Pickle the task function entity during runtime and upload it to a remote location (e.g., S3) when running in a Jupyter Notebook environment.

To implement this, the following modules were modified:
1. New `interactive_mode_enabled` field in`SerializationSettings`:
    - When serializing the task in `flytekit/tools/translator.py::get_serializable_task`, the Python task entity is pickled and uploaded to an S3 bucket. The `FastSerializationSettings` in `SerializationSettings` is updated to ensure that the remote task can download the entity from S3 before execution.
2. Update `DefaultTaskResolver` class in `flyte/core/python_auto_container.py`:
    - When loading the command arguments, we now check if `interactive_mode_enabled` is set in `SerializationSettings`. If so, we add an extra `pkl` argument to signal that the resolver should download the task entity from S3 in `load_task()`.
3. Update `download_distribution` function in `flytekit/tools/fast_registration.py`:
    - The original function extracted tarred files from S3, but for pickled entities, I updated the logic to handle `pkl.gz` file as a special case that can bypass the distribution format check.
4. Update `Promise` class:
    - Pickling `Promise` objects in `partial` functions (commonly used in map tasks) can lead to an infinite loop. To address this, I implemented custom `__getstate__` and `__setstate__` methods to handle the pickling process properly as suggested [here](https://stackoverflow.com/questions/50156118/recursionerror-maximum-recursion-depth-exceeded-while-calling-a-python-object-w).
5. New `interactive_mode_enabled` property in `FlyteRemote`:
    - Adding this property to `FlyteRemote` allows it to be specified once during the construction of the `FlyteRemote` object. Afterward, `FlyteRemote` will handle pickling and uploading tasks during registration.
    - FlyteRemote also provides an `upload_file` function to manage how the Python task entity is uploaded to S3.
 
#### Map task:

For `map_task`, we will only pickle the `actual_task` of the `map_task`, and upload the pickled file to the remote environment. However, when initializing `ArrayNodeMapTask` in the remote, it will try to get the `name` of the array node and call `extract_task_module` function to determine some information about the task. This will cause errors as the pickled entity is not a Python code, so we couldn't determine the module with `inspect.getmodule(f)` in the `_task_module_from_callable` function. As the naming process in initializing `ArrayNodeMapTask` doesn't really need the module name, I'll simply make the function successful if the task is pickled.

Nevertheless, I believe we need to have a more determined behavior for extracting information about pickled tasks or tasks from the Jupyter Notebook cells.

#### TODO
- [x] a new integration unit-test testing the Jupyter Notebook environment.
- [x] Add translator unit-test if `interactive_mode_enabled` => check pickle & no dest-dir
- [x] [0917] Fix Map task

### Setup process

- Adding the following codes into the Jupyter cells:
  ```python
  import nest_asyncio
  nest_asyncio.apply()
  
  remote = FlyteRemote(Config.auto(), default_project="flytesnacks", default_domain="development", interactive_mode_enabled=True)
  ```

### Simple Task & Workflow & map_task:
```python
@task(container_image=img)
def hello(name: str) -> str:
    return f"Hello {name}"

@task(container_image=img)
def world(pre: str) -> str:
    return f"{pre} Welcome to the world!"

@workflow
def wf(name: str) -> str:
    return world(pre=hello(name=name))
```
#### task:
```python
out = remote.execute(hello, inputs={"name": "world"}, version="1.0.1", wait=True)
print(remote.generate_console_url(out))
print(out.outputs)
```
image
image

#### workflow:
```python
out = remote.execute(wf, inputs={"name": "world"}, version="1.1.1", wait=True)
print(remote.generate_console_url(out))
print(out.outputs)
```
image
image

#### map_task:
```python
from flytekit import task, workflow, map_task
from functools import partial

@task(container_image=img)
def fn(x: int, y: int) -> int:
    return x + y

@workflow
def workflow_with_maptask(data: list[int], y: int) -> list[int]:
    partial_fn = partial(fn, y=y)
    return map_task(partial_fn)(x=data)

d: list[int] = [1, 2, 3]
out = remote.execute(workflow_with_maptask, inputs={"data": d, "y": 3}, version="1.2.1", wait=True)
print(out.outputs)
```
image
image

### Run k8s plugin (pytorch)
Moving the code from [flytesnacks/kfpytorch_plugin/pytorch_mnist.py](https://github.com/flyteorg/flytesnacks/blob/master/examples/kfpytorch_plugin/kfpytorch_plugin/pytorch_mnist.py) into a Jupyter notebook cell.
image
image

### Run File Sensor Agent
Follow the link of the [doc](https://docs.flyte.org/en/latest/flytesnacks/examples/sensor/file_sensor_example.html), and copy the code to the Jupyter Notebook.
image
image

## Related PRs

https://github.com/flyteorg/flytekit/pull/2579
https://github.com/flyteorg/flytekit/pull/2618


flytekit/bin/entrypoint.py
```diff
@@ -495,6 +522,8 @@ def _pass_through():
 @click.option("--dynamic-addl-distro", required=False)
 @click.option("--dynamic-dest-dir", required=False)
 @click.option("--resolver", required=False)
+@click.option("--pickled", is_flag=True, default=False, help="Use this to mark if the distribution is pickled.")
+@click.option("--pkl-file", required=False, help="Location where pickled file can be found.")
```
thomasjpfan, line 526: Do we need the `pickled` flag? I think the existance of `pkl-file` is enough to show that the task is pickled.

flytekit/remote/remote.py
```diff
@@ -858,9 +879,17 @@ def register_workflow(
...
 
         self._resolve_identifier(ResourceType.WORKFLOW, entity.name, version, serialization_settings)
-        ident = asyncio.run(
-            self._serialize_and_register(entity, serialization_settings, version, options, default_launch_plan)
-        )
+        try:
+            import nest_asyncio
+
+            nest_asyncio.apply()
+            ident = asyncio.run(
```
thomasjpfan, line 886: Is this because the remote is trying to use Jupyter's event loop? If so, can `FlyteRemote` create it's own event loop to process these async calls?

thomasjpfan (commented):
@Mecoli1219 Thank you for opening this PR. I rather have this in first before we try to think of a new API for execution.

With this PR, what happens if there is a global variable in a notebook?

```python
rng = np.random.default_rng()
X_big = rng.integers(size=(1000000, 1000000))

@task
def hello_world(x: int):
    Y = X_big + x
    ....
```

flytekit/remote/remote.py
thomasjpfan: Actually, can we just use the running loop?

```
def get_working_loop():
    """Returns a running event loop."""
    try:
        return asyncio.get_running_loop()
    except RuntimeError:
        with warnings.catch_warnings():
            warnings.simplefilter("error", DeprecationWarning)
            try:
                return asyncio.get_event_loop_policy().get_event_loop()
            except DeprecationWarning:
                return asyncio.get_event_loop_policy().new_event_loop()
```

Mecoli1219:
> With this PR, what happens if there is a global variable in a notebook?

We use `cloudpickle` to pickle the function. It should handle the global variable serialization (I am not really sure about it but at least I tested it and it works)

flytekit/tools/translator.py
```diff
@@ -164,6 +191,62 @@ def fn(settings: SerializationSettings) -> List[str]:
...
+
+        from flytekit.configuration import FastSerializationSettings
+
+        display_ipython_warning("Jupyter notebook and interactive task support is still alpha.")
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            dest = pathlib.Path(tmp_dir, "pkl.gz")
+            with gzip.GzipFile(filename=dest, mode="wb", mtime=0) as gzipped:
+                cloudpickle.dump(entity, gzipped)
+            display_ipython_warning("Uploading Pickled representation of Task to remote storage...")
```
thomasjpfan, line 242: Can you check the size of `dest` and throw an error if the pickle is too big? I'll start with ~150 MB.

(Not for this PR) A quick followup is to do this check while pickling.

Mecoli1219: @thomasjpfan I just moved to UCSD recently, and I’ll be updating the code this weekend. Once I’m settled in, I’ll have a lot of time to focus on this. I can follow your timeline to push this feature after next week. Feel free to reach out if you’d like to have a 1-on-1.

flytekit/bin/entrypoint.py
```diff
@@ -584,6 +623,8 @@ def handle_sigterm(signum, frame):
 @click.option("--resolver", required=True)
 @click.option("--checkpoint-path", required=False)
 @click.option("--prev-checkpoint", required=False)
+@click.option("--pickled", is_flag=True, default=False, help="Use this to mark if the distribution is pickled.")
```
pingsutw, line 626: I think it's better to add a pickle resolver and add it to the compilation context [here](https://github.com/flyteorg/flytekit/blob/dfe93eb55fadf26190e898a25936dad0d5deaf39/flytekit/core/python_auto_container.py#L119), so that we don't need to change the entrypoint.

flytekit/remote/remote.py
Mecoli1219: @thomasjpfan I think the main problem here is that if we call `asyncio.run` directly. There will have nested asyncio event loops exist (the jupyter one and the `asyncio.run` one). This is not permitted in asyncio package, so we need to use nest_asyncio package to resolve this problem. 

Honestly, I thought that the try & except here is not necessary as `import asyncio` should resolve the problem already.

thomasjpfan: With `get_working_loop`, we never run `asyncio.run`. The pattern would be:

```python
loop = get_working_loop()
loop.run(...)
```

Where `get_working_loop` will either create a loop if there is not one running or return the Jupyter event loop. This way, there is no more nested event loops.

(Note I have not tested this out.)

Mecoli1219: I got it, but there is another problem. We couldn't run a new task in a running event loop, at least the function `loop.run_until_complete()` will throw error since Jupyter event loop is always running.
I also tried `create_task()` function, but not yet figured out how to get the result without changing the function into asynchronous function.

Mecoli1219: @thomasjpfan I couldn't figure out the approach without the `nest_asyncio` package, and I also think we can open another PR to fix this issue. 
As for now, I was thinking of removing the `nest_asyncio` package and asking the user to add this manually 
```
import nest_asyncio
nest_asyncio.apply()
```
at the beginning of their Jupyter Notebook if they want to use it in the Jupyter Notebook environment so that we don't need to add an extra dependency.

thomasjpfan: I'm undecided on this especially because https://github.com/erdewit/nest_asyncio is unmaintained. 

If we can not get this working with Jupyter's event loop, then I'll like to run a flytekit owned event loop on another thread: https://github.com/flyteorg/flytekit/pull/2744

I really do not like mixing async code with sync code. @pingsutw What do you think of having a sync version of `_serialize_and_register` that runs in jupyter? Another options is to revert https://github.com/flyteorg/flytekit/pull/2276.

flytekit/bin/entrypoint.py
Mecoli1219: I removed these two arguments from `pyflyte-execute` since I moved the pickle check into the `task_resolver.load_task`.

flytekit/bin/entrypoint.py
Mecoli1219: I modified the behavior of the default resolver so that it could support the pickled task.

flytekit/remote/remote.py
Mecoli1219: @thomasjpfan Should we include the Flytekit-owned event loop in this PR? For now, it might be better to ask the user to manually add `nest_asyncio`, as shown in the example in this PR’s description, instead of including it in this project. We could then create a separate PR to address this issue.

I’m almost finished with this PR, except for adding a few unit tests. Could you please review the code again and let me know if there’s anything else that needs attention or discussion? Thanks!

Mecoli1219:
@thomasjpfan This PR is ready for your review. Let me know if you have any questions or need further clarification.

pingsutw:
I just added a notebook resolver. Ran following example to test it

```python
from datetime import datetime

remote = FlyteRemote(Config.auto(), default_project="flytesnacks", default_domain="development", interactive_mode_enabled=True)
new_flytekit = "git+https://github.com/flyteorg/flytekit.git@ff8b92ba6925c774377f20fc0667711a1cd89828"
img = ImageSpec(registry="pingsutw", packages=[new_flytekit], apt_packages=["git"])


@task(container_image=img)
def hello(name: str) -> str:
    return f"Hello {name}"


@task(container_image=img)
def world(pre: str) -> str:
    return f"{pre} Welcome to the world!"


@workflow
def wf(name: str) -> str:
    return world(pre=hello(name=name))


out = remote.execute(hello, inputs={"name": "world"}, version="1.0.6", wait=True)
print(out)
```

tests/flytekit/integration/remote/test_remote.py
```diff
@@ -591,6 +592,135 @@ def test_flyteremote_uploads_large_file(gigabytes):
             assert s3_md5_bytes == md5_bytes
 
 
+@mock.patch("flytekit.tools.interactive.ipython_check")
+def test_workflow_remote_func(mock_ipython_check):
+    """Test the logic of the remote execution of workflows and tasks."""
+    mock_ipython_check.return_value = True
```
pingsutw, line 598: Should we set `interactive_mode_enabled` to True for the remote client instead?

Mecoli1219: You are right

flytekit/tools/fast_registration.py
```diff
@@ -226,12 +226,13 @@ def download_distribution(additional_distribution: str, destination: str):
...
     tarfile_name = os.path.basename(additional_distribution)
-    if not tarfile_name.endswith(".tar.gz"):
+    if tarfile_name.endswith(".tar.gz"):
+        # This will overwrite the existing user flyte workflow code in the current working code dir.
+        result = subprocess.run(
+            ["tar", "-xvf", os.path.join(destination, tarfile_name), "-C", destination],
+            stdout=subprocess.PIPE,
+        )
+        result.check_returncode()
+    elif tarfile_name != "pkl.gz":
```
thomasjpfan, line 236: Can this special `"pkl.gz"` value be stored somewhere?

flytekit/core/python_auto_container.py
```diff
@@ -274,6 +281,34 @@ def get_all_tasks(self) -> List[PythonAutoContainerTask]:  # type: ignore
...
 
+class DefaultNotebookTaskResolver(TrackedInstance, TaskResolverMixin):
+    """
+    This resolved is used when the task is defined in a notebook. It is used to load the task from the notebook.
+    """
+
+    def name(self) -> str:
+        return "DefaultNotebookTaskResolver"
+
+    @timeit("Load task")
```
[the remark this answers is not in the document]
pingsutw: We will be able to view the overhead of the loading test in the Flyte deck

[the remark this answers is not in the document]
pingsutw: We already use `timeit` in the default resolver

flytekit/tools/fast_registration.py
Mecoli1219: Do you mean that we should store `pkl.gz` as a global variable somewhere else?

pingsutw (dismissed):
LGTM, thank you @Mecoli1219

flytekit/tools/fast_registration.py
thomasjpfan: Yup. Looks like you already did it in https://github.com/flyteorg/flytekit/pull/2733/commits/1ed6960191ca5a34d2c2e7dd66081a574a40bcc5

thomasjpfan (commented):
With https://github.com/flyteorg/flytekit/pull/2733#issuecomment-2369868663, I am getting this error in a Jupyter notebook:

Screenshot 2024-09-26 at 11 59 40 AM

flytekit/core/tracker.py
```diff
@@ -273,6 +274,13 @@ def _resolve_abs_module_name(self, path: str, package_root: typing.Optional[str]
         if dirname == package_root:
             return basename
 
+        # Execution in a Jupyter notebook, we cannot resolve the module path
+        if not os.path.exists(dirname):
+            logger.warning(f"Directory {dirname} does not exist. It is likely that we are in a Jupyter notebook.")
```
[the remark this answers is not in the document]
Mecoli1219: Sure, I'll make it lru_cache function

Mecoli1219:
@pingsutw I noticed the map_task wasn’t using the DefaultNotebookTaskResolver, so I’ve updated it. Could you please approve the CI test?

Mecoli1219:
> When I run [#2733 (comment)](https://github.com/flyteorg/flytekit/pull/2733#issuecomment-2369868663), I am getting this error in a Jupyter notebook:
> 
> Screenshot 2024-09-26 at 11 59 40 AM

we have to use nest_asyncio for now

flytekit/core/tracker.py
```diff
@@ -273,6 +273,11 @@ def _resolve_abs_module_name(self, path: str, package_root: typing.Optional[str]
...
```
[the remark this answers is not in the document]
Mecoli1219: @thomasjpfan There are two possible cases that `dirname` doesn't exist in the current os. The first one is already mentioned. 
The second one is when we pickled the actual task in `map_task`, the `ArrayNodeMapTask` will try to get the module name [here](https://github.com/flyteorg/flytekit/blob/98d722fec491ecd68ba3ea92dfae9f676d32f677/flytekit/core/array_node_map_task.py#L101-L110) in the remote environment. In this case, since the entity is already pickled, it is not in the Jupyter Notebook environment. I can check whether the special file `PICKLE_FILE_PATH` exists in the current os for the check. WDYT

Mecoli1219:
Update comment:

#### Map task:

For `map_task`, we will only pickle the `actual_task` of the `map_task`, and upload the pickled file to the remote environment. However, when initializing `ArrayNodeMapTask` in the remote, it will try to get the `name` of the array node and call `extract_task_module` function to determine some information about the task. This will cause errors as the pickled entity is not a Python code, so we couldn't determine the module with `inspect.getmodule(f)` in the `_task_module_from_callable` function. As the naming process in initializing `ArrayNodeMapTask` doesn't really need the module name, I'll simply make the function successful if the task is pickled.

Nevertheless, I believe we need to have a more determined behavior for extracting information about pickled tasks or tasks from the Jupyter Notebook cells.

thomasjpfan:
> we have to use nest_asyncio for now if we still want to use asyncio.run during registration.

I do not want to use `nest_asyncio` because https://github.com/erdewit/nest_asyncio is not maintained anymore.

I implemented a way to run in a jupyter notebook without `nest_asyncio` in another project. I'll introduce the technique into `FlyteRemote` in another PR and test that it works with your PR.

kumare3:
@thomasjpfan / @Mecoli1219 i would love to get this merged :). how can we move to get this done?

Mecoli1219:
@kumare3 The code for interactive mode in the Jupyter notebook is ready. However, the current approach requires using `nest_asyncio` to run Flyte’s event loop inside Jupyter’s event loop, which is not ideal since `nest_asyncio` is no longer actively maintained. @thomasjpfan is planning to implement a different approach that avoids using `nest_asyncio` in a separate PR. Once that’s in place, I can update my PR to ensure compatibility and add the necessary tests. 

Let me know if there’s anything else I can assist with to get this merged!

thomasjpfan:
I opened https://github.com/flyteorg/flytekit/pull/2784 to update `FlyteRemote` to use a event loop on another thread.

I tested this PR with my updates to `FlyteRemote` and it works! (No longer requires `nest_asyncio`)

thomasjpfan (approved):

wild-endeavor:
hey @Mecoli1219 few comments on the pr.  first thank you though for taking it across the finish line.  had a few questions and wanted to revisit some things though...
shorter comments:
* I see that you added code and a test for pickling promise objects, but could you explain when this is needed?
* In the `translator.py`... the `_update_serialization_settings_for_ipython` function - can we make it not mutating?  Ideally the serialization settings object is frozen - it's too late to do that, but we can keep new code copy-on-write i think.
* In `tracker.py` in the debug statements, can we add the values of the variables that will be returned?  (like add `basename` to logger.debug).  This file is confusing enough and has caused us enough headaches that more printing is okay. 

more structural stuff:
* I feel like the addition of the `file_uploader` function to the Options dataclass a bit out of place.  If you take a look at that object, it's really meant to be used alongside launch plans.  Having a function, that makes a network call, that serializes tasks, feels pretty odd.
  * Extending that thought though, the addition of a network call in general changes the pattern for serialization a bit.  I don't remember if there are other network calls in `translator` but the original idea at least was that the translating step is only pure functions, and just converts the Python objects into Flyte protos.  Adding the file_uploader now changes that and feels like a bit of a shortcut.  Think about how fast registration works in the `pyflyte package` case.  When you `fast package workflows`, flytekit will create a fast register tar.gz file that contains the code to be injected into the image.  This is the same as `pyflyte run`.  But when you do `package`, this file is not uploaded... it's the responsibility of `flytectl register` to do the upload (and then inject the upload location into the various task definitions [something I think it doesn't do 100% correctly today actually but that's beside the point]).  Could you revisit this part a bit?  Not necessarily saying it must change - but I think worth spending time thinking about, and seeing if you can find a better pattern.
  * Along the same line, could you explain what the issue is for dynamic tasks?  I read the TODO you left.  What backend work needs to happen?  The comment suggests adding another uploader to the flytecontext - would definitely like to avoid this if at all possible.  Having a callable in there also feels strange.

Thank you!

Mecoli1219:
@wild-endeavor Thank you for your comments! I apologize for any inconvenience this PR may have caused. Here are some clarifications and further thoughts based on our discussion:

### Quick Response
> I see that you added code and a test for pickling promise objects, but could you explain when this is needed?

Pickling Promise objects in partial functions (commonly used in map tasks) can lead to an infinite loop. To address this, I implemented custom __getstate__ and __setstate__ methods to handle the pickling process properly as suggested [here](https://stackoverflow.com/questions/50156118/recursionerror-maximum-recursion-depth-exceeded-while-calling-a-python-object-w).

> In tracker.py in the debug statements, can we add the values of the variables that will be returned? (like add basename to logger.debug). This file is confusing enough and has caused us enough headaches that more printing is okay.

I agree that we should add the basename to `logger.debug`

### Discussion

> In the translator.py... the _update_serialization_settings_for_ipython function - can we make it not mutating?

I’m not entirely sure what you mean by “non-mutating” in this context. In [this code snippet](https://github.com/flyteorg/flytekit/blob/79f683b4821475fc7c06c2cd2119c0237f75e29f/flytekit/tools/translator.py#L198C5-L208C22) the serialization settings are being updated as well.

Nevertheless, after revisiting the code, I think there is another possible alternative is that we can pickle the file within `FlyteRemote::execute_local_task` in [remote.py](https://github.com/flyteorg/flytekit/blob/79f683b4821475fc7c06c2cd2119c0237f75e29f/flytekit/remote/remote.py#L1836C1-L1843C67) and set the `fast_serialization` during construction of the `SerializationSettings` if `self.interactive_mode_enabled` is True. 

> structural stuff & file_uploader issue

Currently, file_uploader uses the `FlyteRemote::upload_file` function. If we integrate the uploading logic into `FlyteRemote` itself as mentioned in the previous section, we wouldn’t need to pass `file_uploader` into the `Options` dataclass. 

I believe this restructuring could simplify the code, but I’m not sure if it’s too late to implement such changes at this stage. What are your thoughts?

Mecoli1219:
Update:

I found that pickling the file within `FlyteRemote::execute_local_task` isn’t sufficient because it only makes remote task execution succeed, while workflow execution still fails. The main issue with moving `upload_file` outside of `get_serializable_task()` is that, at the workflow level, the `fast_serialization_setting` will package the entire workflow (and its tasks) into a single tarred module directory, which is then used as an additional distribution in the sandbox.
In the Jupyter Notebook environment, the main challenge is that we don’t have an actual module or directory that we can compress into a `tar.gz` file. All we can do is read the entity from memory and dump it via `cloudpickle`, which raises two concerns:
1. We need to update or add the `fast_serialization_setting` in `SerializationSettings` for each task during workflow registration.
2. We can only modify the serialization setting within the `get_serialization()` function (specifically, `get_serialization_task()`) in `translator.py` during workflow registration (as shown in this PR via `get_serializable_task()`).

I understand the concern around using network calls in translator.py, but I couldn’t find a better solution for now. Nevertheless, I think that using `file_uploader` in the `Options` could be addressed by leveraging the FlyteRemote’s `upload_file` function, as this is the only use case currently.

image

PR #2734, opened 2024-09-07
Deduce cache version iff `cache_version` is not set

eapolinario:
## Tracking issue
https://github.com/flyteorg/flyte/issues/770

We should decrease the burden to use a cache in the case of python tasks.

This PR calls [inspect.getsource](https://docs.python.org/3/library/inspect.html#inspect.getsource) passing the task function as an argument, which loads the actual source code of the function and use that to use generate a string to represent the cache version.

Notice that any change to the function (including adding/removing a parameter, changing the docstring, rearranging/reindenting some code) is going to produce a new value. In other words, the function can be moved around in a file, but its contents cannot change, otherwise a new value is going to be generated.

## How was this patch tested?
Unit tests and local executions.

### DIFF
--- a/flytekit/core/task.py
+++ b/flytekit/core/task.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
 import datetime
+import hashlib
+import inspect
 import os
+import sys
 from functools import update_wrapper
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union, overload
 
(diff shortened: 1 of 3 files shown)


flytekit/core/task.py
```diff
@@ -378,6 +381,21 @@ def wrapper(fn: Callable[P, Any]) -> PythonFunctionTask[T]:
         return wrapper
 
 
+def _deduce_cache_version(fn: Callable[P, Any]) -> str:
```
thomasjpfan, line 384: If the task uses an imported function, any updates to the imported function would not be captured by `inspect.getsource`:

```python
from utils import helper

@task(cache=True)
def my_task(x: int):
    x = my_helper(x)
```

I think this is okay as long as we documented the behavior and what users should do.

pingsutw: Yup. I think it's okay too. Could we also update the docstring for cache_version [here](https://github.com/flyteorg/flytekit/blob/a366653e697cd3adf429a05adde2ac421fbcae99/flytekit/core/base_task.py#L116)?

PR #2735, opened 2024-09-08
GH-5732: Make `TypeEngine.lazy_import_transformers()` thread safe

Tom-Newton:
## Tracking issue
Closes: https://github.com/flyteorg/flyte/issues/5732

Lack of thread safety in `TypeEngine.lazy_import_transformers()` can cause errors when triggering Flyte executions simultaneously from concurrent threads. 

- Add a lock to make it threadsafe. 
  - I considered just moving `cls.has_lazy_import = True` to the end of the function but I decided against it for 2 reasons:
    - Its possible that importing transformers twice could cause issues - though currently I don't think it does. 
    - Apparently python variable assignments are not always atomic https://github.com/google/styleguide/blob/91d6e367e384b0d8aaaf7ce95029514fcdf38651/pyguide.md#218-threading - though in practice I doubt this will be a problem. 
- Write a test

## How was this patch tested?
Wrote a test

- [x] I updated the documentation accordingly. - not applicable 

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -9,8 +9,8 @@ import inspect
 import json
 import mimetypes
 import sys
-import threading
 import textwrap
+import threading
 import typing
 from abc import ABC, abstractmethod
 from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


pingsutw (approved):
Thank you @Tom-Newton

eapolinario (approved):
LGTM.

PR #2737, opened 2024-09-09
eager workflow: use event loop instead of asyncio.run

cosmicBboy:

The `entrypoint._dispatch_execute` function used `asyncio.run` to run the coroutine. The problem with this is that it closes the event loop, which cause the following error related to task data persistence:

```
[1/1] currentAttempt done. Last Error: USER::
[ajwx7zsjbghhfzhpzhnv-f4usdeia-0] terminated with exit code (1). Reason [Error]. Message: 
micromamba/envs/runtime/bin/pyflyte-execute", line 8, in <module>
    sys.exit(execute_task_cmd())
             ^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/flytekit/bin/entrypoint.py", line 528, in execute_task_cmd
    _execute_task(
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/flytekit/bin/entrypoint.py", line 403, in _execute_task
    _dispatch_execute(ctx, load_task, inputs, output_prefix)
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/flytekit/bin/entrypoint.py", line 180, in _dispatch_execute
    ctx.file_access.put_data(ctx.execution_state.engine_dir, output_prefix, is_multipart=True)
  File "/opt/micromamba/envs/runtime/lib/python3.11/site-packages/flytekit/core/data_persistence.py", line 592, in put_data
    raise FlyteAssertion(
flytekit.exceptions.user.FlyteAssertion: USER:AssertionError: error=Failed to put data from /tmp/flytei9z9be0x/local_flytekit/engine_dir to unionmeta://opta-gcp-serverless-1-union-us-central1/cosmicbboy/flytesnacks-development-ajwx7zsjbghhfzhpzhnv/testeagerserverlesssimpleeagerworkflow/data/0 (recursive=True).

Original exception: There is no current event loop in thread 'MainThread'., cause=There is no current event loop in thread 'MainThread'.
Error in sys.excepthook:

Original exception was:
.
```

Execution on [Union serverless](https://serverless.union.ai/org/cosmicbboy/projects/flytesnacks/domains/development/executions/ajwx7zsjbghhfzhpzhnv)

Use `get_event_loop().run_until_complete(<coroutine>)` instead so that the event loop isn't closed.

## How was this patch tested?

Tested on serverless:

```python
import os
from union.remote import UnionRemote
from union._config import _get_config_obj

flytekit = "git+https://github.com/flyteorg/flytekit@688569c37930be2640a418407ed75e6afb6b9037"

image = ImageSpec(
    name="flytekit-eager",
    packages=["flytekit", flytekit],
    apt_packages=["git"],
)


@task(container_image=image)
def add_one(x: int) -> int:
    return x + 1


@task(container_image=image)
def double(x: int) -> int:
    return x * 2


@eager(
    container_image=image,
    remote=UnionRemote(_get_config_obj(None, default_to_union_semantics=False)),
    client_secret_key="serverless-1-cosmicbboy-niels-functional-testing",  # replace with your own secret
    client_secret_env_var="UNION_SERVERLESS_API_KEY",
)
    out = await add_one(x=x)
    if out < 0:
        return -1
    return await double(x=out)

```

Run with:

```
union run --remote test_eager_serverless.py simple_eager_workflow --x 5
```

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -9,6 +9,7 @@ import subprocess
 import sys
 import tempfile
 import traceback
+import warnings
 from sys import exit
 from typing import Callable, List, Optional
 
@@ -68,6 +69,23 @@ def _compute_array_job_index():
     return offset
 
 
+def _get_working_loop():
+    """Returns a running event loop."""
+    try:
+        return asyncio.get_running_loop()
+    except RuntimeError:
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", DeprecationWarning)
+            try:
+                return asyncio.get_event_loop_policy().get_event_loop()
+            # Since version 3.12, DeprecationWarning is emitted if there is no
+            # current event loop.
+            except DeprecationWarning:
+                loop = asyncio.get_event_loop_policy().new_event_loop()
+                asyncio.set_event_loop(loop)
+                return loop
+
+
 def _dispatch_execute(
     ctx: FlyteContext,
     load_task: Callable[[], PythonTask],
@@ -107,7 +125,7 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
-            outputs = asyncio.get_running_loop().run_until_complete(outputs)
+            outputs = _get_working_loop().run_until_complete(outputs)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -107,8 +107,7 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
-            with asyncio.Runner() as runner:
-                outputs = runner.run(outputs)
+            outputs = asyncio.get_event_loop().run_until_complete(outputs)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -107,7 +107,8 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
-            outputs = asyncio.get_event_loop().run_until_complete(outputs)
+            with asyncio.Runner() as runner:
+                outputs = runner.run(outputs)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -107,7 +107,7 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
-            outputs = asyncio.get_event_loop().run_until_complete(outputs)
+            outputs = asyncio.get_running_loop().run_until_complete(outputs)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -107,7 +107,7 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
-            outputs = asyncio.run(outputs)
+            outputs = asyncio.get_event_loop().run_until_complete(outputs)
 
         # Step3a
         if isinstance(outputs, VoidPromise):
--- a/tests/flytekit/unit/experimental/test_eager_workflows.py
+++ b/tests/flytekit/unit/experimental/test_eager_workflows.py
@@ -1,4 +1,5 @@
 import asyncio
+import mock
 import os
 import sys
 import typing
@@ -9,8 +10,13 @@ import pytest
 from hypothesis import given
 
 from flytekit import dynamic, task, workflow
+
+from flytekit.bin.entrypoint import _dispatch_execute
+from flytekit.core import context_manager
+from flytekit.core.promise import VoidPromise
 from flytekit.exceptions.user import FlyteValidationException
 from flytekit.experimental import EagerException, eager
+from flytekit.models import literals as _literal_models
 from flytekit.types.directory import FlyteDirectory
 from flytekit.types.file import FlyteFile
 from flytekit.types.structured import StructuredDataset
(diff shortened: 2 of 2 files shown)


flytekit/bin/entrypoint.py
```diff
@@ -107,7 +107,7 @@ def _dispatch_execute(
         if inspect.iscoroutine(outputs):
             # Handle eager-mode (async) tasks
             logger.info("Output is a coroutine")
-            outputs = asyncio.run(outputs)
+            outputs = asyncio.get_event_loop().run_until_complete(outputs)
```
thomasjpfan, line 110: Is there a good way to add a unit test for this?

eapolinario (commented):
(You shared a serverless execution. How can I view it?)

Based on the exception in the description, can you explain why the [AsyncUnionMetaFS](https://github.com/unionai/unionai/blob/fca561fe6f4b1ab0a6e4f77708f87cb4c2013886/union/filesystems/_unionmetafs.py#L47) filesystem is failing? More specifically, Why aren't we seeing the exception being bubbled up from there? I'm assuming that there's something weird going on with [these two lines](https://github.com/unionai/unionai/blob/fca561fe6f4b1ab0a6e4f77708f87cb4c2013886/union/filesystems/_unionmetafs.py#L55-L56) in the implementation of the unionmeta fs, but I'm not sure what yet.

Is it easy to repro this? If so, can we try using a [Runner](https://docs.python.org/3/library/asyncio-runner.html#runner-context-manager) in the entrypoint instead?

As a minor issue, we have other [uses](https://github.com/search?q=repo%3Aflyteorg%2Fflytekit%20asyncio.run&type=code) of `asyncio.run` in the codebase. Does it make sense to go over each one and ensure that its use is appropriate? This can be done in a separate PR / investigation.

cosmicBboy:
@eapolinario I added a unit test on this PR, the `asyncio.Runner` approach works. (You can repro the Mainthread error if you switch the changed like back to `asyncio.run`.

cosmicBboy:
@eapolinario I can confirm that the error message
```
Original exception: There is no current event loop in thread 'MainThread'., cause=There is no current event loop in thread 'MainThread'.
```
only crops up in serverless (the eager workflow succeeds in flyte sandbox). Going to investigate a solution in the `unionai` repo.

flytekit/bin/entrypoint.py
wild-endeavor: I think `get_running_loop` is preferred over `get_event_loop`, can we try with that?

wild-endeavor: see [comment](https://github.com/python/cpython/issues/96377#issuecomment-1258819222) here for more information, but it's the more thread-safe variant... doesn't really come into play here, but i think better practice.

wild-endeavor:
Okay merging this PR if we need to to unblock something on serverless, but yeah the solution should be investigated on the union side.

I'm still trying to work through some of the async loop semantics, but I think the unit test as it's written is kinda doing the opposite of what we want.  I don't fully understand async loop lifecycles yet and how to get them to not conflict, but I feel like in `entrypoint.py` at least, (and assuming we don't interfere with core functionality like grpc), `_dispatch_execute` should be able to run it's own event loop lifecycle.  That is, we should be able to call `loop.close()` inside of `_dispatch_execute`.  This means that after dispatch execute ends, a loop _shouldn't_ be available.

* @EngHabu do you remember why [those lines](https://github.com/unionai/unionai/blob/66dcde98bb685ad61d77c5c8fa054892acdc146a/union/filesystems/_unionmetafs.py#L55) are in the filesystem?  They're also in the [non-meta fs](https://github.com/unionai/unionai/blob/66dcde98bb685ad61d77c5c8fa054892acdc146a/union/filesystems/_unionfs.py#L102-L103).  If both are created at the same time, won't they conflict?  If you remember the error/repro at all, can you share?  I wanna see if I can play around and understand what's happening.

wild-endeavor:
I think this might be relevant: https://github.com/grpc/grpc/issues/32480#issuecomment-1499972980

Maybe the original issue was that there was no event loop, so we grabbed it from the loop in the grpc.aio channel?

The OP in the issue suggests it might work if we delay instantiation of the channel until we're already in an async context?  Like, don't call `_create_channel` until we're in the body of some async function (which is annoying cuz it means we'll have to add it to every function), and then somewhere in flytekit, set up a loop?

cosmicBboy:
> Why is the unit test you're adding failing in CI?

No idea why it's flaky... failing tests will pass when I retry it enough times.

cosmicBboy:
@eapolinario I changed the unit testing strategy: basically I tried to recapitulate what the UnionFS classes were doing prior to @thomasjpfan's PR: https://github.com/unionai/unionai/pull/393. Just testing that the event loop set in the test is still available (and is equal to) the event loop after the dispatch call.

tests/flytekit/unit/experimental/test_eager_workflows.py
```diff
@@ -275,3 +281,29 @@ async def eager_wf_flyte_directory() -> str:
...
+
+    ctx = context_manager.FlyteContext.current_context()
+    with context_manager.FlyteContextManager.with_context(
+        ctx.with_execution_state(
+            ctx.execution_state.with_params(mode=context_manager.ExecutionState.Mode.TASK_EXECUTION)
+        )
+    ) as ctx:
+        loop = asyncio.new_event_loop()
+        asyncio.set_event_loop(loop)
+        asyncio.events.set_event_loop(loop)
```
thomasjpfan, line 306: Is setting the event loop safe to do in a test?

cosmicBboy: not sure, but perplexity seems to think so: https://www.perplexity.ai/search/is-asyncio-set-event-loop-safe-MXeyb8O3QPyddzGacsrRqA

A relevant reference is: https://github.com/pytest-dev/pytest-asyncio/issues/80

cosmicBboy: even better, the pytest-asyncio library has an `event_loop` fixture

tests/flytekit/unit/experimental/test_eager_workflows.py
```diff
@@ -275,3 +281,26 @@ async def eager_wf_flyte_directory() -> str:
...
+
+    ctx = context_manager.FlyteContext.current_context()
+    with context_manager.FlyteContextManager.with_context(
+        ctx.with_execution_state(
+            ctx.execution_state.with_params(mode=context_manager.ExecutionState.Mode.TASK_EXECUTION)
+        )
+    ) as ctx:
+        _dispatch_execute(ctx, lambda: eager_wf, "inputs path", "outputs prefix")
+        loop_after_execute = asyncio.get_event_loop_policy().get_event_loop()
+        assert event_loop == loop_after_execute
```
wild-endeavor, line 306: assuming the fix goes in on the union side, i still feel like this assert should fail.

wild-endeavor:
do we have to consider this?  https://github.com/fsspec/filesystem_spec/blob/76ca4a68885d572880ac6800f079738df562f02c/fsspec/asyn.py#L128  the windows specific policy.

they're planning on deprecating policies in favor of runners right?  but for older versions of python will we need to do something similar?

tests/flytekit/unit/experimental/test_eager_workflows.py
wild-endeavor: posted some thoughts on the other pr, which basically wraps the whole _dispatch_execute call in an event loop so one is there if necessary.

wild-endeavor (approved):

cosmicBboy:
Can we merge this? @wild-endeavor @eapolinario @thomasjpfan

wild-endeavor:
i was planning on merging on monday if that's okay?  was going to test it some more with the union lib just to make sure if that's alright.

wild-endeavor:
still testing this.  give me until tomorrow - running into other issues and having trouble repro'ing this original issue in serverless.

PR #2742, opened 2024-09-10
[Bug] Set bindings for ArrayNode

pvditt:
## Tracking issue
fixes: https://linear.app/unionai/issue/COR-1822/mapping-over-launch-plans-w-default-inputs-doesnt-compile

Compiling a workflow fails when evaluating an ArrayNode that maps over a LaunchPlan contains a default input. This occurs since we don't set the input bindings for the node.

https://github.com/unionai/flyte/blob/926cadac9a8cf406b85c8e7443668a13a3f60dfe/flytepropeller/pkg/compiler/validators/interface.go#L76-L86

Get errors such as:

```
pyflyte run --remote tests/flytekit/integration/map_ref_lp.py map_regression_line_wf
Running Execution on Remote.
Request rejected by the API, due to Invalid input.
RPC Failed, with Status: StatusCode.INVALID_ARGUMENT
        details: failed to compile workflow for [resource_type:WORKFLOW project:"flytesnacks" domain:"development" name:"tests.flytekit.integration.map_ref_lp.map_regression_line_wf" version:"m6vMatWi-HCjUFD5scx4rQ"] with err failed to compile workflow with err Collected Errors: 3
        Error 0: Code: VariableNameNotFound, Node Id: n0, Description: Variable [val] not found on node [n0].
        Error 1: Code: VariableNameNotFound, Node Id: n0, Description: Variable [x] not found on node [n0].
        Error 2: Code: VariableNameNotFound, Node Id: n0, Description: Variable [y] not found on node [n0].

```

- Update create_and_link_node and create_and_link_node_remote functions to create a node and optionally link it to the workflow context

- clean up some of the setting of bound_inputs for array node since that isn't supported atm

## How was this patch tested?
- added unit test to confirm the serialized array node contains the bindings
- ran E2E
```
@task
def slope(x: list[int], y: list[int]) -> float:
    sum_xy = sum([x[i] * y[i] for i in range(len(x))])
    sum_x_squared = sum([x[i] ** 2 for i in range(len(x))])
    n = len(x)
    return (n * sum_xy - sum(x) * sum(y)) / (n * sum_x_squared - sum(x) ** 2)


@task
def intercept(x: list[int], y: list[int], slope: float) -> float:
    mean_x = sum(x) / len(x)
    mean_y = sum(y) / len(y)
    intercept = mean_y - slope * mean_x
    return intercept


@workflow
def slope_intercept_wf(x: list[int], y: list[int]) -> (float, float):
    slope_value = slope(x=x, y=y)
    intercept_value = intercept(x=x, y=y, slope=slope_value)
    return (slope_value, intercept_value)


@task
def regression_line(val: int, slope_value: float, intercept_value: float) -> float:
    return (slope_value * val) + intercept_value  # y = mx + c


@workflow
def regression_line_wf(val: int = 5, x: list[int] = [-3, 0, 3], y: list[int] = [7, 4, -2]) -> float:
    slope_value, intercept_value = slope_intercept_wf(x=x, y=y)
    return regression_line(val=val, slope_value=slope_value, intercept_value=intercept_value)
```

```
@reference_launch_plan(
    project="flytesnacks",
    domain="development",
    name="advanced_composition.subworkflow.regression_line_wf",
    version="1jtCscLI-_FBiNcmL8Nr4Q",
)
def ref_regression_line_wf(
    val: int, x: list[int], y: list[int]
) -> float:
    return 1.0


@workflow
def map_regression_line_wf() -> list[float]:
    a = [1, 2, 3]
    b = [[-3, 0, 3], [-8, 2, 4], [7, 3, 1]]
    c = [[7, 4, -2], [-2, 4, 7], [3, 6, 4]]
    return map_task(ref_regression_line_wf)(val=a, x=b, y=c)
```

### Setup process
- need to run w/ union's flyte fork

### DIFF
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -50,7 +50,7 @@ class ArrayNode:
         self._concurrency = concurrency
         self._execution_mode = execution_mode
         self.id = target.name
-        self._bindings = []
+        self._bindings: List[_literal_models.Binding] = []
 
         if min_successes is not None:
             self._min_successes = min_successes
@@ -192,13 +192,15 @@ class ArrayNode:
         return self._execution_mode
 
     def __call__(self, *args, **kwargs):
-        ctx = FlyteContext.current_context()
-        # interface mismatch between the target and the actual inputs since we don't overwrite the interface to a list
-        temp = {
+        # interface mismatch between the target and the actual inputs since we don't create a new entity with a
+        # transformed to list interface
+        transformed_kwargs = {
             key: (val[0] if isinstance(val, list) and val else None) if isinstance(val, list) else val
             for key, val in kwargs.items()
         }
-        bound_subnode = create_and_link_node(ctx, entity=self.flyte_entity, *args, **temp)
+        bound_subnode = create_and_link_node(
+            FlyteContext.current_context(), entity=self.flyte_entity, link_node=False, **transformed_kwargs
+        )
         self._bindings = bound_subnode.ref.node.bindings
         return flyte_entity_call_handler(self, *args, **kwargs)
 
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -11,6 +11,7 @@ from flytekit.core.node import Node
 from flytekit.core.promise import (
     Promise,
     VoidPromise,
+    create_and_link_node,
     flyte_entity_call_handler,
     translate_inputs_to_literals,
 )
@@ -49,6 +50,7 @@ class ArrayNode:
         self._concurrency = concurrency
         self._execution_mode = execution_mode
         self.id = target.name
+        self._bindings = []
 
         if min_successes is not None:
             self._min_successes = min_successes
@@ -99,7 +101,7 @@ class ArrayNode:
     @property
     def bindings(self) -> List[_literal_models.Binding]:
         # Required in get_serializable_node
-        return []
+        return self._bindings
 
     @property
     def upstream_nodes(self) -> List[Node]:
@@ -190,6 +192,14 @@ class ArrayNode:
         return self._execution_mode
 
     def __call__(self, *args, **kwargs):
+        ctx = FlyteContext.current_context()
+        # interface mismatch between the target and the actual inputs since we don't overwrite the interface to a list
+        temp = {
+            key: (val[0] if isinstance(val, list) and val else None) if isinstance(val, list) else val
+            for key, val in kwargs.items()
+        }
+        bound_subnode = create_and_link_node(ctx, entity=self.flyte_entity, *args, **temp)
+        self._bindings = bound_subnode.ref.node.bindings
         return flyte_entity_call_handler(self, *args, **kwargs)
 
 
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -193,7 +193,7 @@ class ArrayNode:
 
     def __call__(self, *args, **kwargs):
         ctx = FlyteContext.current_context()
-        if ctx.compilation_state is not None:
+        if ctx.compilation_state:
             # since a new entity with an updated list interface is not created, we have to work around the mismatch
             # between the interface and the inputs
             collection_interface = transform_interface_to_list_interface(self.flyte_entity.python_interface, set())
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -21,6 +21,8 @@ from flytekit.models import literals as _literal_models
 from flytekit.models.core import workflow as _workflow_model
 from flytekit.models.literals import Literal, LiteralCollection, Scalar
 
+ARRAY_NODE_SUBNODE_NAME = "array_node_sub_node"
+
 
 class ArrayNode:
     def __init__(
@@ -208,6 +210,7 @@ class ArrayNode:
                 entity=self.flyte_entity,
                 add_node_to_compilation_state=False,
                 overridden_interface=collection_interface,
+                node_id=ARRAY_NODE_SUBNODE_NAME,
                 **kwargs,
             )
             self._bindings = bound_subnode.ref.node.bindings
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -30,7 +30,6 @@ class ArrayNode:
         concurrency: Optional[int] = None,
         min_successes: Optional[int] = None,
         min_success_ratio: Optional[float] = None,
-        bound_inputs: Optional[Set[str]] = None,
         metadata: Optional[Union[_workflow_model.NodeMetadata, TaskMetadata]] = None,
     ):
         """
@@ -42,7 +41,6 @@ class ArrayNode:
         :param min_successes: The minimum number of successful executions. If set, this takes precedence over
             min_success_ratio
         :param min_success_ratio: The minimum ratio of successful executions.
-        :param bound_inputs: The set of inputs that should be bound to the map task
         :param execution_mode: The execution mode for propeller to use when handling ArrayNode
         :param metadata: The metadata for the underlying entity
         """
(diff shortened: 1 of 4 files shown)


flytekit/core/array_node.py
```diff
@@ -190,6 +192,16 @@ def execution_mode(self) -> _core_workflow.ArrayNode.ExecutionMode:
         return self._execution_mode
 
     def __call__(self, *args, **kwargs):
+        ctx = FlyteContext.current_context()
+        if ctx.compilation_state is not None:
+            # interface mismatch between the target and the actual inputs since we don't create a new entity with a
+            # transformed to list interface
+            transformed_kwargs = {
```
pvditt, line 199: this is fairly hacky. Would love if there's a better way to do this.

flytekit/core/array_node.py
```diff
@@ -49,6 +50,7 @@ def __init__(
         self._concurrency = concurrency
         self._execution_mode = execution_mode
         self.id = target.name
+        self._bindings: List[_literal_models.Binding] = []
```
wild-endeavor, line 53: Make this optional?  and set default to None.

flytekit/core/array_node.py
wild-endeavor: I think this is fine... but can you pull this out into a separate function and add some comments and that also makes it easier to write a unit test for this.

I assume this is why the bindings in the subnode in the test seem to be binding to the first item.  Could you add a comment somewhere explaining why?

Also does this feature work in local execution?  If so, could you add some more units that just exercise local execution?

flytekit/core/promise.py
```diff
@@ -1114,7 +1117,9 @@ def create_and_link_node_from_remote(
         upstream_nodes=upstream_nodes,
         flyte_entity=entity,
     )
-    ctx.compilation_state.add_node(flytekit_node)
+
+    if link_node:
```
wild-endeavor, line 1121: flag is maybe more correctly named as add_node_to_compilation_state or something...

tests/flytekit/unit/core/test_array_node.py
```diff
@@ -45,16 +45,22 @@ def test_lp_serialization(serialization_settings):
 
     wf_spec = get_serializable(OrderedDict(), serialization_settings, grandparent_wf)
     assert len(wf_spec.template.nodes) == 1
-    assert wf_spec.template.nodes[0].array_node is not None
```
wild-endeavor, line 48: can you replace with this block?  some minor changes.
```python
    wf_spec = get_serializable(OrderedDict(), serialization_settings, grandparent_wf)
    assert len(wf_spec.template.nodes) == 1
    top_level = wf_spec.template.nodes[0]
    assert top_level.inputs[0].var == "a"
    assert len(top_level.inputs[0].binding.collection.bindings) == 3
    assert top_level.inputs[1].var == "b"
    assert len(top_level.inputs[1].binding.collection.bindings) == 3
    serialized_array_node = wf_spec.template.nodes[0].array_node
    assert serialized_array_node is not None
    assert serialized_array_node.node is not None
    assert serialized_array_node.node.workflow_node is not None
    assert (
        serialized_array_node.node.workflow_node.launchplan_ref.resource_type
        == identifier_models.ResourceType.LAUNCH_PLAN
    )
    assert serialized_array_node.node.workflow_node.launchplan_ref.name == "tests.flytekit.unit.core.test_array_node.parent_wf"
    assert serialized_array_node._min_success_ratio == 0.9
    assert serialized_array_node._parallelism == 10
    assert len(serialized_array_node.node.inputs) == 2
    assert serialized_array_node.node.inputs[0].var == "a"
    assert serialized_array_node.node.inputs[0].binding.scalar.primitive.integer == 1
    assert serialized_array_node.node.inputs[1].var == "b"
    assert serialized_array_node.node.inputs[1].binding.scalar.primitive.integer == 2
```

wild-endeavor:
also, one more question, so the unit test with all the asserts... that's how propeller expects the node to be?  It's supposed to bound to the first element of the list?

Can we write an example with unions also?  What if we're mapping over a union, and different elements are of different types of the union, does that matter?

pvditt:
> also, one more question, so the unit test with all the asserts... that's how propeller expects the node to be? It's supposed to bound to the first element of the list?
> 
> Can we write an example with unions also? What if we're mapping over a union, and different elements are of different types of the union, does that matter?

@wild-endeavor This was the hacky-ness I was referring to. 

We don't update the interface of the underlying subnode, launch plan in this case, to a list of a type so we bypass this mismatch during [compilation](https://github.com/unionai/flyte/blob/bfb0329a5e418d059667ee8a982d0ed914aad43b/flytepropeller/pkg/compiler/validators/interface.go#L162-#L188). ArrayNode SubNodeSpec input bindings aren't utilized in propeller. 

This input binding would just be used to set a [parameter name to get set for a workflow node's input](https://github.com/unionai/flyte/blob/bfb0329a5e418d059667ee8a982d0ed914aad43b/flytepropeller/pkg/compiler/validators/interface.go#L81). I guess this could also lead to some issues with GateNodes if/when we add support to map over those as well.
 
This definitely could lead to other future issues/confusion. Let me look into updating binding_data_from_python_std to handle the mismatch between a simple type and a collection.

I do have this [1 line compiler update](https://github.com/unionai/flyte/pull/464) to resolve the root problem of ArrayNode mapping launch plans not compiling if default input is not set by setting the ArrayNode inputs on the subnode spec. Ideally I'd like to set the subnode spec's inputs through the SDK rather than doing it when compiling a workflow.

tests/flytekit/unit/core/test_array_node.py
pvditt: yup - just updated these

flytekit/core/array_node.py
pvditt: @wild-endeavor cleaned this up a lot by passing in an "overriden_interface" that is the subnode's interface transformed to a list interface like we do in array node map tasks. By passing in a fake transformed to list interface we can bind all the inputs to the subnode

flytekit/core/array_node.py
pvditt: @wild-endeavor would it make sense to keep this as is since this isn't a passed in value and to have:

```
    @property
    def bindings(self) -> List[_literal_models.Binding]:
        # Required in get_serializable_node
        return self._bindings
```

return an empty list if the bindings aren't set

flytekit/core/array_node.py
pvditt: @wild-endeavor and yup this works in local execution. There is a lone unit test for it - let me add more to it.

flytekit/core/array_node.py
```diff
@@ -137,7 +139,8 @@ def local_execute(self, ctx: FlyteContext, **kwargs) -> Union[Tuple[Promise], Pr
         literals = []
         for i in range(mapped_entity_count):
             single_instance_inputs = {}
-            for k in self.python_interface.inputs.keys():
+            for binding in self.bindings:
+                k = binding.var
                 if k not in self._bound_inputs:
                     single_instance_inputs[k] = kwargs[k][i]
```
pvditt, line 145: could probably iterate through the bindings here instead, but this feels cleaner for now...

flytekit/core/array_node.py
pvditt: @wild-endeavor there are a couple unit tests to test local execute: https://github.com/flyteorg/flytekit/blob/fb55841f8660b2a31e99381dd06e42f8cd22758e/tests/flytekit/unit/core/test_array_node.py#L75 + https://github.com/flyteorg/flytekit/blob/fb55841f8660b2a31e99381dd06e42f8cd22758e/tests/flytekit/unit/core/test_array_node.py#L99

flytekit/core/array_node.py
wild-endeavor: see slack

flytekit/core/promise.py
```diff
@@ -1214,15 +1220,24 @@ def create_and_link_node(
     # These will be our core Nodes until we can amend the Promise to use NodeOutputs that reference our Nodes
     upstream_nodes = list(set([n for n in nodes if n.id != _common_constants.GLOBAL_INPUT_NODE_ID]))
 
+    # TODO: Better naming, probably a derivative of the function name.
+    # if not adding to compilation state, we don't need to generate a unique node id
+    node_id = (
+        f"{ctx.compilation_state.prefix}n{len(ctx.compilation_state.nodes)}"
+        if add_node_to_compilation_state and ctx.compilation_state
+        else ""
```
wild-endeavor, line 1228: can we add an additional optional argument to this function for node_id?  and then let's pass in a constant `"ARRAY_NODE_SUB"` or some other constant string?  at least that way it'll be obvious.

pvditt: sounds good. Updated

wild-endeavor (approved):

PR #2745, opened 2024-09-11
[flyteagent] FileSensor Timeout (Remote Execution Only)

Future-Outlier:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5035

timeout is important for sensor task.

add a parameter `timeout` in `FileSensor` and `BaseSensor`.

## How was this patch tested?
```python
from flytekit.sensor.file_sensor import FileSensor

sensor = FileSensor(name="test_file_sensor", timeout=1)

@task
def t1():
    return

@workflow
def wf():
    sensor(path="/tmp/abcde")
```
### Screenshots
image

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements timeout configuration support across multiple components, integrating it into task metadata with conflict warnings. The BaseSensor includes timeout validation with guards to prevent setting timeout and metadata simultaneously. The changes enhance the sensor testing suite by adding new test cases that validate behavior when timeout and TaskMetadata are set concurrently, improving test coverage and supporting FlyteKit's timeout functionality.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>


Future-Outlier:
Hi, @pingsutw 
Do you think it's a good idea to support timeout by `Resource Wrapper` in `flyteplugins/agent`?
https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/plugins/webapi/agent/plugin.go#L40-L48

Or should we make this feature more generic?

flytekit/core/base_task.py
```diff
@@ -203,12 +203,13 @@ def __init__(
         task_type_version=0,
         security_ctx: Optional[SecurityContext] = None,
         docs: Optional[Documentation] = None,
+        timeout: Optional[Union[datetime.timedelta, int]] = None,
```
blaketastic2, line 206: Would it make more sense to expose `metadata` to the `PythonTask` and have `BaseSensor` initialize the metadata w/ `TaskMetadata(timeout=timeout)`?

Future-Outlier: yes it will make more sense.

flytekit/core/base_task.py
```diff
@@ -212,6 +214,9 @@ def __init__(
         self._task_type_version = task_type_version
         self._security_ctx = security_ctx
         self._docs = docs
+        self._timeout = timeout
+        # put timeout to metadata here
+        # raise error if metadata and timeout are both set
```
thomasjpfan, line 219: I do not like how `timeout` can be defined in two places. If we really want a easy way to set `timeout`, I would do this logic `FileSensor` and not complicate the `Task` API.

davidmirror-ops:
any chance we could revive this?

kumare3:
Yes please

Future-Outlier:
it work again now!
image

flytekit/sensor/base_sensor.py
```diff
@@ -60,12 +62,14 @@ def __init__(
         for k, _ in signature.parameters.items():  # type: ignore
             annotation = type_hints.get(k, None)
             inputs[k] = annotation
+        metadata = TaskMetadata(timeout=timeout)
```
thomasjpfan, line 65: Should we raise a reasonable error in the case where the task metadata is set together with the timeout?

```python
FileSensor(timeout=12, metadata=TaskMetadata(...))
```

Future-Outlier: make sense, will add it

Future-Outlier: > Should we raise a reasonable error in the case where the task metadata is set together with the timeout?
> 
> ```python
> FileSensor(timeout=12, metadata=TaskMetadata(...))
> ```

Thank you!!!

thomasjpfan: Can you add a quick test with `pytest.raises` to check the error?

thomasjpfan (commented):
Minor nit otherwise LGTM

flytekit/sensor/base_sensor.py
Future-Outlier: no problem, will add it now

Future-Outlier:
> Minor nit otherwise LGTM

yes done, thank you!

pingsutw (approved):

PR #2747, opened 2024-09-11
reference lp example for flyteremote

dansola:

We would like added clarification for how to extract I/O from reference launchplans using FlyteRemote

Added an example of a reference launchplan and flyte remote usage to extract output.

## How was this patch tested?

Registered example, ran it and using the code in the example.

### DIFF
--- a/docs/source/design/control_plane.rst
+++ b/docs/source/design/control_plane.rst
@@ -317,6 +317,42 @@ To fetch output of a specific node execution:
 
 :ref:`Node <flyte:divedeep-nodes>` here, can correspond to a task, workflow, or branch node.
 
+Reference launch plan executions
+====
+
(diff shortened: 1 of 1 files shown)


docs/source/design/control_plane.rst
```diff
@@ -317,6 +317,42 @@ To fetch output of a specific node execution:
 
 :ref:`Node <flyte:divedeep-nodes>` here, can correspond to a task, workflow, or branch node.
 
+Reference launch plan executions
+====
```
neverett, line 321: This needs to be as long as the title above -- I think that should clear up the build warning!

wild-endeavor (approved):

pingsutw (approved):

PR #2426, opened 2024-05-16
Propagate custom_info Dict through agent Resource

ddl-ebrown:
See also https://github.com/flyteorg/flyte/pull/5391

## Why are the changes needed?

- The agent defines a Resource return type with values:

   * outputs
   * message
   * log_links
   * phase

   These are all a part of the underlying protobuf contract defined in
   flyteidl.

   However, the message field custom_info from the protobuf is not here

   google.protobuf.Struct custom_info

   https://github.com/flyteorg/flyte/blob/519080b6e4e53fc0e216b5715ad9b5b5270f35c0/flyteidl/protos/flyteidl/admin/agent.proto#L140

   This field was added in https://github.com/flyteorg/flyte/pull/4874
   but never made it into the corresponding flytekit PR
   https://github.com/flyteorg/flytekit/pull/2146

 - It's useful for agents to return additional metadata about the job,
   and it looks like custom_info is the intended location

 - Make a minor refactor to how the agent responds to requests that
   return Resource by implementing to_flyte_idl / from_flyte_idl
   directly

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #2752, opened 2024-09-16
Async type engine

wild-endeavor:
## Changes
This PR converts the core functions in the TypeEngine to async, adding synchronized functions where needed.

* Introduce a new AsyncTypeTransformer abstract class.
* Added ``async_to_literal and `async_to_python_value` to `TypeEngine`.
* Added a ContextExecutor (inside a new folder that has no internal imports other than `flytekit.logger`) that will copy contextvars.  Without this, we would lose the `FlyteContext` information.
* Moved the async loop manager from `tools` to a new folder called `utils` to avoid circular dependencies (tools already import core).
* Reverted part of https://github.com/flyteorg/flytekit/pull/2426/files to not have the Resource dataclass be a model.

### Notes
Noticed a couple issues while developing this.  Most of these notes are more relevant to the pr that introduced a https://github.com/flyteorg/flytekit/pull/2784 already merged.
#### Event loop overriding 
We ran into this first with a bug in [eager mode](https://github.com/flyteorg/flytekit/pull/2737) where we noticed that when certain libraries were used in conjunction with eager mode, those libraries would break.  What was happening was that that library had on load, created and installed an event loop on the main thread.  When flytekit called `asyncio.run` to run an eager workflow, `run` removed and effectively destroyed that event loop, replacing it with its own, and thereafter cleaning it up.  The fact that any library, as well as user code, at any point, is liable to call `asyncio.run` means that at any point, the loop that flytekit relies on can be destroyed if not properly protected.
#### Inherent recursive behavior
The nature of what the type engine does is very recursive.  For example, the Union transformer recursively may call the List transformer which may then call another transformer.  Because we don't want to force users to only write async transformers from now on, and because users may provide for their own list/union/dict-like transformers that recursively call the type engine, a common scenario that we have to account for is
```
synchronous code -> call async code -> call synchronous code -> call async code
```
This seemingly innocuous pattern is actually a bit tricky to account for.  Typically when you call async code for the first time, you'd use for something like `asyncio.run` or `loop.run_until_complete`.  These root level async entrypoints are only available if they haven't been called before (or have finished).  That is there can only be one of these calls, for a given thread, at any point.  At the point of the second async call in the scenario above, the invoking synchronous function cannot call these root async calls again.

To get around this there's two main solutions.
* Have certain synchronous functions be only callable from a non-async context (that is `asyncio.get_running_loop` must raise a `RuntimeError`).  Other synchronous functions, like the second one in the above scenario, must also return an `asyncio.Future`, in addition to the original return type of the function.
* At the point of the second async invocation, spin up a new thread to run the loop and new coroutine.

The advantage of the first approach is single-threaded-ness, along with not introducing additional blocking.  The advantage of the second approach is that you don't have functions that randomly can't be called from async functions, and other functions that randomly return a Union of a future.  The approach that we decided to pursue with the loop manager is the second approach.  Using the loop manager code, instead of calling `asyncio.run` we can call `loop_manager.run_sync` everywhere.  Haven't yet been able to get signal handlers to work to handle loop termination but that should be okay since we don't need that yet.

Still todo (post-merge):
* Integration testing on more complex types
* Additional perf testing

## How was this patch tested?
Unit testing and tested locally and by running workflows on dev cluster.  The scope of this PR is such that it will need a fair bit of alpha candidate testing.  Something that we can do post-merge, esp in conjunction with some of the other larger changes slated for this cycle.


flytekit/core/type_engine.py
```diff
@@ -204,6 +209,82 @@ def __str__(self):
...
+                logger.error(f"Unknown RuntimeError {str(e)}")
+                raise
+
+        if loop:
+            coro = self.async_to_literal(ctx, python_val, python_type, expected)
+            fut = loop.create_task(coro)
+            return fut
+        else:
+            coro = self.async_to_literal(ctx, python_val, python_type, expected)
+            return asyncio.run(coro)
```
wild-endeavor, line 266: everywhere where run is called will need to be swapped out for whatever async running mechanism we decide on.

flytekit/core/array_node_map_task.py
```diff
@@ -29,6 +29,7 @@
 from flytekit.tools.module_loader import load_object_from_module
 from flytekit.types.pickle import pickle
 from flytekit.types.pickle.pickle import FlytePickleTransformer
+from flytekit.utils.asyn import loop_manager
```
thomasjpfan, line 32: Can this be imported as `run_sync` and then use that? (I prefer to hide the loop manager as much as possible)

flytekit/core/promise.py
```diff
@@ -94,16 +96,19 @@ def my_wf(in1: int, in2: int) -> int:
...
             raise
 
     return result
 
 
-def resolve_attr_path_in_promise(p: Promise) -> Promise:
+translate_inputs_to_literals = loop_manager.synced(_translate_inputs_to_literals)
+
+
+async def resolve_attr_path_in_promise(p: Promise) -> Promise:
```
thomasjpfan, line 111: Should we be concerned with backward compatibility with changing a function from sync to async?

flytekit/core/promise.py
```diff
@@ -895,12 +906,21 @@ def binding_from_python_std(
...
-        t_value_type,
-        nodes,
+    try:
+        asyncio.get_running_loop()
+        raise AssertionError("binding_from_python_std cannot be run from within an async call stack")
+    except RuntimeError as e:
+        if "no running event loop" not in str(e):
+            logger.error(f"Unknown RuntimeError {str(e)}")
+            raise
+    binding_data = asyncio.run(
```
thomasjpfan, line 916: What is the issue with using `sync_run` here?

thomasjpfan (commented):
Is there an issue/write up that describes the user or developer facing benefits of having a async type engine?

flytekit/core/promise.py
wild-endeavor: updated

flytekit/core/array_node_map_task.py
wild-endeavor: i guess so.  i was hoping to one day figure out type hinting, which doesn't work today in either case, but i think will never work with the run sync way.  I tried briefly getting functool wraps to work but alas to no avail.

flytekit/core/promise.py
wild-endeavor: this particular function is pretty limited in terms of callees so hopefully that's okay.

flytekit/core/array_node_map_task.py
wild-endeavor: ohh it was missing a paramspec.  it works now. and changed promise.py and base_task.py to just use the run_sync.

wild-endeavor:
@thomasjpfan 
> Is there an issue/write up that describes the user or developer facing benefits of having a async type engine?

Not currently no, I did some preliminary local testing and noticed an improvement so it's something I think is worth doing.  I'm planning on putting together more performance testing in general though for the release notes covering a variety of scenarios.  But ultimately speed is the main driver ofc.

eapolinario (approved):
This is amazing. Really. The first step in the next round of innovation in flytekit. Thank you for working on this.

PR #2755, opened 2024-09-18
Update pyflyte defaults to use --copy behavior

wild-endeavor:
This is the last of a series of PRs to clean up the fast register switches in the various `pyflyte` commands.  The main change was https://github.com/flyteorg/flytekit/pull/2690

This PR changes the default behavior for fast registration (invoked by `run`, `register`, and `package` to use the [new logic](https://github.com/flyteorg/flytekit/blob/fb55841f8660b2a31e99381dd06e42f8cd22758e/flytekit/tools/fast_registration.py#L116) instead.  Previously if the `--copy` command was not specified, the old logic was still used.  Only the `serialize` command after this PR will use the old logic.

* Change messaging, exception raising, and defaults to trigger the new logic.

## How was this patch tested?
Locally tested and local executions run.

## Related PRs
https://github.com/flyteorg/flytekit/pull/2690
https://github.com/flyteorg/flytekit/pull/2716
https://github.com/flyteorg/flytekit/pull/2731

### DIFF
--- a/flytekit/clis/sdk_in_container/package.py
+++ b/flytekit/clis/sdk_in_container/package.py
@@ -1,4 +1,5 @@
 import os
+import sys
 import typing
 
 import rich_click as click
@@ -131,6 +132,13 @@ def package(
     """
     # Ensure that the two flags are consistent
     if fast:
+        if "--copy" in sys.argv:
+            raise click.BadParameter(
+                click.style(
+                    "Cannot use both --fast and --copy flags together. Please move to --copy",
+                    fg="red",
+                )
+            )
         click.secho("The --fast flag is deprecated, please use --copy all instead", fg="yellow")
         copy = CopyFileDetection.ALL
 
--- a/flytekit/clis/sdk_in_container/package.py
+++ b/flytekit/clis/sdk_in_container/package.py
@@ -131,7 +131,7 @@ def package(
     """
     # Ensure that the two flags are consistent
     if fast:
-        click.secho(f"The --fast flag is deprecated, please use --copy all instead", fg="yellow")
+        click.secho("The --fast flag is deprecated, please use --copy all instead", fg="yellow")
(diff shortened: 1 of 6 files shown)


flytekit/clis/sdk_in_container/package.py
```diff
@@ -54,17 +54,18 @@
...
+    help="[Deprecated, see --copy] This flag enables fast packaging, that allows `no container build`"
+    " deploys of flyte workflows and tasks. You should specify --copy all/auto instead"
     " Note this needs additional configuration, refer to the docs.",
 )
 @click.option(
     "--copy",
     required=False,
     type=click.Choice(["all", "auto", "none"], case_sensitive=False),
-    default=None,  # this will be changed to "none" after removing fast option
+    default="none",
```
thomasjpfan, line 65: As long as `--fast` is in the codebase, someone can still run `--fast --copy auto`, which is conflicting/confusing. I like keeping the default for `copy` as `None` and raising the current error.

When we actually remove `--fast` as an option, then I feel safe enough to set `default="none"`.

wild-endeavor: we won't remove it though for a while, at least a year or two... so just keep it around?

eapolinario: can we detect (and fail) in this case? Also, add a message that explains how to fix the issue.

wild-endeavor: also did you mean this for `package` specifically or the commands in general?  if so we can just discard this PR... but that leaves the old logic and doesn't exercise the new logic unless it's opted into.

wild-endeavor: we can detect and fail yeah... but i'd like to exercise the new code

eapolinario (approved):

PR #2761, opened 2024-09-19
add support for mapping over remote launch plans

pvditt:
## Tracking issue
fixes: https://linear.app/unionai/issue/COR-1686/support-mapping-over-flytelaunchplan-remote

Mapping over remote launch plans doesn't work

- add support for mapping over remote launch plans by using create_and_link_node_from_remote
- updated models TypedInterface to transform variables of a type or variables of a collection of type

## How was this patch tested?
- unit testing
- sandbox
- dogfood (https://dogfood-gcp.cloud-staging.union.ai/console/projects/flytesnacks/domains/development/executions/arqcgvlsbz92cdffsw2d/nodeId/n0/nodes)

### Setup process
```
remote = UnionRemote(config=Config.auto(), default_project="flytesnacks", default_domain="development")
lp1 = remote.fetch_launch_plan(name="advanced_composition.subworkflow.regression_line_wf",
                               version="v0.3.288")


@workflow
def test_wf1_remote():

    val = [1, 2, 3]
    x = [[-3, 0, 3], [-8, 2, 4], [7, 3, 1]]
    y = [[7, 4, -2], [-2, 4, 7], [3, 6, 4]]
    map_task(lp1)(val=val, x=x, y=y)
```

### DIFF
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -1,5 +1,5 @@
 import math
-from typing import Any, List, Optional, Set, Tuple, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Set, Tuple, Union
 
 from flyteidl.core import workflow_pb2 as _core_workflow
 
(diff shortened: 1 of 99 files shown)


flytekit/core/array_node.py
```diff
@@ -27,7 +31,7 @@
 class ArrayNode:
     def __init__(
         self,
-        target: LaunchPlan,
+        target: Any,
```
eapolinario, line 34: Can we use `Union` to describe the types? I'd rather keep `ArrayNode` well-typed.

eapolinario (commented):
Thanks for fixing this. Let's try to keep the interface clean by not using `Any` at this point, unless we have a very good reason.

flytekit/core/array_node.py
pvditt: yup - getting an update for this. Was working around a circular imports error. Can use if TYPE_CHECKING

wild-endeavor:
I think there's some circular dependency that needs to happen but I think maybe what's here right now is more than necessary.  let's chat about this some time today or tomorrow?

wild-endeavor (approved):

PR #2762, opened 2024-09-21
[flyteagent] All agents return dict instead of literal map

Future-Outlier:
## Tracking issue
[https://github.com/flyteorg/flyte/issues/3936](https://github.com/flyteorg/flyte/issues/3936)

When users trace the code, they tend to follow patterns we use in Flytekit's agent. By allowing them to use `dict` instead of `LiteralMap`, we can improve their development experience. This way, they don't need to understand the details of how `LiteralMap` works or what a literal is.

In the agent's `get` request:

**Before:**
- Returns a `LiteralMap`.

**After:**
- Returns a `dict`.

Note: We will automatically handle the conversion from `dict` to `LiteralMap` for the user.
- In local execution: 
  [flytekit base_agent.py](https://github.com/flyteorg/flytekit/blob/9bce7c37e950d33688820849b127d1b8ee8af7aa/flytekit/extend/backend/base_agent.py#L315-L316)
- In remote execution: 
  [flytekit agent_service.py](https://github.com/flyteorg/flytekit/blob/9bce7c37e950d33688820849b127d1b8ee8af7aa/flytekit/extend/backend/agent_service.py#L141-L145)

## How was this patch tested?
Through unit tests.

### DIFF
--- a/plugins/flytekit-bigquery/tests/test_agent.py
+++ b/plugins/flytekit-bigquery/tests/test_agent.py
@@ -90,7 +90,7 @@ def test_bigquery_agent(mock_client, mock_query_job):
     resource = agent.get(metadata)
     assert resource.phase == TaskExecution.SUCCEEDED
     assert (
-        resource.outputs.literals["results"].scalar.structured_dataset.uri
+        resource.outputs["results"].uri
         == "bq://dummy_project:dummy_dataset.dummy_table"
     )
     assert resource.log_links[0].name == "BigQuery Console"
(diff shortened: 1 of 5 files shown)


plugins/flytekit-bigquery/tests/test_agent.py
```diff
@@ -90,7 +90,7 @@ def __init__(self):
     resource = agent.get(metadata)
     assert resource.phase == TaskExecution.SUCCEEDED
     assert (
-        resource.outputs.literals["results"].scalar.structured_dataset.uri
+        resource.outputs["results"].uri
```
thomasjpfan, line 93: Technically, if there is another library that depends on `resource.outputs.literals` being defined, then this breaks backward compatibility.

I do not think there are anyone that depends on this, so I am okay with this change.

thomasjpfan (approved):

plugins/flytekit-bigquery/tests/test_agent.py
Future-Outlier: Thank you so much for thinking so carefully

PR #2763, opened 2024-09-22
Adds new pyflyte capability to run arbitrary scripts on remote server.

kumare3:
A lot of repos in the world have simply scripts or torchrun scripts to run. Also some users where flyte is already deployed may want to run a script on remote. This command makes it possible, with a future extension to support arbitrary job type.

No need to import flytekit, just have local flytekit environment and run any python scripts on remote. Example training scripts etc. You can request specific resources too

Example Code:
hello.py
```python
print("Hello world")
```

Can be run
```bash
pyflyte run --remote script hello.py
```

To run on a bigger box `cfg.yaml`
```bash
run_command: |
  python test.py
resources:
  gpu: 1
  mem: 2Gi
  accelerator: nvidia-a10g
secrets:
  - key: openai
```

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -13,7 +13,7 @@ from typing import Iterator, get_args
 
 import rich_click as click
 import yaml
-from click import Context
+from click import Context, Command
 from mashumaro.codecs.json import JSONEncoder
 from rich.progress import Progress
 from typing_extensions import get_origin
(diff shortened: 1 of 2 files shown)


pingsutw (commented):
Some ideas about it.
- We could add a flag (backend) that allows people to run the script on different platform, such as databricks, aws batch, etc.
- It would be great if we could specify python dependencies or conda env name in the config file. When we run the workflow task locally, flytekit will use that conda env to execute the tasks.

kumare3:
@pingsutw we should add no flag. It should use the same Agent and backend plugin framework. We should add support for task_config. For this we should standardize on pydantic for all task configs

PR #2764, opened 2024-09-23
GH-5768: Better pyflyte boolean parsing

Tom-Newton:
## Tracking issue
Closes [flyteorg/flyte#999](https://github.com/flyteorg/flyte/issues/5768)

The current boolean parsing behaviour is confusing for default True arguments. 

Add `--no_...` and `--no-...` variants for boolean flags as per the recommendation at https://click.palletsprojects.com/en/8.1.x/options/#boolean-flags

## How was this patch tested?
Wrote new unit tests to cover it. 

- [x] I updated the documentation accordingly - I'm not aware of any docs that need updating. pyflyte's `--help` should be automatically taken care of by the `click` library. 

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -449,10 +449,14 @@ def to_click_option(
     # If a query has been specified, the input is never strictly required at this layer
     required = False if default_val and isinstance(default_val, ArtifactQuery) else required
 
+    if literal_converter.is_bool():
+        click_cli_paramater_name = f"--{input_name}/--no_{input_name}"
+    else:
+        click_cli_paramater_name = f"--{input_name}"
+
     return click.Option(
-        param_decls=[f"--{input_name}"],
+        param_decls=[click_cli_paramater_name],
         type=literal_converter.click_type,
-        is_flag=literal_converter.is_bool(),
         default=default_val,
         show_default=True,
         required=required,
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -450,7 +450,10 @@ def to_click_option(
     required = False if default_val and isinstance(default_val, ArtifactQuery) else required
 
     if literal_converter.is_bool():
-        click_cli_parameter_names = [f"--{input_name}/--no_{input_name}", f"--{input_name}/--no-{input_name.replace('_', '-')}"]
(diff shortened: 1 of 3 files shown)


flytekit/clis/sdk_in_container/run.py
```diff
@@ -449,10 +449,14 @@ def to_click_option(
     # If a query has been specified, the input is never strictly required at this layer
     required = False if default_val and isinstance(default_val, ArtifactQuery) else required
 
+    if literal_converter.is_bool():
+        click_cli_paramater_name = f"--{input_name}/--no_{input_name}"
```
eapolinario, line 453: Can we use `--no-{input_name}`?

eapolinario (commented):
Thank you. This makes sense. I have only one small comment.

flytekit/clis/sdk_in_container/run.py
Tom-Newton: We can. I opted for `--no_` because `input_name` comes from the python function arguments and is therefore likely to use underscores instead of dashes. On the other hand `--no-` is probably more common amongst general `click` usage. I don't have a strong opinion either way.

Tom-Newton: Please let me know if you would rather I change to `--no-{input_name}`

flytekit/clis/sdk_in_container/run.py
```diff
@@ -449,10 +449,14 @@ def to_click_option(
...
+    else:
+        click_cli_paramater_name = f"--{input_name}"
```
pingsutw, line 455:
```suggestion
        click_cli_parameter_name = f"--{input_name}"
```

flytekit/clis/sdk_in_container/run.py
pingsutw, line 453: are we able to support both? if the input is `my_input: int`, pyflyte should accept both `--no_my_input` and `--no-my-input`.
```suggestion
        click_cli_parameter_name = [f"--{input_name}/--no_{input_name}", f"--{input_name}/--no-{input_name.replace('_', '-')}"]
```

Tom-Newton: It looks like we can :+1:

pingsutw (dismissed):
There is a lint error. otherwise LGTM

eapolinario (approved):

PR #2765, opened 2024-09-23
Enable Spark Fast Register

pingsutw:
## Tracking issue
NA

When we fast-register a Spark task, we will always get a `module not found` error in the Spark executors.

Archive the current directory into a zip file and send it to the spark executor. Exectors will load the module from that zip file.

## How was this patch tested?
```
pyflyte run --remote fast_spark.py my_spark
```

### Setup process
```python
import datetime
import random
from operator import add

import flytekit
from flytekitplugins.spark import Spark
from long_description import hello  # local module 
from misc.ignore_pattens import hello123  # local module

flytekit_hash = "be5d8eac263fc8e5d57de110f9648b696d48f20d"
new_flytekit = f"git+https://github.com/flyteorg/flytekit@{flytekit_hash}"
spark_plugins = f"git+https://github.com/flyteorg/flytekit.git@{flytekit_hash}#subdirectory=plugins/flytekit-spark"
custom_image = ImageSpec(
    # python_version="3.9",
    # base_image="pingsutw/spark:v1",
    registry="ghcr.io/flyteorg",
    # packages=[spark_plugins],
    packages=[spark_plugins, new_flytekit],
    apt_packages=["git"]
)


@task(
    task_config=Spark(
        # This configuration is applied to the Spark cluster
        spark_conf={
            "spark.driver.memory": "1000M",
            "spark.executor.memory": "1000M",
            "spark.executor.cores": "1",
            "spark.executor.instances": "2",
            "spark.driver.cores": "1",
            "spark.jars": "https://storage.googleapis.com/hadoop-lib/gcs/gcs-connector-hadoop3-latest.jar",
        },
        executor_path="/usr/bin/python3",
        applications_path="local:///usr/local/bin/entrypoint.py",
    ),
    limits=Resources(mem="2000M"),
    container_image=custom_image,
)
def hello_spark(partitions: int) -> float:
    print("Starting Spark with Partitionsssss: {}".format(partitions))
    hello()
    hello123()

    n = 1 * partitions
    sess = flytekit.current_context().spark_session
    count = sess.sparkContext.parallelize(range(1, n + 1), partitions).map(f).reduce(add)

    pi_val = 4.0 * count / n
    return pi_val


def f(_):
    x = random.random() * 2 - 1
    y = random.random() * 2 - 1
    return 1 if x**2 + y**2 <= 1 else 0


@task(
    container_image=custom_image,
)
def print_every_time(value_to_print: float, date_triggered: datetime.datetime) -> int:
    print("My printed value: {} @ {}".format(value_to_print, date_triggered))
    hello()
    hello123()

    return 1


@workflow
def my_spark(triggered_date: datetime.datetime = datetime.datetime(2020, 9, 11)) -> float:
    """
    Using the workflow is still as any other workflow. As image is a property of the task, the workflow does not care
    about how the image is configured.
    """
    pi = hello_spark(partitions=1)
    print_every_time(value_to_print=pi, date_triggered=triggered_date)
    return pi
```

### Screenshots
Screenshot 2024-09-23 at 2 03 50 PM

## Related PRs
NA

## Docs link
NA

### DIFF
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -270,6 +270,8 @@ class _ModuleSanitizer(object):
         # Let us remove any extensions like .py
         basename = os.path.splitext(basename)[0]
 
+        # This is an escape hatch for the zipimporter (used by spark).  As this function is called recursively,
+        # it'll eventually reach the zip file, which is not extracted, so we should return.
         if not Path(dirname).is_dir():
             return basename
 
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -274,6 +274,9 @@ class _ModuleSanitizer(object):
         # Let us remove any extensions like .py
         basename = os.path.splitext(basename)[0]
 
+        if not Path(dirname).is_dir():
+            return basename
+
         if dirname == package_root:
             return basename
 
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -113,6 +113,10 @@ class TrackedInstance(metaclass=InstanceTrackingMeta):
         return n
 
     @property
+    def module_file(self) -> str:
+        return str(self._module_file)
+
+    @property
     def lhs(self):
         if self._lhs is not None:
             return self._lhs
--- a/flytekit/core/tracker.py
+++ b/flytekit/core/tracker.py
@@ -113,10 +113,6 @@ class TrackedInstance(metaclass=InstanceTrackingMeta):
         return n
 
     @property
-    def module_file(self) -> str:
-        return str(self._module_file)
-
-    @property
     def lhs(self):
         if self._lhs is not None:
             return self._lhs
--- a/plugins/flytekit-spark/flytekitplugins/spark/task.py
+++ b/plugins/flytekit-spark/flytekitplugins/spark/task.py
@@ -218,8 +218,8 @@ class PysparkFunctionTask(AsyncAgentExecutorMixin, PythonFunctionTask[Spark]):
         self.sess = sess_builder.getOrCreate()
         # self.sess.addArtifacts("fast_spark.py", file=True)
         # self.sess.sparkContext.addPyFile(self.module_file)
-        shutil.make_archive("archive", 'zip', os.getcwd())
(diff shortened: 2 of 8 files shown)


flytekit/core/tracker.py
```diff
@@ -270,6 +270,9 @@ def _resolve_abs_module_name(self, path: str, package_root: typing.Optional[str]
         # Let us remove any extensions like .py
         basename = os.path.splitext(basename)[0]
 
+        if not Path(dirname).is_dir():
```
pingsutw, line 273: since the executor will load the workflow from the zip file. 
image

pingsutw:
cc @fiedlerNr9

plugins/flytekit-spark/flytekitplugins/spark/task.py
```diff
@@ -201,6 +195,21 @@ def pre_execute(self, user_params: ExecutionParameters) -> ExecutionParameters:
             sess_builder = sess_builder.config(conf=spark_conf)
 
         self.sess = sess_builder.getOrCreate()
+
+        if ctx.execution_state and ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION:
+            current_time = time.time()
+            current_dir = os.getcwd()
+            # Set the modification time of all files in the current directory to the current time
+            # since fast register doesn't preserve the modification time of the files and make_archive
```
wild-endeavor, line 203: we arbitrarily set the time in fast register to 1970 just cuz.  it can be any constant.  if we make it 1980, would that make this cleaner?

pingsutw: set it to 1980

flytekit/core/tracker.py
wild-endeavor, line 273:
```suggestion
        # This is an escape hatch for the zipimporter (used by spark).  As this function is called recursively,
        # it'll eventually reach the zip file, which is not extracted, so we should return.
        if not Path(dirname).is_dir():
```

wild-endeavor (approved):

wild-endeavor (approved):

PR #2768, opened 2024-09-26
enabled copy-all for programmatic fast-registration via FlyteRemote

granthamtaylor:

One should be able to utilize `--copy-all` via `FlyteRemote.fast_register_workflow`

I expose `--copy-all` to fast-registration.

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -883,7 +883,7 @@ class FlyteRemote(object):
         :param serialization_settings: The serialization settings to be used
         :param default_launch_plan: This should be true if a default launch plan should be created for the workflow
         :param options: Additional execution options that can be configured for the default launchplan
-        :param fast_package_options: Options to customize copy_all behavior
+        :param fast_package_options: Options to customize copying behavior
         :return:
         """
         if not isinstance(entity, PythonFunctionWorkflow):
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -874,6 +874,7 @@ class FlyteRemote(object):
         version: typing.Optional[str] = None,
         default_launch_plan: typing.Optional[bool] = True,
         options: typing.Optional[Options] = None,
+        copy_all: bool = False,
     ) -> FlyteWorkflow:
         """
         Use this method to register a workflow with zip mode.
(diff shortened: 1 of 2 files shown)


flytekit/remote/remote.py
```diff
@@ -874,6 +874,7 @@ def fast_register_workflow(
         version: typing.Optional[str] = None,
         default_launch_plan: typing.Optional[bool] = True,
         options: typing.Optional[Options] = None,
+        copy_all: bool = False,
```
thomasjpfan, line 877: @wild-endeavor Since this is new API for `fast_register_workflow`, should we just expose `FastPackageOptions` here?

thomasjpfan (commented):
Thank you for the PR! I left a question around `FastPackageOptions` for maintainers to decide which API to go with.

Context: `FastPackageOptions` covers all the features `copy_all` does and more.

flytekit/remote/remote.py
wild-endeavor: yup that sounds good. please do

granthamtaylor: I have updated it and it works as expected. That being said, it is pretty rough compared to just `copy_all=True` or `--copy-all`

```python
from flytekit.tools.fast_registration import FastPackageOptions
remote.fast_register_workflow(
    my_workflow,
    fast_package_options=FastPackageOptions(ignores=[], copy_style="ALL")
)
```


```python
from flytekit.tools.fast_registration import FastPackageOptions
remote.fast_register_workflow(my_workflow, copy_all=True)
```

flytekit/remote/remote.py
```diff
@@ -882,6 +883,7 @@ def fast_register_workflow(
         :param serialization_settings: The serialization settings to be used
         :param default_launch_plan: This should be true if a default launch plan should be created for the workflow
         :param options: Additional execution options that can be configured for the default launchplan
+        :param fast_package_options: Options to customize copy_all behavior
```
thomasjpfan, line 886:
```suggestion
        :param fast_package_options: Options to customize copying behavior
```

thomasjpfan (approved):
Since this is passing a flag into another method, I am okay with not having a unit test for this.

wild-endeavor (approved):

flytekit/remote/remote.py
wild-endeavor: We can pre-create some of these objects and expose them as constants to improve the experience.

eapolinario (approved):

PR #2772, opened 2024-09-27
Fix array node map task for offloaded literal

pmahindrakar-oss:

When the input to a map task array node is offloaded literal then the current condition breaks out and fails the task.
Adds support to check for offloaded literal in array node and download the literal before the condition checks are done 

## How was this patch tested?

Tested this on sandbox by passing an offloaded literal to map task and using the fix from here https://github.com/flyteorg/flyte/pull/5771 and removing the input overwriting and saw flytekit correctly downloaded offloaded metadata for the passed in literal
example used here for integration test:

```
import typing
from flytekit import task, workflow, map_task

@dataclass
# class InnerDC(DataClassJSONMixin):
class InnerDC:
    a: bool
    b: int
    c: float
    d: list[int]
    e: dict[str, int]
    f: FlyteFile

@dataclass
# class DC(DataClassJSONMixin):
class DC:
    name: str
    other: str
    f: FlyteFile
    inner: InnerDC


@task(cache=True, cache_version="1.1")
def f(n: int) -> DC:
    with open("/tmp/abc", "w") as f:
        f.write("wow")
    name = "a" * 500_000
    return DC(
        name=name,
        other=name,
        f=FlyteFile("/tmp/abc"),
        inner=InnerDC(
            a=True,
            b=42,
            c=3.1415,
            d=[1,2,3],
            e={"a": 1, "b": 2},
            f=FlyteFile("/tmp/abc"),
        )
    )

@task(cache=True, cache_version="1.1")
def get_list(n: int) -> list[int]:
    return list(range(n))

@task(cache=True, cache_version="1.1")
def id_dataclass(dc: DC) -> DC:
    with open(dc.f, "r") as f:
        print(f.read())
    return dc


@task(cache=True, cache_version="1.1")
def consume_list_of_objects(xs: list[DC]) -> list[DC]:
    with open(xs[0].f, "r") as f:
        print(f.read())
    return xs[0:2]


@workflow
def working_wf(n: int = 15) -> list[DC]:
    xs=map_task(f)(n=get_list(n=n))
    return consume_list_of_objects(xs=xs)


@workflow
def passing_map_output_wf(n: int = 15) -> list[DC]:
    xs=map_task(f)(n=get_list(n=n))
    # Notice how we pass the offloaded literal to map_task
    return map_task(id_dataclass)(dc=xs) 

@task(cache=True, cache_version="1.1")
def consume_single_element(y: DC) -> DC:
    with open(y.f, "r") as f:
        print(f.read())
    return y

@workflow
def indexing_map_output_wf(n: int = 15):
    a=passing_map_output_wf(n=n)
    consume_single_element(y=a[0])
```

### DIFF
--- a/tests/flytekit/unit/core/test_array_node_map_task.py
+++ b/tests/flytekit/unit/core/test_array_node_map_task.py
@@ -260,8 +260,8 @@ def test_inputs_outputs_length():
     m = map_task(p3)
     assert m.python_interface.inputs == {"a": int, "b": str, "c": float}
     assert (
-         m.name
-         == "tests.flytekit.unit.core.test_array_node_map_task.map_many_inputs_5d2500dc176052a030efda3b8c283f96-arraynode"
+        m.name
+        == "tests.flytekit.unit.core.test_array_node_map_task.map_many_inputs_5d2500dc176052a030efda3b8c283f96-arraynode"
     )
     r_m = ArrayNodeMapTask(many_inputs, bound_inputs={"a", "c", "b"})
     assert str(r_m.python_interface) == str(m.python_interface)
@@ -482,9 +482,9 @@ def test_load_offloaded_literal(tmp_path, monkeypatch):
                 ctx.execution_state.with_params(mode=context_manager.ExecutionState.Mode.TASK_EXECUTION)
             )
     ) as ctx:
-        list_ints = ["a", "b", "c"]
+        list_strs = ["a", "b", "c"]
         lt = TypeEngine.to_literal_type(typing.List[str])
-        to_be_offloaded = TypeEngine.to_literal(ctx, list_ints, typing.List[str], lt)
+        to_be_offloaded = TypeEngine.to_literal(ctx, list_strs, typing.List[str], lt)
         with open(f"{tmp_path}/literal.pb", "wb") as f:
             f.write(to_be_offloaded.to_flyte_idl().SerializeToString())
 
@@ -499,9 +499,11 @@ def test_load_offloaded_literal(tmp_path, monkeypatch):
             "name": literal
         })
 
-        monkeypatch.setenv("BATCH_JOB_ARRAY_INDEX_VAR_NAME", "name")
-        monkeypatch.setenv("name", "0")
-        t = map_task(say_hello)
-        res = t.dispatch_execute(ctx, lm)
-        assert len(res.literals) == 1
-        assert res.literals["o0"].scalar.primitive.string_value == "hello a!"
+        for index, map_input_str in enumerate(list_strs):
+            monkeypatch.setenv("BATCH_JOB_ARRAY_INDEX_VAR_NAME", "name")
+            monkeypatch.setenv("name", str(index))
+            t = map_task(say_hello)
+            res = t.dispatch_execute(ctx, lm)
+            assert len(res.literals) == 1
+            assert res.literals[f"o{0}"].scalar.primitive.string_value == f"hello {map_input_str}!"
+            monkeypatch.undo()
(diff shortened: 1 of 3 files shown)


Future-Outlier (approved):
looks good, can we add a test for this case?

pmahindrakar-oss:
Had added in another PR https://github.com/flyteorg/flytekit/pull/2773 , will merge it in here after confirmation

tests/flytekit/unit/core/test_array_node_map_task.py
```diff
@@ -464,3 +469,39 @@ def wf():
...
+        lm = LiteralMap({
+            "name": literal
+        })
+
+        monkeypatch.setenv("BATCH_JOB_ARRAY_INDEX_VAR_NAME", "name")
+        monkeypatch.setenv("name", "0")
+        t = map_task(say_hello)
+        res = t.dispatch_execute(ctx, lm)
+        assert len(res.literals) == 1
+        assert res.literals["o0"].scalar.primitive.string_value == "hello a!"
```
eapolinario, line 507: Can we verify that all 3 inputs work? This works by setting the `name` env var like this [test is doing](https://github.com/flyteorg/flytekit/pull/2771/files#diff-31caa03908ae2151c0e6160672d02a7e5e7c7a96e3c8a0a811020c744cf56bb2R503-R510).

tests/flytekit/unit/core/test_array_node_map_task.py
```diff
@@ -464,3 +469,39 @@ def wf():
...
+    def say_hello(name: str) -> str:
+        return f"hello {name}!"
+
+    ctx = context_manager.FlyteContextManager.current_context()
+    with context_manager.FlyteContextManager.with_context(
+            ctx.with_execution_state(
+                ctx.execution_state.with_params(mode=context_manager.ExecutionState.Mode.TASK_EXECUTION)
+            )
+    ) as ctx:
+        list_ints = ["a", "b", "c"]
```
eapolinario, line 485: nit: can we rename this variable to `list_strs`?

tests/flytekit/unit/core/test_array_node_map_task.py
```diff
@@ -255,8 +260,8 @@ def many_inputs(a: int, b: str, c: float) -> str:
     m = map_task(p3)
     assert m.python_interface.inputs == {"a": int, "b": str, "c": float}
     assert (
-        m.name
-        == "tests.flytekit.unit.core.test_array_node_map_task.map_many_inputs_5d2500dc176052a030efda3b8c283f96-arraynode"
+         m.name
+         == "tests.flytekit.unit.core.test_array_node_map_task.map_many_inputs_5d2500dc176052a030efda3b8c283f96-arraynode"
```
eapolinario, line 264: why are we changing this?

eapolinario (commented):
just a few minor comments, otherwise, LGTM.

tests/flytekit/unit/core/test_array_node_map_task.py
pmahindrakar-oss: done

tests/flytekit/unit/core/test_array_node_map_task.py
pmahindrakar-oss: done

tests/flytekit/unit/core/test_array_node_map_task.py
pmahindrakar-oss: it seems my editor is changing it. May be we have a formatting guide which i can import to not do this automatically in the future for my Pycharm IDE

eapolinario (approved):

PR #2775, opened 2024-09-27
Do not copy flytekit itself during fast registration

thomasjpfan:

When writing a workflow from the root `flytekit` directory, the flytekit is automatically picked up and bundled it during fast registration. This upload is incomplete because `flytekit` depends on `flytekit.bin.entrypoints`, which is not imported by the workflow but required by the workflow to work. When this partial copy of `flytekit` is copied into `/root`, then will look for `flytekit.bin.entrypoints` here first, but it does not exist.

With this PR, any files that share the root with `flytekit` will not be uploaded.

## How was this patch tested?

Create a workflow in the root directory and run `pyflyte run --remote`.

### DIFF
--- a/flytekit/tools/script_mode.py
+++ b/flytekit/tools/script_mode.py
@@ -14,6 +14,7 @@ from pathlib import Path
 from types import ModuleType
 from typing import List, Optional, Tuple, Union
 
+import flytekit
 from flytekit.constants import CopyFileDetection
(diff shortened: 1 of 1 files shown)


eapolinario (approved):

Future-Outlier:
This is super amazing, thank you @thomasjpfan 

image

PR #2776, opened 2024-09-27
Support default label/annotation for the default launch plan creating from workflow definition

Mecoli1219:
## Tracking issue
Close flyteorg/flyte#5774

As mentioned in the tracking issue, users need to manually adjust launch plans or toggle settings in the Flyte Console, which is prone to accidents. This PR simplifies the process by allowing labels and annotations to be defined directly in the workflow definition, reducing the need for manual interventions.

Right now, fields in the [Options](https://github.com/flyteorg/flytekit/blob/8bcb9d0cd3f2eea179dd35385bbf9ed6eebba840/flytekit/tools/translator.py#L65) class are used to set values for a LaunchPlan during registration or execution overrides. When a workflow is registered, it automatically creates a default LaunchPlan, but there is currently no way to set default values for these fields in the workflow itself. Adding this capability will allow users to customize the default LaunchPlan at the workflow definition level, ensuring consistency and making the configuration process more efficient.

This PR introduces a new `default_options` argument of type `Options` in the workflow decorator. When a default LaunchPlan is created from a workflow, it will now check if `default_options` is provided and use the specified fields (currently only `labels` and `annotations`, but we can add others in the future) as default values. 

## How was this patch tested?
```python
from flytekit import task, workflow, Annotations, Labels
from flytekit.tools.translator import Options

img = "localhost:30000/flytekit:wf_annotation"

@task(container_image=img)
def wf_labeling_task() -> int:
    return 42


@workflow(
    default_options=Options(
        annotations=Annotations({"annotation": "annotation_wf5"}),
        labels=Labels({"label": "labeling_wf5"}),
    )
)
def wf_labeling_wf5() -> int:
    return wf_labeling_task()


if __name__ == "__main__":
    import os

    print(wf_labeling_wf5())

    runner = CliRunner()
    result = runner.invoke(
        pyflyte.main,
        [
            "run",
            "--remote",
            os.path.realpath(__file__),
            "wf_labeling_wf5",
        ],
    )
    print(result.output)
```

### Setup process
1. Build the image and replace the `img` variable in the code above.
2. Run the code directly or use `pyflyte run`.

### Screenshots
- Successfully execution from FlyteConsole
    image
- Default annotation and labels
   image
- Successfully annotating and labeling
    image

### DIFF
--- a/flytekit/core/workflow.py
+++ b/flytekit/core/workflow.py
@@ -6,7 +6,7 @@ import typing
 from dataclasses import dataclass
 from enum import Enum
 from functools import update_wrapper
-from typing import Any, Callable, Coroutine, Dict, List, Optional, Tuple, Type, Union, cast, overload
+from typing import TYPE_CHECKING, Any, Callable, Coroutine, Dict, List, Optional, Tuple, Type, Union, cast, overload
 
 from typing_inspect import is_optional_type
 
@@ -59,6 +59,9 @@ from flytekit.models import literals as _literal_models
 from flytekit.models.core import workflow as _workflow_model
 from flytekit.models.documentation import Description, Documentation
 from flytekit.types.error import FlyteError
+
+if TYPE_CHECKING:
+    from flytekit.tools.translator import Options
 
 GLOBAL_START_NODE = Node(
     id=_common_constants.GLOBAL_INPUT_NODE_ID,
@@ -194,6 +197,7 @@ class WorkflowBase(object):
         python_interface: Interface,
         on_failure: Optional[Union[WorkflowBase, Task]] = None,
         docs: Optional[Documentation] = None,
+        default_options: Optional[Options] = None,
         **kwargs,
     ):
         self._name = name
@@ -208,6 +212,7 @@ class WorkflowBase(object):
         self._on_failure = on_failure
         self._failure_node = None
         self._docs = docs
+        self._default_options = default_options
 
         if self._python_interface.docstring:
             if self.docs is None:
@@ -271,6 +276,10 @@ class WorkflowBase(object):
     @property
     def failure_node(self) -> Optional[Node]:
         return self._failure_node
+
+    @property
+    def default_options(self) -> Optional[Options]:
+        return self._default_options
 
     def __repr__(self):
         return (
@@ -661,6 +670,7 @@ class PythonFunctionWorkflow(WorkflowBase, ClassStorageTaskResolver):
         docstring: Optional[Docstring] = None,
         on_failure: Optional[Union[WorkflowBase, Task]] = None,
         docs: Optional[Documentation] = None,
+        default_options: Optional[Options] = None,
     ):
         name, _, _, _ = extract_task_module(workflow_function)
         self._workflow_function = workflow_function
@@ -677,6 +687,7 @@ class PythonFunctionWorkflow(WorkflowBase, ClassStorageTaskResolver):
             python_interface=native_interface,
             on_failure=on_failure,
             docs=docs,
+            default_options=default_options,
         )
         self.compiled = False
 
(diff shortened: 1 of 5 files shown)


flytekit/core/workflow.py
```diff
@@ -60,6 +60,9 @@
 from flytekit.models.documentation import Description, Documentation
 from flytekit.types.error import FlyteError
 
+if TYPE_CHECKING:
+    from flytekit.tools.translator import Options
```
Future-Outlier, line 64: Why this need `if TYPE_CHECKING`?

Future-Outlier (commented):
LGTM, can you also test it on serverless?
I am not familiar enough to approve this PR, but it looks good to me to solve the issue's problem.
Thank you.

flytekit/core/workflow.py
Mecoli1219: This is used to prevent circular import if two modules are mutually dependent.

Mecoli1219:
@Future-Outlier I tested it on serverless and it ran successfully, but I am not sure how to check out the labels and annotations for serverless.
image

Future-Outlier (dismissed):

flytekit/core/workflow.py
wild-endeavor: this is just a tiny dataclass that depends only on the model files... can we just move it out somewhere please?  then we don't need this type checking.

wild-endeavor (commented):
thanks!  just move that one thing plz

flytekit/core/workflow.py
Mecoli1219: @wild-endeavor I’m planning to move Options into a new file (flytekit/core/options.py). Should we provide a shortcut for users to easily import it using:
```
from flytekit import Options
```
Currently, users have to import it using:
```
from flytekit.core.options import Options
```
or even
```
from flytekit.tools.translator import Options
```
The latter feels odd since I’m relocating Options from that file, but many existing docs still use this import.

Mecoli1219 (commented):
@

wild-endeavor:
yeah, if it's something that's user facing, then it should be exposed at the top level.  at least for something as basic as this.

Custom types like flytefile are okay being in the types package and such.


Question though, what did you mean by `from flytekit.core.options import Options`... that doesn't work today.

After moving this object, let's continue to also import it into `translator.py` though, just to not break people.  In addition to main flytekit init i mean.

Thank you.

wild-endeavor:
oh you moved it into core.options.  i think that might still be circular.  I think you can move it out of core and into another folder.

Mecoli1219:
Sure I'll move it back to `flytekit.tool`

Mecoli1219:
@wild-endeavor  What do you mean by it being circular? If we move Options into the core, the dependency path from `translator.py` would be:
```
flytekit/tool/translator.py -> flytekit/core/options -> flytekit/models
```
And for the workflow, the path would be similar:
```
flytekit/core/workflow.py -> flytekit/core/option -> flytekit/models
```
Based on this, I don’t see any circular imports in these cases. However, I might be missing something—please let me know if I’ve overlooked any dependency. Also, since we currently import `Options` in `translator.py`, we can continue supporting `from flytekit.tools.translator import Options` without breaking user code. That said, I think it would be more appropriate to import it directly from the core going forward.

If you prefer, I can move `Options` back to `flytekit/tools/options.py` and import it into `flytekit/tools/translator.py`. But my main concern is that, if we want to expose `Options` at the top level, it doesn’t seem like any functions from `flytekit/tools/` are currently exported from [flytekit/init.py](https://github.com/flyteorg/flytekit/blob/master/flytekit/__init__.py), and I’m not sure why. Any insights on this would be helpful!

wild-endeavor (approved):

wild-endeavor:
sorry, thank you for the explanation and the pr!

PR #2789, opened 2024-10-07
[BUG] `BQToPandasDecodingHandler` only reads partial data when dataset > 100MB

mao3267:
## Tracking issue

Closes flyteorg/flyte#3339

While `BQToPandasDecodingHandler` reads data using the `_read_from_bq` function, its current implementation only reads from a single stream (the first stream). Since we do not set the `max_stream_count` to 1, querying larger datasets causes multiple streams to be used automatically. As a result, data from other streams will not be loaded into the dataframe.

1. In the `_read_from_bq` function, we read from all streams rather than only the first stream.

## How was this patch tested?
1. Query large datasets with local execution. 

- `bq_test.py`

```py
import pandas as pd
from flytekit import StructuredDataset, kwtypes, task, workflow
from flytekitplugins.bigquery import BigQueryConfig, BigQueryTask
from typing_extensions import Annotated

DogeCoinDataset = Annotated[StructuredDataset, kwtypes(hash=str, size=int, block_number=int)] 

bigquery_task_templatized_query = BigQueryTask(
    name="sql.bigquery.w_io",
    # Define inputs as well as their types that can be used to customize the query.
    inputs=kwtypes(version=int),
    output_structured_dataset_type=DogeCoinDataset,
    task_config=BigQueryConfig(ProjectID="flyte-437212", Location="US"),
    query_template="SELECT * FROM `bigquery-public-data.crypto_dogecoin.transactions` WHERE version = @version LIMIT 1000000",
)

@task
def convert_bq_table_to_pandas_dataframe(sd: DogeCoinDataset) -> pd.DataFrame:
    return sd.open(pd.DataFrame).all()

@workflow
def full_bigquery_wf(version: int) -> pd.DataFrame:
    sd = bigquery_task_templatized_query(version=version)
    return convert_bq_table_to_pandas_dataframe(sd=sd)
```

To run the code above, I used `pyflyte run bq_test.py full_bigquery_wf --version 2`. The limit in the SQL query is set to 1,000,000 to ensure the result is large enough. By running both the original and modified versions, we can observe the difference.

- Original Query Results
 
image

- Modified Query Results

image

#### Other Cases

- **Small size query**
```sql
SELECT * FROM `bigquery-public-data.crypto_dogecoin.transactions` WHERE version = @version LIMIT 10
```
Run with:
```py
pyflyte run --remote local_test/bqdecoder.py full_bigquery_wf --version 2 
```
image

- **Single binary BQ Agent** 
    - Small size (10 rows)

```sql
SELECT * FROM `bigquery-public-data.crypto_dogecoin.transactions` WHERE version = @version LIMIT 10
```

Run with:

```py
pyflyte run --remote local_test/bqdecoder.py full_bigquery_wf --version 2  
```

image

Read the parquet file with pandas:

image

- Large size (500,000 rows)

```sql
SELECT * FROM `bigquery-public-data.crypto_dogecoin.transactions` WHERE version = @version LIMIT 500000
```

Run with:
```py
pyflyte run --remote local_test/bqdecoder.py full_bigquery_wf --version 2 
```

image

Read the parquet file with pandas:

image

### Screenshots
image

The data is read by three streams. As a result, we received approximately one-third of the expected number of rows in the original version.

- Observations
    - 1,000,000 rows in `bigquery-public-data.crypto_dogecoin.transactions` are approximately 2,721 MB. Therefore, a stream may be able to read about 900 MB.
    - The cost of testing is significant.

```sql
WITH query_results AS (
    SELECT *
    FROM `bigquery-public-data.crypto_dogecoin.transactions`
    WHERE version = @version LIMIT 1000000
)

SELECT
    SUM(BYTE_LENGTH(TO_JSON_STRING(query_results))) / 1048576 AS size_in_mb
FROM
    query_results;
```
image

### Setup process
```
git clone https://github.com/flyteorg/flytekit.git
gh pr checkout 2789
make setup && pip install -e .
```

## Related PRs
None

## Docs link
TODO

### DIFF
--- a/flytekit/types/structured/bigquery.py
+++ b/flytekit/types/structured/bigquery.py
@@ -50,12 +50,21 @@ def _read_from_bq(
     requested_session = types.ReadSession(table=table, data_format=types.DataFormat.ARROW, read_options=read_options)
     read_session = client.create_read_session(parent=parent, read_session=requested_session)
 
-    stream = read_session.streams[0]
-    reader = client.read_rows(stream.name)
     frames = []
-    for message in reader.rows().pages:
-        frames.append(message.to_dataframe())
-    return pd.concat(frames)
+    for stream in read_session.streams:
+        reader = client.read_rows(stream.name)
+        for message in reader.rows().pages:
+            frames.append(message.to_dataframe())
+
+    if len(frames) > 0:
+        df = pd.concat(frames)
+    else:
+        import pyarrow
+
+        schema = pyarrow.ipc.read_schema(pyarrow.py_buffer(read_session.arrow_schema.serialized_schema))
+        df = schema.empty_table().to_pandas()
+
+    return df
 
 
 class PandasToBQEncodingHandlers(StructuredDatasetEncoder):
--- a/flytekit/types/structured/bigquery.py
+++ b/flytekit/types/structured/bigquery.py
@@ -59,9 +59,7 @@ def _read_from_bq(
     if len(frames) > 0:
         df = pd.concat(frames)
     else:
-        import pyarrow
-
-        schema = pyarrow.ipc.read_schema(pyarrow.py_buffer(read_session.arrow_schema.serialized_schema))
+        schema = pa.ipc.read_schema(pa.py_buffer(read_session.arrow_schema.serialized_schema))
         df = schema.empty_table().to_pandas()
 
     return df


Future-Outlier (commented):
Is there a way to mock case like 100mb?
if not, it looks pretty good.
Have you ever thought about any edge cases?

mao3267:
> Is there a way to mock case like 100mb?
if not, it looks pretty good.
Have you ever thought about any edge cases?

Will try on small and large datasets (>>100MB), and also run with single binary agent.

Future-Outlier (dismissed):
LGTM, very nice PR

flytekit/types/structured/bigquery.py
```diff
@@ -50,12 +50,21 @@ def _read_from_bq(
...
-    return pd.concat(frames)
+    for stream in read_session.streams:
+        reader = client.read_rows(stream.name)
+        for message in reader.rows().pages:
+            frames.append(message.to_dataframe())
+
+    if len(frames) > 0:
+        df = pd.concat(frames)
+    else:
+        import pyarrow
```
wild-endeavor, line 62: should be able to use `pa` right?

kumare3:
Wow this is a very good bug find. Thank you

mao3267:
@Future-Outlier In my observation, a single stream can read around 900 MB at this time. Therefore, I am wondering if adding tests with queries larger than 900 MB would incur significant costs. Btw, it seems that test on small-sized queries is already added [here](https://github.com/flyteorg/flytekit/blob/51751ebb5b2aaa62264e094b8987d86214350283/plugins/flytekit-bigquery/tests/test_bigquery.py#L58-L72).

pingsutw (commented):
thank you for fixing it

pingsutw (approved):

PR #2798, opened 2024-10-10
Allow empty lists and dicts in LaunchPlan creation

mark-thm:

When building a `flytekit.LaunchPlan` my default inputs may contain an empty list, but the code in `isinstance_generic` may only pass if at least one element of a list is present. Changes since flytekit 1.13.5 cause a regression in allowing this empty list.

Modify `isinstance_generic` to short-circuit loops/exceptions for lists, tuples, sets, and dicts if these collections are empty.

## How was this patch tested?

The changes are very small and verified by inspection. There are no existing tests for `isinstance_generic` to wire into.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -167,22 +167,22 @@ class TypeTransformer(typing.Generic[T]):
 
         # Optionally check the type of elements if it's a collection like list or dict
         if origin in {list, tuple, set}:
-            if len(obj) == 0:
-                return
-            for item in obj:
-                self.assert_type(args[0], item)
-                return
-            raise TypeTransformerFailedError(f"Not all items in '{obj}' are of type {args[0]}")
+            try:
+                for item in obj:
+                    self.assert_type(args[0], item)
+            except TypeTransformerFailedError:
+                # nicer error than a single value failing
+                raise TypeTransformerFailedError(f"Not all items in '{obj}' are of type {args[0]}")
 
         if origin is dict:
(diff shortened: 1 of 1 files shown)


flytekit/core/type_engine.py
```diff
@@ -167,12 +167,16 @@ def isinstance_generic(self, obj, generic_alias):
 
         # Optionally check the type of elements if it's a collection like list or dict
         if origin in {list, tuple, set}:
+            if len(obj) == 0:
+                return
             for item in obj:
                 self.assert_type(args[0], item)
                 return
             raise TypeTransformerFailedError(f"Not all items in '{obj}' are of type {args[0]}")
```
mark-thm, line 175: it's also not currently the case that this block asserts that _every_ `item` in `obj` is of type `args[0]`, rather just that the first item is of the correct type

pingsutw: But someone could add two items with different types to the list, right?

pingsutw (approved):

pingsutw:
Thank you for fixing it, btw.

flytekit/core/type_engine.py
mark-thm: Yes, I left the bulk of the logic unchanged. If you like I can fix this to do what it seems like it was trying to do…

mark-thm: updated to fix the logic to check every item.

mark-thm:
This seems to have been superseded by #2813

PR #2799, opened 2024-10-10
Revise Pickle Remote Task for Jupyter Notebook Environment

Mecoli1219:
## Tracking PR
https://github.com/flyteorg/flytekit/pull/2733

As discussed in the comments of flyteorg/flytekit#2733, the current design of the Jupyter Notebook is not well considered.
image

- [x] Add `basename` in `logger.debug`
- [x] Remove `__setstate__` and `__getstate__` as pickling `Promise` object is not necessary as expected
- [x] Redesign the interactive mode logic
  - [x] Pickling in the workflow level, so that we don't need extra kwargs in `Options`.
  - [x] Following with the previous point, the `interactive_mode_enabled` field in `SerailizationSettings` is not required either.
  - [x] Make SerializationSettings non-mutating by initializing the fast_registration_setting during construction in `remote.py`.
  - [x] Update `DefaultNotebookTaskResolver` to enable extracting the task entity from the pickled object. 
      - I updated the pickled object from entity to a dictionary of entity, where the key of the dictionary is the name of the task entity. 
  - [x] Automatic versioning for the interactive mode if the version is not provided in `remote.execute(...)` command.

### Setup process
Testing code is exactly the same as flyteorg/flytekit#2733

### Screenshots
| version 1 | version 2-1 | version 2-2 (run again without modification)|
| - | - | - |
| image | image | image |
| image | image | image |

Rerunning the code without modification and explicitly specify the `version` value will use the latest version in interactive mode. Note that `wf` workflow will register `hello` task, so the version of `hello` in second round is equal to the version of `wf` in the first round.

## Related PRs
https://github.com/flyteorg/flytekit/pull/2733

### DIFF
--- a/flytekit/core/python_auto_container.py
+++ b/flytekit/core/python_auto_container.py
@@ -292,16 +292,18 @@ class DefaultNotebookTaskResolver(TrackedInstance, TaskResolverMixin):
 
     @timeit("Load task")
     def load_task(self, loader_args: List[str]) -> PythonAutoContainerTask:
+        _, entity_name, *_ = loader_args
         import gzip
 
         import cloudpickle
 
         with gzip.open(PICKLE_FILE_PATH, "r") as f:
-            return cloudpickle.load(f)
+            entity_dict = cloudpickle.load(f)
+        return entity_dict[entity_name]
 
     def loader_args(self, settings: SerializationSettings, task: PythonAutoContainerTask) -> List[str]:  # type:ignore
-        _, m, t, _ = extract_task_module(task)
-        return ["task-module", m, "task-name", t]
+        n, _, _, _ = extract_task_module(task)
+        return ["entity-name", n]
 
     def get_all_tasks(self) -> List[PythonAutoContainerTask]:  # type: ignore
         raise NotImplementedError
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -1052,6 +1052,16 @@ class FlyteRemote(object):
         # and does not increase entropy of the hash while making it very inconvenient to copy-and-paste.
         return base64.urlsafe_b64encode(h.digest()).decode("ascii").rstrip("=")
 
+    def _get_image_names(self, entity: typing.Union[PythonAutoContainerTask, WorkflowBase]) -> typing.List[str]:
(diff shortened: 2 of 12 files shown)


flytekit/remote/remote.py
```diff
@@ -2526,3 +2539,49 @@ def download(
...
+                        entity.set_resolver(default_notebook_task_resolver)
+                        pickled_target_dict[entity.name] = entity
+            elif isinstance(entity, WorkflowBase):
+                for task in entity.nodes:
+                    queue.append(task)
+            elif isinstance(entity, CoreNode):
+                queue.append(entity.flyte_entity)
+        return pickled_target_dict
+
+    def _pickle_entity(self, entity: typing.Any):
```
thomasjpfan, line 2568:
```suggestion
    def _pickle_and_upload_entity(self, entity: typing.Any) -> FastSerializationSettings:
```

flytekit/core/python_auto_container.py
```diff
@@ -292,16 +292,18 @@ def name(self) -> str:
...
     @timeit("Load task")
     def load_task(self, loader_args: List[str]) -> PythonAutoContainerTask:
+        _, entity_name, *_ = loader_args
         import gzip
 
         import cloudpickle
 
         with gzip.open(PICKLE_FILE_PATH, "r") as f:
-            return cloudpickle.load(f)
+            entity_dict = cloudpickle.load(f)
```
thomasjpfan, line 301: This means each task will load a bigger pickle object than it needs to and that each task will need more memory to load everything up.

But given how normal fast-register creates a tarball with all the task I am okay with this.

flytekit/remote/remote.py
Mecoli1219: updated

Mecoli1219:
@wild-endeavor I updated the automatic versioning followed with the same logistic as [FlyteRemote.register_script](https://github.com/flyteorg/flytekit/blob/51751ebb5b2aaa62264e094b8987d86214350283/flytekit/remote/remote.py#L1124C9-L1145C14). If we are in interactive mode, we will automatically compute a hashed `version` for it if version is not explicitly specified. Otherwise, we will follow the original behavior in `FlyteRemote`.

wild-endeavor (approved):
thanks!!

PR #2800, opened 2024-10-10
Pull secrets from environment when running locally

thomasjpfan:

When running locally, it is not natural to have a `_FSEC_` prefix on your secrets.

With this PR, the `_FSPEC_` prefix is not used anymore for local execution.

## How was this patch tested?

I added unit tests to this PR to check the new behavior and ran this test:

```python
from flytekit import task, Secret
from flytekit import current_context


@task(
    secret_requests=[Secret(key="apikey", group="mygroup")],
    container_image="localhost:30000/flytekit:dev",
)
def check_secret() -> str:
    ctx = current_context()
    secret = ctx.secrets.get(key="apikey", group="mygroup")
    return secret
```

```bash
export MYGROUP_APIKEY=my-secret
pyflyte run --remote main.py check_secret
```

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -346,14 +346,10 @@ class SecretsManager(object):
     def __init__(self, secrets_cfg: typing.Optional[SecretsConfig] = None):
         if secrets_cfg is None:
             secrets_cfg = SecretsConfig.auto()
-            is_local_execution = os.getenv("FLYTE_INTERNAL_EXECUTION_ID") is None
-            if is_local_execution:
-                self._env_prefix = ""
-            else:
-                self._env_prefix = secrets_cfg.env_prefix.strip()
 
         self._base_dir = secrets_cfg.default_dir.strip()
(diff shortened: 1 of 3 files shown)


flytekit/core/context_manager.py
```diff
@@ -373,11 +374,22 @@ def get(
         if not get_plugin().secret_requires_group():
             group, group_version = None, None
 
-        env_var = self.get_secrets_env_var(group, key, group_version)
+        env_prefixes = [self._env_prefix]
+
+        # During local execution check for the key without a prefix
+        is_local_execution = os.getenv("FLYTE_INTERNAL_EXECUTION_ID") is None
```
pmahindrakar-oss, line 380: could we convert this to utility function as i think it could be useful in other places to know if you are in local mode or not . But don't we have --remote flag that we can use to determine the same

pmahindrakar-oss (approved):

pingsutw (approved):

PR #2804, opened 2024-10-10
add links to register

dansola:

When entities are registered there is no helpful link to the UI like there is when entities are run. This helps users navigate to the entity in the UI after they register something in the CLI.

When registering we fetch the entity with flyte remote and then get the console url for the entity.

## How was this patch tested?

`union register docs_examples/flyte_file.py`

Before:
```
Successfully serialized 4 flyte objects
[✔] Registration flyte_file.t1 type TASK successful with version GxQbXiMnIdiRYRxg6I_B2Q
[✔] Registration flyte_file.t2 type TASK successful with version GxQbXiMnIdiRYRxg6I_B2Q
[✔] Registration flyte_file.wf type WORKFLOW successful with version GxQbXiMnIdiRYRxg6I_B2Q
[✔] Registration flyte_file.wf type LAUNCH_PLAN successful with version GxQbXiMnIdiRYRxg6I_B2Q
Successfully registered 4 entities
```

After:
```
Serializing and registering 4 flyte entities
[✔] Task: flyte_file.t1
[✔] Task: flyte_file.t2
[✔] Workflow: flyte_file.wf
[✔] Launch Plan: flyte_file.wf
Successfully registered 4 entities
```
Note that entity names are clickable and underlined (see image). 
image

### DIFF
--- a/flytekit/tools/repo.py
+++ b/flytekit/tools/repo.py
@@ -13,6 +13,8 @@ from flytekit.constants import CopyFileDetection
 from flytekit.core.context_manager import FlyteContextManager
 from flytekit.loggers import logger
 from flytekit.models import launch_plan, task
+from flytekit.models import launch_plan as launch_plan_models
+from flytekit.models.admin import workflow as admin_workflow_models
 from flytekit.models.core.identifier import Identifier
 from flytekit.remote import FlyteRemote
 from flytekit.remote.remote import RegistrationSkipped, _get_git_repo_url
@@ -199,7 +201,7 @@ def list_packages_and_modules(
     return pkgs_and_modules
 
 
-def secho(i: Identifier, state: str = "success", reason: str = None, op: str = "Registration"):
+def secho(i: Identifier, state: str = "success", reason: str = None, op: str = "Registration", console_url: str = None):
     state_ind = "[ ]"
     fg = "white"
     nl = False
@@ -207,7 +209,7 @@ def secho(i: Identifier, state: str = "success", reason: str = None, op: str = "
         state_ind = "\r[✔]"
         fg = "green"
         nl = True
-        reason = f"successful with version {i.version}" if not reason else reason
+        reason = f"successful: {console_url}" if not reason else reason
     elif state == "failed":
         state_ind = "\r[x]"
         fg = "red"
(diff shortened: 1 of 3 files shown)


flytekit/tools/repo.py
```diff
@@ -59,12 +63,22 @@ def serialize_get_control_plane_entities(
...
     :param pkgs: Dot-delimited Python packages/subpackages to look into for serialization.
     :param local_source_root: Where to start looking for the code.
+    :param is_registration: Whether this is happening for registration.
     """
     settings.source_root = local_source_root
     ctx_builder = FlyteContextManager.current_context().with_serialization_settings(settings)
     with FlyteContextManager.with_context(ctx_builder) as ctx:
         registrable_entities = get_registrable_entities(ctx, options=options)
-        click.secho(f"Successfully serialized {len(registrable_entities)} flyte objects", fg="green")
+        if is_registration:
```
thomasjpfan, line 72: Since `is_registration` is only for printing, can this printing logic be moved up to the caller?

flytekit/tools/repo.py
```diff
@@ -21,6 +22,8 @@
 from flytekit.tools.serialize_helpers import get_registrable_entities, persist_registrable_entities
 from flytekit.tools.translator import FlyteControlPlaneEntity, Options
 
+REGISTRATION = "Registration"
```
thomasjpfan, line 25: Is this being used anywhere?

thomasjpfan (commented):
Now that we have links, do you think printing the version itself is useful?

flytekit/tools/repo.py
dansola: ah, I'll remove that

dansola:
> Now that we have links, do you think printing the version itself is useful?

I don't think the versions are necessary if we have links. With this change we don't show them.

flytekit/tools/repo.py
dansola: sure! good call

wild-endeavor (approved):

wild-endeavor:
> > Now that we have links, do you think printing the version itself is useful?
> 
> I don't think the versions are necessary if we have links. With this change we don't show them.

Let's see how much we miss them.  Might be nice to see it once.

PR #2806, opened 2024-10-11
ensure a space is added if both args are set in ImageSpec

blaketastic2:
If both `--index-url` and `--extra-index-url` are set, a space is missing between the 2, resulting in an error.

## How was this patch tested?

Added unit test and visualized the docker command output.

### DIFF
--- a/flytekit/image_spec/default_builder.py
+++ b/flytekit/image_spec/default_builder.py
@@ -145,15 +145,20 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
     requirements_uv_path = tmp_dir / "requirements_uv.txt"
     requirements_uv_path.write_text("\n".join(requirements))
 
-    pip_extra_args = ""
+    pip_extra_args = []
 
     if image_spec.pip_index:
-        pip_extra_args += f"--index-url {image_spec.pip_index}"
+        pip_extra_args.append(f"--index-url {image_spec.pip_index}")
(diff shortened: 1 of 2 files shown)


flytekit/image_spec/default_builder.py
```diff
@@ -145,16 +145,21 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
...
         extra_urls = [f"--extra-index-url {url}" for url in image_spec.pip_extra_index_url]
-        pip_extra_args += " ".join(extra_urls)
+        pip_extra_args.extend(extra_urls)
+
+    pip_extra_args = " ".join(pip_extra_args)
 
     uv_python_install_command = UV_PYTHON_INSTALL_COMMAND_TEMPLATE.substitute(PIP_EXTRA=pip_extra_args)
 
+    # Normalize whitespace to remove any extra spaces
+    uv_python_install_command = " ".join(uv_python_install_command.split())
```
thomasjpfan, line 161: I prefer not to do any formatting if we do not need to. Was the reason to normalize the whitespace for better printing?

thomasjpfan (commented):
Thank you for the PR!

flytekit/image_spec/default_builder.py
blaketastic2: Yea, it prints an extra space, but should not have any impact on the actual command. I'll remove.

thomasjpfan (approved):

PR #2808, opened 2024-10-11
Fix mypy errors caught in 1.11.2

eapolinario:
mypy 1.11.2 catches new errors in flytekit:
```
❯ make lint
mypy flytekit/core
flytekit/core/type_engine.py:439: note: By default the bodies of untyped functions are not checked, consider using --check-untyped-defs  [annotation-unchecked]
flytekit/core/type_engine.py:942: error: Argument 1 to "append" of "list" has incompatible type "tuple[Any, type[Any]]"; expected "tuple[Any,
GenericAlias]"  [arg-type]
flytekit/core/type_engine.py:951: error: Argument 1 to "append" of "list" has incompatible type "tuple[Any, type[Any]]"; expected "tuple[Any, GenericAlias]"  [arg-type]
flytekit/core/type_engine.py:2166: error: Argument 1 to "append" of "list" has incompatible type "tuple[Any, type[Any]]"; expected "tuple[Any, GenericAlias]"  [arg-type]
flytekit/core/python_auto_container.py:211: error: Need type annotation for "env" (hint: "env: dict[<type>, <type>] = ...")  [var-annotated]
flytekit/core/array_node_map_task.py:83: error: Need type annotation for "transformer"  [var-annotated]
Found 5 errors in 3 files (checked 48 source files)
make: *** [lint] Error 1
```

This PR fixes the new errors and opportunistically updates the version of ruff in precommit.

Notice that these types added to handle dataclasses only paper over the problem. We have some code that doesn't make any sense in the python type system, for example https://github.com/flyteorg/flytekit/blob/d22e8a82321f47e6d1b73814d4ca852205abd767/flytekit/core/type_engine.py#L2197. Ideally we should rip this all of this dataclass code out of flytekit.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -16,6 +16,7 @@ import typing
 from abc import ABC, abstractmethod
 from collections import OrderedDict
 from functools import lru_cache
+from types import GenericAlias
 from typing import Any, Dict, List, NamedTuple, Optional, Type, cast
 
 import msgpack
@@ -434,7 +435,7 @@ class DataclassTransformer(TypeTransformer[object]):
 
     """
 
-    def __init__(self):
+    def __init__(self) -> None:
         super().__init__("Object-Dataclass-Transformer", object)
         self._decoder: Dict[Type, JSONDecoder] = dict()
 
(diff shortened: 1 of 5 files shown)


flytekit/core/type_engine.py
```diff
@@ -2176,9 +2192,7 @@ def generate_attribute_list_from_dataclass_json(schema: dict, schema_name: typin
     return attribute_list
 
 
-def convert_marshmallow_json_schema_to_python_class(
-    schema: dict, schema_name: typing.Any
-) -> Type[dataclasses.dataclass()]:  # type: ignore
+def convert_marshmallow_json_schema_to_python_class(schema: dict, schema_name: typing.Any) -> type:
```
eapolinario, line 2195: https://github.com/python/cpython/issues/102699 is a discussion about having a baseclass for dataclasses that could be used here instead of `Type[dataclasses.dataclass()]` which is invalid.

wild-endeavor (approved):

PR #2812, opened 2024-10-13
added node index to current_context

Atharva1723:
## Tracking issue
Closes flyteorg/flyte#5534
In Flyte, when working with map tasks, outputs often risk being overwritten if they share the same static name. For example, if multiple nodes generate output files with the same name, only the last file would remain, as previous ones would be overwritten. By adding the array node index, it becomes possible to uniquely name these outputs, avoiding conflicts.

**The key goal of this PR is to include the index of an array node in map_tasks within the current_context() returned value. This feature is especially useful when multiple nodes within a map task could potentially overwrite outputs**

### DIFF
--- a/flytekit/bin/entrypoint.py
+++ b/flytekit/bin/entrypoint.py
@@ -300,7 +300,9 @@ def setup_execution(
     if checkpoint_path is not None:
         checkpointer = SyncCheckpoint(checkpoint_dest=checkpoint_path, checkpoint_src=prev_checkpoint)
         logger.debug(f"Checkpointer created with source {prev_checkpoint} and dest {checkpoint_path}")
-
+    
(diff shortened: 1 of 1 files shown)


pingsutw (commented):
mind adding a small unit test in this file?
https://github.com/flyteorg/flytekit/blob/f394bc95b94798e856649a2b07e2d87528ebd2cb/tests/flytekit/unit/bin/test_python_entrypoint.py#L342-L341

pingsutw (commented):
Could you run `make fmt` locally to fix the lint error?

Atharva1723:
Will write an unit test and will run formatter too

Atharva1723:
i have run make fmt

PR #2815, opened 2024-10-15
Improved Type engine for generic types and performance

kumare3:
Improves the core type engine system for
1. Faster performance when scanning class tree
2. Better support for UnionType
3. Better support for JSONIterator
4. Cleaner, simpler and more maintainable logic
5. Matrix test for types
6. Better pep604 support

### DIFF
--- a/tests/flytekit/unit/interaction/test_click_types.py
+++ b/tests/flytekit/unit/interaction/test_click_types.py
@@ -518,7 +518,7 @@ def test_pickle_type():
         t.convert("typing:not_exists", None, None)
 
     # test that it can load a variable from a module
-    with tempfile.NamedTemporaryFile("w", dir=".", suffix=".py", delete=True) as f:
+    with tempfile.NamedTemporaryFile("w", dir=".", suffix=".py", delete=False) as f:
         f.write("a = 1")
         f.flush()
         # find the base name of the file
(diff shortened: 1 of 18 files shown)


tests/flytekit/unit/interaction/test_click_types.py
```diff
@@ -498,3 +499,30 @@ class Datum:
...
+    with pytest.raises(click.BadParameter):
+        t.convert("typing:not_exists", None, None)
+
+    # test that it can load a variable from a module
+    with tempfile.NamedTemporaryFile("w", dir=".", suffix=".py", delete=True) as f:
+        f.write("a = 1")
+        f.flush()
+        # find the base name of the file
+        basename = os.path.basename(f.name).split(".")[0]
+        assert t.convert(f"{basename}:a", None, None) == 1
```
eapolinario, line 526: Windows doesn't let the same file be opened twice, which causes this [permission denied issue](https://github.com/flyteorg/flytekit/actions/runs/11372614166/job/31637345293?pr=2815#step:7:2869). The easiest thing to do is to sequence the calls to open+close+unlink. Something like:
```suggestion
    tmp_f = tempfile.NamedTemporaryFile(dir=".", suffix=".py", delete=False)
    f = open(tmp_f.name, "w")
    f.write("a = 1")
    f.close()
    # find the base name of the file
    basename = os.path.basename(tmp_f.name).split(".")[0]
    assert t.convert(f"{basename}:a", None, None) == 1
    # Not sure if need to delete this file, it's a temp file anyway, but leaving here for completeness
    os.unlink(tmp_f.name)
```

wild-endeavor (approved):

Future-Outlier (approved):

pingsutw (approved):

PR #2821, opened 2024-10-16
Flytekitplugin pandera update: use entrypoint and structured dataset

cosmicBboy:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5851

The pandera flytekit plugin is out-dated and needs to be maintained.

- Adds an `entrypoint` config in `setup.py` for auto-registration of the type transformer
- Rename `schema.py -> pandas_transformer.py`
- Create a basics deck renderer `PandasReportRenderer`
- Add configuration `ValidationConfig` with `on_error` option. This is so that `on_error="warn"` can create a Flyte deck that's actually rendered in the UI. Ideally `on_error="raise"` will also render a deck, but right now there's a limitation where decks aren't rendered with failed tasks.

## How was this patch tested?

Update the unit tests, run with `pytest test` in the plugin directory.


plugins/flytekit-pandera/flytekitplugins/pandera/pandas_transformer.py
```diff
@@ -0,0 +1,144 @@
...
+            for arg in args:
+                if isinstance(arg, ValidationConfig):
+                    config = arg
+                    break
+
+        try:
+            type_args = typing.get_args(t)
+        except AttributeError:
+            # for python < 3.8
+            type_args = getattr(t, "__args__", None)
```
eapolinario, line 49: Do we need this? We [dropped](https://github.com/flyteorg/flytekit/pull/1521) support for python 3.7 around the 1.4.0 release.

plugins/flytekit-pandera/flytekitplugins/pandera/pandas_transformer.py
```diff
@@ -0,0 +1,144 @@
...
+
+        renderer = PandasReportRenderer(title=f"Pandera Report: {schema.name}")
+        try:
+            val = schema.validate(df, lazy=True)
+        except (pandera.errors.SchemaError, pandera.errors.SchemaErrors) as exc:
+            if config.on_error == "raise":
+                raise exc
+            elif config.on_error == "warn":
+                print(ctx.execution_state)
+                warnings.warn(str(exc), RuntimeWarning)
```
eapolinario, line 134: Should this happen in a call to flytekit's `logger.warning` ?

plugins/flytekit-pandera/flytekitplugins/pandera/pandas_transformer.py
```diff
@@ -0,0 +1,144 @@
...
+        renderer = PandasReportRenderer(title=f"Pandera Report: {schema.name}")
+        try:
+            val = schema.validate(python_val, lazy=True)
+        except (pandera.errors.SchemaError, pandera.errors.SchemaErrors) as exc:
+            if config.on_error == "raise":
+                raise exc
+            elif config.on_error == "warn":
+                warnings.warn(str(exc), RuntimeWarning)
+                html = renderer.to_html(exc)
+                Deck(renderer._title, html)
```
eapolinario, line 100: > Ideally on_error="raise" will also render a deck, but right now there's a limitation where decks aren't rendered with failed tasks.

we're fixing this in Streaming decks, so how about we future-proof this by adding the Deck in both cases?

pingsutw (approved):

PR #2829, opened 2024-10-17
Async/data persistence

wild-endeavor:
A previous [PR](https://github.com/flyteorg/flytekit/pull/2752) added async versions of the type engine.  This change leverages that by converting flytekit's main I/O layer to async as well.

* Added a new function to get the async version of an fsspec filesystem. This takes care of calling fsspec with the current loop.
* Fix to `get_filesystem` to take into account passed in kwargs even if no special if condition handling is present (this affected using https filesystem).
* Added async versions of the popular functions.
* Updated some type transformers to async.

## How was this patch tested?
Tested locally and then tested on internal clusters.  Performance benefit is mainly noticeable across large number of small files.  A task that produces a list of 1000 files, 2MB each, decreases from around 135s to about 40s on a t3a.xlarge.  No improvement is seen for folders because those were already async.


flytekit/types/pickle/pickle.py
```diff
@@ -52,31 +52,31 @@ def python_type(cls) -> typing.Type:
...
         uri = os.path.join(local_dir, local_path)
         with open(uri, "w+b") as outfile:
             cloudpickle.dump(python_val, outfile)
 
-        return ctx.file_access.put_raw_data(uri)
+        return await ctx.file_access.async_put_raw_data(uri)
 
     @classmethod
-    def from_pickle(cls, uri: str) -> typing.Any:
+    async def from_pickle(cls, uri: str) -> typing.Any:
```
wild-endeavor, line 66: @pingsutw are these public facing functions?

pingsutw: not public facing

flytekit/core/data_persistence.py
```diff
@@ -208,8 +211,15 @@ def get_filesystem(
         storage_options = get_fsspec_storage_options(
             protocol=protocol, anonymous=anonymous, data_config=self._data_config, **kwargs
         )
+        kwargs.update(storage_options)
 
-        return fsspec.filesystem(protocol, **storage_options)
+        return fsspec.filesystem(protocol, **kwargs)
+
+    def get_async_filesystem_for_path(self, path: str = "", anonymous: bool = False, **kwargs) -> AsyncFileSystem:
```
kumare3, line 218:
```suggestion
    async def get_async_filesystem_for_path(self, path: str = "", anonymous: bool = False, **kwargs) -> AsyncFileSystem:
```

thomasjpfan:
Can you share the code that you used to benchmark the performance?

wild-endeavor:
> Can you share the code that you used for benchmarking?

https://github.com/unionai/debugyt/blob/master/user/ytong/src/yt_dbg/data_perf/measure_wfs.py
@thomasjpfan

eapolinario:
kf-pytorch tests are hanging (notice how they ran for 5h+): https://github.com/flyteorg/flytekit/actions/runs/11451023024/job/31860491868?pr=2829#step:7:1

I kicked off a [re-run](https://github.com/flyteorg/flytekit/actions/runs/11451023024/job/31896882107?pr=2829) but it seems like it's also hanging.

eapolinario (approved):
This is amazing.

PR #2671, opened 2024-08-09
Deprecate scopes

pingsutw:
## Tracking issue
NA

## Why are the changes needed?
Error messages are not clear in both remote and local execution.

## What changes were proposed in this pull request?
- Add `FlyteUserRuntimeException`. it is thrown when a user code raises an exception.
- Non-FlyteUserRuntimeException errros are system errors
- remove `system_entry_point` and `user_entry_point` in the code
- deprecate FlyteScopedException

## How was this patch tested?
unit test / remote

### Setup process

```python
from flytekit import task, workflow, ImageSpec

new_flytekit = "git+https://github.com/flyteorg/flytekit@14ab38ca17d3b9a5bb70e5ae9a61955b01f573e4"

sd_finetuning_image = ImageSpec(
    registry="pingsutw",
    apt_packages=["git"],
    packages=[
        new_flytekit,
    ],
    builder="envd",
)


@task(enable_deck=True, container_image=sd_finetuning_image)
def t1(a: int) -> int:
    raise ValueError("error")
    return a + 1


@workflow
def wf() -> int:
    return t1(a=3)
```

### Screenshots
Before:
- User error
Screenshot 2024-08-10 at 12 35 01 AM

- System error
Screenshot 2024-08-10 at 2 50 54 AM

- Local error
Screenshot 2024-08-10 at 2 33 12 AM

- User error
Screenshot 2024-08-10 at 2 14 31 AM

- System error
Screenshot 2024-08-10 at 2 47 06 AM

- Local error
Screenshot 2024-08-10 at 2 33 01 AM

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off.

## Related PRs
NA

## Docs link
NA

PR #2831, opened 2024-10-18
handle case where error may not have args

blaketastic2:
Running a task locally, if an exception is thrown that does not have args, an `IndexError: tuple index out of range` is actually thrown.

A simple example that causes the issue:
```python
@task
def task_1(val: int) -> None:
    assert val > 0
```
If this assert fails and an `AssertionError` is thrown, it will not have any `args` and therefore cause the above behaviour.

## How was this patch tested?

Parameterized an existing test w/ more cases.

## Related PRs

The issue was first introduced here - https://github.com/flyteorg/flytekit/pull/2671

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -743,9 +743,7 @@ class PythonTask(TrackedInstance, Task, Generic[T]):
                 raise
             except Exception as e:
                 if is_local_execution:
-                    e.args = (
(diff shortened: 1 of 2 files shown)


thomasjpfan (approved):

pingsutw (approved):

pingsutw:
thank you @blaketastic2

PR #2833, opened 2024-10-18
create flytekit.cwd, add support for pathlib.Path paths in FlyteFile …

granthamtaylor:
Users may access the current working directory within a task like so:

```python
flytekit.cwd()
```

This will return a Path object for convenience, such that one may create a FlyteFile / FlyteDirectory with ease:

```python
FlyteFile(flytekit.cwd() / "my_file.txt")
```

This is significantly less cumbersome than the current approach:
```python

FlyteFile(f"{flytekit.current_context().working_directory}/my_file.txt")

Additionally, this commit adds support for using a `pathlib.Path` object to create a FlyteFile / FlyteDirectory, which, while was documented to work, resulted in unexpected behavior.

Tests have been added for this new functionality.

### DIFF
--- a/flytekit/__init__.py
+++ b/flytekit/__init__.py
@@ -205,6 +205,7 @@ Documentation
 
 import os
 import sys
+from pathlib import Path
 from typing import Generator
 
 from rich import traceback
@@ -279,6 +280,15 @@ def current_context() -> ExecutionParameters:
 def new_context() -> Generator[FlyteContext, None, None]:
     return FlyteContextManager.with_context(FlyteContextManager.current_context().new_builder())
 
+def cwd() -> Path:
+    """
+    Use this method to get a pathlib.Path of the current working directory of the task execution.
+
+    Usage
+    .. code-block:: python
+        path = flytekit.cwd() / 'my_file.txt'
+    """
+    return Path(FlyteContextManager.current_context().working_directory)
 
 def load_implicit_plugins():
     """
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: flytekit/types/directory/types.py [hunk 1]
```diff
@@ -168,7 +168,7 @@ def __init__(
         """
         # Make this field public, so that the dataclass transformer can set a value for it
         # https://github.com/flyteorg/flytekit/blob/bcc8541bd6227b532f8462563fe8aac902242b21/flytekit/core/type_engine.py#L298
-        self.path = path
+        self.path = path if not isinstance(pathlib.Path, str) else str(path)
```
thomasjpfan, line 171: Is `not isinstance(pathlib.Path, str)` intended? It is always True. The type `pathlib.Path` is never a `str`.

Also, doing any conversion here is backward incompatible. If a user passed in a `Path` object before, then doing `str(path)` will change the type.

Do we need `str(path)` for this PR?

flytekit/__init__.py
```diff
@@ -280,6 +281,17 @@ def new_context() -> Generator[FlyteContext, None, None]:
     return FlyteContextManager.with_context(FlyteContextManager.current_context().new_builder())
 
 
+def cwd() -> Path:
```
thomasjpfan, line 284: Do you thinking this will be confusing for users? The "current working directory" is usually `/root`, but this "working directory" is a random temporary directory.

PR #2834, opened 2024-10-18
Catch mistake in structured dataset

wild-endeavor:
Adds an error message to try to catch a common mistake users make, when they request a transformer for the `StructuredDataset` type, but just pass in the raw dataframe.


eapolinario (approved):
That's great.

PR #2845, opened 2024-10-21
Union/enum handling

wild-endeavor:
The union transformer should not be calling out to the super's invocation of `assert_type`, it should be calling the assert on the underlying subtype's transformer.

Fetch each subtype's transformer in the Union transformer and call assert type.

* Also fixed an issue in the dataclass transformer where a KeyError was being raised instead of a type transformer exception.

## How was this patch tested?
Local unit testing, ran workflow on local cluster, tested with flyte remote from jupyter.

### DIFF
--- a/tests/flytekit/unit/core/test_unions.py
+++ b/tests/flytekit/unit/core/test_unions.py
@@ -1,5 +1,6 @@
 import typing
 from dataclasses import dataclass
+from enum import Enum
 
 import pytest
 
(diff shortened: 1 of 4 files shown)


tests/flytekit/unit/core/test_unions.py
```diff
@@ -0,0 +1,93 @@
...
+
+    lt = TypeEngine.to_literal_type(Color)
+    guessed = TypeEngine.guess_python_type(lt)
+    tf = TypeEngine.get_transformer(guessed)
+    tf.assert_type(guessed, guessed("two"))
+    tf.assert_type(Color, Color.GREEN)
+
+
+@pytest.mark.sandbox_test
+def test_with_remote():
```
eapolinario, line 79: Do we have to make this test run in CI? I don't think we're running any of the tests marked as `sandbox_test` currently.

wild-endeavor: no this is just to have for local testing

eapolinario (approved):

PR #2848, opened 2024-10-21
Restrict Python Version Mismatch between Pickled Object and Remote Envrionment

Mecoli1219:
## Tracking Issues
https://github.com/flyteorg/flyte/issues/5907

In the Jupyter Notebook environment, we are using `cloudpickle` to package the entity at the binary level. When we load the pickled object with a different Python version, it will throw an unexpected error. Therefore, we should restrict loading pickled objects with the mismatched Python version.

Let's assume the Python version for the pickled object is X. 
1. If we load it in a Python version smaller than X, it is possible to crash. Therefore, we need to handle loading exceptions.
2. If we successfully load the object, we need to check whether the Python version is matched. I added a new metadata in the pickled object to store the Python version when pickling, and this will be used for validation after loading.
3. Enable logging error messages when trying to extract the output from remote execution so that the user doesn't need to go to Console to get the error message.

## How was this patch tested?
I set the Python version of Jupyter Notebook to 3.11 and built 3 images with different Python versions for remote execution, including 3.9, 3.11, and 3.12.

### Screenshots
- Remote Python version = 3.12
image
- Remote Python version = 3.11
image
- Remote Python version = 3.9
image

### DIFF
--- a/flytekit/core/python_auto_container.py
+++ b/flytekit/core/python_auto_container.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import importlib
 import re
 from abc import ABC
+from dataclasses import dataclass
 from typing import Callable, Dict, List, Optional, TypeVar, Union
 
 from flyteidl.core import tasks_pb2
@@ -282,6 +283,32 @@ class DefaultTaskResolver(TrackedInstance, TaskResolverMixin):
 default_task_resolver = DefaultTaskResolver()
 
 
+@dataclass
+class PickledEntityMetadata:
+    """
+    Metadata for a pickled entity containing version information.
+
+    Attributes:
+        python_version: The Python version string (e.g. "3.12.0") used to create the pickle
+    """
+
+    python_version: str
+
+
+@dataclass
+class PickledEntity:
+    """
+    Represents the structure of the pickled object stored in the .pkl file for interactive mode.
+
+    Attributes:
+        metadata: Metadata about the pickled entities including Python version
+        entities: Dictionary mapping entity names to their PythonAutoContainerTask instances
+    """
+
+    metadata: PickledEntityMetadata
+    entities: Dict[str, PythonAutoContainerTask]
+
+
 class DefaultNotebookTaskResolver(TrackedInstance, TaskResolverMixin):
     """
     This resolved is used when the task is defined in a notebook. It is used to load the task from the notebook.
@@ -300,7 +327,7 @@ class DefaultNotebookTaskResolver(TrackedInstance, TaskResolverMixin):
 
         try:
             with gzip.open(PICKLE_FILE_PATH, "r") as f:
-                entity_dict = cloudpickle.load(f)
+                loaded_data = cloudpickle.load(f)
         except TypeError:
             raise RuntimeError(
                 "The Python version is smaller than the version used to create the pickle file. "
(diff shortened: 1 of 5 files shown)


Mecoli1219:
cc @thomasjpfan

flytekit/remote/remote.py
```diff
@@ -208,8 +208,14 @@ def _get_pickled_target_dict(
     :param root_entity: The entity to get the pickled target for.
         :return: hashed bytes and the pickled target dictionary.
     """
+    import sys
+
     queue: typing.List[typing.Union[WorkflowBase, PythonTask, CoreNode]] = [root_entity]
-    pickled_target_dict = {}
+    pickled_target_dict = {
+        "metadata": {
```
wild-endeavor, line 215: but this shares a namespace with task names right?  so if someone makes a task where `entity.name` is "metadata" then this metadata will be lost and when we go to look for it we'll get a KeyError or something.

I don't think it needs to be here though, this is a pretty standard thing for all tasks.  There's a couple places I think we can fit this information.
* RuntimeMetadata's `flavor` [field](https://github.com/flyteorg/flyte/blob/13b3d82c94fb61047a49aa2093a2b17fb6a7c792/flyteidl/protos/flyteidl/core/tasks.proto#L86).  I don't think this is being used right now.  The name "runtime" is not correct.  It's more like "compile-time", but that's a different problem.
* Something in these [tags](https://github.com/flyteorg/flyte/blob/13b3d82c94fb61047a49aa2093a2b17fb6a7c792/flyteidl/protos/flyteidl/core/tasks.proto#L128).  To my knowledge these tags aren't really being used right now.

I think i have a slight preference for the first one.  what do you think @eapolinario?

Mecoli1219: I am not sure the first approach will be enough though. It is possible that we have to extend it to more fields (like maybe the package version?) in the future, and the first approach will have only 1 string that can store this value.

I think that this metadata is primarily used for interactive mode since only pickling objects will potentially cause the version mismatch problem. When creating a pickled object, the versions should follow with this object. I am thinking of moving the entities dictionary one layer down, so that user could have define function with name `metadata`
```
{
  "metadata": {
    "python_version": "3.12",
    ...
  },
  "entity_dict": {
    "task1": ...
    ...
  }
}
```

wild-endeavor: your call.  I like the idea of adding the `entity_dict` key in general.  I think we should do that anyways.

Your call on the metadata.  The issue with just adding fields is that this risks becoming an undocumented API, so I was thinking we'd just use the existing flavor until we needed more than one field.  If you go the pickled dict route though, just make sure it's well documented/commented.

Mecoli1219: I see your concern. How about creating a dataclass to define the pickled object? In that way, people could know what is inside this object.

Regarding the metadata, the pickling only affects the Python side and doesn't need to be stored in the backend. The main issue this PR addresses is the Python version mismatch between pickled files and remote execution. By sending the Python version inside the pickled file, we can ensure the correctness of the version. Also as far as I researched, I think that `runtimeMetadata` is currently used for flytekit's language and version information rather than the Python version.

wild-endeavor: sounds good!  just make sure using dataclass doesn't break the pickling.

Mecoli1219:
@wild-endeavor I push the `dataclass` version and test it with the task, workflow, map_task examples. My Python version for Jupyter Notebook is `3.12.0`.
- Cluster: `3.12.7`
  image
- Cluster: `3.11.0`
  image

flytekit/core/python_auto_container.py
```diff
@@ -294,12 +321,41 @@ def name(self) -> str:
...
 
-        with gzip.open(PICKLE_FILE_PATH, "r") as f:
-            entity_dict = cloudpickle.load(f)
-        return entity_dict[entity_name]
+        try:
+            with gzip.open(PICKLE_FILE_PATH, "r") as f:
+                loaded_data = cloudpickle.load(f)
+        except TypeError:
+            raise RuntimeError(
+                "The Python version is smaller than the version used to create the pickle file. "
```
eapolinario, line 333: nit: is the backwards compatibility a guarantee that cloudpickle gives? (As in, cloudpickle version X will always deserialize cloudpickle Y if X > Y). If that's not stated anywhere I'd just say that versions are different.

Mecoli1219: No, I don't think `cloudpickle` has explicitly promised backward compatibility (but `pickle` does). I'll update that.

eapolinario (approved):

PR #2852, opened 2024-10-23
add class methods, unit tests for flytefile and flytedirectory

granthamtaylor:
I feel like it is verbose to get the working directory (`flytekit.current_context().working_directory`) in a task. I find myself doing this very, very frequently, primarily to write files. I use the working directory so that files won’t get written to my local working directory during local execution.

I had thought about this for a while, and the more I thought about what exactly I wanted to accomplish, I feel like it should be accomplished with a few new classmethod for `FlyteFile` / `FlyteDirectory`...

-----

This task...

```python
@task
def my_task() -> FlyteFile:
    local_path = os.path.join(current_context().working_directory, "data.txt")
    with open(local_path, mode="w") as file:
        file.write("Here is some sample data.")
    return FlyteFile(path=local_path)
```

Should ideally be replaced with something like this:

```python
@task
def my_task() -> FlyteFile:

    ff = FlyteFile.new('myfile.txt')

    with open(ff, mode="w") as file:
        file.write("Here is some sample data.")

    return ff
```

In this case, `FlyteFile.new` will create a FlyteFile with a path of the filename inside of the working directory (`current_context().working_directory`).

-----

The same goes for FlyteDirectory...

```python
@task
def my_task() -> FlyteDirectory:

    p = os.path.join(current_context().working_directory, "my_new_directory")

    os.makedirs(p)

    with open(os.path.join(p, "file_1.txt"), 'w') as file:
        file.write("This is file 1.")

    return FlyteDirectory(p)
```

Should be replaced with:

```python
@task
def my_task() -> FlyteDirectory:

    dir = FlyteDirectory.new('my_new_directory')

    with open(dir / "file_1.txt", 'w') as file:
        file.write("This is file 1.")

    return dir
```

In this case, `FlyteDirectory.new` will create the directory `my_new_directory` under `current_context().working_directory`, and ensure that it exists (`os.makedirs(dir.path)`)

Additionally, I am overloading the `__truedivide__` operator to `FlyteDirectory` for convenience:

```python

class FlyteDirectory:

    ...

    def __truedivide__(self, other: str | os.PathLike) -> Path:
        """
        This is a convenience method to allow for easy concatenation of paths.
        """
        return Path(self.path) / other
```

### DIFF
--- a/flytekit/types/directory/types.py
+++ b/flytekit/types/directory/types.py
@@ -206,6 +206,23 @@ class FlyteDirectory(SerializableType, DataClassJsonMixin, os.PathLike, typing.G
         remote_path = ctx.file_access.generate_new_custom_path(alt=alt, stem=stem)
         return cls(path=remote_path)
 
+    @classmethod
+    def new(cls, dirname: str | os.PathLike) -> FlyteFile:
+        """
+        Create a new FlyteDirectory object in current Flyte working directory.
+        """
+
+        if os.path.isabs(dirname):
+            raise ValueError("Path should be relative.")
+
+        ctx = FlyteContextManager.current_context()
+
+        path = os.path.join(ctx.user_space_params.working_directory, dirname)
+
+        os.makedirs(path, exist_ok=False)
+
+        return cls(path=path)
+
     def __class_getitem__(cls, item: typing.Union[typing.Type, str]) -> typing.Type[FlyteDirectory]:
         if item is None:
             return cls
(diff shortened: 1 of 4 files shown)


flytekit/types/directory/types.py
```diff
@@ -206,6 +206,23 @@ def new_remote(cls, stem: typing.Optional[str] = None, alt: typing.Optional[str]
...
+        """
+
+        if os.path.isabs(dirname):
+            raise ValueError("Path should be relative.")
+
+        ctx = FlyteContextManager.current_context()
+
+        path = os.path.join(ctx.user_space_params.working_directory, dirname)
+
+        os.makedirs(path, exist_ok=False)
```
eapolinario, line 222: I love the `exist_ok=False`.

eapolinario (approved):

cosmicBboy (approved):

PR #2853, opened 2024-10-23
add `__hash__` method to `FlyteFile` to fix bug during interactive mode

granthamtaylor:
This task will fail only in a notebook. It will not fail when running locally as a python script, running locally via pyflyte, or running remotely.

```python
import flytekit as fk

@fk.task
def write_file(message: str) -> fk.FlyteFile:

    ff = fk.FlyteFile(path='myfile.txt')

    with open(ff, mode="w") as file:
        file.write(message)

    return ff

write_file(message='hello world')
```

```
> TypeError: Error encountered while executing '968396757.write_file':
  unhashable type: 'FlyteFile'
```

This can be fixed by adding the `__hash__` method to `FlyteFile`. I am creating the hash from the serialized representation of the `FlyteFile`.

### DIFF
--- a/flytekit/types/file/file.py
+++ b/flytekit/types/file/file.py
@@ -372,6 +372,9 @@ class FlyteFile(SerializableType, os.PathLike, typing.Generic[T], DataClassJSONM
     def __str__(self):
         return self.path
 
+    def __hash__(self):
+        return hash(self._serialize()["path"])
+
 
 class FlyteFilePathTransformer(AsyncTypeTransformer[FlyteFile]):
     def __init__(self):


granthamtaylor:
> The root cause is in Jupyter's interactiveshell.py, where it checks whether the opened file is one of stdin, stdout, stderr. During this check, the file is compared against a set (or dictionary keys) that contains these standard I/O objects.

> Since sets (and dictionary keys) rely on hashing for membership checks, Jupyter attempts to hash the opened file. FlyteFile is not hashable, so it throws an error.

flytekit/types/file/file.py
```diff
@@ -372,6 +372,9 @@ def __repr__(self):
     def __str__(self):
         return self.path
 
+    def __hash__(self):
+        return hash(self._serialize()["path"])
```
thomasjpfan, line 376: Can we add a small test where we place `FlyteFile` into a set or a dict? (To emulate what Jupyter does without having to have add a new full integration test)

thomasjpfan (approved):

wild-endeavor:
@Mecoli1219 did some digging around this and found that jupyter is checking to see if this is stdin, out, err.
image (5)

I think this is fine, but in the interest of rigor, i don't see a harm in including all the fields of the object (excluding the downloader function but including the downloaded bit) as part of the hash computation.  what do you think @Mecoli1219?

Also this change is not needed for the FlyteDirectory type right?  That also has a `__fspath__` function.

Mecoli1219:
I am not really sure, but I think that there wouldn't be a use case where there are two FlyteFiles in the same process that have the same path with different downloader functions. If there exists, that would be really weird. Let's say we create two FlyteFile:
```
f1 = FlyteFile("test.txt", downloader=...)
f2 = FlyteFile("test.txt", downloader=...)
```
The order of executing `f1.download()` and `f2.download()` will cause result different.

If the previous statement is right, we would only have two FlyteFiles with the same downloader function, and this also means that two FlyteFiles should be the same FlyteFile. However, it is possible that the downloaded bits are different in two instances (Just think that one calls `download()` and another doesn't).  If we consider the downloaded bit in the hash function, this would probably cause two files to be not the same after hashing.

So I don't think I'll include the downloaded bit here, but I am not entirely familiar with the usage of FlyteFile, this is just my current thought.

granthamtaylor:
> Also this change is not needed for the FlyteDirectory type right? That also has a __fspath__ function.

I tried a number of operations and I couldn't produce any unexpected behavior.

Yes, you are right, I should have used a more comprehensive hash function.

PR #2855, opened 2024-10-24
Async/Batching of coroutines

wild-endeavor:
Because of the unbounded nature of two of the transformers, we noticed higher levels of memory usage when running tests.  This runs the coroutines in batches.

For the list and dict transformers only, instead of doing `asyncio.gather` on everything at once, run them via fsspec's helper function that batches coroutines.  This has the added benefit that it makes the batch size controllable through fsspec's config mechanism.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1563,7 +1563,7 @@ class ListTransformer(AsyncTypeTransformer[T]):
         else:
             t = self.get_sub_type(python_type)
             lit_list = [asyncio.create_task(TypeEngine.async_to_literal(ctx, x, t, expected.collection_type)) for x in python_val]
-            lit_list = await _run_coros_in_chunks(lit_list)
+            lit_list = await _run_coros_in_chunks(lit_list, batch_size=20)
             # lit_list = await asyncio.gather(*lit_list)
 
         return Literal(collection=LiteralCollection(literals=lit_list))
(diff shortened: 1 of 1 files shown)


flytekit/core/type_engine.py
```diff
@@ -22,6 +22,7 @@
 import msgpack
 from dataclasses_json import DataClassJsonMixin, dataclass_json
 from flyteidl.core import literals_pb2
+from fsspec.asyn import _run_coros_in_chunks  # pylint: disable=W0212
```
thomasjpfan, line 25: Can we bring in this code from fsspec? I rather not depend on private API from another library.

eapolinario: fsspec is not going anywhere. I don't understand why that particular method is private, but that function existed in its current form [since 2021](https://github.com/fsspec/filesystem_spec/blame/master/fsspec/asyn.py#L205-L233). IMO it's safe to assume that's going to live there for a long time.

wild-endeavor: asked them to make it public: https://github.com/fsspec/filesystem_spec/issues/1740.  i think we should just use the private function until they make it public.  if they refuse, we can always copy later.  If we copy, i'll also be copying in the config code cuz I think it is nice to have some user facing controls around that.

wild-endeavor: @thomasjpfan?  is this satisfactory?

eapolinario (approved):

PR #2857, opened 2024-10-24
Remove pickle batching

wild-endeavor:
Removing support for the BatchSize in the list of flyte pickle use-case.  This is a breaking change, but we think this feature was rarely used, and should be less needed now that the uploading/downloading of pickles is done through async.


eapolinario:
flytesnacks CI checks master are failing due to this: https://github.com/flyteorg/flytesnacks/actions/runs/12448000284/job/34752111784.

PR #2866, opened 2024-10-25
Fix Jupyter Versioning 

Mecoli1219:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5881

We should not create new tasks for each updated code in Jupyter Notebook's cells. We should aggregate the same tasks or workflows and update the version only.

1. Create a new hash function for the pickled entity, and compute the version if the version is not specified before fetching the entities. Therefore, if we don't update the entity, we will use the one with the same version. However, if we update the code (to be specific, re-executing the unchanged cell will also count as an update), the version will be different, and new version will be registered.
2. Make the module names of entities from Jupyter Notebook empty string so that it doesn't create multiple tasks for the same task.
3. I found that sometimes Jupyter Notebook's virtual file is actually an accessible file, like during debugging mode (I didn't see any related doc for Jupyter Notebook yet).

### Setup process
Run some examples, the first two screenshots are the executions without updating the code. The last one is the execution where the entity definitions are re-executed (Note that I didn't update the code inside).
### Screenshots
- 1st time
  image
- 2nd time
  image
- 3rd time (Re-execute definition)
  image
- The versions
  image

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -200,13 +200,15 @@ def _get_git_repo_url(source_path: str):
         return ""
 
 
-def _get_pickled_target_dict(root_entity: typing.Union[WorkflowBase, PythonTask]) -> typing.Tuple[bytes, typing.Dict[str, PythonAutoContainerTask]]:
+def _get_pickled_target_dict(
+    root_entity: typing.Union[WorkflowBase, PythonTask],
+) -> typing.Tuple[bytes, typing.Dict[str, PythonAutoContainerTask]]:
     """
     Get the pickled target dictionary for the entity.
     :param root_entity: The entity to get the pickled target for.
         :return: hashed bytes and the pickled target dictionary.
     """
-    queue = [root_entity]
+    queue: typing.List[typing.Union[WorkflowBase, PythonTask, CoreNode]] = [root_entity]
     pickled_target_dict = {}
     while queue:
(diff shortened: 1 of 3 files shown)


kumare3:
this is pretty cool. I would need help in reviewing this. cc @wild-endeavor ?

wild-endeavor (dismissed):

flytekit/remote/remote.py
```diff
@@ -229,7 +231,8 @@ def _get_pickled_target_dict(root_entity: typing.Union[WorkflowBase, PythonTask]
                 queue.append(task)
         elif isinstance(entity, CoreNode):
             queue.append(entity.flyte_entity)
-    return pickled_target_dict
+    md5_bytes = hashlib.md5(cloudpickle.dumps(pickled_target_dict)).digest()
```
thomasjpfan, line 234: With this change, I think we end up calling `cloudpickle.dumps(pickled_dict)` twice, here and in `_pickle_and_upload_entity`. Is there a way to only pickle once?

Mecoli1219: that's true. We can store the `pickled_target_dict` output at L1990, and send it into `_pickle_and_upload_entity()`. In `_pickle_and_upload_entity()`, it could compute pickle dict if `pickled_target_dict` is None

wild-endeavor (approved):

PR #2869, opened 2024-10-26
improved output handling in notebooks

kumare3:
This adds native capability of retrieving outputs in the right type - for example a dataframe or a tensor as the notebook remembers the original interface.

For example if you have a method that produces a dataframe
Screenshot 2024-10-25 at 5 26 22 PM

it can be directly accessed in correct type - it is also possible to retrieve it as another type like - polars dataframe etc
Screenshot 2024-10-25 at 5 25 59 PM

and can be passed to a downstream function

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -1388,7 +1388,12 @@ class FlyteRemote(object):
             exec_id = WorkflowExecutionIdentifier(
                 project=project or self.default_project, domain=domain or self.default_domain, name=execution_name
             )
-        execution = FlyteWorkflowExecution.promote_from_model(self.client.get_execution(exec_id), remote=self)
+        type_hints = None
+        if entity.python_interface:
+            type_hints = entity.python_interface.outputs
+        execution = FlyteWorkflowExecution.promote_from_model(
+            self.client.get_execution(exec_id), remote=self, type_hints=type_hints
+        )
 
         if wait:
             return self.wait(execution)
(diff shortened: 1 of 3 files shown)


wild-endeavor (approved):

flytekit/remote/remote.py
```diff
@@ -2080,6 +2087,7 @@ def execute_local_launch_plan(
         domain = resolved_identifiers.domain
         try:
             flyte_launchplan: FlyteLaunchPlan = self.fetch_launch_plan(**resolved_identifiers_dict)
+            flyte_launchplan._python_interface = entity.python_interface
```
thomasjpfan, line 2090: When a function adjusts another objects private attribute, it feels like an anti-pattern. 

I'll be okay with either:

1. `RemoteEntity._python_interface` to `RemoteEntity.python_interface`
2. Adding a `setter`:

```python
class RemoteEntity:
    @python_interface.setter
    def python_interface(self, value):
        self._python_interface = value
```

This way, we can write `flyte_launchplan.python_interface = entity.python_interface` here.

kumare3: ok will do

kumare3: I actually completely agree with you, I hate modifying private properties. Sadly this was the previous case. But I now made sure and do not see it used anywhere, so updated it everywhere.

kumare3:
@wild-endeavor / @thomasjpfan / @pingsutw Please approve. Addressed the comments

pingsutw (approved):

PR #2870, opened 2024-10-27
Remove misleading address field from Ray Task

Sovietaced:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/5877

The documentation implies that you can run a ray job on an existing cluster by configuring the address but this is not  true since the task still requires users to configure details for a new and dynamically created Ray Cluster. And ray jobs will run on the dynamically created Ray Cluster even if an address is configured. 

## Related PRs
https://github.com/flyteorg/flytesnacks/pull/1765

### DIFF
--- a/plugins/flytekit-ray/flytekitplugins/ray/task.py
+++ b/plugins/flytekit-ray/flytekitplugins/ray/task.py
@@ -43,7 +43,6 @@ class RayJobConfig:
     head_node_config: typing.Optional[HeadNodeConfig] = None
     enable_autoscaling: bool = False
     runtime_env: typing.Optional[dict] = None
-    address: typing.Optional[str] = None
     shutdown_after_job_finishes: bool = False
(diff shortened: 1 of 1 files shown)


wild-endeavor (commented):
This is backwards incompatible right?

plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -65,7 +64,7 @@ def __init__(self, task_config: RayJobConfig, task_function: Callable, **kwargs)
         self._task_config = task_config
 
     def pre_execute(self, user_params: ExecutionParameters) -> ExecutionParameters:
-        init_params = {"address": self._task_config.address}
```
pingsutw, line 68: thanks. I think this only works for local execution. For those people who still want to use it, they can use `RAY_ADDRESS` env.

pingsutw (approved):

PR #2872, opened 2024-10-28
Offload literals

eapolinario:
## Tracking issue
RFC https://github.com/flyteorg/flyte/pull/5103

We need to be able to offload literals from flytekit.

This PR adds support for offloading of literals gated by the environment variables `_F_L_MIN_SIZE_MB` and `_F_L_MAX_SIZE_MB`, the minimal literal size for offloading and the max size, respectively.

Since we need to use the output name (i.e. "o0", "o1", etc) in the name of the offloaded literal, we use the entrypoint (since it's there that we have natural access to the metadata bucket) to write the offloaded literals to files named "o0_offloaded_metadata.pb", "o1_offloaded_metadata.pb", etc.  

### How hashing is handled
We need a stable representation of offloaded literals for caching purposes. In golang we use [jsonpb.Marshaler](https://github.com/flyteorg/flyte/blob/13b3d82c94fb61047a49aa2093a2b17fb6a7c792/flytestdlib/pbhash/pbhash.go#L32) + [objecthash](https://github.com/benlaurie/objecthash) to generate a stable hash for a proto message, but unfortunately there's no readily available equivalent in python. Instead, we rely on the stable json representation of the proto Literal and use SHA256 to produce the hash. 

In case a task makes use of [HashMethod](https://docs.flyte.org/en/latest/user_guide/development_lifecycle/caching.html#caching-of-non-flyte-offloaded-objects), then we use that hash in the offloaded literal.

## How was this patch tested?
Unit tests and testing locally in sandbox.


flytekit/bin/entrypoint.py
```diff
@@ -137,7 +137,41 @@ def _dispatch_execute(
...
+
+                    # TODO: hash calculation
+
+                    offloaded_filename = f"{k}_offloaded_metadata.pb"
+
+                    offloaded_literal = _literal_models.Literal(
+                        offloaded_metadata=_literal_models.LiteralOffloadedMetadata(
+                            uri=f"{ctx.user_space_params.output_metadata_prefix}/{offloaded_filename}",
+                            size_bytes=lit.ByteSize(),
+                            # TODO: do I have to set the inferred literal type?
```
eapolinario, line 165: @pmahindrakar-oss , in what conditions we have to set this?

flytekit/bin/entrypoint.py
```diff
@@ -137,7 +137,41 @@ def _dispatch_execute(
...
+                lit = v.to_flyte_idl()
+                if lit.ByteSize() >= max_offloaded_size:
+                    raise ValueError(
+                        f"Literal {k} is too large to be offloaded. Max literal size is {max_offloaded_size} whereas the literal size is {lit.ByteSize()} bytes"
+                    )
+
+                if lit.ByteSize() >= min_offloaded_size:
+                    logger.debug(f"Literal {k} is too large to be inlined, offloading to metadata bucket")
+
+                    # TODO: hash calculation
```
eapolinario, line 157: In golang we use [jsonpb.Marshaler](https://github.com/flyteorg/flyte/blob/13b3d82c94fb61047a49aa2093a2b17fb6a7c792/flytestdlib/pbhash/pbhash.go#L32) + [objecthash](https://github.com/benlaurie/objecthash) to generate a stable hash for a proto message. 

We'll have to figure out a similar structure for the offloaded literals. The objecthash repo provides an [python](https://github.com/benlaurie/objecthash/blob/master/objecthash.py) implementation, but I couldn't find a package in pypi (so we might need to copy the code with attribution if we go that route).

flytekit/bin/entrypoint.py
pmahindrakar-oss, line 165: We do need to add inferred type, since during consumption we validate the input type

https://github.com/unionai/flyte/blob/5e9ec71a3b55003955642aeeaaf6dbeb43b3a279/flytepropeller/pkg/compiler/validators/utils.go#L271-L272

https://github.com/unionai/flyte/blob/2c070c301ccb37c5a336dd7022a1d9846d9bb173/flyteadmin/pkg/manager/impl/validation/execution_validator.go#L100

flytekit/bin/entrypoint.py
```diff
@@ -137,7 +137,41 @@ def _dispatch_execute(
             logger.warning("Task produces no outputs")
             output_file_dict = {_constants.OUTPUT_FILE_NAME: _literal_models.LiteralMap(literals={})}
         elif isinstance(outputs, _literal_models.LiteralMap):
-            output_file_dict = {_constants.OUTPUT_FILE_NAME: outputs}
+            offloaded_literals: Dict[str, _literal_models.Literal] = {}
+            literal_map_copy = {}
+
+            min_offloaded_size = int(os.environ.get("FK_L_MIN_SIZE_MB", "10")) * 1024 * 1024
+            max_offloaded_size = int(os.environ.get("FK_L_MAX_SIZE_MB", "1000")) * 1024 * 1024
```
pmahindrakar-oss, line 144: Do we use the same names here that we use in propeller
https://github.com/unionai/flyte/blob/c10b55373e9dea2bcce493f978499648c3f89d8f/flytepropeller/pkg/controller/config/config.go#L204-L207

flytekit/bin/entrypoint.py
pmahindrakar-oss, line 157: +1

flytekit/interaction/string_literals.py
```diff
@@ -61,6 +61,9 @@ def literal_string_repr(lit: Literal) -> typing.Any:
         return [literal_string_repr(i) for i in lit.collection.literals]
     if lit.map:
         return {k: literal_string_repr(v) for k, v in lit.map.literals.items()}
+    if lit.offloaded_metadata:
+        # TODO: load literal from offloaded literal?
+        return f"Offloaded literal metadata: {lit.offloaded_metadata}"
```
pmahindrakar-oss, line 66: Is there an example we can try to see if we need to load the literal here

flytekit/bin/entrypoint.py
pmahindrakar-oss: This is needed when we have launcplan execution created by propeller 
eg : here 
https://demo.hosted.unionai.cloud/console/projects/flytesnacks/domains/development/executions/akzthr8gxbdvpnq6f5lr/nodes

```[UserError] failed to launch workflow, caused by: rpc error: code = InvalidArgument desc = invalid input input wrong type. Expected collection_type:{simple:STRING}, but got collection_type:{}```

Code link in my comments below https://github.com/flyteorg/flytekit/pull/2872/files#r1819666906

File: flytekit/bin/entrypoint.py (hunk 3 above)
Comment (eapolinario) [reply flytekit/bin/entrypoint.py:144]: we'll need to update the [injection of environment variables ](https://github.com/flyteorg/flyte/blob/master/flyteplugins/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go#L141)to match those names.

The need for short env vars is described https://github.com/flyteorg/flyte/issues/5665.

File: flytekit/interaction/string_literals.py (hunk 4 above)
Comment (eapolinario) [reply flytekit/interaction/string_literals.py:66]: this is relevant for the `pyflyte fetch` command:
```
❯ pyflyte --config ~/.flyte/config-sandbox.yaml fetch flyte://v1/flytesnacks/development/asvskwn766f5v492pzgt/n0-0-n1/o
Fetching data from flyte://v1/flytesnacks/development/asvskwn766f5v492pzgt/n0-0-n1/o...
╭───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ {                                                                                                                                                                                                                                                                                             │
│     'o0': [                                                                                                                                                                                                                                                                                   │
│         'Offloaded literal metadata: Flyte Serialized object (LiteralOffloadedMetadata):\n  uri: s3://my-s3-bucket/metadata/propeller/flytesnacks-development- [...]\n  size_bytes: 39936015',                                                                                                │
│         'Offloaded literal metadata: Flyte Serialized object (LiteralOffloadedMetadata):\n  uri: s3://my-s3-bucket/metadata/propeller/flytesnacks-development- [...]\n  size_bytes: 39936015',                                                                                                │
...
│         'Offloaded literal metadata: Flyte Serialized object (LiteralOffloadedMetadata):\n  uri: s3://my-s3-bucket/metadata/propeller/flytesnacks-development- [...]\n  size_bytes: 39936015'                                                                                                 │
│     ]                                                                                                                                                                                                                                                                                         │
│ }                                                                                                                                                                                                                                                                                             │
╰───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯

```

flytekit/bin/entrypoint.py
pmahindrakar-oss: Ok with keeping the current ones .

flytekit/bin/entrypoint.py
pmahindrakar-oss: Wont we still need this in the new isInstance form https://github.com/flyteorg/flyte/pull/5909/files#diff-af9c7a5cd03e0e860c155861edcf0b96863e8781526f51d1b13c7f9278ffa72fR415

flytekit/bin/entrypoint.py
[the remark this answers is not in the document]
eapolinario: Since the hash calculation in entrypoint only happens after dispatch_execute (in other words, if a user defined a `HashMethod` then we can reuse it here). I just pushed an update that implements this idea.

flytekit/bin/entrypoint.py
eapolinario: Let's separate those two discussions. We might be able to remove `LiteralTypeForLiteral`, but that shouldn't block us from getting the actual literal type from the typed interface in flytekit and use it here.

Also, if/when we end up removing `LiteralTypeForLiteral` we should use a different field to store the offloaded literal literal type and not use `inferred` in the name since it'll be the exact type.

eapolinario: I commented in about separating those discussions and how we use the task typed interface to set `inferred_type` with the appropriate literal type.

pmahindrakar-oss (approved):

eapolinario:
hold on, I had a bad merge.

wild-endeavor (dismissed):
+1 with dm

wild-endeavor (approved):

wild-endeavor (approved):

PR #2875, opened 2024-10-29
Add memray plugin

fiedlerNr9:

- Enables memray profiling on Flyte task level
- renders memray report into Flytedeck

- Adding memray flytekit plugin

## How was this patch tested?
- unit tests
- tested local & remote run

### Setup process
```python
from flytekit import workflow, task, ImageSpec
from flytekitplugins.memray import memray_profiling
import time


flytekit_hash = "82d5ac739f5f02998edb9538c58cf93c8f6e501b"
flytekitplugins_memray = f"git+https://github.com/flyteorg/flytekit.git@{flytekit_hash}#subdirectory=plugins/flytekit-memray"

image = ImageSpec(
    name="memray_demo",
    python_version="3.11.10",
    apt_packages=["git"],
    packages=[flytekitplugins_memray],
    registry="ghcr.io/fiedlernr9",
)


def generate_data(n: int):
    leak_list = []
    for _ in range(n):  # Arbitrary large number for demonstration
        large_data = " " * 10**6  # 1 MB string
        leak_list.append(large_data)  # Keeps appending without releasing
        time.sleep(0.1)  # Slow down the loop to observe memory changes


@task(container_image=image, enable_deck=True)
@memray_profiling(memray_html_reporter="table")
def memory_usage(n: int) -> str:
    generate_data(n=n)

    return "Well"


@task(container_image=image, enable_deck=True)
@memray_profiling(trace_python_allocators=True, memray_reporter_args=["--leaks"])
def memory_leakage(n: int) -> str:
    generate_data(n=n)

    return "Well"


@workflow
def wf(n: int = 500):
    memory_usage(n=n)
    memory_leakage(n=n)
```

### Flamegraph
image

### Table
image


eapolinario (commented):
This is wicked cool!

Can you add `flytekit-memray` to https://github.com/flyteorg/flytekit/blob/master/.github/workflows/pythonbuild.yml#L319-L364 ?

plugins/flytekit-memray/flytekitplugins/memray/profiling.py
```diff
@@ -0,0 +1,74 @@
...
+
+import memray
+from flytekit import Deck
+from flytekit.core.utils import ClassDecorator
+
+
+class memray_profiling(ClassDecorator):
+    def __init__(
+        self,
+        task_function: Optional[Callable] = None,
```
thomasjpfan, line 13: Can we allow for `native_traces`, `trace_python_allocators`, `follow_fork`, `memory_interval_ms` from `memray.Tracker` to be passed through here?

With `trace_python_allocators=True`, I do not think we need to set `PYTHONMALLOC` in the environment anymore.

fiedlerNr9: Nice catch! Added these 👍

fiedlerNr9:
> This is wicked cool!
> 
> Can you add `flytekit-memray` to https://github.com/flyteorg/flytekit/blob/master/.github/workflows/pythonbuild.yml#L319-L364 ?

✅

plugins/flytekit-memray/README.md
```diff
@@ -0,0 +1,55 @@
...
+```python
+from flytekit import workflow, task, ImageSpec
+from flytekitplugins.memray import memray_profiling
+import time
+
+
+image = ImageSpec(
+    name="memray_demo",
+    packages=["flytekitplugins_memray"],
+    env={"PYTHONMALLOC": "malloc"},
```
thomasjpfan, line 22: Instead of hard coding this into the environment, can we now `trace_python_allocators=True`?

fiedlerNr9: I tested that and its not throwing any warnings but the results look different though:
Thats the task I tested without having the env variable set:
``` python
@task(container_image=image, enable_deck=True)
@memray_profiling(trace_python_allocators=True, memray_reporter_args=["--leaks"])
def memory_leakage(n: int) -> str:
    generate_data(n=n)

    return "Well"
```

Thats the result:
image

Not sure if this is expected

thomasjpfan: For completeness, what do you see when you set `trace_python_allocators=False`?

fiedlerNr9: This, which is expected i guess
image

thomasjpfan: It's weird that it gives two different flamegraphs. The new flamegraph makes more sense to me because the tracker is wrapping the user code and you can clearly see the `generete_data`.

I can not really see where the `generate_data` is on your original flamegraph.

fiedlerNr9: Oh man, I mixed up my demos. Its looking exactly the same with `trace_python_allocaters=False` and having the env variable set. Sorry for the confusion, I will update the Readme shortly
image

plugins/flytekit-memray/tests/test_memray_profiling.py
```diff
@@ -0,0 +1,39 @@
+from unittest.mock import Mock, patch
+import pytest
+from flytekit import task, current_context
+from flytekitplugins.memray import memray_profiling
+
+
+@task(enable_deck=True)
+@memray_profiling
```
thomasjpfan, line 8: **Not actionable for this PR** I wish there was a way to ensure that `enable_deck=True` when using `memray_profiling`. Otherwise, we just add overhead without any reports.

@eapolinario @pingsutw What do you think of making `deck_fields=None` and set `enable_decks=True`?

https://github.com/flyteorg/flytekit/blob/master/flytekit/core/task.py#L203-L210

eapolinario: What if we added a way to carry over task kwargs from successive flytekit decorators? Something like https://github.com/flyteorg/flytekit/pull/2911.

plugins/flytekit-memray/flytekitplugins/memray/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
+.. currentmodule:: flytekitplugins.wandb
+
+This package contains things that are useful when extending Flytekit.
+
+.. autosummary::
+   :template: custom.rst
+   :toctree: generated/
+
+   wandb_init
```
eapolinario, line 10: replace mentions to wandb

fiedlerNr9: What exactly is supposed to be in this comment section of `__init__.py`?
I planned to just get rid of it actually?

eapolinario: This is exposed as the landing page for the plugin in the docs, e.g. https://docs.flyte.org/en/latest/api/flytekit/plugins/wandb.html

fiedlerNr9: Good catch - updated!

ppiegaze (dismissed):

plugins/flytekit-memray/flytekitplugins/memray/__init__.py
fiedlerNr9: Removed adding docs from this PR, since we need a published version of thi plugin on pypi to continue with the process of adding docs to flytesnacks & flyte. Will follow up with docs as soon this is merged. 
Ok with you @eapolinario & can you please have a last review if times allows

plugins/flytekit-memray/setup.py
```diff
@@ -0,0 +1,37 @@
...
+    name=microlib_name,
+    version=__version__,
+    author="flyteorg",
+    author_email="admin@flyte.org",
+    description="This package enables memory profiling for tasks with memray",
+    namespace_packages=["flytekitplugins"],
+    packages=[f"flytekitplugins.{PLUGIN_NAME}"],
+    install_requires=plugin_requires,
+    license="apache2",
+    python_requires=">=3.8",
```
eapolinario, line 21: we [recently dropped](https://github.com/flyteorg/flytekit/pull/2909) support for python 3.8. Can you also remove it from here?

eapolinario (approved):

PR #2878, opened 2024-10-30
[Core feature] Enforce Keyword-Only Arguments in SecretsManager.get Method and Add Deprecation Warning

10sharmashivam:
## Tracking issue

Reference [Issue](https://github.com/flyteorg/flyte/issues/5773) 

The changes are needed to enforce keyword-only arguments in the SecretsManager.get method, addressing a common usability issue where users inadvertently use positional arguments, which could lead to confusion or errors in secret retrieval. Additionally, a deprecation warning will be raised whenever positional arguments are passed.

1.	Updated the SecretsManager.get method to require keyword arguments by adding * to the method signature. This change enforces that all arguments must be provided as keywords.
2.	Introduced a deprecation warning for the use of positional arguments. If the method is called with positional arguments, a Deprecation Warning is raised to inform users of the impending removal of support for this usage.

## How was this patch tested?

•	test_get_with_keyword_arguments: Verified that calling the get method with keyword arguments works as expected.
•	test_get_with_positional_arguments: Confirmed that calling the get method with positional arguments raises the appropriate Deprecation Warning.

- [ X] All new and existing tests passed.
- [ X] All commits are signed off.

### DIFF
--- a/flytekit/core/context_manager.py
+++ b/flytekit/core/context_manager.py
@@ -19,6 +19,7 @@ import pathlib
 import tempfile
 import traceback
 import typing
+import warnings
 from contextlib import contextmanager
 from contextvars import ContextVar
 from dataclasses import dataclass, field
@@ -359,11 +360,19 @@ class SecretsManager(object):
 
     def get(
         self,
+        *,
         group: Optional[str] = None,
         key: Optional[str] = None,
         group_version: Optional[str] = None,
         encode_mode: str = "r",
     ) -> str:
+        # Raise a warning if this method is called without keyword arguments
+        if (group is not None and key is not None) and (not isinstance(group, str) or not isinstance(key, str)):
+            warnings.warn(
+            "The use of positional arguments for SecretsManager.get is deprecated. "
+            "Please use keyword arguments instead. This will be removed in a future version.",
+            DeprecationWarning
+        )
         """
         Retrieves a secret using the resolution order -> Env followed by file. If not found raises a ValueError
         param encode_mode, defines the mode to open files, it can either be "r" to read file, or "rb" to read binary file


flytekit/core/context_manager.py
```diff
@@ -359,11 +360,19 @@ def __getattr__(self, item: str) -> _GroupSecrets:
 
     def get(
         self,
+        *,
```
thomasjpfan, line 363: Adding a `*` right away is backward breaking. This means that old code using position arguments like `secret.get("my_group", "key")` will stop working. 

I implemented a similar deprecation logic in scikit-learn here: https://github.com/scikit-learn/scikit-learn/blob/44017bb4a09359c22d4810c870fd69fbd5673e63/sklearn/utils/validation.py#L33 that you copy into flytekit.

The decorator can be used like this, and it'll raise the correct warning based on where the `*` is, but also continue to support positional arguments.

```python
@_deprecate_positional_args(version="...")
def get(self, *, ...)
```

10sharmashivam: Hi @thomasjpfan, Thank you for the feedback! I’ll implement the decorator for handling deprecation warnings, similar to the logic used in scikit-learn, to maintain backward compatibility. I’ll update the PR shortly with these changes.

10sharmashivam: Hi @thomasjpfan, I’ve implemented the decorator based on the structure and logic from the scikit-learn reference to ensure backward compatibility, issuing a FutureWarning for positional arguments as discussed. While the core logic seems aligned, it didn’t fully pass tests, even with debug statements added for troubleshooting.

Any feedback or guidance on fine-tuning or spotting potential issues with the test outcomes would be a great help!

Regards

eapolinario (commented):
@10sharmashivam , can you add a few unit tests just to confirm the behavior?

10sharmashivam:
> @10sharmashivam , can you add a few unit tests just to confirm the behavior?

Hi @eapolinario 

I’ll add the unit tests to confirm the decorator’s behavior and will update the PR soon.

10sharmashivam:
> @10sharmashivam , can you add a few unit tests just to confirm the behavior?

Hi @eapolinario @thomasjpfan,

I’ve implemented the decorator based on the logic and approach suggested by @thomasjpfan, referencing [sklearn](https://github.com/scikit-learn/scikit-learn/blob/44017bb4a09359c22d4810c870fd69fbd5673e63/sklearn/utils/validation.py#L33). The goal is to ensure backward compatibility while issuing a `FutureWarning` for positional arguments as discussed. 

To validate the behavior, I temporarily added debug print statements to `_deprecate_positional_args` to confirm it is being triggered during tests. I’ve also included unit tests (@eapolinario as as suggested), but despite these efforts, the tests are not fully passing, as the `FutureWarning` is not being captured as expected.

Any feedback or guidance on fine-tuning the decorator or identifying potential issues with the tests would be greatly appreciated!

Thanks in advance for your help!

Regards

flytekit/core/context_manager.py
davidmirror-ops: cc @eapolinario

PR #2889, opened 2024-11-01
Default nb task resolver msg

cosmicBboy:

It's useful to the know the unpickled object and its type when loading the task with the default notebook task resolver. This just eases the debugging experience.

Updates the error message when the unpickled file is of the unexpected type.


eapolinario (commented):
Can we move the printing of `loaded_data` to the end of the string? I'm concerned about objects that don't have a well-behaving string representation making this error difficult to read.

eapolinario (approved):

thomasjpfan (approved):

PR #2890, opened 2024-11-01
Kill the vscode server itself when resume the task

pingsutw:
## Tracking issue
NA

When we resume the task in the vscode, it will send a `SIGTERM` to flytekit (main process), then flytekit will [kill](https://github.com/flyteorg/flytekit/blob/5bf0acd22623f02293a75d41952dd4ec26e025ae/flytekit/interactive/vscode_lib/decorator.py#L61) the vscode server. However, if the main process is down for some reason, the VSCode server won't shut down even if we resume the task.

Kill the vscode server itself when resuming the task.

## How was this patch tested?
```python
from union.actor import ActorEnvironment
from flytekit import logger
from flytekit.interactive import vscode

new_flytekit = "git+https://github.com/flyteorg/flytekit.git@3a9afb1a88753fe3c783da91fc501b8fc7838fb8"
image_spec = ImageSpec(name="actor", registry="pingsutw", packages=["panda", "numpy", "union", new_flytekit], apt_packages=["git"], env={"FLYTE_SDK_LOGGING_LEVEL": "10"})


actor1 = ActorEnvironment(
    name="foo17",
    container_image=image_spec,
    replica_count=1,
    ttl_seconds=600,
    requests=Resources(cpu="2", mem="2000Mi"),
)


@actor1.task(enable_deck=True)
@vscode(port=8080)
def t1(a: int = 3, b: int = 4) -> int:
    print("print---------------------------------------------")
    # vs = os.getenv("_F_E_VS", "None")
    # logger.info(f"_F_E_VS: {vs}")
    logger.info("logger-----------------------------------")
    raise ValueError("This is an error")
    return a + b

@workflow
def wf():
    t1()
```

### DIFF
--- a/flytekit/interactive/vscode_lib/decorator.py
+++ b/flytekit/interactive/vscode_lib/decorator.py
@@ -67,7 +67,7 @@ def exit_handler(
     delta = 0
 
     logger.info("waiting for task to resume...")
-    while child_process.is_alive() and not resume_task.is_set():
+    while child_process.is_alive():
         if not os.path.exists(HEARTBEAT_PATH):
             delta = time.time() - start_time
             logger.info(f"Code server has not been connected since {delta} seconds ago.")
@@ -464,7 +464,7 @@ class vscode(ClassDecorator):
         prepare_resume_task_python(child_process.pid)
 
         # 6. Register the signal handler for task resumption. This should be after creating the subprocess so that the subprocess won't inherit the signal handler.
-        signal.signal(signal.SIGTERM, resume_task_handler)
+        # signal.signal(signal.SIGTERM, resume_task_handler)
 
         return exit_handler(
             child_process=child_process,
--- a/flytekit/interactive/vscode_lib/decorator.py
+++ b/flytekit/interactive/vscode_lib/decorator.py
@@ -439,6 +439,7 @@ class vscode(ClassDecorator):
         # 2. Prepare the interactive debugging Python script and launch.json.
         prepare_interactive_python(self.task_function)  # type: ignore
 
+        logger.info(f"pid, {os.getpid()}")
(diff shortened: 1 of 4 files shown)


flytekit/interactive/vscode_lib/decorator.py
```diff
@@ -195,14 +197,14 @@ def download_vscode(config: VscodeConfig):
 
     # If the code server already exists in the container, skip downloading
     executable_path = shutil.which(EXECUTABLE_NAME)
-    if executable_path is not None:
+    if executable_path is not None or os.path.exists(DOWNLOAD_DIR):
```
pingsutw, line 200: For the case where the file has been downloaded but is not in the `$PATH`.

flytekit/interactive/vscode_lib/decorator.py
```diff
@@ -459,8 +443,14 @@ def execute(self, *args, **kwargs):
         )
         child_process.start()
 
-        # 6. Register the signal handler for task resumption. This should be after creating the subprocess so that the subprocess won't inherit the signal handler.
-        signal.signal(signal.SIGTERM, resume_task_handler)
+        # 3. Prepare the interactive debugging Python script and launch.json.
+        prepare_interactive_python(self.task_function)  # type: ignore
+
+        # 4. Prepare the task resumption Python script
+        prepare_resume_task_python(child_process.pid)
```
thomasjpfan, line 450: @pingsutw For my knowledge, is passing the `child_process.pid` the major part of this fix?

pingsutw: yup, otherwise, the rust worker in the actor task will be killed when resuming the task

thomasjpfan (approved):

PR #2910, opened 2024-11-06
Improve error message for nested tasks

pingsutw:
## Tracking issue
NA

Update error message for nested task and workflow

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -1413,7 +1413,7 @@ def flyte_entity_call_handler(
         or ctx.execution_state.mode == ExecutionState.Mode.LOCAL_TASK_EXECUTION
     ):
         logger.error(
-            "You are not supposed to nest @Task/@Workflow because the nested task or workflow will run in the same container."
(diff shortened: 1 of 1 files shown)


flytekit/core/promise.py
```diff
@@ -1412,7 +1412,9 @@ def flyte_entity_call_handler(
         ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION
         or ctx.execution_state.mode == ExecutionState.Mode.LOCAL_TASK_EXECUTION
     ):
-        logger.error("You are not supposed to nest @Task/@Workflow inside a @Task!")
+        logger.error(
+            "You are not supposed to nest @Task/@Workflow because the nested task or workflow will run in the same container."
```
thomasjpfan, line 1416: Nit: Do you think these should be lowercased?

```suggestion
            "You are not supposed to nest @task/@workflow because the nested task or workflow will run in the same container."
```

eapolinario (approved):

PR #2912, opened 2024-11-07
initial make auto cache plugin

dansola:

Make caching easier to use in flytekit by reducing cognitive burden of specifying cache versions

`AutoCache` protocol added to flytekit with a `salt` argument and a `get_version` method. The task function is passed to `get_version` in the task decorator to determine a cache version. Then the new cache version follows the same path a user created cache version would.

`auto_cache` plugin created which will contain implementations of `AutoCache`. The first one is `CacheFunctionBody` which just checks the function body and ignores formatting and comments. 

## How was this patch tested?

Unit tests added which verify versions are consistent and change when we expect. Since changes to the function name will cause a different hash, we move dummy functions to a separate directory and import them so we can keep the name the same but test that the hash changes with the contents change.

### DIFF
--- a/flytekit/core/auto_cache.py
+++ b/flytekit/core/auto_cache.py
@@ -1,0 +1,37 @@
+from typing import Any, Callable, Protocol, runtime_checkable
+
+
+@runtime_checkable
+class AutoCache(Protocol):
+    """
+    A protocol that defines the interface for a caching mechanism
+    that generates a version hash of a function based on its source code.
+
+    Attributes:
+        salt (str): A string used to add uniqueness to the generated hash. Default is "salt".
+
+    Methods:
+        get_version(func: Callable[..., Any]) -> str:
+            Given a function, generates a version hash based on its source code and the salt.
+    """
+
+    def __init__(self, salt: str = "salt") -> None:
+        """
+        Initialize the AutoCache instance with a salt value.
+
+        Args:
+            salt (str): A string to be used as the salt in the hashing process. Defaults to "salt".
+        """
+        self.salt = salt
+
+    def get_version(self, func: Callable[..., Any]) -> str:
+        """
+        Generate a version hash for the provided function.
+
+        Args:
+            func (Callable[..., Any]): A callable function whose version hash needs to be generated.
+
+        Returns:
+            str: The SHA-256 hash of the function's source code combined with the salt.
+        """
+        ...
--- a/flytekit/core/task.py
+++ b/flytekit/core/task.py
@@ -5,6 +5,7 @@ import os
 from functools import update_wrapper
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union, overload
 
+from flytekit.core.auto_cache import AutoCache
 from flytekit.core.utils import str2bool
 
 try:
@@ -99,7 +100,7 @@ FuncOut = TypeVar("FuncOut")
 def task(
     _task_function: None = ...,
     task_config: Optional[T] = ...,
-    cache: bool = ...,
+    cache: Union[bool, list[AutoCache]] = ...,
     cache_serialize: bool = ...,
     cache_version: str = ...,
     cache_ignore_input_vars: Tuple[str, ...] = ...,
@@ -137,7 +138,7 @@ def task(
 def task(
     _task_function: Callable[P, FuncOut],
     task_config: Optional[T] = ...,
-    cache: bool = ...,
+    cache: Union[bool, list[AutoCache]] = ...,
     cache_serialize: bool = ...,
     cache_version: str = ...,
     cache_ignore_input_vars: Tuple[str, ...] = ...,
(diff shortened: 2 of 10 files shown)


dansola:
@eapolinario @cosmicBboy Here is a draft PR doing roughly what we talked about. Would be great to get your opinions before I implement more hashing methods. A couple questions:
1. Is `flytekit/core/task.py` the right place to hash the task function body? i.e in the decorator right before we pass the cache version to `TaskMetadata`.
2. Do we still like the idea of using protocols and having a different implantation for different things like the function body, imports, imagespec, external packages, etc.? Right now `cache` will take a `list[AutoCache]` and we combine the hashes (we can hash them later so the string isn't long), but is the most user friendly as compared to some dataclass config that has a series of boolean arguments for example?

flytekit/core/auto_cache.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Attributes:
+        salt (str): A string used to add uniqueness to the generated hash. Default is "salt".
+
+    Methods:
+        get_version(func: Callable[..., Any]) -> str:
+            Given a function, generates a version hash based on its source code and the salt.
+    """
+
+    def __init__(self, salt: str = "salt") -> None:
```
cosmicBboy, line 18: nit: can the default value be `""`?

flytekit/core/task.py
```diff
@@ -343,10 +344,16 @@ def launch_dynamically():
     """
 
     def wrapper(fn: Callable[P, Any]) -> PythonFunctionTask[T]:
+        if isinstance(cache, list) and all(isinstance(item, AutoCache) for item in cache):
+            cache_versions = [item.get_version() for item in cache]
+            task_hash = "".join(cache_versions)
```
cosmicBboy, line 349: @eapolinario does the cache version string have a character limit? If so we may need to re-hash the concatenated hashes.

dansola: Ah I had a feeling this would be an issue. I just left the long join for simplicity and to not introduce a hashing function in `flytekit/core/task.py` , but happy to hash that.

dansola:
> Just putting some thoughts down:
> 
> From the current design the order of the cache objects matters, which I'm okay with... I'd like to document the pattern to define one caching policy and re-using it like so:
> 
> ```python
> cache_policy = [CacheFunctionBody(), CachePrivateModules(), ...]
> 
> @task(cache=cache_policy)
> def my_task(): ...
> ```
> 
> I wonder if it would be useful to expose a `Salt` object in the case where I want to invalidate a list of cache objects without having to change the string on one of them arbitrarily:
> 
> ```python
> #  multiple cache objects with separate salts
> cache_policy = [CacheFunctionBody(salt="a"), CachePrivateModules(salt="a"), ...]
> 
> # I just want to invalidate the cache, but it feels weird and arbitrary to pick one object to bump
> cache_policy = [CacheFunctionBody(salt="a"), CachePrivateModules(salt="b"), ...]
> ```
> 
> Doing this feels more natural
> 
> ```python
> cache_policy = [CacheFunctionBody(), CachePrivateModules(), Salt("a")]
> 
> # bump cache
> cache_policy = [CacheFunctionBody(), CachePrivateModules(), Salt("b")]
> ```
> 
> I still think it's nice to have a `salt` argument in the `AutoCache.__init__` method for the case of a single cache object

Thanks so much! Definitely agree re. a single salt. One idea I had was to use a dataclass where there is a list of `AutoCache` as one parameter and a salt string as another parameter. In this case some hashing would have to occur outside of AutoCache implementations but maybe that's not a big deal, especially if we want to hash the hashed of each individual implementation. 

Also, I like that pattern of defining the cache policy as a constant that should be defined outside of the task decorator and reused. +1 to that.

flytekit/core/task.py
```diff
@@ -343,10 +344,16 @@ def launch_dynamically():
     """
 
     def wrapper(fn: Callable[P, Any]) -> PythonFunctionTask[T]:
+        if isinstance(cache, list) and all(isinstance(item, AutoCache) for item in cache):
+            cache_versions = [item.get_version() for item in cache]
```
dansola, line 348: need to pass the function here

cosmicBboy:
@dansola another idea is to have a `CachePolicy` class that takes a list of cache classes plus `salt`:

```python
class CachePolicy:
    def __init__(self, *args, salt=""):
        self.cache_objects = args
        self.salt = salt

    def get_version(self, func):
        # use self.cache_objects to generate hashes

@task(cache=CachePolicy(CacheFunctionBody(), CachePrivateModules(), salt="mysalt"))
def task_fn(): ...
```

cosmicBboy:
also if we go with the `CachePolicy` idea, we won't need to handle a `list[AutoCache]` objects in the `@task` decorator.

dansola:
> also if we go with the `CachePolicy` idea, we won't need to handle a `list[AutoCache]` objects in the `@task` decorator.

Yup I really like that idea. I added it to the PR where I make a new implementation for image caching here: https://github.com/flyteorg/flytekit/pull/2944

Right now all the PRs are branching off each other. It's probably cleaner to consolidate.

PR #2918, opened 2024-11-11
[Flytekit][BUG] Failed to serialize FlyteTypes within Union alongside other non-None variants

mao3267:
## Tracking issue
Closes flyteorg/flyte#5988

When using strings as input for `FlyteTypes`, we need to convert them into `FlyteTypes` within the `_make_dataclass_serializable` function. However, this function currently fails for Union types that include a `FlyteType` alongside other non-None variants.

1. Modify the if statement to handle Union types with non-None variants for `FlyteTypes` as well

## How was this patch tested?
1. Test simple example on local and remote

```py
from flytekit import task
from typing import Union
from flytekit.image_spec import ImageSpec


flytekit_hash = "d4171c3cd79ed406988d4213e4e79ee2b0bd2a93"
flytekit = f"git+https://github.com/mao3267/flytekit.git@{flytekit_hash}"

image_spec = ImageSpec(
    name="serialize-union-1",
    packages=[flytekit],
    builder="default",
    registry="localhost:30000",
    apt_packages=["git", "gh"],
)

@dataclass
class DC:
    ff: Union[int, FlyteFile]

def t_dc() -> DC:
    return DC(ff="s3://path")
```

2. Add unit tests in test_type_engine.py

### Setup process
```
git clone https://github.com/flyteorg/flytekit.git
gh checkout pr 2918
pip install -e .
```
### Screenshots
1. Test simple example on local and remote

- Local

image

- Remote

image

## Related PRs
None

## Docs link
None

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -784,7 +784,7 @@ class DataclassTransformer(TypeTransformer[object]):
         from flytekit.types.structured import StructuredDataset
 
         # Handle Optional and Union Types
-        if UnionTransformer.is_optional_type(python_type) or _is_union_type(python_type):
+        if _is_union_type(python_type):
 
             def get_expected_type(python_val: T, types: tuple) -> Type[T | None]:
                 if len(set(types) & {FlyteFile, FlyteDirectory, StructuredDataset}) > 1:
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -783,8 +783,8 @@ class DataclassTransformer(TypeTransformer[object]):
         from flytekit.types.file import FlyteFile
         from flytekit.types.structured import StructuredDataset
 
-        # Handle Optional
-        if UnionTransformer.is_optional_type(python_type):
+        # Handle Optional and Union Types
+        if UnionTransformer.is_optional_type(python_type) or _is_union_type(python_type):
 
             def get_expected_type(python_val: T, types: tuple) -> Type[T | None]:
                 if len(set(types) & {FlyteFile, FlyteDirectory, StructuredDataset}) > 1:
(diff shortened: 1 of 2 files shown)


tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -968,6 +968,7 @@ class TestFileStruct(DataClassJsonMixin):
         h_prime: typing.Optional[FlyteFile] = None
         i: typing.Optional[A] = None
         i_prime: typing.Optional[A] = field(default_factory=lambda: A(a=99))
+        j: typing.Union[int, FlyteFile] = None
```
thomasjpfan, line 971: Should `None` be a part of this Union?

flytekit/core/type_engine.py
```diff
@@ -783,8 +783,8 @@ def t1() -> DC:
         from flytekit.types.file import FlyteFile
         from flytekit.types.structured import StructuredDataset
 
-        # Handle Optional
-        if UnionTransformer.is_optional_type(python_type):
+        # Handle Optional and Union Types
+        if UnionTransformer.is_optional_type(python_type) or _is_union_type(python_type):
```
wild-endeavor, line 787: if we have the second one we don't need the first one right?

mao3267: Yep, the second one is a superset. Thanks for reminding me.

tests/flytekit/unit/core/test_type_engine.py
mao3267: It should not be. Sorry for not noticing that. Thanks~

Future-Outlier (approved):
This is really a good catch, thank you.

eapolinario (approved):

PR #2923, opened 2024-11-13
Copy ca-certificates from micromamba

thomasjpfan:

For macOS using Rosetta, sometimes installing `ca-certificates` or running `update-ca-certificates` can run into errors.

1. With this PR, we copy over the certificates from the `micromamba` image, which avoids installing `ca-certificate` and running `update-ca-certificates` all together.
2. Without the apt command + `update-ca-certifcates` the builds are faster.

## How was this patch tested?

I ran a simple script with the sandbox:

```python

image_spec = ImageSpec(
    name="flytekit-check-2",
    packages=["numpy"],
    registry="localhost:30000",
)


def wow() -> float:
    import numpy as np
    return np.ones((30, 10)).sum().item()
```


eapolinario (approved):
I'm all for more reproducibility and better perf.

pingsutw (approved):

PR #2924, opened 2024-11-13
Decouple ray submitter, worker, and head resources

Sovietaced:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/5666

These changes update the flytekit-ray package such that users can configure pod specs for worker and head workloads. 

Updating the version of flyteidl and plumb k8s pods through worker and head config. 

## How was this patch tested?
See unit tests.

## Related PRs
https://github.com/flyteorg/flyte/pull/5933

### DIFF
--- a/flytekit/models/task.py
+++ b/flytekit/models/task.py
@@ -1018,7 +1018,7 @@ class K8sPod(_common.FlyteIdlEntity):
 
     def to_flyte_idl(self) -> _core_task.K8sPod:
         return _core_task.K8sPod(
-            metadata=self._metadata.to_flyte_idl(),
+            metadata=self._metadata.to_flyte_idl() if self.metadata else None,
             pod_spec=_json_format.Parse(_json.dumps(self.pod_spec), _struct.Struct()) if self.pod_spec else None,
             data_config=self.data_config.to_flyte_idl() if self.data_config else None,
         )
--- a/plugins/flytekit-ray/flytekitplugins/ray/models.py
+++ b/plugins/flytekit-ray/flytekitplugins/ray/models.py
@@ -3,6 +3,7 @@ import typing
 from flyteidl.plugins import ray_pb2 as _ray_pb2
 
 from flytekit.models import common as _common
+from flytekit.models.task import K8sPod
 
 
 class WorkerGroupSpec(_common.FlyteIdlEntity):
@@ -13,12 +14,14 @@ class WorkerGroupSpec(_common.FlyteIdlEntity):
         min_replicas: typing.Optional[int] = None,
         max_replicas: typing.Optional[int] = None,
         ray_start_params: typing.Optional[typing.Dict[str, str]] = None,
+        k8s_pod: typing.Optional[K8sPod] = None,
     ):
         self._group_name = group_name
         self._replicas = replicas
         self._max_replicas = max(replicas, max_replicas) if max_replicas is not None else replicas
         self._min_replicas = min(replicas, min_replicas) if min_replicas is not None else replicas
         self._ray_start_params = ray_start_params
+        self._k8s_pod = k8s_pod
 
     @property
     def group_name(self):
@@ -60,6 +63,14 @@ class WorkerGroupSpec(_common.FlyteIdlEntity):
         """
         return self._ray_start_params
 
+    @property
+    def k8s_pod(self):
+        """
+        Additional pod specs for the worker node pods.
+        :rtype: K8sPod
+        """
+        return self._k8s_pod
+
     def to_flyte_idl(self):
         """
         :rtype: flyteidl.plugins._ray_pb2.WorkerGroupSpec
@@ -70,6 +81,7 @@ class WorkerGroupSpec(_common.FlyteIdlEntity):
             min_replicas=self.min_replicas,
             max_replicas=self.max_replicas,
             ray_start_params=self.ray_start_params,
+            k8s_pod=self.k8s_pod.to_flyte_idl() if self.k8s_pod else None,
         )
 
     @classmethod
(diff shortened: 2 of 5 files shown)


plugins/flytekit-ray/flytekitplugins/ray/models.py
```diff
@@ -1,10 +1,170 @@
...
 
+from flyteidl.core import tasks_pb2 as _core_task
 from flyteidl.plugins import ray_pb2 as _ray_pb2
+from google.protobuf import json_format as _json_format
+from google.protobuf import struct_pb2 as _struct
 
 from flytekit.models import common as _common
 
 
+class K8sObjectMetadata(_common.FlyteIdlEntity):
```
Sovietaced, line 12: I took these from `flytekit/models/task.py` instead of reusing them. I'm happy to just reuse the other models but I figured we might want to remove data loading config since that doesn't make any sense here yet.

plugins/flytekit-ray/flytekitplugins/ray/models.py
```diff
@@ -1,10 +1,170 @@
...
+    def from_flyte_idl(cls, pb2_object: _core_task.IOStrategy):
+        if pb2_object is None:
+            return None
+        return cls(
+            download_mode=pb2_object.download_mode,
+            upload_mode=pb2_object.upload_mode,
+        )
+
+
+class DataLoadingConfig(_common.FlyteIdlEntity):
```
Sovietaced, line 78: Do we think we'd need this for the ray plugin?

plugins/flytekit-ray/flytekitplugins/ray/models.py
```diff
@@ -84,30 +255,42 @@ def from_flyte_idl(cls, proto):
...
+        k8s_pod: typing.Optional[K8sPod] = None,
     ):
         self._ray_start_params = ray_start_params
+        self._k8s_pod = k8s_pod
 
     @property
     def ray_start_params(self):
         """
-        The ray start params of worker node group.
+        The ray start params of head node group.
```
Sovietaced, line 274: lil typo

plugins/flytekit-ray/flytekitplugins/ray/models.py
kumare3: we do not

plugins/flytekit-ray/flytekitplugins/ray/models.py
```diff
@@ -1,10 +1,170 @@
...
+            io_strategy=IOStrategy.from_flyte_idl(pb2.io_strategy) if pb2.HasField("io_strategy") else None,
+        )
+
+
+class K8sPod(_common.FlyteIdlEntity):
+    def __init__(
+        self,
+        metadata: K8sObjectMetadata = None,
+        pod_spec: typing.Dict[str, typing.Any] = None,
+        data_config: typing.Optional[DataLoadingConfig] = None,
```
kumare3, line 129: drop this. Also why add this K8s pod. We dont need that, as Ray config will simply float like a json. So just use K8s pod object?

kumare3: we can keep the k8s pod too, i see that you are actually setting the pod properties. 
This way we could also use pod template?

plugins/flytekit-ray/flytekitplugins/ray/task.py
```diff
@@ -89,7 +92,9 @@ def get_custom(self, settings: SerializationSettings) -> Optional[Dict[str, Any]
         ray_job = RayJob(
             ray_cluster=RayCluster(
                 head_group_spec=(
-                    HeadGroupSpec(cfg.head_node_config.ray_start_params) if cfg.head_node_config else None
+                    HeadGroupSpec(cfg.head_node_config.ray_start_params, cfg.head_node_config.k8s_pod)
```
kumare3, line 95: IMO [this](https://github.com/flyteorg/flyte/blob/f20b8aa082820c56c0c721670b54148afa7d36a4/flyteidl/protos/flyteidl/plugins/ray.proto#L10) is not great
cc @EngHabu @pingsutw 
I would have loved us to model it more like a json so that modifying it would be faster without needing a protobuf change. 

But i see what you are doing now

plugins/flytekit-ray/flytekitplugins/ray/models.py
Sovietaced: Yeah we had a discussion about this on the backend change here: https://github.com/flyteorg/flyte/pull/5933#discussion_r1820096649

It started as just adding support for resources but we realized it would be more flexible if we added support for something similar to a pod template since that is ultimately what the [kuberay contract](https://ray-project.github.io/kuberay/reference/api/#headgroupspec) is.

plugins/flytekit-ray/flytekitplugins/ray/models.py
eapolinario: We're trying to get rid of these model files, so let's try to contain their existence to `flytekit/models`. In this specific case, we don't anticipate that they will diverge (due to the Lindy effect), so reusing the existing model file is the way to go.

plugins/flytekit-ray/flytekitplugins/ray/task.py
kumare3: but i donot want to block. I see that we have backend ray plugin proto already.

flytekit/models/task.py
```diff
@@ -1018,7 +1018,7 @@ def data_config(self) -> typing.Optional[DataLoadingConfig]:
 
     def to_flyte_idl(self) -> _core_task.K8sPod:
         return _core_task.K8sPod(
-            metadata=self._metadata.to_flyte_idl(),
+            metadata=self._metadata.to_flyte_idl() if self.metadata else None,
```
Sovietaced, line 1021: Handles cases where metadata is `None`

plugins/flytekit-ray/flytekitplugins/ray/models.py
Sovietaced: Updated in 185ba394f08fbfe195b2a569a5e312292f41ad3f

plugins/flytekit-ray/flytekitplugins/ray/task.py
Sovietaced: I appreciate it thank you.

eapolinario (approved):

PR #2928, opened 2024-11-13
add recursive cache for private modules

dansola:

See https://github.com/flyteorg/flytekit/pull/2912 for more context.

Make caching easier to use in flytekit by reducing cognitive burden of specifying cache versions.

In this change we add a class to recursively explore the any functions and class methods defined in the users repo and called within a function of interest. The hash of each user defined function or method computed and hashed together to generate a single unique hash.

This covers the following scenarios:
- Nesting of functions in functions in functions, and so on.
- Imports that happen at the global level or within a task/function/method defnition.
- Imports where the module is imported and the function or class is referenced. e.g. `import my_module`, `print(my_module.var)`
- Imports where function or class itself is imported. e.g `from my_module import var`
- If user defined class is used, we recursively hash all the methods.
- If an external library is used (as determined by site-packages), we stop the recursive search and don't hash that external functions contents

## How was this patch tested?

Since `get_version` is already tested in https://github.com/flyteorg/flytekit/pull/2912, we just ensure that the correct callables are extracted for hashing. We set up a mock package and make sure all of the above conditions are satisfied.

Will eventually separate tests out so failures can be more easily diagnosed.


dansola:
cc @eapolinario @cosmicBboy @granthamtaylor

This is the second caching method added so far in the series of draft PRs for auto caching.

PR #2929, opened 2024-11-14
Adds experimental support for uv.lock in ImageSpec

thomasjpfan:

uv and `uv.lock` is becoming a commonly used tool for managing python environments. This PR adds experimental support for `uv.lock` into `ImageSpec`.

This PR enables `requirements="uv.lock"` support in `ImageSpec`. Using `uv.lock` shows a warning stating that `uv.lock` support is experimental since `uv sync` is experimental, so our integration may change.

## How was this patch tested?

```bash
mkdir uv-check
cd uv-check
uv init
uv add flytekit numpy

# activate python environment with this PR
pyflyte run --remote main.py my_task
```

where `main.py` is:

```python

image_spec = ImageSpec(
    name="uv-wow",
    requirements="uv.lock",
    registry="localhost:30000",
    env={"abc": "xyz125"},
)


def my_task() -> int:
    import numpy as np

    return np.asarray([1, 2, 3]).sum().item()
```


flytekit/image_spec/default_builder.py
```diff
@@ -114,26 +128,54 @@ def _is_flytekit(package: str) -> bool:
...
-    """Populate tmp_dir with Dockerfile as specified by the `image_spec`."""
-    base_image = image_spec.base_image or "debian:bookworm-slim"
+def prepare_uv_lock_command(image_spec: ImageSpec, pip_install_args: List[str], tmp_dir: Path) -> str:
+    # uv sync is experimental, so our uv.lock support is also experimental
+    # the parameters we pass into install args could be different
+    warnings.warn("uv.lock support is experimental", UserWarning)
 
-    requirements = []
+    if image_spec.packages is not None:
+        msg = "uv.lock is not compatible with packages, please include all your dependencies in uv.lock"
```
eapolinario, line 137: nit: should this mention that support for `uv.lock` files and packages is mutually exclusive instead?

eapolinario (dismissed):
I wish we had a better way to signal that uv lock files could be used in imagespec that didn't involve overloading the `requirements` parameter. Can't think of anything better though, so let's get this out.

eapolinario:
Can you merge master to fix the errors in CI?

eapolinario (approved):

PR #2942, opened 2024-11-20
Add error message when docker isn't running

dansola:

When a user isn't running docker, it is unclear when they see the logs:
```
Running Execution on Remote.
Failed to check if the image exists with error:
 Error while fetching server API version: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
Flytekit assumes the image ghcr.io/dansola/test-image:ppjeS9H6wavg5FvXbowKEQ already exists.
```

Ideally we'd have more clear logs when we can narrow down that the issue is that docker isn't running.

Added additional logging.

## How was this patch tested?

Did not run docker, ran pyflyte run --remote with an ImangeSpec, observed the log message:
```
Running Execution on Remote.
Unable to connect to Docker daemon. Please ensure Docker is installed and running.
You can start Docker by:
  - Windows/Mac: Start the Docker Desktop application
  - Linux: Run 'sudo systemctl start docker'
Flytekit assumes the image ghcr.io/dansola/test-image:ORI1BZVtVKVludLUJsq_Rg already exists.
```

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -227,7 +227,7 @@ class ImageSpec:
         Return None if failed to check if the image exists due to the permission issue or other reasons.
         """
         import docker
-        from docker.errors import APIError, ImageNotFound
+        from docker.errors import APIError, DockerException, ImageNotFound
 
         try:
             client = docker.from_env()
@@ -277,6 +277,16 @@ class ImageSpec:
                     f"You can upgrade the package by running:\n"
                     f"    pip install --upgrade docker"
                 )
+
+            if isinstance(e, DockerException) and "Error while fetching server API version" in str(e):
+                click.secho(
+                    "Unable to connect to Docker daemon. Please ensure Docker is installed and running.\n"
+                    "You can start Docker by:\n"
+                    "  - Windows/Mac: Start the Docker Desktop application\n"
+                    "  - Linux: Run 'sudo systemctl start docker'",
+                    fg="red",
+                )
+                return None
 
             click.secho(f"Failed to check if the image exists with error:\n {e}", fg="red")
             return None


flytekit/image_spec/image_spec.py
```diff
@@ -278,6 +278,16 @@ def exist(self) -> Optional[bool]:
                     f"    pip install --upgrade docker"
                 )
 
+            if isinstance(e, DockerException) and "Error while fetching server API version" in str(e):
+                click.secho(
```
thomasjpfan, line 282: If the image exists in your `ghcr.io` registry and docker is not running, then that should still work. This message tells a user to install Docker, but it is not strictly required.

dansola: Yup. But I don't think we want to hide the fact that docker should generally be running. If it's not clear, the user will eventually run into images that don't exist in their registry and they will get image pull failures they don't understand. This happened today.

dansola: maybe I should change the "error message" wording in the PR title, but this change doesn't block images being found and used if the docker is off but the image exists

flytekit/image_spec/image_spec.py
```diff
@@ -278,6 +278,16 @@ def exist(self) -> Optional[bool]:
                     f"    pip install --upgrade docker"
                 )
 
+            if isinstance(e, DockerException) and "Error while fetching server API version" in str(e):
```
pingsutw, line 281: nit: I'd like to see the original error still.
```suggestion
            click.secho(f"Failed to check if the image exists with error:\n {e}", fg="red")
            if isinstance(e, DockerException) and "Error while fetching server API version" in str(e):
```

flytekit/image_spec/image_spec.py
thomasjpfan: Let's write out these details in the message. Something like: "Docker is not running and flytekit will assume that the image exists in your registry. If you want to build the image, please install docker: ..."

dansola: So I was going to change the `click.secho(f"Flytekit assumes the image {img_name} already exists.", fg="blue")` in `should_build()`, but that could occur even if docker is running, for example an `APIError` that returns `None`. 

The current suggestion actually mentions that docker isn't running if it catches the exact error docker throws, then the `Flytekit assumes the image ghcr.io/dansola/test-image:ORI1BZVtVKVludLUJsq_Rg already exists.` log still comes up after. 

Do you think it's not clear that those two are associated? Happy to keep it general and just toss in the mention of docker in `should_build()`, but I just don't want to sound that alarm if it isn't the case.

flytekit/image_spec/image_spec.py
dansola: Thanks! I'll make sure the original error is still shown.

PR #2947, opened 2024-11-22
Add support for refresh_token to device flow auth

Sovietaced:
## Tracking issue
Closes flyteorg/flyte#6044

Device flow auth did not support refresh tokens which makes the pyflyte user experience less enjoyable since you may need to login often. 

Passes along the refresh token from the token response to the code which writes it to the keyring. Also updates the device flow authenticator to attempt to refresh the access token.

## How was this patch tested?
Unit tests as well as in our production environment. 

### DIFF
--- a/flytekit/clients/auth/auth_client.py
+++ b/flytekit/clients/auth/auth_client.py
@@ -303,6 +303,7 @@ class AuthorizationClient(metaclass=_SingletonPerEndpoint):
         """
         response_body = auth_token_resp.json()
         refresh_token = None
+        expires_in = None
         id_token = None
         if "access_token" not in response_body:
             raise ValueError('Expected "access_token" in response from oauth server')
--- a/flytekit/clients/auth/authenticator.py
+++ b/flytekit/clients/auth/authenticator.py
@@ -9,7 +9,7 @@ import requests
 
 from . import token_client
 from .auth_client import AuthorizationClient
-from .exceptions import AccessTokenNotFoundError, AuthenticationError
+from .exceptions import AccessTokenNotFoundError, AuthenticationError, AuthenticationPending
 from .keyring import Credentials, KeyringStore
 
 
@@ -272,9 +272,6 @@ class DeviceCodeAuthenticator(Authenticator):
         session: typing.Optional[requests.Session] = None,
     ):
         cfg = cfg_store.get_client_config()
-        self._auth_client = None
-        self._authorization_endpoint = cfg.authorization_endpoint
-        self._redirect_uri = cfg.redirect_uri
         self._audience = audience or cfg.audience
         self._client_id = cfg.client_id
         self._device_auth_endpoint = cfg.device_authorization_endpoint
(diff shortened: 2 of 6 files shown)


Sovietaced:
I think this still needs a little work.

eapolinario:
what's missing, @Sovietaced ?

Sovietaced:
> what's missing, @Sovietaced ?

Need to use the auth client to get a new access token. I'll ping you when I'm done. Working on it now.

flytekit/clients/auth/auth_client.py
```diff
@@ -303,6 +303,7 @@ def _credentials_from_response(self, auth_token_resp) -> Credentials:
         """
         response_body = auth_token_resp.json()
         refresh_token = None
+        expires_in = None
```
Sovietaced, line 306: There was a warning about this possibly not being defined in my editor due to the if statement below.

flytekit/clients/auth/token_client.py
```diff
@@ -119,7 +119,11 @@ def get_token(
         raise AuthenticationError("Status Code ({}) received from IDP: {}".format(response.status_code, response.text))
 
     j = response.json()
-    return j["access_token"], j["expires_in"]
+    refresh_token = None
```
thomasjhuang, line 122: One thing to note is that you can actually receive the refresh token in the `body`. In our internal implementation we add a check like this:
```python
    if refresh_token:
        body["refresh_token"] = refresh_token
```
and then later:
```python
    j = response.json()
    # Check if refresh_token already existed and was sent with the body, or if we are getting it for the first time
    if refresh_token or "refresh_token" in j and j["refresh_token"] is not None:
        return j["access_token"], j["refresh_token"], j["expires_in"]
    else:
        logging.debug("Refresh token was not received in the response, only storing access token")
        return j["access_token"], None, j["expires_in"]
```

Sovietaced: Not sure what you mean, wouldn't that be adding the refresh token into the request payload?

thomasjhuang: So what this would mean is that you fetch the pre-existing refresh token if there is one, alongside the request payload.

flytekit/clients/auth/authenticator.py
```diff
@@ -293,7 +296,36 @@ def __init__(
...
+                session=self._session,
+                refresh_access_token_params={},
+            )
+
     def refresh_credentials(self):
+        self._initialize_auth_client()
+        if self._creds:
+            """We have an access token so lets try to refresh it"""
+            try:
+                self._creds = self._auth_client.refresh_access_token(self._creds)
```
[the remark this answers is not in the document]
Sovietaced: Based on the spec you linked that seems like it should be ok? I validated this on our setup and it worked fine for us and we use `offline_access`.

> Typically this will not be included in the request, and if omitted, the service should issue an access token with the same scope as was previously issued.

flytekit/clients/auth/token_client.py
Sovietaced: I see. I think I'm confused because the code that I modified that uses this currently just passes grant type `DEVICE_CODE`. The code path that I'm using to refresh the token (which I copied from the PKCE auth) just seems to make a request directly to the token API and that has worked for us.

Sovietaced: I think I could reuse the token client if I made that change you suggested, let me try that.

Sovietaced: That seems to work. Added in fe8277b45fdba47c4224b97dad5460cf36af757d

flytekit/clients/auth/authenticator.py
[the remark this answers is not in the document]
thomasjhuang: Ah okay, non-issue then. I do recall encountering issues where not including this scope caused problems, but I wonder if that has to do with Azure AD, which could be different than Okta.

thomasjhuang (approved):
lgtm!

wild-endeavor (approved):
can't repro for some reason, but validated that device flow continues to work.  something wrong with our idp setup i think.  thanks!

PR #2949, opened 2024-11-22
[WIP] allow python interface to get the default inputs from launch plan

KyleShao1016:
## Tracking issue
Closes flyteorg/flyte#6030

Initially, the launch plan only retrieved the interface from the workflow. However, this approach could not get the default inputs of the launch plan. Adding a new argument, python_interface, to the launch plan resolves the issue where inputs could not be processed directly from the launch plan.

1. adding a new parameter python_interface
2. return this parameter when launch plan is instantiate


Mecoli1219 (changes requested):
@KyleShao1016 Three things to update:
1. Sign off your commit 
2. It seems like some of your updates break some unit tests. You can either update the broken unit tests or your code to make sure the original unit tests are passed. You can check it with `make test` command.
3. Create some unit tests related to the bug you fixed. They could be added somewhere like here: https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/core/test_launch_plan.py.

PR #2859, opened 2024-10-24
Type Mismatching while Serializing Dataclass with Union 

mao3267:
## Tracking issue
Closes flyteorg/flyte#5910

## Why are the changes needed?
While converting dataclasses from Python types to literal types through the `to_literal` function, the dataclass will pass through the `_make_dataclass_serializable` function to ensure it can be serialized later. For dataclasses with `Union` properties, under the `if UnionTransformer.is_optional_type(python_type):` statement, it uses `get_args(python_type)[0]` as the type of the property without verifying its compatibility. This will cause an error if the first argument is not the expected type.

## What changes were proposed in this pull request?
1. Do typing matching instead of using the first type from `get_args(python_type)`
2. Add unit tests for more possible union types

## How was this patch tested?
1. Run on local and remote execution

- Union[None, FlyteFile]
`flytefile_serialize.py`
```py
from dataclasses import dataclass
from flytekit import task
from flytekit.types.file import FlyteFile
from typing import Union
from flytekit.image_spec import ImageSpec

flytekit_hash = "f647bd3a1b727082210454c3f5d1e652a29b1083"
flytekit = f"git+https://github.com/mao3267/flytekit.git@{flytekit_hash}"

image_spec = ImageSpec(
    name="serialize-union-1",
    packages=[flytekit],
    builder="default",
    registry="localhost:30000",
    apt_packages=["git", "gh"],
)

@dataclass
class InnerDC:
    ff: Union[None, FlyteFile]

@dataclass
class DC:
    inner_dc: InnerDC

@task(container_image=image_spec)
def t_dc() -> DC:
    return DC(inner_dc=InnerDC(ff="s3://path"))
```
   - Union[None, int, str] 
`serialize_union.py`
```py
from dataclasses import dataclass
from flytekit import task
from typing import Union
from flytekit.image_spec import ImageSpec

flytekit_hash = "f647bd3a1b727082210454c3f5d1e652a29b1083"
flytekit = f"git+https://github.com/mao3267/flytekit.git@{flytekit_hash}"

image_spec = ImageSpec(
    name="serialize-union-1",
    packages=[flytekit],
    builder="default",
    registry="localhost:30000",
    apt_packages=["git", "gh"],
)

@dataclass
class InnerDC:
    ff: Union[None, int, str]

@dataclass
class DC:
    inner_dc: InnerDC

@task(container_image=image_spec)
def t_dc() -> DC:
    return DC(inner_dc=InnerDC(ff="string"))
```
[...]

PR #2952, opened 2024-11-25
[Bug][TypeEngine] Unit tests for dataclass in union with more than two non-None variants

mao3267:
## Tracking issue

Related to flyteorg/flyte#5986

After mashumaro release version `3.15`, it supports messagepack robust decoding with the correct type in Union. Therefore, we would like to add a unit test to show it works. Also, we should fix errors to support changes in mashumaro to version `3.15`.

1. Update the unit test for dataclass in Union to include more than two non-None variants.
2. Fix bugs while generating attribute list from `dataclass_json` in `generate_attribute_list_from_dataclass_json` function.
    - Initially, we thought we should use `additionalProperties` to generate dataclass properties. However, both dataclass and dictionary type can have an `additionalProperties` field. This mismatch causes the bug of generating an incorrect attribute list for dictionaries.

## How was this patch tested?
1. Passing existing and modified unit tests.

### Setup process
```
git clone https://github.com/flytorg/flytekit.git
gh pr checkout 2952
pip install -e .
```
### Screenshots
None

## Related PRs
https://github.com/flyteorg/flytekit/pull/2859
https://github.com/Fatal1ty/mashumaro/pull/256

## Docs link
None

## Note
1. We can not distinguish between dict and dataclass with json schema. So the current implementation will fail to handle dataclass with additionalProperties.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -2332,9 +2332,12 @@ def generate_attribute_list_from_dataclass_json(schema: dict, schema_name: typin
                     )
                 )
             elif property_val.get("additionalProperties"):
-                attribute_list.append(
-                    (property_key, Dict[str, _get_element_type(property_val["additionalProperties"])])  # type: ignore[misc,index]
-                )
+                if property_val["additionalProperties"] == {}:
+                    attribute_list.append((property_key, dict))  # type: ignore
+                else:
+                    attribute_list.append(
+                        (property_key, Dict[str, _get_element_type(property_val["additionalProperties"])])  # type: ignore[misc,index]
+                    )
         # Handle int, float, bool or str
         else:
             attribute_list.append([property_key, _get_element_type(property_val)])  # type: ignore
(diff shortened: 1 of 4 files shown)


flytekit/core/type_engine.py
```diff
@@ -2333,10 +2333,8 @@ def generate_attribute_list_from_dataclass_json(schema: dict, schema_name: typin
                 )
             elif property_val.get("additionalProperties"):
                 attribute_list.append(
-                    (property_key, typing.cast(GenericAlias, _get_element_type(property_val["additionalProperties"]))),
+                    (property_key, Dict[str, _get_element_type(property_val["additionalProperties"])])  # type: ignore[misc,index]
                 )
-            else:
-                attribute_list.append((property_key, Dict[str, _get_element_type(property_val)]))  # type: ignore[misc,index]
         # Handle int, float, bool or str
```
Future-Outlier, line 2338: Can you explain more why they can be merged together to `Dict[str, _get_element_type(property_val["additionalProperties"])])`?

mao3267: Both `dict` and `typing.Dict` include the `additionalProperties` field. The key difference is that the `additionalProperties` value from `dict` is an empty dictionary, whereas the one from `typing.Dict` contains type information. I plan to make a change to address this distinction.

Future-Outlier (approved):
You are amazing, thank you so much Vincent.

PR #2963, opened 2024-11-27
Always print traceback

pingsutw:
## Tracking issue
NA

It’s difficult to identify which function failed when encountering a gRPC error

Always print the traceback after we get a grpc error.

## How was this patch tested?
Raise a grpc error in the create_execution method in raw.py


wild-endeavor (commented):
let me test this next time i see an issue

PR #2974, opened 2024-12-04
Add cache key metadata

eapolinario:
## Tracking issue
https://linear.app/unionai/issue/DX-1202/add-cacke-key-attributes-to-typeannotation

This is the flytekit PR mentioned in https://github.com/flyteorg/flyte/pull/6078.

We introduce a top-level dictionary to `TypeAnnotation` under a special key called `cache-key-metadata`. 

The three transformers that produce msgpack-binary idl literals (i.e. `DictTransformer`, `DataclassTransformer`, and `PydanticTransformer`) will now emit a `TypeAnnotation` that contains that special key.

## How was this patch tested?
unit tests and sandbox.

### DIFF
--- a/flytekit/models/annotation.py
+++ b/flytekit/models/annotation.py
@@ -42,6 +42,19 @@ class TypeAnnotation:
 
         return cls(annotations=_json_format.MessageToDict(proto.annotations))
 
+    @classmethod
+    def merge_annotations(cls, annotation: "TypeAnnotation", other_annotation: "TypeAnnotation") -> "TypeAnnotation":
+        """
+        Merges two annotations together. If the same key exists in both annotations, the value in the other annotation
+        will be used.
+        :param TypeAnnotation annotation: The first annotation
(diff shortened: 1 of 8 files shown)


flytekit/models/annotation.py
```diff
@@ -42,6 +42,19 @@ def from_flyte_idl(cls, proto):
 
         return cls(annotations=_json_format.MessageToDict(proto.annotations))
 
+    @classmethod
+    def merge_annotations(cls, annotation: "TypeAnnotation", other_annotation: "TypeAnnotation") -> "TypeAnnotation":
```
wild-endeavor, line 46: let's move this out of the model in the future... keep the models as easy to delete as possible.

wild-endeavor (approved):

flytekit/models/annotation.py
eapolinario: good call.

PR #2154, opened 2024-01-31
Support overridding cache configuration

hamersaw:
## Tracking issue
_NA_

## Why are the changes needed?
Enables cache overrides.

## What changes were proposed in this pull request?
^^^

## How was this patch tested?
Locally.

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Related PRs
_NA_

## Docs link
_NA_

PR #2975, opened 2024-12-04
Fix Bug: with_overrides for cache doesn't work

hebiao064:

I tried the with_overrides to override cache configuration for a local flyte @task, but it didn't work as expected.

Code:
```python
@task(
    cache=False,
    cache_version="v1",
    cache_serialize=False,
)
def to_be_overridden_task(a: str) -> str:
    return f"*~*~*~{a}*~*~*~"

def to_be_overridden_task_wrapper(a: str, cache: bool = False, cache_version: str = "", cache_serialize: bool = False) -> str:
    return to_be_overridden_task(a=a).with_overrides(
        cache=cache,
        cache_version=cache_version,
        cache_serialize=cache_serialize
    )

@workflow
def my_overridden_wf(a: str):
    return to_be_overridden_task_wrapper(a=a, cache=True, cache_version="foo", cache_serialize=True)
```

After pyflyte run, I found the `cache` is still `False` and `cache_version` still `""`

1. When cache overrides specified, we also override the flyteentity's metadata
2. Reorder the serialization order to serialize workflow first hence with_overrides will be called first, otherwise the task will be always registered as original task configuration without overrides.

## How was this patch tested?
make test with a new test case

Also tested in our company's fork E2E with many workflows submission.

### Screenshots
Before:
Screenshot 2024-12-03 at 5 57 29 PM
After:
Screenshot 2024-12-03 at 5 57 22 PM

## Related PRs

https://github.com/flyteorg/flytekit/pull/2154


flytekit/core/node.py
```diff
@@ -194,6 +194,8 @@ def with_overrides(
 
         if name is not None:
             self._metadata._name = name
+            if self.run_entity and hasattr(self.run_entity, "metadata"):
```
pingsutw, line 197: We should not override the run_entity.

pingsutw: Suppose you have two tasks in a workflow, like below. The cache version of the second task will be overridden as well.
```
@workflow
def wf():
  t1().with_override(cache_version="v2")
  t1()
```

pingsutw: You can pass the node info to `local_execute`, and use the cache version in the node info to retrieve the data from the cache.
https://github.com/flyteorg/flytekit/blob/935be5c28dccbf860f7188357edc3bba9c7c70f1/flytekit/core/base_task.py#L306

hebiao064: @pingsutw Could I know how to do this?

"You can pass the node info to local_execute`

hebiao064: I think the problem is: 

In `node.py`, `self._metadata._cacheable` won't be affecting the actual Task entity.

That's why we want to modify the Task's entity's metadata.

But I acknowledge the issue that the two tasks example woulld be problematic. I am not sure how `local_config` can help with it.

PR #2982, opened 2024-12-05
Support with_overrides setting metadata for map_task subnode instead of parent node

pvditt:
## Tracking issue
fixes: https://linear.app/unionai/issue/COR-2498/with-overrides-sets-metadata-for-parent-instead-of-subnode-for-map

with_overrides doesn't work for map_tasks 

create a new field in array node map task to explicitl

## How was this patch tested?
- added unit tests
- tested workflows locally that overrode timeouts, cache, container image, etc

### Setup process

```
from flytekit import task, workflow, map_task, reference_launch_plan, reference_task, ImageSpec, LaunchPlan, Resources, TaskMetadata
from datetime import timedelta


@task(
    cache=True,
    cache_version="v1.0",
    timeout=timedelta(seconds=20),
    container_image="pvditt/flytekit:15",
    interruptible=False,
    retries=12,
)
def basic_cache(num: str) -> str:
    return f"Hi: {num[0]} ugh-1"


@workflow()
def map_metadata_wf() -> list[str]:
    a = ["1", "2", "3", "4", "5"]
    return map_task(basic_cache)(num=a).with_overrides(
        cache_version="v1.1", 
        timeout=timedelta(seconds=10), 
        container_image="pvditt/flytekit:20",
        interruptible=True,
        retries=10,
    )

```

### Screenshots

image

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
Fixed a critical bug in map_tasks where metadata was incorrectly being set on parent nodes instead of subnodes. Implemented dedicated sub_node_metadata field and refactored override logic for proper application of settings (timeout, cache, container image) to mapped subtasks. Simplified metadata access by removing getter method and using direct property access while maintaining backward compatibility.
<br>
<br>
<br>
<br>
</div>


pvditt:
noticed a separate issue: https://github.com/flyteorg/flyte/issues/6153 - will get a quick fix for this after this is merged

flytekit/core/array_node_map_task.py
```diff
@@ -137,15 +140,15 @@ def python_interface(self):
...
         return NodeMetadata(
             name=self.name,
         )
 
-    def construct_sub_node_metadata(self) -> NodeMetadata:
-        nm = super().construct_node_metadata()
-        nm._name = self.name
-        return nm
+    def get_sub_node_metadata(self) -> NodeMetadata:
+        return self.sub_node_metadata
```
thomasjpfan, line 151: The pattern in flytekit is to use a property + a private attribute:

```python
@property
def sub_node_metadata(self) -> NodeMetadata:
    return self._sub_node_metadata
```

thomasjpfan (approved):
Minor nit, otherwise LGTM

flytekit/core/array_node_map_task.py
pvditt: thanks for pointing this out. just updated

thomasjpfan (approved):

PR #2983, opened 2024-12-05
Move digest computation

thomasjpfan:

During fast registration, the digest computation happens even if one of the branches do not use it.

This PR moves the digest computation to the branch that uses it.

## How was this patch tested?

The current unit tests should cover this, since the behavior has not changed.


wild-endeavor:
delete line 172 🙏

wild-endeavor (approved):

PR #2984, opened 2024-12-05
Follow up to 2980 - Do not copy socket files

eapolinario:
Follow up to https://github.com/flyteorg/flytekit/pull/2980, but this time we not only skip socket files from the digest calculation but we also do not copy them to the tar file.

Skip Unix Domain Files

## How was this patch tested?
Unit tests and local sandbox. Pass `--copy all` to `pyflyte run` and verified that it succeeds:
```
❯ pyflyte run --remote --copy all workflows/example.py wf
```

### DIFF
--- a/flytekit/tools/script_mode.py
+++ b/flytekit/tools/script_mode.py
@@ -5,6 +5,7 @@ import hashlib
 import os
 import shutil
 import site
+import stat
 import sys
 import tarfile
 import tempfile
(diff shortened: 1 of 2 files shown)


flytekit/tools/script_mode.py
```diff
@@ -172,6 +173,10 @@ def list_all_files(source_path: str, deref_symlinks, ignore_group: Optional[Igno
             if ignore_group:
                 if ignore_group.is_ignored(abspath):
                     continue
+            # Skip socket files
```
wild-endeavor, line 176: can we move this above the ignore?

eapolinario: done.

wild-endeavor (approved):

PR #2985, opened 2024-12-06
Use `sys.base_prefix` and `sys.prefix` to filter out modules in fast registry

thomasjpfan:

The currently filter is not good enough because it includes files python's `lib` directory.

In this PR, we use `sys.prefix` to filter out everything in the python environment. For virtual envs, `sys.base_prefix` points to the base python environment (The one the virtual environment was created from)

## How was this patch tested?
A unit test was added check the behavior of `list_imported_modules_as_files`.

### DIFF
--- a/flytekit/tools/script_mode.py
+++ b/flytekit/tools/script_mode.py
@@ -207,7 +207,7 @@ def list_imported_modules_as_files(source_path: str, modules: List[ModuleType]) 
     files = []
     flytekit_root = os.path.dirname(flytekit.__file__)
 
-    invalid_directories = [flytekit_root, sys.prefix, sys.base_exec_prefix] + site_packages
+    invalid_directories = [flytekit_root, sys.prefix, sys.base_prefix] + site_packages
 
     for mod in modules:
         try:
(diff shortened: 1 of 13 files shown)


flytekit/tools/script_mode.py
```diff
@@ -183,6 +183,15 @@ def list_all_files(source_path: str, deref_symlinks, ignore_group: Optional[Igno
     return all_files
 
 
+def _file_is_in_directory(file: str, directory: str) -> bool:
+    """Return True if file is in directory and in and of it's children."""
```
eapolinario, line 187: nit: typo: it's -> its

flytekit/tools/script_mode.py
```diff
@@ -183,6 +183,15 @@ def list_all_files(source_path: str, deref_symlinks, ignore_group: Optional[Igno
...
+    try:
+        return os.path.commonpath([file, directory]) == directory
+    except ValueError:
+        # ValueError is raised by windows if the paths are not from the same drive
```
eapolinario, line 191: can we add a `logger.debug` to this case?

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #2994, opened 2024-12-10
vscode decorator for the dynamic task

pingsutw:
## Tracking issue
NA

To support `vscode` decorator for the dynamic tasks.

Set the correct resolver in the current context, so flytekit can compile the vscode task

## How was this patch tested?
```python
from flytekit import task, workflow, ImageSpec, Resources, dynamic
from flytekit.interactive import vscode

new_flytekit = "git+https://github.com/flyteorg/flytekit@cedb8188f6fa86481064fb9022b16c9f5d369f4d"
image_spec = ImageSpec(registry="ghcr.io/unionai-oss",
                       packages=[new_flytekit, "kubernetes", "mypy"],
                       apt_packages=["git"])


@task(container_image=image_spec, enable_deck=True, requests=Resources(cpu="2", mem="2000Mi"))
def train():
    print("forward", flush=True)
    print("backwarddddf")


@dynamic(container_image=image_spec)
@vscode
def d1():
    print("dynamic", flush=True)
    train()


@workflow
def wf():
    train()
    d1(
```

### Screenshots
Screenshot 2024-12-10 at 12 48 29 PM


flytekit/core/context_manager.py
```diff
@@ -735,7 +735,9 @@ def new_compilation_state(self, prefix: str = "") -> CompilationState:
         Creates and returns a default compilation state. For most of the code this should be the entrypoint
         of compilation, otherwise the code should always uses - with_compilation_state
         """
-        return CompilationState(prefix=prefix)
+        from flytekit.core.python_auto_container import default_task_resolver
+
+        return CompilationState(prefix=prefix, task_resolver=default_task_resolver)
```
eapolinario, line 740: why do we need this if the correct task resolver is being set in `compile` in flytekit/core/workflow.py?

pingsutw: I set the task resolver in the `ctx.with_compilation_state(...)`, but there are some places where `new_compilation_state()` get called, potentially causing the loss of the resolver if it's not set.

PR #2998, opened 2024-12-12
Remove open_async usage in put raw data

wild-endeavor:
Syntax was always wrong.  This is not how open_async is used.

See https://github.com/fsspec/s3fs/issues/871, which may be resolved [here](https://github.com/fsspec/s3fs/pull/885).

### Polars
This also affected polars.  In testing, noticed part of the logic was incorrect - updated and added a test to exercise.

Replace with a normal call - this means that writes for bytes and bytes-like objects will be blocking.  Given this seems to be fundamentally missing from fsspec at this time I think it's okay.

## How was this patch tested?
Unit tested and local sandbox.

### DIFF
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -199,8 +199,7 @@ class FileAccessProvider(object):
         elif protocol == "s3":
             s3kwargs = s3_setup_args(self._data_config.s3, anonymous=anonymous)
             s3kwargs.update(kwargs)
-            fs = fsspec.filesystem(protocol, **s3kwargs)  # type: ignore
-            return fs
+            return fsspec.filesystem(protocol, **s3kwargs)  # type: ignore
         elif protocol == "gs":
             if anonymous:
                 kwargs["token"] = _ANON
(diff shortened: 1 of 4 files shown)


eapolinario (commented):
This looks reasonable. 

I don't understand the integration test failures. This change shouldn't have any impact in the minio credentials. Any idea what's going on there?

wild-endeavor:
nope...  `botocore.exceptions.NoCredentialsError: Unable to locate credentials`  i don't see how that's related to anything at all.

flytekit/core/data_persistence.py
```diff
@@ -425,45 +424,30 @@ async def async_put_raw_data(
...
         if isinstance(lpath, bytes):
-            fs = await self.get_async_filesystem_for_path(to_path)
-            if isinstance(fs, AsyncFileSystem):
-                async with fs.open_async(to_path, "wb", **kwargs) as s:
-                    s.write(lpath)
-            else:
-                with fs.open(to_path, "wb", **kwargs) as s:
-                    s.write(lpath)
-
+            fs = self.get_filesystem_for_path(to_path)
```
eapolinario, line 427: Can you leave a comment and mention the fsspec gh issue?

eapolinario (dismissed):

eapolinario (approved):

PR #3003, opened 2024-12-14
WIP - Lazy Evaluation

pingsutw:
## Tracking issue
NA

## How was this patch tested?
```python


@task(container_image="cr.flyte.org/flyteorg/flytekit:py3.10-1.13.0", retries=5, enable_deck=True)
def say_hello() -> str:
    return "Hello, World!"


@workflow
def wf():
    t1 = say_hello()
    t2 = say_hello().with_overrides(cache=True)
    print("t1: ", t1)
    print("t2: ", t2)
```

### DIFF
--- a/flytekit/core/base_task.py
+++ b/flytekit/core/base_task.py
@@ -193,6 +193,94 @@ class IgnoreOutputs(Exception):
     pass
 
 
+class Run(object):
+    def __init__(self, ctx, task, inputs):
+        self._ctx = ctx
+        self._task = task
+        self._inputs = inputs
+        self._has_run: bool = False
+        self._outputs: "Literal" = None
+
+    @property
+    def outputs(self) -> "Literal":
+        return self._outputs
+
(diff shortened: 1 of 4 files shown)


flytekit/core/base_task.py
```diff
@@ -193,6 +193,94 @@ class IgnoreOutputs(Exception):
...
+            )
+        except TypeTransformerFailedError as exc:
+            exc.args = (f"Failed to convert inputs of task '{task.name}':\n  {exc.args[0]}",)
+            raise
+
+        input_literal_map = _literal_models.LiteralMap(literals=literals)
+
+        local_config = LocalConfig.auto()
+        # TODO: use node metadata here
+        print("node_metadata", node_metadata)
```
pingsutw, line 227: use node metadata here

eapolinario:
Need to handle the case of the "dangling" node, i.e. how to force the evaluation of nodes that are not an upstream dependency of the graph.

PR #3022, opened 2024-12-26
Store protos in local cache

eapolinario:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/6125

Values stored in the local cache in flytekit<1.13.6 cannot be used in flytekit>=1.13.6.

This PR changes the values stored in the local cache to be the serialized bytes of the Literal proto message instead of the model file Literal object. This way if we end up adding new fields to the `Literal` proto message we are protected from the error described in the github issue being fixed by this PR.

Notice that this change makes values produced by the new versions of flytekit to be backwards incompatible (i.e. values written to the local cache will not be read by older versions of flytekit).

## How was this patch tested?
Unit tests and local runs across a diverse set of versions.


flytekit/models/literals.py
```diff
@@ -979,7 +979,15 @@ def offloaded_metadata(self) -> Optional[LiteralOffloadedMetadata]:
...
         This value holds metadata about the offloaded literal.
         """
-        return self._offloaded_metadata
+        # The following check might seem non-sensical, since `_offloaded_metadata` is set in the constructor.
+        # This is here to support backwards compatibility caused by the local cache implementation. Let me explain.
+        # The local cache pickles values and unpickles them. When unpickling, the constructor is not called, so there
+        # are cases where the `_offloaded_metadata` is not set (for example if you cache a value using flytekit<=1.13.6
+        # and you load that value later using flytekit>1.13.6).
+        # In other words, this is a workaround to support backwards compatibility with the local cache.
+        if hasattr(self, "_offloaded_metadata"):
```
pingsutw, line 988: > if isinstance(serialized_obj, ModelLiteralMap):
            return ModelLiteralMap.from_flyte_idl(ModelLiteralMap.to_flyte_idl(serialized_obj))

Q: we double deserialize the cached value. Won't it have `_offloaded_metadata` in the literal map?

pingsutw (approved):

flytekit/models/literals.py
eapolinario: We need this because `ModelLiteralMap.to_flyte_idl` uses the `offloaded_metadata` property [directly](https://github.com/flyteorg/flytekit/blob/cc9f96171a08339b9f5f63e89ddb2079f18fab8e/flytekit/models/literals.py#L1002). The idea is that by forcing it to go through `ModelLiteralMap.from_flyte_idl` we invoke the constructor which sets the [`_offloaded_metadata` field](https://github.com/flyteorg/flytekit/blob/master/flytekit/models/literals.py#L1013-L1015) (which bulletproofs the literal model file from similar changes going forward).

pingsutw: I see, thank you

PR #3028, opened 2025-01-02
feat: add options to output registered entity summary

paullongtan:


flytekit/tools/repo.py
```diff
@@ -350,30 +362,60 @@ def _raw_register(cp_entity: FlyteControlPlaneEntity):
...
+        supported_format = ["json", "yaml"]
+        if summary_format not in supported_format:
+            raise ValueError(f"Unsupported file format: {summary_format}")
+
+        if summary_dir is not None:
+            os.makedirs(summary_dir, exist_ok=True)
+        else:
+            summary_dir = os.getcwd()
+
+        summary_file = f"registration_summary.{summary_format}"
```
Mecoli1219, line 411: Does this mean that the other registration may overwrite the summary? I think we could name the file with other unique names (ie. tmp file, py file name, wf name, version number)

tests/flytekit/unit/cli/pyflyte/test_register.py
```diff
@@ -163,3 +164,75 @@ def test_non_fast_register_require_version(mock_client, mock_remote):
...
+    with runner.isolated_filesystem():
+        out = subprocess.run(["git", "init"], capture_output=True)
+        assert out.returncode == 0
+        os.makedirs("core4", exist_ok=True)
+        with open(os.path.join("core4", "sample.py"), "w") as f:
+            f.write(sample_file_contents)
+            f.close()
+        result = runner.invoke(pyflyte.main, ["register", "--summary-dir", "summaries", "core4"])
+        assert result.exit_code == 2
+        print(result.output)
```
Mecoli1219, line 188:
```suggestion

```

flytekit/tools/repo.py
```diff
@@ -350,30 +362,60 @@ def _raw_register(cp_entity: FlyteControlPlaneEntity):
                     print_registration_status(
                         i, console_url=console_url, verbosity=verbosity, activation=print_activation_message
                     )
+                    result["status"] = "success"
 
                 except Exception as e:
                     if not skip_errors:
                         raise e
                     print_registration_status(og_id, success=False)
+                    result["status"] = "failed"
```
Mecoli1219, line 371: If it fails, what will the values of other keys be?

Mecoli1219 (commented):
Nice first PR! I think you can add more unit tests.
1. yaml
2. testing registration without specifying `--summary-dir`

tests/flytekit/unit/cli/pyflyte/test_register.py
paullongtan: Done.

paullongtan:
> Nice first PR! I think you can add more unit tests.
> 
> 1. yaml
> 2. testing registration without specifying `--summary-dir`

Thanks! 
I have added unit tests for both yaml and registration without specifying --summary-dir.

flytekit/tools/repo.py
paullongtan: The values of the other keys(id, type, version) are pre-computed before registration. Thus, the values will not be empty even if the registration fails.

The values are from:

result = {
            "id": og_id.name,
            "type": og_id.resource_type_name(),
            "version": og_id.version,
            "status": "skipped",  # default status
        }
where og_id is the id of the entity's template / entity itself.

flytekit/tools/repo.py
paullongtan: Yes! The summary will be overwritten if the registration is rerun.

paullongtan:
New updates:
1. Added option `--quiet` to pyflyte register to mute all output to CLI
2. Updated registration summary to output directly to the CLI instead of saving to a file to ensure scalability
  - uses the --quiet flag to ensure proper output format
  - drops option `--summary-dir` for pyflyte register

Note:
1. I have temporarily commented out the log for config file overwriting in `pyflyte.py` and log for packages used in `validate_package` function in `utils.py`. This is to prevent these logs from breaking the CLI output when outputting registration summary to the CLI and ensure no CLI outputs when using the `--quiet` flag for register.

PR #3047, opened 2025-01-10
adjust COPY behavior for files and directories to match desired struc…

samhita-alla:

This update modifies the COPY logic to ensure files and directories are copied directly to /root/ without preserving the entire source path structure.

- Ensures files are copied directly to /root/ without preserving the full source path structure.
- If a directory is copied, only its contents are placed in /root/ (instead of copying the directory itself).

Generated COPY commands:

```
COPY --chown=flytekit services/AgentService/main.py /root/
COPY --chown=flytekit common/utils.py /root/
COPY --chown=flytekit configs/ /root/configs/
COPY --chown=flytekit services/configs /root/configs/
```

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR optimizes Docker image building by simplifying COPY command behavior in the image specification builder. The changes streamline how files are copied directly to the /root/ directory, eliminating unnecessary path nesting and ensuring a cleaner directory structure. The modification prevents full source path preservation, making Docker images more maintainable. A security concern was identified regarding potential path traversal issues in the implementation.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/flytekit/image_spec/default_builder.py
+++ b/flytekit/image_spec/default_builder.py
@@ -332,7 +332,7 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
             if src_path.is_absolute() or ".." in src_path.parts:
                 raise ValueError("Absolute paths or paths with '..' are not allowed in COPY command.")
 
-            dst_path = tmp_dir / src_path.name  # Use only the file or directory name for the destination
+            dst_path = tmp_dir / src_path
             dst_path.parent.mkdir(parents=True, exist_ok=True)
 
             if src_path.is_dir():
--- a/flytekit/image_spec/default_builder.py
+++ b/flytekit/image_spec/default_builder.py
@@ -332,15 +332,15 @@ def create_docker_context(image_spec: ImageSpec, tmp_dir: Path):
             if src_path.is_absolute() or ".." in src_path.parts:
                 raise ValueError("Absolute paths or paths with '..' are not allowed in COPY command.")
 
-            dst_path = tmp_dir / src_path
+            dst_path = tmp_dir / src_path.name
             dst_path.parent.mkdir(parents=True, exist_ok=True)
 
             if src_path.is_dir():
                 shutil.copytree(src_path, dst_path, dirs_exist_ok=True)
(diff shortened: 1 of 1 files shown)


thomasjpfan (commented):
I'm unsure about doing this. For example, this would work locally but not remotely if the structure is not preserved:

```python
MY_LOCAL_FILE = "my_dir/data/my_file.txt"

image = ImageSpec(
    copy=[MY_LOCAL_FILE]
)

@task(container_image=image)
def look_at_file() -> str:
    with open(MY_LOCAL_FILE, "r") as f:
        return f.read()
```

The other issue is with files with the same name, but in different directories:

```python
image = ImageSpec(
    copy=["dir1/data/my_file.txt", "dir2/my_file.txt"]
)
```

samhita-alla:
@thomasjpfan how about we let the user specify the destination path?

thomasjpfan:
> how about we let the user specify the destination path? i can modify the code accordingly.

I'm okay with this, but I want to keep support for `list[str]`. Setting the destintation can end up being very verbose when setting many files.

> the file will be copied to /root, so the path must be: with open(f"/root/{MY_LOCAL_FILE}", "r") as f:

I just tried this:

```python

MY_FILE = "hello.txt"

image = ImageSpec(copy=[MY_FILE], registry="localhost:30000")


@task(container_image=image)
def read() -> str:
    with open(MY_FILE) as f:
        return f.read()
```

where `hello.txt` is next to my workflow and it works.

> i also think we shouldn't set /root as the default because, if the base image has a working directory like /app, we would need to copy the files or directories there instead, right?

For the image builder, it hard codes the workdir as `/root`. So even if the base image is `/app`, the image builder changes it to `/root`. The question becomes, "should we do this"? My concern is when the base image working directory is `/`, in that case, I rather change it to `/root`.

samhita-alla:
> where hello.txt is next to my workflow and it works.

hmm.. could be because the working directory is set to root by default.

> My concern is when the base image working directory is /, in that case, I rather change it to /root.

since image spec doesn't support setting a custom workdir, i agree that we need to default to `/root`.

i’ll close this issue for now. i think specifying destination paths is something we can plan for down the road.

PR #3049, opened 2025-01-11
Prevent register_launch_plan from re-registering already registered workflow

wild-endeavor:
## Tracking issue
Linking this to https://github.com/flyteorg/flyte/issues/6062 which is related.

Please see [comment](https://github.com/flyteorg/flyte/issues/6062) in the linked issue, copied here:

Assume in the past `my_wf` has already been registered with version `ksD2LBYDu5PrNtNnnpd95Q`

Now if you try to do:
```
from wfs import my_wf

lp = LaunchPlan.get_or_create(workflow=my_wf, name="other_lp_for_hello2", default_inputs={})
remote = FlyteRemote(...)
remote.register_launch_plan(entity=lp, version="ksD2LBYDu5PrNtNnnpd95Q")
...
```

The issue is that `register_launch_plan` will currently try to register the underlying workflow again.  This would be fine, except that the serialization settings that are used may be completely different.  Even if it's exactly the same, this is a waste of time because the best case scenario is that Admin returns already exists for the workflow and every underlying task (and subwf, etc.).

We should check first to see if the workflow already exists with that version, and only register the launch plan if the workflow already exists.

Skipping re-registration of the workflow does mean that users who were relying on this re-registration (and were somehow reconstructing serialization settings to be identical) to detect changes in workflow/task structure will no longer see errors, but that seems too far-fetched.

* Check if the underlying workflow exists, and if so only register the launch plan.
* Also pass through an argument (`create_default_launchplan`) in `_serialize_and_register` which for some reason wasn't used.  Just a nit, default was always just `True`.

## How was this patch tested?
Tested with local sandbox.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR streamlines the launch plan registration process by implementing checks to prevent redundant workflow registrations. The changes optimize the registration flow by adding existence verification and proper handling of the create_default_launchplan parameter. The implementation focuses on efficiency and maintainability through improved registration logic.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -1316,7 +1316,7 @@ class FlyteRemote(object):
                 launch_plan_model, serialization_settings, version, create_default_launchplan=False
             )
             if ident is None:
-                raise ValueError("Failed to register launch plan")
+                raise ValueError("Failed to register launch plan, identifier returned was empty...")
         else:
             # Register the launch and everything under it
             ident = run_sync(
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -870,7 +870,12 @@ class FlyteRemote(object):
                 loop.run_in_executor(
                     None,
                     functools.partial(
-                        self.raw_register, cp_entity, serialization_settings, version, og_entity=task_entity
+                        self.raw_register,
+                        cp_entity,
+                        serialization_settings,
+                        version,
+                        create_default_launchplan=create_default_launchplan,
+                        og_entity=task_entity,
                     ),
                 )
             )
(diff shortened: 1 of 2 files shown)


flytekit/remote/remote.py
```diff
@@ -1271,16 +1299,35 @@ def register_launch_plan(
...
+        ):
+            # Underlying workflow, exists, only register the launch plan itself
+            launch_plan_model = get_serializable(
+                OrderedDict(), settings=serialization_settings, entity=entity, options=options
+            )
+            ident = self.raw_register(
+                launch_plan_model, serialization_settings, version, create_default_launchplan=False
+            )
+            if ident is None:
+                raise ValueError("Failed to register launch plan")
```
pingsutw, line 1319: Should we raise this error in `self.raw_register`? Also, are we able to add the error message here? Something like `Failed to register launch plan with err: {err}`

pingsutw (dismissed):

flytekit/remote/remote.py
wild-endeavor: i say we keep this here until we can make the raw_register signature not an optional.

pingsutw (approved):

PR #3057, opened 2025-01-14
Fix: Always propagate pytorch task worker process exception timestamp to task exception

fg91:

Since RFC https://github.com/flyteorg/flyte/pull/5598, flytepropeller can collect error files from multiple worker pods of a distributed task in order to determine the root cause error by identifying the exception with the earliest timestamp. Before, only a single error file was used even if a task launched multiple pods that each could produce an error file.

Generally, the timestamp of the container error is measured [here](https://github.com/flyteorg/flytekit/blob/a2bcf95048962b77ba2e67052b8f4a8c956fe81f/flytekit/bin/entrypoint.py#L382) in the pod entrypoint.

In the case of the pytorch elastic plugin where each worker pod can launch multiple worker processes, [torch elastic launch determines the worker process exception with the earliest timestamp](https://github.com/flyteorg/flytekit/blob/a2bcf95048962b77ba2e67052b8f4a8c956fe81f/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L472). The earliest worker process timestamp determined by torch elastic launch is considered the timestamp with which the entire pytorch elastic task pod failed [here](https://github.com/flyteorg/flytekit/blob/a2bcf95048962b77ba2e67052b8f4a8c956fe81f/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L476).

However, we only propagate the timestamp of the worker process exception to the task exception in the case of [recoverable errors](https://github.com/flyteorg/flytekit/blob/a2bcf95048962b77ba2e67052b8f4a8c956fe81f/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L476). In the case of[ non-recoverable errors](https://github.com/flyteorg/flytekit/blob/a2bcf95048962b77ba2e67052b8f4a8c956fe81f/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L478) we don't propagate the timestamp and instead rely on a new timestamp being measured in the pod entrypoint (see above).

I observed that sometimes the timestamps measured in the pod entrypoint are roughly the same even though the timestamps in the worker processes in the different pods definitely differed. This suggests that in these cases torch distributed still manages to synchronize after one worker experienced an exception so that the pod entrypoints execute [`get_container_error_timestamp(...)`](https://github.com/flyteorg/flytekit/blob/a2bcf95048962b77ba2e67052b8f4a8c956fe81f/flytekit/bin/entrypoint.py#L363) roughly at the same time. The fix is obvious: propagate the worker process exception timestamp to the task exception also for non-recoverable errors.

Added unit tests.

## Related PRs

https://github.com/flyteorg/flytekit/pull/2797 
 <div id='description'>
<h3>Summary by Bito</h3>
Updates exception handling in PyTorch elastic task tests to use FlyteUserRuntimeException instead of RuntimeError. Improves worker process exception timestamp propagation for both recoverable and non-recoverable scenarios. Enhances test expectations and imports required exception classes.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/flytekit/exceptions/user.py
+++ b/flytekit/exceptions/user.py
@@ -15,14 +15,15 @@ class FlyteUserException(_FlyteException):
 class FlyteUserRuntimeException(_FlyteException):
     _ERROR_CODE = "USER:RuntimeError"
 
-    def __init__(self, exc_value: Exception):
+    def __init__(self, exc_value: Exception, timestamp: typing.Optional[float] = None):
         """
         FlyteUserRuntimeException is thrown when a user code raises an exception.
 
         :param exc_value: The exception that was raised from user code.
+        :param timestamp: The timestamp as fractional seconds since epoch when the exception was raised.
         """
         self._exc_value = exc_value
-        super().__init__(str(exc_value))
+        super().__init__(str(exc_value), timestamp=timestamp)
 
     @property
     def value(self):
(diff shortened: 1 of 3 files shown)


thomasjpfan (approved):
LGTM

PR #3076, opened 2025-01-22
Improve Union Transformer Ambiguous Error Message

Future-Outlier:

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1887,6 +1887,7 @@ class UnionTransformer(AsyncTypeTransformer[T]):
     ) -> typing.Union[Literal, asyncio.Future]:
         python_type = get_underlying_type(python_type)
 
+        potential_types = []
         found_res = False
         is_ambiguous = False
         res = None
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
```diff
@@ -1906,14 +1907,15 @@ async def async_to_literal(
...
+                potential_types.append(t)
             except Exception as e:
                 logger.warning(
                     f"UnionTransformer failed attempt to convert from {python_val} to {t} error: {e}",
                 )
                 continue
 
         if is_ambiguous:
-            raise TypeError("Ambiguous choice of variant for union type")
+            raise TypeError(f"Ambiguous choice of variant for union type.\n" f"Potential types: {potential_types}\n")
```
thomasjpfan, line 1918: Can we add more to this error message stating why they are ambiguous? Something like:

> These types are structurally the same, because it's attributes have the same names and associated types.

Otherwise, it's hard to tell from the error what is wrong.

thomasjpfan (commented):
For your example:

```python
@dataclass
class A:
    a: int

@dataclass
class B:
    a: int
```

what should we recommend the user to make it work?

flytekit/core/type_engine.py
Future-Outlier: love u thomas, it's upadted

pingsutw (approved):

PR #3080, opened 2025-01-23
Fix async coroutine limit not respected and add s3/gcs chunk size

wild-endeavor:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6188

Please see issue.

Because flytekit loops run on a different thread, the moment the asyncio.Task is created, it starts running, rendering any batching of coroutines moot.

In testing, we also noticed that the s3fs and gcsfs default write chunk sizes were 50MB.  We felt that was a bit high and in testing noticed no performance decrease (actually saw possibly non-statistically significant increase) in performance.  We are therefore changing it by default to 25MB.  Also this setting and the number of coroutine batches the list & dict transformers run will be configurable via environment variables.

The new env vars are:
* `_F_P_WRITE_CHUNK_SIZE` - This is the number of bytes and by default is set to `"26214400"` (25 * 2^20)
* `_F_TE_MAX_COROS` - This is the number of coroutines the type engine will use for lists/dicts and is set to `"10"`.

## How was this patch tested?
Tested on our byoc clusters, and by the OSS OP.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request introduces improvements to the Flytekit library by adding configurable chunk sizes for S3 and GCS uploads and enhancing coroutine batching in the type engine. The chunk size for S3/GCS uploads is now set to 25MB by default and can be configured via an environment variable.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -363,8 +363,6 @@ class FileAccessProvider(object):
         """
         file_system = await self.get_async_filesystem_for_path(to_path)
         from_path = self.strip_file_header(from_path)
-        import os
-
         if recursive:
             # Only check this for the local filesystem
             if file_system.protocol == "file" and not file_system.isdir(from_path):
--- a/flytekit/core/data_persistence.py
+++ b/flytekit/core/data_persistence.py
@@ -52,6 +52,8 @@ _ANON = "anon"
 
 Uploadable = typing.Union[str, os.PathLike, pathlib.Path, bytes, io.BufferedReader, io.BytesIO, io.StringIO]
 
+_S3_WRITE_SIZE_CHUNK_BYTES = int(os.environ.get("FLYTE_S3_WRITE_CHUNKSIZE_BYTES", "26214400"))  # 25 * 2**20
+
 
 def s3_setup_args(s3_cfg: configuration.S3Config, anonymous: bool = False) -> Dict[str, Any]:
     kwargs: Dict[str, Any] = {
@@ -340,8 +342,8 @@ class FileAccessProvider(object):
         from_path = self.strip_file_header(from_path)
         import os
 
-        cs = os.environ.get("chunksize", "8388608")  # 8 * 2**20
-        kwargs["chunksize"] = int(cs)
+        if to_path.startswith("s3"):
+            kwargs["chunksize"] = _S3_WRITE_SIZE_CHUNK_BYTES
         if recursive:
             # Only check this for the local filesystem
             if file_system.protocol == "file" and not file_system.isdir(from_path):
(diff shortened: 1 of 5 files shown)


flytekit/core/data_persistence.py
```diff
@@ -52,6 +52,8 @@
 
 Uploadable = typing.Union[str, os.PathLike, pathlib.Path, bytes, io.BufferedReader, io.BytesIO, io.StringIO]
 
+_WRITE_SIZE_CHUNK_BYTES = int(os.environ.get("_F_P_WRITE_CHUNK_SIZE", "26214400"))  # 25 * 2**20
```
eapolinario, line 55: Can you leave a comment? 

Also, why the `_` in the constant name?

tests/flytekit/unit/core/test_list.py
```diff
@@ -0,0 +1,89 @@
...
+    TypeEngine,
+)
+from flytekit.models.literals import (
+    Literal,
+    Primitive,
+    Scalar,
+)
+from flytekit.models.types import LiteralType, SimpleType
+
+T = typing.TypeVar("T")
```
eapolinario, line 19: We don't need this, right?

flytekit/core/data_persistence.py
```diff
@@ -108,6 +110,29 @@ def get_fsspec_storage_options(
...
+    method_name = method_name.replace("_", "")
+    if isinstance(protocol, tuple):
+        protocol = protocol[0]
+
+    if protocol == "s3":
+        if method_name == "put":
+            kwargs["chunksize"] = _WRITE_SIZE_CHUNK_BYTES
+    if protocol == "gs":
+        if method_name == "put":
+            kwargs["chunksize"] = _WRITE_SIZE_CHUNK_BYTES
```
eapolinario, line 131: what do you think of 
```suggestion
    # TODO: Comment explaining the issue
    if method_name == "put" and protocol in ["s3", "gs"]:
            kwargs["chunksize"] = _WRITE_SIZE_CHUNK_BYTES
```

eapolinario (commented):
What about https://github.com/flyteorg/flytekit/blob/master/flytekit/core/base_task.py#L638 ?

flytekit/core/data_persistence.py
wild-endeavor: it's fairly internal to this file, thought the _ would be nice.  i'll add a comment.

tests/flytekit/unit/core/test_list.py
wild-endeavor: removed

eapolinario (approved):

PR #3081, opened 2025-01-23
Add pip_extra_args

amitani:
## Tracking issue

Related to https://github.com/flyteorg/flyte/issues/6119

We have a few path dependencies in `pyproject.toml` and `poetry.lock`, and only want to install packages from external sources, which requires adding `--no-directory` to `poetry install`. This is a standard practice when installing dependencies without code.
See
https://python-poetry.org/docs/faq/#poetry-busts-my-docker-cache-because-it-requires-me-to-copy-my-source-files-in-before-installing-3rd-party-dependencies

`pip_extra_args` is a specifiable parameter for `ImageSpec`, and appended when running `poetry install`, `uv sync` and `pip install`.

## How was this patch tested?

I tested locally with a poetry project with internal path dependency by running `pyflyte build`. Planning to add unit tests once the implementation get LGTM.

## Related PRs

https://github.com/flyteorg/flytekit/pull/3025

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
Added new pip_extra_args parameter to ImageSpec class for enhanced package installation flexibility, enabling custom pip installation arguments for path dependencies in poetry projects. Enhanced test suite implementation to validate pip_extra_args parameter, including test cases for --no-install-package flag and path dependency handling with --no-directory option. Tests ensure proper integration with both poetry and pip installation workflows.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -48,6 +48,7 @@ class ImageSpec:
         platform: Specify the target platforms for the build output (for example, windows/amd64 or linux/amd64,darwin/arm64
         pip_index: Specify the custom pip index url
         pip_extra_index_url: Specify one or more pip index urls as a list
+        pip_extra_args: Specify one or more extra pip install arguments as a space-delimited string
         registry_config: Specify the path to a JSON registry config file
         entrypoint: List of strings to overwrite the entrypoint of the base image with, set to [] to remove the entrypoint.
         commands: Command to run during the building process
@@ -82,6 +83,7 @@ class ImageSpec:
     platform: str = "linux/amd64"
     pip_index: Optional[str] = None
     pip_extra_index_url: Optional[List[str]] = None
+    pip_extra_args: Optional[str] = None
     registry_config: Optional[str] = None
     entrypoint: Optional[List[str]] = None
     commands: Optional[List[str]] = None
--- a/flytekit/image_spec/default_builder.py
+++ b/flytekit/image_spec/default_builder.py
@@ -216,6 +216,9 @@ def prepare_python_install(image_spec: ImageSpec, tmp_dir: Path) -> str:
         extra_urls = [f"--extra-index-url {url}" for url in image_spec.pip_extra_index_url]
         pip_install_args.extend(extra_urls)
 
+    if image_spec.pip_extra_args:
+        pip_install_args.append(image_spec.pip_extra_args)
+
     requirements = []
     if image_spec.requirements:
         requirement_basename = os.path.basename(image_spec.requirements)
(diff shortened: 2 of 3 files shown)


flytekit/image_spec/image_spec.py
```diff
@@ -82,6 +83,7 @@ class ImageSpec:
     platform: str = "linux/amd64"
     pip_index: Optional[str] = None
     pip_extra_index_url: Optional[List[str]] = None
+    pip_extra_args: Optional[List[str]] = None
```
thomasjpfan, line 86: Can we make this a single string and not a list?

```suggestion
    pip_extra_args: Optional[str] = None
```

thomasjpfan (commented):
Can you update a list to make sure that the new argument is used?

flytekit/image_spec/image_spec.py
amitani: I made the update to take a space-delimited string. What's the rationale behind this? `pip_install_args` is joined with " " later, so taking a list and extending the list seems more aligned with how it is implemented. 

For example, when adding multiple command line arguments, it is done by extending the list of multiple elements, not a single element with multiple command line arguments concatenated with spaces.
https://github.com/flyteorg/flytekit/blob/5a2c44f1976665c420541d3385bf946cf959b2f6/flytekit/image_spec/default_builder.py#L195

(a separate topic but it may be worth using `shlex` instead of `" ".join`)

wild-endeavor (dismissed):

flytekit/image_spec/image_spec.py
thomasjpfan: I see that using a list is more aligned with the implementation.

I was mostly thinking about the developer experience. If one wants to add many arguments, they can use a single string:

```python
image = ImageSpec(
    pip_extra_args="--prerelease --all-extras --compile-bytecode --only-binary"
)
```

It takes a little more work to make it a list of strings:

```python
image = ImageSpec(
    pip_extra_args=["--prerelease", "--all-extras", "--compile-bytecode", "--only-binary"]
)
```

Do you have a preference when it comes to developer experience?

amitani: Good point. Given that these args are not parsed, a single string is simpler and less confusing.

thomasjpfan (approved):

ttngu207:
Hi @amitani and development team,

This PR is great! However I think there's a missing step in adding this new argument `pip_extra_args` into the `_SUPPORTED_IMAGE_SPEC_PARAMETERS` ([here](https://github.com/flyteorg/flytekit/blob/d42a0024436a3bd70b3abddc12b2cd9541dbb67f/flytekit/image_spec/default_builder.py#L450))
Thus I believe this new argument is being completely ignored by the `DefaultImageBuilder`.

I'm not quite sure where to raise this point as there's no `Issues` for this repo. Let me know if I'm completely wrong here or I should raise this point elsewhere (instead of on this closed PR).

Thanks again

amitani:
@ttngu207 Good catch! It was my oversight. 
It looks like `_SUPPORTED_IMAGE_SPEC_PARAMETERS` is only used to warn if any other parameters are set, and they aren't actually ignored. I'm using this with the default image builder and it has been working.
It would be great if you create a PR to add this field to `_SUPPORTED_IMAGE_SPEC_PARAMETERS`.

PR #3093, opened 2025-01-25
Adding missing fields to FlyteTask remote entity

UmerAhmad:
## Tracking issue
flyteorg/flyte#6192

Missing fields from FlyteTask entity from remote, for example, we need k8s_pod for template data analysis, but it only includes container as of now.

Adding the following fields to FlyteTask init, and promotion from model (also config):

security_context
config
k8s_pod
sql
extended_resources

## How was this patch tested?
unit tests

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
Enhanced FlyteTask remote entity with security_context, k8s_pod, sql, and extended_resources support, including default values and validation. Improved type system with strict type hint matching and added webhook task functionality with rate limiting for worker queues. Implemented shared memory support and enhanced caching capabilities across components. Added comprehensive test coverage and validation improvements for Ray tasks, remote execution, and type handling.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>


tests/flytekit/unit/remote/test_remote.py
```diff
@@ -547,14 +547,21 @@ def wf1(name: str = "union") -> float:
     flyte_remote.register_script(wf1)
 
 
-@mock.patch("flytekit.remote.remote.FlyteRemote.client")
-def test_local_server(mock_client):
+@pytest.fixture()
```
UmerAhmad, line 550: When using fetch_task, an error will be thrown if sql, k8s_pod, and container are not in an exclusive disjunction (XOR). Originally I just patched this for the test it was failing in, test_register_task_with_node_dependency_hints, which when calling register subsequently calls the fetch and causes a fail. But, I think a global patch with this fixture makes more sense in the case for any future cases where the case might reoccur.

tests/flytekit/unit/remote/test_remote.py
```diff
@@ -547,14 +547,21 @@ def wf1(name: str = "union") -> float:
...
+def mock_flyte_remote_client():
+    with patch("flytekit.remote.remote.FlyteRemote.client") as mock_flyte_remote_client:
+        mock_flyte_remote_client.get_task.return_value.closure.compiled_task.template.sql = None
+        mock_flyte_remote_client.get_task.return_value.closure.compiled_task.template.k8s_pod = None
```
UmerAhmad, line 554: When using fetch_task, an error will be thrown if sql, k8s_pod, and container are not in an exclusive disjunction (XOR). Originally I just patched this for the test it was failing in, test_register_task_with_node_dependency_hints, which when calling register subsequently calls the fetch and causes a fail. But, I think a global patch with this fixture makes more sense in the case for any future cases where the case might reoccur.

eapolinario (changes requested):
the spark unit test [failure](https://github.com/flyteorg/flytekit/actions/runs/13210255872/job/37490857239?pr=3093) is real. Can you take a look?

UmerAhmad:
Patched! - learned that the plugins had a separate local suite, thanks for pointing it out. And, it looks like the sagemaker test is globally failing.

eapolinario:
> Patched! - thanks for pointing it out, saw that plugins had a separate suite locally. And, it looks like the sagemaker test is globally failing.

Correct. The sagemaker failures are not related to your PR.

eapolinario (approved):

PR #3094, opened 2025-01-25
[WIP] Rename container_image to image for improved UX

JiangJiaWei1103:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6140

To enhance the user experience, the concept of `container` should be abstracted from flytekit users. 

At this early stage, we propose to focus on the user-facing codebase and support both `image` and `container_image` for backward compatibility. There exist three cases:
1. Both `image` and `container_image` are specified: Raise an error
2. `container_image` is used: Warn users about the future deprecation
3. `image` is used: Use `image` directly

### DIFF
--- a/flytekit/core/task.py
+++ b/flytekit/core/task.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import datetime
 import inspect
 import os
+import warnings
 from functools import partial, update_wrapper
 from typing import Any, Callable, Dict, Iterable, List, Optional, Tuple, Type, TypeVar, Union, overload
 
@@ -104,6 +105,7 @@ def task(
     interruptible: Optional[bool] = ...,
     deprecated: str = ...,
     timeout: Union[datetime.timedelta, int] = ...,
+    image: Optional[Union[str, ImageSpec]] = ...,
     container_image: Optional[Union[str, ImageSpec]] = ...,
     environment: Optional[Dict[str, str]] = ...,
     requests: Optional[Resources] = ...,
@@ -143,6 +145,7 @@ def task(
     interruptible: Optional[bool] = ...,
     deprecated: str = ...,
     timeout: Union[datetime.timedelta, int] = ...,
+    image: Optional[Union[str, ImageSpec]] = ...,
     container_image: Optional[Union[str, ImageSpec]] = ...,
     environment: Optional[Dict[str, str]] = ...,
     requests: Optional[Resources] = ...,
@@ -181,6 +184,7 @@ def task(
     interruptible: Optional[bool] = None,
     deprecated: str = "",
     timeout: Union[datetime.timedelta, int] = 0,
+    image: Optional[Union[str, ImageSpec]] = None,
     container_image: Optional[Union[str, ImageSpec]] = None,
     environment: Optional[Dict[str, str]] = None,
     requests: Optional[Resources] = None,
@@ -272,7 +276,7 @@ def task(
                        indicates that the task is active and not deprecated
     :param timeout: the max amount of time for which one execution of this task should be executed for. The execution
                     will be terminated if the runtime exceeds the given timeout (approximately).
-    :param container_image: By default the configured FLYTE_INTERNAL_IMAGE is used for every task. This directive can be
+    :param image: By default the configured FLYTE_INTERNAL_IMAGE is used for every task. This directive can be
                 used to provide an alternate image for a specific task. This is useful for the cases in which images
                 bloat because of various dependencies and a dependency is only required for this or a set of tasks,
                 and they vary from the default.
@@ -282,13 +286,13 @@ def task(
                     # Use default image name `fqn` and alter the tag to `tag-{{default.tag}}` tag of the default image
                     # with a prefix. In this case, it is assumed that the image like
                     # flytecookbook:tag-gitsha is published alongwith the default of flytecookbook:gitsha
-                    @task(container_image='{{.images.default.fqn}}:tag-{{images.default.tag}}')
+                    @task(image='{{.images.default.fqn}}:tag-{{images.default.tag}}')
                     def foo():
                         ...
 
                     # Refer to configurations to configure fqns for other images besides default. In this case it will
                     # lookup for an image named xyz
-                    @task(container_image='{{.images.xyz.fqn}}:{{images.default.tag}}')
+                    @task(image='{{.images.xyz.fqn}}:{{images.default.tag}}')
                     def foo2():
                         ...
     :param environment: Environment variables that should be added for this tasks execution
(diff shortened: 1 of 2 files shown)


flytekit/core/task.py
```diff
@@ -584,6 +606,24 @@ async def eager_workflow(x: int) -> int:
             async def eager_workflow(x: int) -> int:
                 ...
     """
+    # Rename the `container_image` parameter to `image` for improved user experience.
```
thomasjpfan, line 609: Should `image` be added to `PythonAutoContainerTask` and we do the validation there? This way `@eager` (through `EagerAsyncPythonFunctionTask`) and `@task` will get the check.

For backward compatible, I would have `PythonAutoContainerTask._image`, but have `._container_image` and `container_image` be a mirror of it:

```python
class PythonAutoContainerTask:
    def __init__(self, container_image, image):
        # validate image and container_image
        self._image = container_image
        
    @property
    def container_image(self):
        return self._image

    @property
    def _container_image(self):
        return self._image
        
    @_container_image.setter
    def _container_image(self, value):
        self._image = value

    @property
    def image(self):
        return self._image
```

JiangJiaWei1103: Both work fine for me. Thanks for your suggestion--I learned a lot!

I have two questions:
1. Is it common practice to make a class’s internal attribute (like `_container_image` here) accessible and modifiable from outside the class?
    * I assume we do this to avoid breaking users’ code.
2. Would it be good to add a `setter` for `image`?

thomasjpfan: > I assume we do this to avoid breaking users’ code.

I'm mostly trying to keep backward compatibility. I see bunch of code paths that does `obj._container_image = `. (I do not really like this practice, but that's the current state of the code base)

> Would it be good to add a setter for image?

I'll say it's not worth it here. If we keep the `obj._container_image =` practice, then future users can do `obj._image = `.

pingsutw (approved):

flytekit/core/task.py
JiangJiaWei1103: Thanks for the clarification! Everything is now perfectly clear.

flytekit/core/python_auto_container.py
```diff
@@ -133,9 +152,23 @@ def __init__(
...
+        """Deprecated, please use `image` instead."""
+        return self._image
+
+    @property
+    def _container_image(self) -> Optional[Union[str, ImageSpec]]:
+        """Deprecated, please use `image` instead."""
+        return self._image
+
+    @_container_image.setter
+    def _container_image(self, image: Optional[Union[str, ImageSpec]]):
```
thomasjpfan, line 170:
```suggestion
    def _container_image(self, image: Optional[Union[str, ImageSpec]]):
        """Deprecated, please use `image` instead."""
        # This setter is for backward compatibility, so that setting `_container_image`
        # will adjust the new `_image` parameter directly.
```

thomasjpfan (approved):
Minor nit on adding a comment

flytekit/core/python_auto_container.py
JiangJiaWei1103: Hi Thomas,

I’ve just added i. Thanks for your guidance!

thomasjpfan (dismissed):

flytekit/core/task.py
```diff
@@ -104,6 +105,7 @@ def task(
     interruptible: Optional[bool] = ...,
     deprecated: str = ...,
     timeout: Union[datetime.timedelta, int] = ...,
+    image: Optional[Union[str, ImageSpec]] = ...,
```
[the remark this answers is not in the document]
eapolinario: we need to be backwards compatible.

eapolinario:
shoulnd't we also support `image` in [with_overrides](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/node.py#L189) too?

JiangJiaWei1103:
Hi @eapolinario,

Thanks for the suggestion! We focus on `@task`, `@dynamic`, and `@eager` in this PR, as approved by @thomasjpfan.

Supporting `image` in `with_overrides` is also a great enhancement. Would you prefer modifying it in this PR, or should we explore it in a follow-up PR? Thanks!

flytekit/core/python_auto_container.py
```diff
@@ -42,6 +43,7 @@ def __init__(
         name: str,
         task_config: T,
         task_type="python-task",
+        image: Optional[Union[str, ImageSpec]] = None,
         container_image: Optional[Union[str, ImageSpec]] = None,
```
wild-endeavor, line 47: @eapolinario and I were talking about this and we were thinking we might want to put this behind kwargs instead.  Could we try that please?

Like delete this line... but leave it in the docs below, maybe move it to the bottom of the params, and check for it explicitly in the code to see if it's set (like `kwargs["container_image"]`).  This way we can hide it from the user, they don't see it in the IDE etc.

cc @thomasjpfan

pingsutw: > put this behind kwargs instead
+1

flytekit/core/python_auto_container.py
```diff
@@ -90,7 +93,23 @@ def __init__(
...
+        # Rename the `container_image` parameter to `image` for improved user experience.
+        # Currently, both `image` and `container_image` are supported to maintain backward compatibility.
+        # For more details, please refer to https://github.com/flyteorg/flyte/issues/6140.
+        if image is not None and container_image is not None:
+            raise ValueError(
+                "Cannot specify both image and container_image. "
+                "Please use image because container_image is deprecated and will be removed in the future."
+            )
+        elif container_image is not None:
+            warnings.warn(
```
wild-endeavor, line 105: should we just make this info?  warning is a little too annoying I think.

thomasjpfan: I feel like this is not annoying enough. `DeprecationWarning` does not really show up with default filter warnings. (`FutureWarning` would be more annoying)

If we really want users to change their code, then I think warnings should be a bit annoying. Because when we remove this feature, breaking code will be even more annoying.

flytekit/core/python_auto_container.py
thomasjpfan: It's a bit non-standard, but I am okay with hiding it in `kwargs`, but keeping it in the docs.

flytekit/core/python_auto_container.py
```diff
@@ -57,7 +59,8 @@ def __init__(
         :param name: unique name for the task, usually the function's module and name.
         :param task_config: Configuration object for Task. Should be a unique type for that specific Task.
         :param task_type: String task type to be associated with this Task
-        :param container_image: String FQN for the image.
+        :param image: String FQN for the image.
```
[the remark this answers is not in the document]
JiangJiaWei1103: Hi Yee,

Thanks for catching that and providing the correction!

flytekit/core/python_auto_container.py
JiangJiaWei1103: I've hidden `container_image` behind `kwargs` and kept it at the bottom of docstring. Thanks a lot!

wild-endeavor:
@JiangJiaWei1103 could you change `with_overrides` as well in this pr please?

Also, could you remove `container_task` from the `@task` decorator (and the `@overload`s also) please?  If you merge in master, you'll see that @eapolinario has added `**kwargs` [to the decorator](https://github.com/flyteorg/flytekit/blob/338dd6d129ff1c2728d505e7a858ed50ad2e795c/flytekit/core/task.py#L208) already.

Thank you!  Let's merge after this last change

JiangJiaWei1103:
Hi Yee,

I’ll get it done this weekend. Thanks for the suggestion!

JiangJiaWei1103:
The `container_image` parameter is now encapsulated within `kwargs` for the `with_overrides` method of both `Node` and `Promise`. All changes have been validated through enhanced and newly added tests.

### Follow-ups
I’m curious to know if it’s also necessary to modify `TaskNodeOverrides` as shown below:
https://github.com/flyteorg/flytekit/blob/66d4aedc7098b4a900f29fcc8c48763aad76ec79/flytekit/models/core/workflow.py#L615-L626

Thanks!

PR #3095, opened 2025-01-25
[WIP] Rename ephemeral_storage to disk for simplicity

JiangJiaWei1103:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6142

The attribute `ephemeral_storage` of `Resources` (i.e., K8s pod resources) is too complicated and should be simplfied.

At this stage, we propose supporting both `disk` and `ephemeral_storage` with the following behavior:
1. `ephemeral_storage` will remain supported as an alias for backward compatibility and developer-facing code
2. Users are encouraged to adopt `disk` for future configurations

The behavior for different configurations is summarized as follows:
1. Both `disk` and `ephemeral_storage` are specified: Raise an error to avoid ambiguity
2. Only `ephemeral_storage` is specified: Issue a warning about future deprecation but continue to use it as is
3. Only `disk` is specified: Automatically set `ephemeral_storage` to the value of `disk`

### DIFF
--- a/flytekit/core/resources.py
+++ b/flytekit/core/resources.py
@@ -1,3 +1,4 @@
+import warnings
 from dataclasses import dataclass, fields
 from typing import Any, List, Optional, Union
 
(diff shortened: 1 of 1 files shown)


JiangJiaWei1103:
I’d like to explain why I made the newly added `disk` field init-only.

Consider the following example:
https://github.com/flyteorg/flytekit/blob/92ddd0e3b402c755a62bde03f62db8cd80e46439/tests/flytekit/unit/core/test_resources.py#L103-L107

Here, a `Resources` instance is first serialized into a JSON string and then deserialized back into a `Resources` instance. If `disk` were treated as a normal field, both `disk` and `ephemeral_storage` would be included in the JSON output. However, this would violate the rule that these two fields cannot be specified simultaneously.

To prevent this issue, I made `disk` an init-only field. I’d love to hear your thoughts—is this approach a reasonable solution for now, or would you suggest a better way to handle this? Thanks a lot!

PR #3097, opened 2025-01-27
don't override timeout on with_overrides if not specified

pvditt:
## Tracking issue
fixes: https://github.com/flyteorg/flyte/issues/6153

with_overrides set timeout to default of 0 if timeout param is not specified

- differentiate between timeout not getting set vs explicitly set to None for with_overrides

## How was this patch tested?
added unit tests

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR fixes a critical bug in the with_overrides() method where timeouts were incorrectly defaulting to 0. The implementation introduces a sentinel object (renamed to TIMEOUT_OVERRIDE_SENTINEL for Python naming conventions) to distinguish between explicit timeout settings and unspecified cases. The changes modify the Node class timeout handling logic and improve differentiation between unset timeout and explicit None timeout cases, with comprehensive test coverage for timeout override scenarios.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>


flytekit/core/node.py
```diff
@@ -127,7 +129,7 @@ def metadata(self) -> _workflow_model.NodeMetadata:
     def _override_node_metadata(
         self,
         name,
-        timeout: Optional[Union[int, datetime.timedelta]] = None,
+        timeout: Optional[Union[int, datetime.timedelta, object]] = timeout_override_sentinel,
```
thomasjpfan, line 132: Can we use `None` as the sentinel? I feel like setting `timeout to datetime.timedelta()` was a bug.

pvditt: @thomasjpfan 

> Can we use None as the sentinel? 

What do you mean by this?

I wanted to handle the case of .with_overrides(timeout=None) to zero out the timeout

thomasjpfan: Ah, I think it's a bit strange to have `.with_overrides(timeout=None)` mean "zero out the timeout".

Okay, lets be safe and use the sentinel the way you already have it.

flytekit/core/node.py
```diff
@@ -47,6 +47,8 @@ class Node(object):
     ID, which from the registration step
     """
 
+    timeout_override_sentinel = object()
```
thomasjpfan, line 50: Nit: I usually see python sentinels as constants, thus all capitalized: 

```suggestion
    TIMEOUT_OVERRIDE_SENTINEL = object()
```

pvditt: sounds good - just updated

flytekit/core/node.py
```diff
@@ -142,14 +144,16 @@ def _override_node_metadata(
...
-            node_metadata._timeout = datetime.timedelta()
-        elif isinstance(timeout, int):
-            node_metadata._timeout = datetime.timedelta(seconds=timeout)
-        elif isinstance(timeout, datetime.timedelta):
-            node_metadata._timeout = timeout
-        else:
-            raise ValueError("timeout should be duration represented as either a datetime.timedelta or int seconds")
+        if timeout is not Node.timeout_override_sentinel:
+            if timeout is None:
+                node_metadata._timeout = 0
```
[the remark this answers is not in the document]
thomasjpfan: @pvditt For typing, I think this needs to be a timedelta object.

eapolinario (approved):

PR #3109, opened 2025-02-05
Flytekit remote updates for notebooks and versions

kumare3:
Problems:
1. Version had to be provided when registering certain entities from the notebook
2. Execute worked, but register did not work in the same way
3. In cases when same entity is registered in non interactive mode, remote entity was auto selected to be latest registered if version field was empty. This is not explicitly called `latest` as a special version
4. Makes version handling uniform across all entities 

Example cases, now works
Screenshot 2025-02-04 at 9 35 13 PM

Screenshot 2025-02-04 at 9 35 47 PM
Screenshot 2025-02-04 at 9 36 12 PM

### DIFF
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -123,6 +123,8 @@ ExecutionDataResponse = typing.Union[WorkflowExecutionGetDataResponse, NodeExecu
 
 MOST_RECENT_FIRST = admin_common_models.Sort("created_at", admin_common_models.Sort.Direction.DESCENDING)
 
+LATEST_VERSION_STR = "latest"
+
 
 class RegistrationSkipped(Exception):
     """
@@ -161,12 +163,14 @@ def _get_entity_identifier(
     name: str,
     version: typing.Optional[str] = None,
 ):
+    if version is None or version == LATEST_VERSION_STR:
+        version = _get_latest_version(list_entities_method, project, domain, name)
     return Identifier(
         resource_type,
         project,
         domain,
         name,
-        version if version is not None else _get_latest_version(list_entities_method, project, domain, name),
+        version,
     )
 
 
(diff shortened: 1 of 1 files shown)


wild-endeavor:
There's a breaking change here I wanted to highlight which may or may not be okay.  I think it's technically more correct, but it is a change in behavior.

Assuming you have a simple file like this
```python
from flytekit import task, workflow, ImageSpec, FlyteRemote, Config

image = ImageSpec(
    name="yt_public",
    builder="default",
    registry="ghcr.io/wild-endeavor",
    packages=["flytekit==v1.15.0b2"],
)

@task(container_image=image, enable_deck=True)
def print_hello():
    print("hello")

@workflow
def wf():
    print_hello()

if __name__ == "__main__":
    remote = FlyteRemote(Config.for_sandbox(), default_project="flytesnacks", default_domain="development")
    remote.register_workflow(wf, version="from_main_v4_pr")
```

the behavior on master is as follows
* if the user does `pyflyte register`, fast register is used, and the code is copied into the tarball.  Code is _not_ built into the image.
* if the user does `python file.py`, fast register is not used, and the code _is_ built into the image.

The reason the code is built into the image in the second case is because the code currently is setting the project root.  When the register workflow call triggers the downstream register task call, it gets set on the image spec, causing the user code to be built in.

With this pr, without setting project root, pyflyte register remains the same, but `python file.py` will no longer copy code - when the workflow is run, the task will fail.

If the user wants to make this work, they'd have to change to `fast_register_workflow`, which actually feels more correct.

I think we can make a better experience actually - the register_workflow function right now never does fast register... even if the user passes in a serialization settings object with fast register enabled, so there's some discrepancy here to clean up.

pingsutw (approved):

wild-endeavor:
@kumare3 pushed a nit to get the monodocs build to pass.  should we merge?  Keep in mind there is a small breaking change (detailed above).  I'm okay with this given the fact that this function is a bit broken to begin with (fast settings not being respected), and the behavior now actually might better align with expectations.

wild-endeavor (approved):

PR #3123, opened 2025-02-10
Fix dict annotations

wild-endeavor:
## Tracking issue
Internal ticket only.

Customer reported issue with internal model registry.  The issue though is that annotations don't work well with dictionary return types.

Most of the changes here are in the dict transformer.
* The `extract_types_or_metadata` function is inconsistent, returning different things in different scenarios - sometimes it would return annotations, sometimes it would return the key/value type of a dictionary.  This has been changed to always return the latter, or an empty tuple `()` in the case of an untyped dictionary.  In all cases now, the caller interprets the result as a tuple of key type,value type.  Renaming function to just `extract_types`.
* the `is_pickle` function returned both a boolean, as well as types, but the latter isn't really necessary.  Changing this to just return the pickle bool.
* `dict_types` is a function that was almost the same as `extract_types_or_metadata`, removing in favor of that one and manually extracting the annotation args where needed.
* Remove a duplicate test

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances dictionary type handling in the type system by refactoring the DictTransformer class and simplifying type extraction functionality. The extract_types_or_metadata function has been renamed to extract_types, and the is_pickle function has been streamlined. The changes improve consistency in dictionary type handling and include removal of redundant functionality.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -947,7 +947,7 @@ async def binding_data_from_python_std(
             if transformer_override and hasattr(transformer_override, "extract_types_or_metadata"):
                 _, v_type = transformer_override.extract_types_or_metadata(t_value_type)  # type: ignore
             else:
-                _, v_type = DictTransformer.extract_types(t_value_type)  # type: ignore
+                _, v_type = DictTransformer.extract_types(cast(typing.Type[dict], t_value_type))
             m = _literals_models.BindingDataMap(
                 bindings={
                     k: await binding_data_from_python_std(
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -947,7 +947,7 @@ async def binding_data_from_python_std(
             if transformer_override and hasattr(transformer_override, "extract_types_or_metadata"):
                 _, v_type = transformer_override.extract_types_or_metadata(t_value_type)  # type: ignore
             else:
-                _, v_type = DictTransformer.extract_types_or_metadata(t_value_type)  # type: ignore
+                _, v_type = DictTransformer.extract_types(t_value_type)  # type: ignore
             m = _literals_models.BindingDataMap(
                 bindings={
                     k: await binding_data_from_python_std(
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -2105,7 +2105,7 @@ class DictTransformer(AsyncTypeTransformer[dict]):
     @staticmethod
     def is_pickle(python_type: Type[dict]) -> bool:
         _origin = get_origin(python_type)
-        metadata = []
+        metadata: typing.Tuple = ()
         if _origin is Annotated:
             metadata = get_args(python_type)[1:]
 
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -878,7 +878,7 @@ class DataclassTransformer(TypeTransformer[object]):
             return list(map(lambda x: self._fix_val_int(ListTransformer.get_sub_type(t), x), val))
 
         if isinstance(val, dict):
-            ktype, vtype = DictTransformer.extract_types_or_metadata(t)
+            ktype, vtype = DictTransformer.extract_types(t)
             # Handle nested Dict. e.g. {1: {2: 3}, 4: {5: 6}})
             return {
                 self._fix_val_int(cast(type, ktype), k): self._fix_val_int(cast(type, vtype), v) for k, v in val.items()
@@ -2018,7 +2018,7 @@ class DictTransformer(AsyncTypeTransformer[dict]):
         super().__init__("Typed Dict", dict)
 
     @staticmethod
-    def extract_types_or_metadata(t: Optional[Type[dict]]) -> typing.Tuple:
+    def extract_types(t: Optional[Type[dict]]) -> typing.Tuple:
         _origin = get_origin(t)
         _args = get_args(t)
         if _origin is not None:
@@ -2034,7 +2034,7 @@ class DictTransformer(AsyncTypeTransformer[dict]):
             if _origin is dict and _args is not None:
                 return _args  # type: ignore
             elif _origin is Annotated:
-                return DictTransformer.extract_types_or_metadata(_args[0])
+                return DictTransformer.extract_types(_args[0])
             else:
                 raise ValueError(f"Trying to extract dictionary type information from a non-dict type {t}")
         return None, None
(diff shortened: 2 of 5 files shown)


flytekit/core/promise.py
```diff
@@ -947,7 +947,7 @@ async def binding_data_from_python_std(
             if transformer_override and hasattr(transformer_override, "extract_types_or_metadata"):
                 _, v_type = transformer_override.extract_types_or_metadata(t_value_type)  # type: ignore
             else:
-                _, v_type = DictTransformer.extract_types_or_metadata(t_value_type)  # type: ignore
+                _, v_type = DictTransformer.extract_types(t_value_type)  # type: ignore
```
eapolinario, line 950: can we remove the `# type: ignore` now?

flytekit/core/type_engine.py
```diff
@@ -2099,31 +2103,24 @@ async def dict_to_binary_literal(
             raise TypeTransformerFailedError(f"Cannot convert `{v}` to Flyte Literal.\n" f"Error Message: {e}")
 
     @staticmethod
-    def is_pickle(python_type: Type[dict]) -> typing.Tuple[bool, Type]:
-        base_type, *metadata = DictTransformer.extract_types_or_metadata(python_type)
+    def is_pickle(python_type: Type[dict]) -> bool:
```
eapolinario, line 2106: the change of return type in `is_pickle` is the best change in this PR.

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -2863,23 +2863,6 @@ def test_get_underlying_type(t, expected):
...
-                Annotated[typing.Dict[str, str], kwtypes(allow_pickle=True)],
-                (typing.Dict[str, str], kwtypes(allow_pickle=True)),
-        ),
-        (typing.Dict[Annotated[str, "a-tag"], int], (Annotated[str, "a-tag"], int)),
-    ],
-)
-def test_dict_get(t, expected):
-    assert DictTransformer.extract_types_or_metadata(t) == expected
-
-
```
eapolinario, line 2882: nit: can you remove the [other one](https://github.com/flyteorg/flytekit/blob/master/tests/flytekit/unit/core/test_generice_idl_type_engine.py#L2861-L2875) instead?

eapolinario (approved):
this looks good. Left a few nits, but won't block on any of them.

tests/flytekit/unit/core/test_type_engine.py
wild-endeavor: sure

eapolinario (approved):

flytekit/core/type_engine.py
```diff
@@ -2031,8 +2031,12 @@ def extract_types_or_metadata(t: Optional[Type[dict]]) -> typing.Tuple:
                         raise ValueError(
                             f"Flytekit does not currently have support for FlyteAnnotations applied to dicts. {t} cannot be parsed."
                         )
-            if _origin in [dict, Annotated] and _args is not None:
+            if _origin is dict and _args is not None:
                 return _args  # type: ignore
+            elif _origin is Annotated:
+                return DictTransformer.extract_types(_args[0])
+            else:
+                raise ValueError(f"Trying to extract dictionary type information from a non-dict type {t}")
```
Future-Outlier, line 2039: Just use debugger to run the unit test of this PR, this make sense to me.

Future-Outlier: checking other usecase in annotation, put `tuple` in it

Future-Outlier: oh I see, tuple type is restricted now

tests/flytekit/unit/core/test_annotated_bindings.py
```diff
@@ -339,3 +343,77 @@ def t0(ii: Annotated[MyStr, tf]) -> None:
             print(dynamic_job_spec)
 
     del TypeEngine._REGISTRY[MyStr]
+
+
```
Future-Outlier, line 347: Let's add test when Tuple IDL is supported!
https://github.com/flyteorg/flyte/pull/5699

```python

def test_pydantic_annotated_tuple():
    from flytekit.core.artifact import Artifact
    from flytekit import task, dynamic
    from flytekit.configuration import SerializationSettings, ImageConfig, Image
    from flytekit.core.context_manager import FlyteContextManager, ExecutionState, FlyteContext
    from typing import Annotated, Union, Tuple

    # Local Pydantic model for demonstration
    class Config2(BaseModel):
        param: int = 42

    @task
    def get_config() -> Config2:
        return Config2()

    # We'll reuse the Prophet class already defined above in the same file.
    # If needed, ensure it's defined at a higher scope or imported in your actual code.
    def dt_function():
        # Create Pydantic object
        cfg = get_config()
        # Create a dummy Prophet dict
        best_models = {"model_1": Prophet(param=0.1)}
        # Return a tuple of (Config2, dict[str, Prophet])
        return cfg, best_models

    a = Artifact(name="my_tuple_model")

    # A dynamic workflow returning an Annotated tuple
    @dynamic
    def dt_tuple() -> Annotated[Tuple[Config2, dict[str, Prophet]], a]:
        return dt_function()

    # Serialization settings similar to other tests
    ss = SerializationSettings(
        project="test_proj",
        domain="test_domain",
        version="abc",
        image_config=ImageConfig(Image(name="name", fqn="image", tag="name")),
        env={},
    )

    # Compile the dynamic workflow into a spec and ensure it behaves as expected
    with FlyteContextManager.with_context(
        FlyteContextManager.current_context().with_serialization_settings(ss)
    ) as ctx:
        new_exc_state = ctx.execution_state.with_params(mode=ExecutionState.Mode.TASK_EXECUTION)
        with FlyteContextManager.with_context(ctx.with_execution_state(new_exc_state)):
            dynamic_job_spec_tuple = dt_tuple.compile_into_workflow(ctx, dt_tuple._task_function)

    # We expect exactly one node in the spec (this dynamic call itself),
    # and a single output binding containing our tuple
    assert len(dynamic_job_spec_tuple.nodes) == 1
    assert len(dynamic_job_spec_tuple.outputs) == 1

    # Here, we can check that the output is stored as a collection of two items
    bd = dynamic_job_spec_tuple.outputs[0].binding
    assert bd.collection is not None
    assert len(bd.collection.bindings) == 2

```

flytekit/core/type_engine.py
Future-Outlier: the if else statement in this function looks super weird, trying to fix it

PR #3125, opened 2025-02-11
Bug when calling wf in eager task with thread

wild-endeavor:
Unable to run subworkflows, as part of eager workflows, via to_thread.  This PR only fixes the ability to run subworkflows on the backend.  Local execution still needs additional fixes.

* A return was missing from the call handler, leading to local runs of subworkflows locally run by eager to run multiple things.
* The compile step has been taken out of the workflow call handler.
* In local testing found that the behavior of the reconcile function in the worker doesn't handle the re-run case where the execution is already found. The reconcile function now will always just sync the execution, and then handle state management appropriately.
  * Additional terminal states are mapped to failed in the update object.
  * asserts were hard to debug - these have been changed to explicit assertion errors.

The issues encountered while debugging this...
* Local execute in my test case (not included) was only succeeding maybe 1 out of every 6-7 times.
* The compile step that was moved into the conditional was causing failures all the time.

My running theory for this is because the context manager is not thread safe.  Need to investigate further how to_thread's contextvar copy works.  Context copying is needed because we need to accurately represent the state upon entering the thread, but removing the context frames seems to be polluting each other's context.

## How was this patch tested?
Tested locally and on Union internal clusters.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances subworkflow execution in eager workflows via to_thread by implementing improved error handling and state management. The changes include adding a missing return statement in async call handling, better error messaging for worker queue assertions, and proper terminal state handling. Additional improvements include fixes to flytekit version specification in image builds, enhanced thread safety in the context manager, and making workflow compilation conditional. Documentation improvements and comprehensive test additions for workflow execution phases were also made.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/core/promise.py
+++ b/flytekit/core/promise.py
@@ -1482,7 +1482,7 @@ def flyte_entity_call_handler(
         # call the blocking version of the async call handler
         # This is a recursive call, the async handler also calls this function, so this conditional must match
         # the one in the async function perfectly, otherwise you'll get infinite recursion.
-        loop_manager.run_sync(async_flyte_entity_call_handler, entity, **kwargs)
+        return loop_manager.run_sync(async_flyte_entity_call_handler, entity, **kwargs)
 
     if ctx.execution_state and (
         ctx.execution_state.mode == ExecutionState.Mode.TASK_EXECUTION
--- a/flytekit/core/python_function_task.py
+++ b/flytekit/core/python_function_task.py
@@ -607,3 +607,28 @@ class EagerAsyncPythonFunctionTask(AsyncPythonFunctionTask[T], metaclass=FlyteTr
                 # now have to fail this eager task, because we don't want it to show up as succeeded.
                 raise FlyteNonRecoverableSystemException(base_error)
             return result
+
+    def run(self, remote: "FlyteRemote", ss: SerializationSettings, **kwargs):
+        ctx = FlyteContextManager.current_context()
+        # tag is the current execution id
+        # root tag is read from the environment variable if it exists, if not, it's the current execution id
+        if not ctx.user_space_params or not ctx.user_space_params.execution_id:
+            raise AssertionError("User facing context and execution ID should be present when not running locally")
+        tag = ctx.user_space_params.execution_id.name
+        root_tag = os.environ.get(EAGER_ROOT_ENV_NAME, tag)
+
+        # Prefix is a combination of the name of this eager workflow, and the current execution id.
+        prefix = self.name.split(".")[-1][:8]
+        prefix = f"e-{prefix}-{tag[:5]}"
+        prefix = _dnsify(prefix)
+        # Note: The construction of this object is in this function because this function should be on the
+        # main thread of pyflyte-execute. It needs to be on the main thread because signal handlers can only
+        # be installed on the main thread.
+        c = Controller(remote=remote, ss=ss, tag=tag, root_tag=root_tag, exec_prefix=prefix)
+        handler = c.get_signal_handler()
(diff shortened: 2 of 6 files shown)


flytekit/core/promise.py
```diff
@@ -1482,7 +1482,7 @@ def flyte_entity_call_handler(
         # call the blocking version of the async call handler
         # This is a recursive call, the async handler also calls this function, so this conditional must match
         # the one in the async function perfectly, otherwise you'll get infinite recursion.
-        loop_manager.run_sync(async_flyte_entity_call_handler, entity, **kwargs)
+        return loop_manager.run_sync(async_flyte_entity_call_handler, entity, **kwargs)
```
eapolinario, line 1485: this would be wreaking havoc before, right?

tests/flytekit/unit/core/test_worker_queue.py
```diff
@@ -249,3 +250,47 @@ def t1() -> str:
...
+        id=identifier.WorkflowExecutionIdentifier("project", "domain", "exec-name"),
+        spec=ExecutionSpec(
+            identifier.Identifier(identifier.ResourceType.LAUNCH_PLAN, "project", "domain", "name", "version"),
+            ExecutionMetadata(ExecutionMetadata.ExecutionMode.MANUAL, "tester", 1),
+        ),
+        closure=ExecutionClosure(
+            phase=phase,
+            started_at=datetime.datetime(year=2024, month=1, day=2, tzinfo=datetime.timezone.utc),
+            duration=datetime.timedelta(seconds=10),
+            # abort_metadata=AbortMetadata(cause="cause", principal="testuser"),
```
eapolinario, line 283: nit: remove it

flytekit/core/python_function_task.py
```diff
@@ -607,3 +607,32 @@ async def run_with_backend(self, **kwargs):
                 # now have to fail this eager task, because we don't want it to show up as succeeded.
                 raise FlyteNonRecoverableSystemException(base_error)
             return result
+
+    def run(self, remote: "FlyteRemote", ss: SerializationSettings, **kwargs):  # type: ignore[name-defined]
```
eapolinario, line 611: why are we ignoring the type hint?

eapolinario, line 611: where/how is this called?

wild-endeavor: recursive import... didn't see the need to import flyteremote for type checking just yet since it's only for a test/dev function for now.

wild-endeavor: see comment... it's not yet, but it's super helpful for running eager locally, but spawning tasks against a live backend.  used in conjunction with this code here: https://github.com/unionai/debugyt/blob/ccd5dc4648bd223fca9f9957a4bc07bff642aab6/user/ytong/src/async_eager/threaded_oss.py#L169  I was going to formalize it (so users can run eager locally but spawn tasks against serverless) but it hasn't gotten any buy in.  so a test function it remains for now.

flytekit/core/promise.py
wild-endeavor: yes, but only for workflows... for tasks it kinda just falls through

flytekit/core/python_function_task.py
eapolinario: Thanks for explaining, even after reading the comments I was having trouble visualizing how we were going to call it, but the example clarified the intended use.

eapolinario (approved):

PR #3129, opened 2025-02-12
Cache polish work - including the introduction of cache policies to the new top level Cache object

eapolinario:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6143

Over the years we received some feedback complaining about the discoverability of the cache-related functionality, this is finally being addressed in this PR. This PR also opens the door for a more complex caching mechanism to be applied to each `@task`.

We create a new object called `Cache` which is groups all of the caching functionality, including the ability to set the version, [cache serialization](https://docs.flyte.org/en/latest/user_guide/development_lifecycle/cache_serializing.html),  the ability to ignore certain inputs in the cache key calculation (via setting `cache_ignore_input_vars`), and the set of cache policies applied. Here's the full definition of `Cache`:

```python
class Cache:
    version: Optional[str] = None
    serialize: bool = False
    ignored_inputs: Union[Tuple[str, ...], str] = ()
    salt: str = ""
    policies: Optional[Union[List[CachePolicy], CachePolicy]] = None
```

The first three fields are well known and map directly to existing cache-related concepts, now grouped in a single object. As for `policies` and `salt`, those are new concepts, the idea is to provide more complicated caching behavior based on the function definition. A `CachePolicy` is a protocol that specifies how versions are calculated. It exposes a single method called `get_version` that's called to produce a value, that we call `version`. 

The `Cache` object applies the policies in the order in which they are defined to generate the final cache version. The process of generating the final value involves concatenating each version produced by a policy and then using SHA-256 to produce the final hexdigest (i.e. 64 hex characters) value. 

Also worth mentioning that we require the definition of policies in case a `Cache` object is used.

All cache-related parameters are now officially deprecated (although they will keep being supported, potentially for a really long time). In order to make the migration seamless, the `cache` parameter in the `@task` decorator is now `Union[bool, Cache]` and mapping the old names to the new fields in `Cache` should be fairly simple. 

As a final design decision, we also added a hook to help define a list of default policies. This might suit the more advanced users in case they need to maintain a central list of cache policies somewhere.

## How was this patch tested?
Unit tests and local sandbox.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements a comprehensive refactoring of Flyte tasks' caching functionality, transitioning from explicit cache parameters to a kwargs-based approach with a new Cache class. The changes include enhanced pod template support, improved cache policy validation, and authentication system refinements. The implementation features updated error messaging for cache policy requirements and modified task decorator for pod template parameter handling, while maintaining backward compatibility through policy-based version generation.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>


flytekit/configuration/plugin.py
```diff
@@ -103,6 +108,11 @@ def get_auth_success_html(endpoint: str) -> Optional[str]:
         """Get default success html. Return None to use flytekit's default success html."""
         return None
 
+    @staticmethod
+    def get_default_cache_policies() -> List[CachePolicy]:
+        """Get default cache policies for tasks."""
+        return []
+
```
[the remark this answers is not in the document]
eapolinario: the suggestion does not make any sense.

wild-endeavor (approved):

flytekit/core/task.py
```diff
@@ -343,7 +346,35 @@ def launch_dynamically():
     :param pickle_untyped: Boolean that indicates if the task allows unspecified data types.
     """
 
-    def wrapper(fn: Callable[P, Any]) -> PythonFunctionTask[T]:
+    def wrapper(fn: Callable[P, FuncOut]) -> PythonFunctionTask[T]:
+        nonlocal cache, cache_serialize, cache_version, cache_ignore_input_vars
```
thomasjpfan, line 350: Can we avoid `nonlocal` here?

flytekit/core/task.py
```diff
@@ -343,7 +346,35 @@ def launch_dynamically():
...
+                raise ValueError(
+                    "cache_serialize, cache_version, and cache_ignore_input_vars are deprecated. Please use Cache object"
+                )
+
+            cache_version = cache.get_version(VersionParameters(func=fn, container_image=container_image))
+            cache_serialize = cache.serialize
+            cache_ignore_input_vars = cache.get_ignored_inputs()
+            cache = True
+
+        if cache_serialize is None:
```
thomasjpfan, line 371: It's weird to have `cache_version` being set above, but then immediately overwritten. One needs to look a few lines above to see that these parameters can not be set.

Can we at least have a check here to make it obvious that this is only used for is for the bool case?

```python
is_bool_cache = isinstance(cache, bool) and cache is True and cache_version is None
if is_bool_cache:
    cache = Cache(...)

if isinstance(cache, Cache):
    ....

if is_bool_cache:
    if cache_serialize is None:
    ...
```

flytekit/core/cache.py
```diff
@@ -0,0 +1,90 @@
...
+
+@dataclass
+class VersionParameters(Generic[P, FuncOut]):
+    """
+    Parameters used for version hash generation.
+
+    param func: The function to generate a version for. This is an optional parameter and can be any callable
+                 that matches the specified parameter and return types.
+    :type func: Optional[Callable[P, FuncOut]]
+    :param container_image: The container image to generate a version for. This can be a string representing the
```
thomasjpfan, line 21: Should we consider `pod_template` too?

flytekit/core/task.py
eapolinario: can you expand a bit on why? 

We need to modify these variables inside the `wrapper` function and in python variable assignment implies the creation of a new variable unless such variable is declared as nonlocal.

flytekit/core/task.py
eapolinario: The `TaskMetadata` dataclass does not take `Optional` for the cache-related fields, which means that we either (1) modify those types in the dataclass declaration or (2) set default values prior to invoking the constructor. I decided to go with (2) because having those fields as Optional means that I'd have to check for the nil case everywhere (which is pretty annoying).

Also worth noting that this initialization covers the common case where the cache is not used at all (so not only the positive bool case).

flytekit/core/cache.py
eapolinario: that's reasonable.

flytekit/core/task.py
thomasjpfan: I always found `nonlocal` to be a bad code smell. I think we can avoid it like this:

```patch
diff --git a/flytekit/core/task.py b/flytekit/core/task.py
index 910858d6f..c0fc38e89 100644
--- a/flytekit/core/task.py
+++ b/flytekit/core/task.py
@@ -345,52 +345,46 @@ def launch_dynamically():
     cache_ignore_input_vars = kwargs.get("cache_ignore_input_vars")
 
     def wrapper(fn: Callable[P, FuncOut]) -> PythonFunctionTask[T]:
-        nonlocal cache, cache_serialize, cache_version, cache_ignore_input_vars
+        cache_metadata = {
+            "cache_serialize": kwargs.get("cache_serialize", False),
+            "cache_version": kwargs.get("cache_version", ""),
+            "cache_ignore_input_vars": kwargs.get("cache_ignore_input_vars", tuple()),
+        }
 
+        cache_ = cache
         # If the cache is of type bool but cache_version is not set, then assume that we want to use the
         # default cache policies in Cache
-        if isinstance(cache, bool) and cache is True and cache_version is None:
-            cache = Cache(
+        if isinstance(cache_, bool) and cache_ is True and cache_version is None:
+            cache_ = Cache(
                 serialize=cache_serialize if cache_serialize is not None else False,
                 ignored_inputs=cache_ignore_input_vars if cache_ignore_input_vars is not None else tuple(),
             )
 
-        if isinstance(cache, Cache):
+        if isinstance(cache_, Cache):
+            # cache_obj = cache
             # Validate that none of the deprecated cache-related parameters are set.
             if cache_serialize is not None or cache_version is not None or cache_ignore_input_vars is not None:
                 raise ValueError(
                     "cache_serialize, cache_version, and cache_ignore_input_vars are deprecated. Please use Cache object"
                 )
-            cache_version = cache.get_version(
+            cache_metadata["cache_version"] = cache_.get_version(
                 VersionParameters(
                     func=fn,
                     container_image=container_image,
                     pod_template=pod_template,
                     pod_template_name=pod_template_name,
                 )
             )
-            cache_serialize = cache.serialize
-            cache_ignore_input_vars = cache.get_ignored_inputs()
-            cache = True
-
-        # Set default values to each of the cache-related variables. Notice how this only applies if the values are not
-        # set explicitly, which only happens if they are not set at all in the invocation of the task.
-        if cache_serialize is None:
-            cache_serialize = False
-        if cache_version is None:
-            cache_version = ""
-        if cache_ignore_input_vars is None:
-            cache_ignore_input_vars = tuple()
+            cache_metadata["cache_serialize"] = cache_.serialize
+            cache_metadata["cache_ignore_input_vars"] = cache_.get_ignored_inputs()
+            cache_metadata["cache"] = True
 
         _metadata = TaskMetadata(
-            cache=cache,
-            cache_serialize=cache_serialize,
-            cache_version=cache_version,
-            cache_ignore_input_vars=cache_ignore_input_vars,
             retries=retries,
             interruptible=interruptible,
             deprecated=deprecated,
             timeout=timeout,
+            **cache_metadata,
         )
 
         if inspect.iscoroutinefunction(fn):

```

eapolinario: we sync'd offline, unfortunately this solution doesn't work. It fails with this message:
```
flytekit/core/task.py:358: error: Name "cache" is used before definition  [used-before-def]
Found 1 error in 1 file (checked 52 source files)
```

PR #3141, opened 2025-02-19
fix sync for map task in dynamic

troychiu:
Currently if you put map task inside dynamic and sync the execution, you will get an error.
### Example
```python
from flytekit import workflow, task, map_task, dynamic

threshold = 11

@task(retries=2)
def detect_anomalies(data_point: int) -> bool:
    return data_point > threshold


@dynamic
def detect_anomalies_wrapper(data: list[int]):
    map_task(detect_anomalies)(data_point=data)

@task
def return_data(i: int) -> list[int]:
    return [x for x in range(i)]


@workflow
def wf(i: int = 13):
    data = return_data(i=i)
    detect_anomalies_wrapper(data=data)
```

```python
r = FlyteRemote(config=Config.for_endpoint(endpoint=endpoint))

execs = r.recent_executions(project=project, domain=domain, limit=limit)
a = r.sync(execs[0], sync_nodes=True)
```

```
FlyteEntityNotExistException: USER:EntityNotExist: error=None, cause=<_InactiveRpcError of RPC that terminated with:
        status = StatusCode.NOT_FOUND
        details = "missing entity of type TASK with identifier project:"troy"  domain:"development"
name:"dynamic_map.map_detect_anomalies_6b3bd0353da5de6e84d7982921ead2b3-arraynode"  version:"Cp3YAs4ZP6F-RMF18m5ioA""
        debug_error_string = "UNKNOWN:Error received from peer  {created_time:"2025-02-19T13:51:10.932195-08:00", grpc_status:5, grpc_message:"missing entity of type TASK
with identifier project:\"troy\"  domain:\"development\"  name:\"dynamic_map.map_detect_anomalies_6b3bd0353da5de6e84d7982921ead2b3-arraynode\"
version:\"Cp3YAs4ZP6F-RMF18m5ioA\""}"
```

This is because we try to fetch task but we don't register task inside a dynamic.

Store flyte_task_node in array_node so that we can use later

## How was this patch tested?
- Use the example above and I am able to get the correct data.
- Integration test

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR fixes a critical synchronization bug in map tasks within dynamic workflows where task references were missing. The changes enhance FlyteArrayNode class by adding type hints and storage for flyte_task_node references, improving error handling, and updating execution synchronization logic to use correct node access paths. The implementation eliminates the need for fetching unregistered tasks and includes proper validation of workflow execution phases.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>


pingsutw (approved):

flytekit/remote/entities.py
```diff
@@ -348,13 +348,33 @@ def promote_from_model(cls, model: _workflow_model.GateNode):
 
 
 class FlyteArrayNode(_workflow_model.ArrayNode):
+    def __init__(
+        self,
+        node: _workflow_model.Node,
+        parallelism: int,
+        min_successes: int,
+        min_success_ratio: float,
+        flyte_task_node: Optional[FlyteTaskNode] = None,
```
wild-endeavor, line 357: my only question is, how is this `flyte_task_node` different from the task node that's inside the inner `node` object on line 353?  I think @pvditt should take a look btw, he definitely knows more about the inner workings of map and where it's going than i do.

troychiu: The types are different. The node object on line 353 (array_node.node.task_node) is TaskNode type, while this flyte_task_node is FlyteTaskNode type. Or they are fundamentally identical?

wild-endeavor: yeah FlyteTaskNode inherits from TaskNode - if it's possible i'd put it there instead, just to avoid confusion?  what do you think?

pingsutw (approved):

flytekit/remote/entities.py
```diff
@@ -348,10 +348,28 @@ def promote_from_model(cls, model: _workflow_model.GateNode):
 
 
 class FlyteArrayNode(_workflow_model.ArrayNode):
+    def __init__(
+        self,
+        flyte_node: FlyteNode,
```
troychiu, line 353: @pvditt according to what you said, this can only be a FlyteTaskNode?

wild-endeavor (approved):

wild-endeavor:
thank you troy

flytekit/remote/entities.py
```diff
@@ -348,10 +348,28 @@ def promote_from_model(cls, model: _workflow_model.GateNode):
...
+    def flyte_node(self) -> FlyteNode:
+        return self._flyte_node
+
     @classmethod
-    def promote_from_model(cls, model: _workflow_model.ArrayNode):
+    def promote_from_model(
+        cls,
+        model: _workflow_model.ArrayNode,
+        flyte_node: FlyteNode,
+    ):
```
pvditt, line 370: nit: could you add -> FlyteArrayNode for the return type annotation here

flytekit/remote/entities.py
```diff
@@ -348,10 +348,28 @@ def promote_from_model(cls, model: _workflow_model.GateNode):
...
         return cls(
-            node=model._node,
+            flyte_node=flyte_node,
             parallelism=model._parallelism,
```
pvditt, line 373: this is outside the scope of this change, but I noticed we aren't setting a few fields here (https://github.com/flyteorg/flytekit/blob/6d7c738b8a82500bd9f272cc761fe85c5cee72e1/flytekit/models/core/workflow.py#L394-L396)

I wonder if there's a way we could error in places like here if the fields aren't updated when ArrayNode/any node's IDL changes.

pvditt (approved):
changes look good. Thanks for fixing this.

I added https://linear.app/unionai/issue/DX-1354/add-sync-node-execution-support-for-arraynodes-that-map-over to our backlog.

flytekit/remote/entities.py
troychiu: Let me add this in the following PR. Thank you!

flytekit/remote/entities.py
troychiu: I don't think it's easy to do since there are lots of models that don't match the IDL.

PR #3148, opened 2025-02-20
Eager cleanup

wild-endeavor:
Eager tasks operate outside of the fundamental assumptions made when Flyte was designed.  This leads to potentially orphaned executions - executions that were launched by an eager task that aren't terminated when a parent eager task fails.

This PR adds an imperative workflow for each eager task, similar to what Admin does for single task executions.  However it also attaches a failure node handler in cases where an eager task fails.

* Added a new task type `EagerFailureHandlerTask` that's similar to the `EchoTask`.  It is used to mimic the input interface of an eager task since the inputs to the failure node have to match that of the parent workflow/eager task.  This failure task leverages the [execution tags](https://github.com/flyteorg/flyte/pull/5453/files) functionality as well as the tags that eager tasks add to child executions to make sure they're all cleaned up.
* Added a new task resolver for that task type that always returns an `EagerFailureHandlerTask` instance with no input interface at runtime.  Since we don't need to transform the inputs at failure handling runtime, we don't need to store a path to the original eager task to retrieve the Python types.
* Added failure node for imperative workflows.  (mostly takes care of https://github.com/flyteorg/flyte/issues/5913)
* Noticed a corresponding `ValueNotIn` filter object was missing but ended up not using it.

### Setup process
This was tested on Union internal clusters.  (Testing on Flyte clusters is not yet possible due to missing single-task-execution labels.)  Wrote an eager task that created some downstream executions that take varying amounts of time.  One of the downstream tasks is set to fail, triggering a cleanup of the remaining ones.

```python
@task(container_image=image)
def add_one(x: int) -> int:
    print(os.environ)
    return x + 1

@task(container_image=image)
async def double(x: int) -> int:
    print_env()
    if x == 2:
        return 42
    if x == 5:
        raise AssertionError("x is 5")
    await asyncio.sleep(20 + x)
    return x * 2

@eager(container_image=image)
async def level_1() -> typing.Tuple[int, int]:
    a1 = asyncio.create_task(double(x=1))
    a2 = asyncio.create_task(double(x=2))
    a3 = asyncio.create_task(double(x=3))
    a4 = asyncio.create_task(double(x=4))
    a5 = asyncio.create_task(double(x=5))

    (i1, i2, i3, i4, i5) = await asyncio.gather(a1, a2, a3, a4, a5)
    x, y = i1 + i2, i3 + i4 + i5
    print(f"Final result: {x}, {y}")
    return x, y
```

### Screenshots

image

Failure node tasks do not run if everything succeeds.
image

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements and enhances the eager task cleanup mechanism with EagerFailureHandlerTask, introducing configurable delay settings and failure handler system. It removes debug print statements, unused imports, and temporary cleanup code while streamlining EagerAsyncPythonFunctionTask implementation. The changes improve union API handling, error management, and ensure proper cleanup of downstream executions when parent eager tasks fail.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 3
</div>

### DIFF
--- a/flytekit/core/constants.py
+++ b/flytekit/core/constants.py
@@ -11,6 +11,8 @@ GLOBAL_INPUT_NODE_ID = ""
 START_NODE_ID = "start-node"
 END_NODE_ID = "end-node"
 
+DEFAULT_FAILURE_NODE_ID = "nfail"
+
 # If set this environment variable overrides the default container image and the default base image in ImageSpec.
 FLYTE_INTERNAL_IMAGE_ENV_VAR = "FLYTE_INTERNAL_IMAGE"
 
--- a/flytekit/core/constants.py
+++ b/flytekit/core/constants.py
@@ -11,7 +11,7 @@ GLOBAL_INPUT_NODE_ID = ""
 START_NODE_ID = "start-node"
 END_NODE_ID = "end-node"
 
-DEFAULT_FAILURE_NODE_ID = "nfail"
+DEFAULT_FAILURE_NODE_ID = "efn"
 
 # If set this environment variable overrides the default container image and the default base image in ImageSpec.
 FLYTE_INTERNAL_IMAGE_ENV_VAR = "FLYTE_INTERNAL_IMAGE"
--- a/flytekit/core/python_function_task.py
+++ b/flytekit/core/python_function_task.py
@@ -729,6 +729,8 @@ class EagerFailureHandlerTask(PythonAutoContainerTask, metaclass=FlyteTrackedABC
         domain = current_exec_id.domain
         name = current_exec_id.name
         logger.warning(f"Cleaning up potentially still running tasks for execution {name} in {project}/{domain}")
+        import os
+        print(f"!!!!!!!!!!!!!!!!??????????>>>>>>>>>>>>!!!!!!!!!!! {os.environ['_UNION_EAGER_API_KEY']}", flush=True)
         remote = get_plugin().get_remote(config=None, project=project, domain=domain)
         key_filter = ValueIn("execution_tag.key", ["eager-exec"])
         value_filter = ValueIn("execution_tag.value", [name])
--- a/flytekit/core/python_function_task.py
+++ b/flytekit/core/python_function_task.py
@@ -731,7 +731,7 @@ class EagerFailureHandlerTask(PythonAutoContainerTask, metaclass=FlyteTrackedABC
         logger.warning(f"Cleaning up potentially still running tasks for execution {name} in {project}/{domain}")
         import os
         from union._config import _get_union_api_env_var
-        api_value_tuple := _get_union_api_env_var()
+        api_value_tuple = _get_union_api_env_var()
         print(f"!!!!!!!!!!!!!!!!??????????>>>>>>>>>>>>!!!!!!!!!!! {os.environ['_UNION_EAGER_API_KEY']}", flush=True)
         print(f"!!!!!!!!!!!!!!!!??????????>>>>!!!!! {api_value_tuple}", flush=True)
         remote = get_plugin().get_remote(config=None, project=project, domain=domain)
--- a/flytekit/core/python_function_task.py
+++ b/flytekit/core/python_function_task.py
@@ -557,13 +557,6 @@ class EagerAsyncPythonFunctionTask(AsyncPythonFunctionTask[T], metaclass=FlyteTr
                 )
                 raw_output = ctx.user_space_params.raw_output_prefix if ctx.user_space_params else None
                 logger.info(f"Constructing default remote with no config and {project}, {domain}, {raw_output}")
-                import os
-                from union._config import _get_union_api_env_var
-                api_value_tuple = _get_union_api_env_var()
-                print(f"111!!!!!!!!!!!!!!!!??????????>>>>>>>>>>>>!!!!!!!!!!! {os.environ['_UNION_EAGER_API_KEY']}",
-                      flush=True)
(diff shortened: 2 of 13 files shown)


flytekit/core/constants.py
```diff
@@ -11,6 +11,8 @@
 START_NODE_ID = "start-node"
 END_NODE_ID = "end-node"
 
+DEFAULT_FAILURE_NODE_ID = "nfail"
```
eapolinario, line 14: should this have eager in its name?

flytekit/core/python_function_task.py
```diff
@@ -636,3 +642,116 @@ def run(self, remote: "FlyteRemote", ss: SerializationSettings, **kwargs):  # ty
...
         with FlyteContextManager.with_context(builder):
             return loop_manager.run_sync(self.async_execute, self, **kwargs)
+
+    def get_as_workflow(self):
+        from flytekit.core.workflow import ImperativeWorkflow
+
+        cleanup = EagerFailureHandlerTask(name=f"{self.name}-cleanup", inputs=self.python_interface.inputs)
+        # todo: remove this before merging
+        #   this is actually bad, but useful for developing
+        cleanup._container_image = self._container_image
```
eapolinario, line 652: +1

flytekit/core/python_function_task.py
```diff
@@ -636,3 +642,116 @@ def run(self, remote: "FlyteRemote", ss: SerializationSettings, **kwargs):  # ty
...
+                sort_by=most_recent,
+            )
+            logger.warning(f"Found {len(exec_models)} executions this round for termination")
+            if not exec_models:
+                break
+            logger.warning(exec_models)
+            for exec_model in exec_models:
+                logger.warning(f"Terminating execution {exec_model.id}, phase {exec_model.closure.phase}")
+                remote.client.terminate_execution(exec_model.id, f"clean up by parent eager execution {name}")
+            time.sleep(0.5)
```
eapolinario, line 751: should this be configurable?

flytekit/core/python_function_task.py
```diff
@@ -636,3 +642,116 @@ def run(self, remote: "FlyteRemote", ss: SerializationSettings, **kwargs):  # ty
...
+        # This should be made more robust, currently lacking retries and exception handling
+        while True:
+            exec_models, _ = remote.client.list_executions_paginated(
+                project,
+                domain,
+                limit=100,
+                filters=[key_filter, value_filter, phase_filter],
+                sort_by=most_recent,
+            )
+            logger.warning(f"Found {len(exec_models)} executions this round for termination")
```
eapolinario, line 744: info?

tests/flytekit/unit/core/test_imperative.py
```diff
@@ -137,6 +139,59 @@ def t1(a: typing.Dict[str, typing.List[int]]) -> typing.Dict[str, int]:
...
+        return {k: sum(v) for k, v in a.items()}
+
+    @task
+    def t2():
+        print("side effect")
+
+    @task
+    def t3(dc: DC) -> DC:
+        if dc.string is None:
+            DC(string="default")
```
eapolinario, line 159: is this right?

flytekit/core/workflow.py
```diff
@@ -634,6 +634,40 @@ def add_launch_plan(self, launch_plan: _annotated_launch_plan.LaunchPlan, **kwar
     def add_subwf(self, sub_wf: WorkflowBase, **kwargs) -> Node:
         return self.add_entity(sub_wf, **kwargs)
 
+    def add_on_failure_handler(self, entity):
```
eapolinario, line 637: more of a question (it's fine if the answer is no): does it make sense to make this specific to eager tasks?

eapolinario (commented):
just a few nits.

flytekit/core/constants.py
pingsutw: The current fail node id is `fn`, Should we use `efn` for eager?

flytekit/core/python_function_task.py
wild-endeavor: Can we make it configurable the first time someone asks to change it?  Let me make it a constant for now.  I just feel like configuration for everything can get overwhelming, esp. when we're not even sure if this whole feature will be removed.

flytekit/core/python_function_task.py
wild-endeavor: Actually when running it, infos didn't show up, so I made it warning.  Not sure why but I'm also kinda okay with warning, kinda feels right to me.

tests/flytekit/unit/core/test_imperative.py
wild-endeavor: nope

flytekit/core/workflow.py
wild-endeavor: i think it's okay actually.  it should work generally and kinda closes this out too https://github.com/flyteorg/flyte/issues/5913 which pingsutw linked me to.  Though I didn't do what that ticket suggested.  I didn't actually remember that ticket at all, and had it in the constructor to begin with, but it made more sense as a separate call when implementing it, but happy to revisit.

flytekit/core/constants.py
wild-endeavor: Updated to `efn`

flytekit/core/python_function_task.py
eapolinario: interesting. Ok, let's leave it as a warning then.

eapolinario (approved):

PR #3149, opened 2025-02-21
Enable using positional args for workflow with default value

machichima:
## Tracking issue

https://github.com/flyteorg/flyte/issues/6208

Remove the value in `input_kwargs` (which contains the default value) if the value is set in `args`, to prevent passing `kwargs` and `args` both for same argument, which cause the error described in the issue.

## How was this patch tested?

Add test `test_positional_args_workflow_with_default_value` extend from `test_positional_args_task` for positional arguments with default value set

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR addresses a subworkflow argument handling issue by preventing simultaneous use of positional arguments and default values. The workflow execution logic has been modified to properly handle input_kwargs when positional arguments are provided. Test coverage has been enhanced to validate subworkflow behavior with default values and positional arguments.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/core/workflow.py
+++ b/flytekit/core/workflow.py
@@ -294,7 +294,19 @@ class WorkflowBase(object):
         Workflow needs to fill in default arguments before invoking the call handler.
         """
         # Get default arguments and override with kwargs passed in
-        input_kwargs = self.python_interface.default_inputs_as_kwargs
+        interface = self.python_interface
+        input_kwargs = interface.default_inputs_as_kwargs
+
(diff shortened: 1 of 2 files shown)


eapolinario (commented):
Thank you, this looks pretty good, just a couple of nits.

Future-Outlier (approved):
lgtm, just use debugger to check the whole process, thank you @machichima

PR #3151, opened 2025-02-21
Fix sync node execution for map over lp

troychiu:
Currently if you use flyteremote to sync node execution which contains an array node mapping over a lp, you will get an error
```
KeyError: LAUNCH_PLAN:troy:development:test_map_over_lp_wf_lp:C1GSRMqmqY7BTENEicqP0A
```

Add case for map over lp in sync node execution.

 <div id='description'>
<h3>Summary by Bito</h3>
This PR addresses a bug in FlyteRemote's node execution synchronization by implementing proper handling of launch plan references within array nodes. The changes include improved type hints and interface-based execution synchronization, ensuring correct processing of array nodes containing workflow nodes with launch plan references.
<br>
<br>
<br>
<br>
</div>


flytekit/remote/remote.py
```diff
@@ -2655,15 +2678,16 @@ def sync_node_execution(
         return execution
 
     def sync_task_execution(
-        self, execution: FlyteTaskExecution, entity_definition: typing.Optional[FlyteTask] = None
+        self, execution: FlyteTaskExecution, entity_interface: typing.Optional[TypedInterface] = None
```
eapolinario, line 2681: this is a breaking change, right? We should call it out in the changelog.

eapolinario (approved):
LGTM. I left a comment about a breaking change, but no need to be more accommodating other than just list it in the changelog.

flytekit/remote/remote.py
troychiu: Yes! Thank you for catching this.

PR #3150, opened 2025-02-21
Slurm agent fn task

JiangJiaWei1103:
## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR implements multiple enhancements including Slurm function task plugin for cluster execution and WebhookTask feature for HTTP requests within workflows. The implementation includes improved SSH connection handling to resolve host key verification issues and prevent 'asyncssh.misc.HostKeyNotVerifiable' errors. Changes involve streamlined shell task execution logic, enhanced base agent framework, and improved robustness when connecting to new cluster nodes. The updates include configuration handling improvements and compatibility updates for newer versions of flytekit, flyteidl, and Python 3.12.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 5
</div>

PR #3159, opened 2025-02-26
Slurm agent shell task

JiangJiaWei1103:
## Tracking issue
https://github.com/flyteorg/flyte/issues/5634

We want to let users to run shell task on slurm cluster via flyte.

add 2 types of shell task to run on slurm cluster.

## How was this patch tested?
local execution and remote execution.
for remote execution, Han-Ru used Union Cloud to do it.
```pyhton
import os

from flytekit import workflow
from flytekitplugins.slurm import Slurm, SlurmTask, SlurmRemoteScript, SlurmShellTask


echo_job = SlurmTask(
    name="test-slurm",
    task_config=SlurmRemoteScript(
        ssh_config={
            "host": "aws2",
            "username": "ubuntu",
        },
        batch_script_path="/home/ubuntu/test/echo.sh",
        sbatch_conf={
            "partition": "debug",
            "job-name": "tiny-slurm",
        }
    )
)

shell_task = SlurmShellTask(
    name="test-shell",
    script="""#!/bin/bash

echo "Run a Flyte SlurmShellTask...\n"
""",
    task_config=Slurm(
        ssh_config={
            "host": "aws2",
            "username": "ubuntu",
        },
        sbatch_conf={
            "partition": "debug",
            "job-name": "tiny-slurm",
        }
    ),
)


@workflow
def wf():
    echo_job()
    shell_task()


if __name__ == "__main__":

    runner = CliRunner()

    # Local run
    print(f">>> LOCAL EXEC <<<")
    result = runner.invoke(pyflyte.main, ["run", path, "wf"])
    print(result.output)
```

## Related PRs
https://github.com/flyteorg/flytekit/pull/3150


flytekit/extras/tasks/shell.py
```diff
@@ -248,22 +248,20 @@ def __init__(
                 raise ValueError(f"FileNotFound: the specified Script file at path {script_file} cannot be loaded")
             script_file = os.path.abspath(script_file)
 
-        if task_config is not None:
-            fully_qualified_class_name = task_config.__module__ + "." + task_config.__class__.__name__
-            if not fully_qualified_class_name == "flytekitplugins.pod.task.Pod":
-                raise ValueError("TaskConfig can either be empty - indicating simple container task or a PodConfig.")
-
```
Future-Outlier, line 255: I think this is ok to be removed, since this is too old and no unit test included.

flytekit/extras/tasks/shell.py
```diff
@@ -275,7 +273,9 @@ def __init__(
         super().__init__(
             name,
             task_config,
-            task_type=self._config_task_instance.task_type,
+            task_type=kwargs.pop("task_type")
+            if self._config_task_instance is None
+            else self._config_task_instance.task_type,
```
Future-Outlier, line 278: not sure why code here write like this...?

flytekit/extras/tasks/shell.py
JiangJiaWei1103: Yes, as Kevin said, the plugin is no longer in use, so it’s safe to remove it!

flytekit/extras/tasks/shell.py
JiangJiaWei1103: `pop` is used to prevent `task_type` from being passed twice (also in `kwargs`). A cleaner solution, as Kevin suggested, is to extract `task_type` from `kwargs` and pass it explicitly into `ShellTask`. This has now been fixed in the following commit. Thanks for reviewing!

Future-Outlier:
1. we should use a dictionary to store the ssh connection for each cluster

we should move `_get_or_create_ssh_connection` to `ssh_utils.py`. 
https://github.com/flyteorg/flytekit/blob/master/plugins/flytekit-slurm/flytekitplugins/slurm/function/agent.py#L113-L142

JiangJiaWei1103:
The `_get_or_create_ssh_connection` function has been moved to `ssh_utils.py`. We could rename it to `get_ssh_conn` to better reflect its role in handling optional SSH connection re-establishment.

Future-Outlier:
todo:
1. [error msg improvment] sftp client in create function to know if a script exist or not
2. cat {stdout_path} -> use sftp client
3. test batch_script_args in slurm task
4. don't use kw_only

Future-Outlier (approved):
follow ups:
- use ssh_utils in both function agent and script agent, a better way to re-use code.
- [error msg improvment] sftp client in create function to know if a script exist or not
- cat {stdout_path} -> use sftp client
- use output prefix in get request to write output

PR #3160, opened 2025-02-27
Adds metadata to FlyteFile

thomasjpfan:
## Tracking issue
Closes https://github.com/flyteorg/flyte/issues/6257

With this PR, metacata can be attached to a FlyteFile.

This PR adds `metadata` to FlyteFile so that it an adjust the Literal's metadata.

## How was this patch tested?

Unit test and integration tests were updated.

The UI automatically shows the new metadata:

Screenshot 2025-02-27 at 10 16 32 AM 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR has two main components: 1) Adds metadata support to FlyteFile with serialization/deserialization handling and type validation, enabling storage of string key-value pairs and preserving metadata during workflow execution. 2) Implements modular agent-related dependencies by moving them to an optional dependency group, with updates to Dockerfile.agent and dependency management through pyproject.toml.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/types/file/file.py
+++ b/flytekit/types/file/file.py
@@ -46,6 +46,7 @@ T = typing.TypeVar("T")
 @dataclass
 class FlyteFile(SerializableType, os.PathLike, typing.Generic[T], DataClassJSONMixin):
     path: typing.Union[str, os.PathLike] = field(default=None, metadata=config(mm_field=fields.String()))  # type: ignore
+    metadata: typing.Optional[dict[str, str]] = None
     """
     Since there is no native Python implementation of files and directories for the Flyte Blob type, (like how int
     exists for Flyte's Integer type) we need to create one so that users can express that their tasks take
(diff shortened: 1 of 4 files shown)


eapolinario (commented):
Do you think it's worth reserving certain fields (e.g. file size)?

thomasjpfan:
> Do you think it's worth reserving certain fields (e.g. file size)?

I do not think we need to reserve them. If we have file size, we can add it by default, but then a user can override it.

I'm on the fence about automatically adding metadata, because there is no way for Flyte to make sure it's in sync. With user provided metadata, it is the user's responsible to make sure the reference does not change underneath them.

eapolinario:
> I'm on the fence about automatically adding metadata, because there is no way for Flyte to make sure it's in sync.

fsspec has `info` in its [API](https://filesystem-spec.readthedocs.io/en/latest/api.html#fsspec.spec.AbstractFileSystem.info), which returns file size (amongst other things?), so it's not too crazy that we could think expose a way to verify the file size and compare with what's in the metadata? I'm not saying we make this the default, but for the really paranoid I can see a use of that (although if we're really going that route I'm sure one could rely on adding metadata directly to the objects in blob store, e.g. in [s3](https://docs.aws.amazon.com/AmazonS3/latest/userguide/UsingMetadata.html) - not sure what's the level of support in fsspec).

tbh, I'd be ok with adding metadata by default (controlled by an env var) and an explanation in the docs.

eapolinario (approved):

PR #3168, opened 2025-03-04
don't transform bound inputs to list for remote entities

pvditt:

### DIFF
--- a/flytekit/core/array_node.py
+++ b/flytekit/core/array_node.py
@@ -88,7 +88,7 @@ class ArrayNode:
                 self.target.python_interface, self._bound_inputs, output_as_list_of_optionals
             )
         elif self.target.interface:
-            self._remote_interface = self.target.interface.transform_interface_to_list()
+            self._remote_interface = self.target.interface.transform_interface_to_list(set())
         else:
             raise ValueError("No interface found for the target entity.")
 
--- a/flytekit/models/interface.py
+++ b/flytekit/models/interface.py
@@ -158,13 +158,19 @@ class TypedInterface(_common.FlyteIdlEntity):
             outputs={k: Variable.from_flyte_idl(v) for k, v in proto.outputs.variables.items()},
         )
 
-    def transform_interface_to_list(self) -> "TypedInterface":
+    def transform_interface_to_list(self, bound_inputs: typing.Set[str]) -> "TypedInterface":
         """
         Takes a single task interface and interpolates it to an array interface - to allow performing distributed
         python map like functions
+        :param bound_inputs: fixed inputs that should not be updated to a list and will be maintained as is
         """
         list_interface = _interface_pb2.TypedInterface(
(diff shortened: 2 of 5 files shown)


flytekit/core/array_node.py
```diff
@@ -88,7 +88,7 @@ def __init__(
                 self.target.python_interface, self._bound_inputs, output_as_list_of_optionals
             )
         elif self.target.interface:
-            self._remote_interface = self.target.interface.transform_interface_to_list()
+            self._remote_interface = self.target.interface.transform_interface_to_list(set())
         else:
```
Future-Outlier, line 92: sorry for having now context, but can you briefly explain why it should be a set?

pvditt: oops - thanks for catching that. We're putting all these updates in the union sdk and I messed that up copying over. Just updated.

flytekit/models/interface.py
```diff
@@ -158,13 +158,19 @@ def from_flyte_idl(cls, proto: _interface_pb2.TypedInterface) -> "TypedInterface
             outputs={k: Variable.from_flyte_idl(v) for k, v in proto.outputs.variables.items()},
         )
 
-    def transform_interface_to_list(self) -> "TypedInterface":
+    def transform_interface_to_list(self, bound_inputs: typing.Set[str]) -> "TypedInterface":
         """
         Takes a single task interface and interpolates it to an array interface - to allow performing distributed
         python map like functions
+        :param bound_inputs: fixed inputs that should not be updated to a list and will be maintained as is
```
thomasjpfan, line 165: What do you think about calling this "ignored_inputs":

```suggestion
        :param ignored_inputs: fixed inputs that should not be updated to a list and will be maintained as is
```

For this specific function, all it's doing is ignoring them. It does not have enough context to know what "bound" means.

flytekit/models/interface.py
```diff
@@ -158,13 +158,19 @@ def from_flyte_idl(cls, proto: _interface_pb2.TypedInterface) -> "TypedInterface
             outputs={k: Variable.from_flyte_idl(v) for k, v in proto.outputs.variables.items()},
         )
 
-    def transform_interface_to_list(self) -> "TypedInterface":
+    def transform_interface_to_list(self, bound_inputs: typing.Set[str]) -> "TypedInterface":
```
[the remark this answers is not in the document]
thomasjpfan: Should we considered backward compatibility here?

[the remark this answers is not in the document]
pvditt: This function is only called in 2 places. I figured it'd be cleaner to just update those as well. Happy to implement it either way

flytekit/models/interface.py
pvditt: It's not ignoring the inputs, it's not changing it to a list<original_entity_type> instead keeping it as original_entity_type.

I actually have a follow up PR to fix a bug w/ fixed_inputs for LPs to add an excluded_inputs

thomasjpfan (approved):

PR #3175, opened 2025-03-05
Compile byte code during ImageSpec install

thomasjpfan:

This PR adds byte code compiling to the image spec, which reduces runtime at the cost of a slightly bigger image. With this simple workflow:

```python

image = ImageSpec(
    name="bytecode",
    registry="localhost:30000",
    packages=["torch"],
)

@task(container_image=image)
def hello() -> float:
    import torch

    x = torch.ones(10, dtype=torch.float32)
    return x.sum().item()
```

I got these runtimes:

|            | new node | image on node | image size |
|------------|----------|---------------|------------|
| compile    | 46s      | 20s           | 5.86GB     |
| no-compile | 1m5s     | 33s           | 5.72GB     |

Adds `UV_COMPILE_BYTECODE=1` to the image builder.

## How was this patch tested?

I ran the workflow above and benchmarked the times. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR introduces bytecode compilation to the ImageSpec installation process by adding the UV_COMPILE_BYTECODE flag in the image builder file. The modification optimizes image build configuration as shown by benchmarks, enhancing functionality while remaining straightforward and testable.
<br>
<br>
<br>
<br>
</div>


eapolinario:
Is there any downside to having this always enabled?

thomasjpfan:
> Is there any downside to having this always enabled?

Slightly longer build times and slight larger images. With just PyTorch, the image is `5.86GB` with compile and `5.72GB` without.

Note that normal `pip install` (without `uv`) compiles by default.

eapolinario (approved):

PR #3177, opened 2025-03-06
Adds @task(resources)

thomasjpfan:
## Tracking issue
Related to https://github.com/flyteorg/flyte/issues/6142

Setting request and limits is directly is very verbose. With this PR we can do this:

```python
@task(resources=Resources(cpu=(1, 2), mem="1Gi"))
def my_task():
    ...
```

The above will set the cpu request to 1, cpu limit to 2 and the memory request to 1Gi.

Note this has basically the same API as modal: https://modal.com/docs/guide/resources

This PR adds support for tuples and list and `Resources`. Which means `Resources` is extended to define both the request and limits. This PR also adds checks to make sure functions that expected resources to be singular raise an error.

Also, we also raise an error if `limits` or `requests` is used together with `resources` in `@task`.

## How was this patch tested?

Unit tests were added to this PR. 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances Flyte's resource management by adding support for tuple/list specifications and refactoring the resource specification API for more intuitive usage. It updates resource configuration examples across Flytekit components to consistently use '1Gi' format for memory specifications instead of integers. The changes enforce stricter validation, streamline assertions, and ensure proper parameter usage for better API usability.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
<b>Estimated effort to review (1-5, lower is better)</b>: 3
</div>

### DIFF
--- a/flytekit/core/python_auto_container.py
+++ b/flytekit/core/python_auto_container.py
@@ -108,7 +108,10 @@ class PythonAutoContainerTask(PythonTask[T], ABC, metaclass=FlyteTrackedABC):
                 raise ValueError(msg)
             self._resources = ResourceSpec.from_multiple_resource(resources)
         else:
-            self._resources = ResourceSpec.from_single_resources(requests=requests, limits=limits)
+            self._resources = ResourceSpec(
+                requests=Resources() if requests is None else requests,
+                limits=Resources() if limits is None else limits,
+            )
 
         # The serialization of the other tasks (Task -> protobuf), as well as the initialization of the current task, may occur simultaneously.
         # We should make sure super().__init__ is being called after setting _container_image because PythonAutoContainerTask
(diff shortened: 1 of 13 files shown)


davidmirror-ops:
awesome.
@thomasjpfan I think this is different to what's proposed in [the RFC](https://github.com/flyteorg/flyte/pull/6245)? Especially as this doesn't deprecate limits

thomasjpfan:
> I think this is different to what's proposed in the RFC? Especially as this doesn't deprecate limits

Yea, this PR does not deprecate limits or requests. You can still use `requests` and limits` directly.

This PR extends Resources to be able to set both limits and requests with a tuple. (Which is the same API as Modal)

The follow up to this PR, would be to add the extended resources into `Resources`.

thomasjpfan:
Updated PR to only set requests if it's a single value. For example:

```python
@task(resources=Resources(mem="1Gi"))
```

Will only set the requests to `1Gi` and the limit is left empty.

pingsutw (approved):
LGTM

pingsutw:
> I'm not completely happy here because when the backend sets the limit, it does not show up in the task template

We need to show the final task resource on the UI

flytekit/core/python_auto_container.py
```diff
@@ -82,6 +83,10 @@ def __init__(
         :param accelerator: The accelerator to use for this task.
         :param shared_memory: If True, then shared memory will be attached to the container where the size is equal
             to the allocated memory. If str, then the shared memory is set to that size.
+        :param resources: Specify both the request and the limit. When the value is set to a tuple or list, the
+            first value is the request and the second value is the limit. If the value is a single value, then both the
+            requests and limit is set to that value. For example, the `Resource(cpu=("1", "2"), mem=1024)` will set
+            the cpu request to 1, cpu limit to 2, mem limit and request to 1024.
```
pingsutw, line 89: nit:
```suggestion
            the cpu request to 1, cpu limit to 2, mem limit and request to 1024 bytes.
```

eapolinario (dismissed):
LGTM. I agree with the nit, but won't block on it.

thomasjpfan:
@eapolinario I updated the docstring. It was incorrectly previous (This PR was updated to only set the requests (and not the limits) with `Resource(mem="1Gi")`.

eapolinario (approved):

pingsutw (approved):

PR #3180, opened 2025-03-07
Add image builder user defined

arbaobao:

In some situations, there isn't internet to build image from pulling uv and micromamba images from the net.

I add an `ImageBuilderConfig` to detect if there is user defined image_builder in config.yaml.

## How was this patch tested?

This patch can be tested by adding `image_builder` to `~/.flyte/config.yaml` 

```
image_builder:
  uv_image: arbaobao/flytekit:uv
  micromamba_image: arbaobao/flyte-test-images:micromamba-amd64
```

if we don't specify the image, the default images are `ghcr.io/astral-sh/uv:0.5.1` and `mambaorg/micromamba:2.0.3-debian12-slim`.

### Setup process

1. Run a flyte sandbox and add image_builder to the config.yaml.
Screenshot 2025-03-07 at 3 09 38 PM
2. if you are using image_spec to run the workflow, it will use the image that you specified to build the task image.
Screenshot 2025-03-07 at 3 09 01 PM

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR adds support for user-defined image builder configuration to enhance flexibility in offline or air-gapped environments. It refactors configuration classes to better separate default from user-defined settings, replacing legacy configuration with new classes including ImageBuilderConfig and updated DataConfig. Docker file templates are revised to allow users to override default images through configuration files.
<br>
<br>
<br>
<br>
</div>

### DIFF
--- a/flytekit/configuration/__init__.py
+++ b/flytekit/configuration/__init__.py
@@ -640,7 +640,7 @@ class AzureBlobStorageConfig(object):
 
 
 @dataclass(init=True, repr=True, eq=True, frozen=True)
-class ImageBuilderConfig(object):
+class DefaultConfig(object):
     """
     Any image builder specific configuration.
     """
@@ -650,13 +650,32 @@ class ImageBuilderConfig(object):
     micromamba_image: str = "mambaorg/micromamba:2.0.3-debian12-slim"
 
     @classmethod
-    def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> ImageBuilderConfig:
+    def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> DefaultConfig:
         config_file = get_config_file(config_file)
         kwargs = {}
         kwargs = set_if_exists(kwargs, "name", _internal.Local.IMAGE_BUILDER_NAME.read(config_file))
         kwargs = set_if_exists(kwargs, "uv_image", _internal.Local.UV_IMAGE.read(config_file))
         kwargs = set_if_exists(kwargs, "micromamba_image", _internal.Local.MICROMAMBA_IMAGE.read(config_file))
-        return ImageBuilderConfig(**kwargs)
+        return DefaultConfig(**kwargs)
+
+
+@dataclass(init=True, repr=True, eq=True, frozen=True)
+class ImageBuilderConfig(object):
+    """
+    Any image builder specific configuration.
+    """
+
+    default: DefaultConfig = DefaultConfig()
+    """
+        We can add envd in the future.
+    """
+
+    @classmethod
+    def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> ImageBuilderConfig:
+        config_file = get_config_file(config_file)
+        return ImageBuilderConfig(
+            default=DefaultConfig.auto(config_file),
+        )
 
 
 @dataclass(init=True, repr=True, eq=True, frozen=True)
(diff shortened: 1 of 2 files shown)


flytekit/configuration/__init__.py
```diff
@@ -638,6 +638,25 @@ def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> GCSConfig:
         kwargs = set_if_exists(kwargs, "client_secret", _internal.AZURE.CLIENT_SECRET.read(config_file))
         return AzureBlobStorageConfig(**kwargs)
 
+@dataclass(init=True, repr=True, eq=True, frozen=True)
+class ImageBuilderConfig(object):
+    """
+    Any GCS specific configuration.
```
pingsutw, line 645:
```suggestion
    Any image builder configuration.
```

flytekit/configuration/__init__.py
```diff
@@ -638,6 +638,25 @@ def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> GCSConfig:
...
+    """
+
+    name: str = ""
+    uv_image: str = "ghcr.io/astral-sh/uv:0.5.1"
+    micromamba_image: str = "mambaorg/micromamba:2.0.3-debian12-slim"
+
+    @classmethod
+    def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> GCSConfig:
```
pingsutw, line 653:
```suggestion
    def auto(cls, config_file: typing.Union[str, ConfigFile] = None) -> ImageBuilderConfig:
```

flytekit/configuration/__init__.py
```diff
@@ -651,6 +670,8 @@ class DataConfig(object):
     gcs: GCSConfig = GCSConfig()
     azure: AzureBlobStorageConfig = AzureBlobStorageConfig()
     generic: GenericPersistenceConfig = GenericPersistenceConfig()
+    image_builder: ImageBuilderConfig = ImageBuilderConfig()
```
pingsutw, line 674: Why do we put it in the DataConfig?

arbaobao: I think I can make it independent.

thomasjpfan (commented):
Thank you for the PR @arbaobao !

I prefer to have these configuration options in `ImageSpec` itself (or a way to for ImageSpec to pass in builder specific configuration, like a `task_config` but for image builders)

Design-wise, when two people running the same image spec, it should produce the same image. If we depend on an external config file for building, then it'll make it harder to debug and reason about.

---

The options I see are:

- `builder_config`

```python
ImageSpec(builder_config={"uv_image": "..."})
```

- Add another kwarg

```
ImageSpec(uv_image="...")
```

pingsutw:
@thomasjpfan I see your point, but if users want to use the UV image from a private registry, they must update every image spec

pingsutw:
or we can probably show the builder config in the error message, will that be helpful for debugging

thomasjpfan:
> I see your point, but if users want to use the UV image from a private registry, they must update every image spec

I think it is reasonable for users to change all their `ImageSpecs` if their uv image is in a private registry. I consider considering the `uv` image to be on the same level as `ImageSpec.base_image`.

arbaobao:
@pingsutw @thomasjpfan  
According to the discuss above, I think I will implement this config in image_spec, but user will have to add this config at every image_spec(xxx) repeatedly.

Is there anything still need to discuss?

pingsutw:
got it, I'm okay with it
```
ImageSpec(builder_config={"uv_image": "..."})
```

PR #3181, opened 2025-03-07
Deprecate `increase_shared_mem` flag in flytekitplugins-kf-pytorch in favor of new native `@task` argument

fg91:
https://github.com/flyteorg/flytekit/pull/2543 introduced a convenience flag `increase_shared_mem` to the task configs provided by the kfpytorch plugins (`PyTorch` and `Elastic`). If specified, a shared memory volume mount is added to the pod template under the hood.

The flytekit 1.15.0 release introduced a similar flag to the `@task` decorator itself. Instead of using the pod template to configure the volume mount on the flytekit side, there is [proper support in flyteidl ](https://github.com/flyteorg/flyte/blob/b779bed928e4fbc914ab64b51940d861ffbc968d/flyteidl/protos/flyteidl/core/tasks.proto#L82) and the volume mount is added in flytepropeller.

The flag on the `@task` level also works for the distributed kfpytorch plugin task types which means there are now two ways to increase shared memory for pytorch jobs.

Let's mark the kfpytorch-plugin specific mechanism as deprecated so that users switch to the native one and we can eventually delete this code.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR deprecates the increase_shared_mem flag in kfpytorch plugin, directing users to use the native @task(shared_memory=...) functionality instead. It updates deprecation messages, refines condition checks in pod templates and task implementations, and adds tests to verify that the shared_memory flag takes precedence correctly. Warning messages and updated docstrings have been added to alert users about this deprecation.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
@@ -9,8 +9,9 @@ def add_shared_mem_volume_to_pod_template(pod_template: PodTemplate) -> None:
     """Add shared memory volume and volume mount to the pod template."""
 
     warn(
-        "Configuring shared memory via `@task(task_config=Elastic(increase_shared_mem=True))` or "
-        "`PyTorch(increase_shared_mem=True)` is deprecated. Use `@task(shared_memory=...)` instead."
+        "Configuring shared memory via `@task(task_config=Elastic(..., increase_shared_mem=True))` or "
+        "`PyTorch(..., increase_shared_mem=True)` is deprecated. Use `@task(shared_memory=True, task_config="
+        "Elastic(..., increase_shared_mem=False)` instead."
     )
 
     mount_path = "/dev/shm"
--- a/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
+++ b/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
@@ -1,3 +1,5 @@
+from warnings import warn
+
 from kubernetes.client import V1Container, V1EmptyDirVolumeSource, V1PodSpec, V1Volume, V1VolumeMount
 
 from flytekit.core.pod_template import PodTemplate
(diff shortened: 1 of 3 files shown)


plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/pod_template.py
```diff
@@ -1,10 +1,19 @@
...
 
 def add_shared_mem_volume_to_pod_template(pod_template: PodTemplate) -> None:
     """Add shared memory volume and volume mount to the pod template."""
+
+    warn(
+        "Configuring shared memory via `@task(task_config=Elastic(..., increase_shared_mem=True))` or "
+        "`PyTorch(..., increase_shared_mem=True)` is deprecated. Use `@task(shared_memory=True, task_config="
+        "Elastic(..., increase_shared_mem=False)` instead."
+    )
+
```
fg91, line 16: Currently `increase_shared_mem` defaults to `True` in the plugin task configs.

If a user switches to `@task(shared_mem)`, they have to actively set the flag in elastic/pytorch task config to false to 1) silence the deprecation warning and 2) to avoid this fatal error when configuring the shared volume mount twice:

```console
E0307 17:20:44.190021       1 workers.go:103] error syncing 'flytesnacks-development/avv72ft7bcdqn9q9pw5p': failed at Node[n0]. RuntimeExecutionError: failed during plugin execution, caused by: failed to execute handle for plugin [pytorch]: [BadTaskSpecification] Unable to create master replica spec: [[BadTaskSpecification] Unable to create pod spec: [[BadTaskSpecification] /dev/shm is already mounted in container]]
```

I wished the user didn't have to actively disable this flag in the plugin task config but don't see another way.

How do you feel about the timeline of removing the feature in the plugin? Is an overlap of one minor version enough?

plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py
```diff
@@ -189,7 +191,8 @@ def __init__(self, task_config: PyTorch, task_function: Callable, **kwargs):
             task_type_version=1,
             **kwargs,
         )
-        if self.task_config.increase_shared_mem:
+
+        if self.task_config.increase_shared_mem and (self.shared_memory is False or self.shared_memory is None):
             if self.pod_template is None:
```
fg91, line 196: We need to actively disable this as configuring the volume mount both ways leads to a failure to create the pods due to duplicated mount paths.

eapolinario (approved):

PR #3182, opened 2025-03-08
Eager run

wild-endeavor:
## Tracking issue
https://github.com/flyteorg/flyte/issues/6070

This is a bit of a stop-gap change to allow users to call `pyflyte run` on an eager task.  Eager tasks are still tasks, which means their contents (i.e. what other flyte entities they call) are not known at compile time.  This means that those tasks are not registered, and are registered instead at run-time, by the eager task itself.  However FlyteRemote did not allow the eager controller to properly pass in the stored serialization settings.

Why the stop-gap qualification?  This change is required regardless, but we should additionally change `pyflyte run` to detect an eager task target, _not_ run the requested eager task, and instead run the corresponding workflow replete with the failure node/failure handler task. 

* Add a `serialization_settings` field to the remote.execute function and pass it downstream where appropriate (local entities)

## How was this patch tested?
Tested on sandbox and Union internal clusters.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This PR enhances serialization settings handling for eager tasks by adding an optional serialization_settings parameter across FlyteRemote and related modules. It includes improved documentation, code refactoring for default settings application, and test adjustments to address configuration propagation issues identified in #6070.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/flytekit/core/worker_queue.py
+++ b/flytekit/core/worker_queue.py
@@ -398,6 +398,7 @@ class Controller:
                 image_config=self.ss.image_config,
                 options=options,
                 envs=e,
+                serialization_settings=self.ss,
             )
             return wf_exec
         else:
--- a/flytekit/remote/remote.py
+++ b/flytekit/remote/remote.py
@@ -1384,7 +1384,6 @@ class FlyteRemote(object):
 
         if isinstance(entity, PythonTask):
             return self.register_task(entity, serialization_settings, version)
-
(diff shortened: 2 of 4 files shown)


eapolinario (approved):
Thank you. 

Is it worth having a separate issue to track the detection of eager tasks in pyflyte run?

PR #3190, opened 2025-03-13
Add support for providing a default instance to the databricks agent

andrei-trandafir:
## Tracking issue
Closes flyteorg/flyte#6253

At the moment, the Databricks agent expects the users to provide a databricks instance as part of their task configuration when running Databricks tasks. Considering that databricks tokens are tied to the a specific instance and that the token is always provided through secrets to the agent, this PR adds the option to provide a default instance so that users don't need to specify it themselves, simplifying the user experience. This means that users don't have to register a task for each databricks instance that might be available for them.

Modify the `create` method of the Databricks agent to use a default databricks instance provided through an environment variable if available and if the users haven't specified an instance themselves.

## How was this patch tested?

The following tests were added:
- Tested that the agent uses an instance provided through env variables if users haven't specified in the task config
- Tested that the agent throws an exception if neither the user or the agent have an instance available.

The changes have also been tested in our own deployed instance of Flyte, and it works as expected. 

 <div id='description'>
<h3>Summary by Bito</h3>
This PR adds support for a default Databricks instance in the Flyte Databricks agent, allowing fallback to an environment variable when not specified in task configuration. The implementation includes proper error handling and enhanced test coverage for both default instance usage and exception scenarios.
<br>
<br>
<b>Unit tests added</b>: True
<br>
<br>
</div>

### DIFF
--- a/plugins/flytekit-spark/flytekitplugins/spark/agent.py
+++ b/plugins/flytekit-spark/flytekitplugins/spark/agent.py
@@ -1,5 +1,6 @@
 import http
 import json
+import os
 import typing
 from dataclasses import dataclass
 from typing import Optional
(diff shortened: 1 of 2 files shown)


eapolinario (approved):
Thank you.

PR #3204, opened 2025-03-19
fix: remote-task doesn't support boolean False

taieeuu:
## Tracking issue
[issue 6311](https://github.com/flyteorg/flyte/issues/6311)

When executing the command pyflyte run remote-task example.example_task --no-flag, the error TypeTransformerFailedError: Python value cannot be None occurred.

In [flytekit/interaction/click_types.py at line 527,](https://github.com/flyteorg/flytekit/blob/cd6bd01ad0ba6688afc71a33a59ece53f90e841a/flytekit/interaction/click_types.py#L527) the code forcibly returns None, causing subsequent conversion failures. Moreover, when running the command pyflyte run --remote example.py example_task --no-flag, the program then checks if the parameter is of boolean type and assigns it a value of False, as implemented in [flytekit/clis/sdk_in_container/run.py at line 696.](https://github.com/flyteorg/flytekit/blob/cd6bd01ad0ba6688afc71a33a59ece53f90e841a/flytekit/clis/sdk_in_container/run.py#L696).

Furthermore, I'm not sure whether it would be better to return value [here](https://github.com/flyteorg/flytekit/blob/cd6bd01ad0ba6688afc71a33a59ece53f90e841a/flytekit/interaction/click_types.py#L527).

## How was this patch tested?

```
from flytekit import task
import os


@task()
def example_task(flag: bool) -> bool:
    return flag

if __name__ == "__main__":

    runner = CliRunner()
    result = runner.invoke(pyflyte.main, ["run", "remote-task", "example.example_task", "--no-flag"])
```
image

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request fixes a bug in remote-task execution by ensuring boolean parameters default to False instead of None, preventing type conversion errors. It also enhances the test suite to cover scenarios involving boolean flags for better validation and includes additional tests for boolean flag handling, improving the reliability of the functionality.

</div>


wild-endeavor:
Still trying to understand why we return None.  It also works if you just comment out those two lines in click_types that return None.  Need to take a further look to understand why it's not later getting filled in.

flytekit/clis/sdk_in_container/run.py
```diff
@@ -473,9 +473,6 @@ def to_click_option(
         is_remote=run_level_params.is_remote,
     )
 
-    if literal_converter.is_bool() and not default_val:
-        default_val = False
-
     description_extra = ""
     if literal_var.type.simple == SimpleType.STRUCT:
         if default_val and not isinstance(default_val, ArtifactQuery):
```
machichima, line 478: I think not setting `default_val` to `False` make this part not reachable for boolean type, which might be the cause of your CI test failure

pingsutw: You can set the default value to `False` [here](https://github.com/flyteorg/flytekit/blob/25eca93ace9ad07f7ef25730fab0d9c848c8e6e7/flytekit/clis/sdk_in_container/run.py#L504-L508) if click type is boolean

taieeuu: I’ve discovered that this CI failure is caused by incompatible packages—most likely the ONNX package. In my original local environment, all tests passed, but after upgrading to the current version, it’s now failing with the same error as on CI, and other people’s PRs are encountering the same issue.

taieeuu:
I’ve also identified another issue. When using `remote-launchplan` or `remote-workflow`, the same error occurs if default_val and value are the same value. So I think this line is needed:

`if v is None and entity.interface.inputs.get(k).type.simple == SimpleType.BOOLEAN:`

wild-endeavor:
Sorry been off doing something else.  Can you tell me what happens if we remove those [two lines](https://github.com/flyteorg/flytekit/blob/d42a0024436a3bd70b3abddc12b2cd9541dbb67f/flytekit/interaction/click_types.py#L527) in click types?  some additional unnecessary work is done?  can you point me to where?

wild-endeavor:
Hey @taieeuu finally forced myself to make some time to take a look another look at this.

I noticed a couple things.
* The `to_click_option` [option](https://github.com/flyteorg/flytekit/blob/5301e3947f75b05fb7b5a901fadae17ed091c764/flytekit/clis/sdk_in_container/run.py#L512) is not a flag for bools... it should probably set `is_flag` equal to true in the constructor.  If there is no default value in the signature, it should be false.  If there is a default value then it should be whatever the value is, true or false.

That is, if I have
```
lp = LaunchPlan.create("blah", default_inputs: {"my_flag": False})
```
we should make a click option that's a flag that's default off.  And the opposite if the default value is True.

* Secondly, there's effectively an API now for how click is passing information to the remote.execute call.  `click_types` will [skip the variable](https://github.com/flyteorg/flytekit/blob/78b27c5d0010200ad3e64ac540a04c506b25f994/flytekit/interaction/click_types.py#L539) if the default matches, which is fine, but remote doesn't know that.  FlyteRemote doesn't know that `None` means look up the value from the default inputs (and I don't think it should).  Can we maybe just drop None values [here](https://github.com/flyteorg/flytekit/blob/78b27c5d0010200ad3e64ac540a04c506b25f994/flytekit/clis/sdk_in_container/run.py#L873)?  At this point, `ctx.params` is the map of keyword to literal.  We can drop None's here before passing to `run_remote`.  This api of "None means it matches the default so we don't have to do anything" is then contained to click handling at least.

Thoughts?

taieeuu:
Hi! @wild-endeavor Thanks so much for taking the time to review this logic!

I completely agree with what you said - I’ve been trying to find a way to access the default value of a parameter, and I didn’t realize we could filter them before calling run_remote. Thanks for your help — and for all the other guidance too!

I’d also like to double-check if my understanding is correct:

1. If default_val is a boolean and no default is explicitly set, Flyte will assign a default value internally.
2. In click.Option, we need to set is_flag=True so that Click properly treats the parameter as a boolean flag.
3. Before calling run_remote, we should filter out parameters that are None and replace them with their corresponding default values to avoid passing None to the remote side.

Please let me know if there’s anything I’ve misunderstood. Thanks again!

PR #3243, opened 2025-04-27
Feat: Allow setting k8s labels and annotations in task decorator

fg91:

### DIFF
--- a/flytekit/models/task.py
+++ b/flytekit/models/task.py
@@ -185,6 +185,7 @@ class TaskMetadata(_common.FlyteIdlEntity):
         cache_ignore_input_vars,
         is_eager: bool = False,
         generates_deck: bool = False,
+        metadata: typing.Optional["K8sObjectMetadata"] = None,
     ):
         """
         Information needed at runtime to determine behavior such as whether or not outputs are discoverable, timeouts,
@@ -208,6 +209,7 @@ class TaskMetadata(_common.FlyteIdlEntity):
         :param pod_template_name: The name of the existing PodTemplate resource which will be used in this task.
         :param cache_ignore_input_vars: Input variables that should not be included when calculating hash for cache.
         :param is_eager:
+        :param metadata: Kubernetes metadata for the task.
         """
         self._discoverable = discoverable
         self._runtime = runtime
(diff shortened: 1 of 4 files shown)


flytekit/models/task.py
```diff
@@ -185,6 +185,7 @@ def __init__(
         cache_ignore_input_vars,
         is_eager: bool = False,
         generates_deck: bool = False,
+        metadata: typing.Optional["K8sObjectMetadata"] = None,
```
Future-Outlier, line 188: should the naming here be `k8s_object_metadata`?

wild-endeavor:
@fg91 can you remind me does flyteidl==1.15.4b0 have what this pr needs?  if so want to bump the requirement?  if not i can cut another idl beta.  (just to get tests passing)

flytekit/models/task.py
fg91: Changed.

fg91:
> @fg91 can you remind me does flyteidl==1.15.4b0 have what this pr needs? if so want to bump the requirement? if not i can cut another idl beta. (just to get tests passing)

Yes, this beta release contains the required change, I bumped the version.

Some other tests are failing with sth seemingly unrelated though 🤔

Sovietaced (approved):

fg91:
> I honestly do not love adding `labels` and `annotations` to task. This implies that every task can have it, which is not true, only k8s pods can have it. This causes problems when folks are using databricks, memverge, slurm etc etc

Result of [this](https://github.com/flyteorg/flyte/issues/6238#issuecomment-2762182977) discussion but I'm happy to implement it another way! How would you expose it?

> only k8s pods can have it
Couldn't the same be said about pod template which is a task arg too?

kumare3:
It can be said about pod_template, by the name `pod_template` and this does go into the `k8spod` attribute. 
Ok lets keep Labels and annotations, but lets not call is K8sObjectMetadata. Lets keep it as labels and annotations in TaskMetadata and the plugin is responsible to put it into k8s.

Moreover, lets document it saying, labels and annotations are optional parameters, that may not be applied in all cases.

kumare3:
I thought a lot more and let's keep what you have. +1

Sorry to keep you waiting. Please document

kumare3 (approved):

PR #3248, opened 2025-05-14
Fixes to dataclass transformer union handling

salotz:
### Motivating use case

I was attempting to use the Python SDK in Python code to run tasks/workflows. When dealing with a case like this:

```python
@dataclass
class FlyteBlobComplex:

    blob: FlyteFile
    param: float

@task()
def smoke_blob_complex_inputs(val: FlyteBlobComplex) -> bool:

    val.blob.download()
    check_file(Path(val.blob.path))
    
    return True
```

I ran into problems with supplying the proper inputs to `FlyteRemote.execute` for the `FlyteFile` attributes.

After some troubleshooting I found a set of bugs surrounding how types and unions are handled in the `DataclassTransformer` class to be the culprit.

At a high level I should be able to supply inputs like this:

```python
flyte_remote.execute(
    smoke_blob_complex_inputs,
    inputs={
        "val" : {
            "blob" : {"path" : "gs://mybucket/path/to/obj"},
            "param" : 1.0,
        },
    },
)
```

However this yielded this error:

```
flytekit.core.type_engine.TypeTransformerFailedError: The original fields are missing the following keys from the dataclass fields: ['metadata']
```

In this PR I fixed two issues related to handling of union types in the `DataclassTransformer` class that were manifesting in the behavior above.

First, when types of dataclass fields were being retrieved with `dataclasses.fields` method (on the `FlyteFile` object, which I suppose is also a dataclass but probably a special case (?)) the `f.type` was somehow being cast to a string such that you would get `'Optional[dict[str,str]]` for the `metadata` field. When this was checked for if it was an optional type this would fail as the checker doesn't understand strings, only real type objects. (NOTE that this is a type mismatch that could probably be figured out statically).

So replacing `dataclasses.fields` with `typing.get_type_hints` properly resolves the types as type objects solves this issue.

Second, further down in the code when dealing with non-Optional Unions values are only checked if they are optional types and if `type(value) == expected_type` where expected type might be `Union[str, Pathlike]` like in the case of the `path` attribute of the `FlyteFile`.

To address this I added some code to check if the value is any of the types in the union.

These fixes solve my immediate issue but I still get the feeling that there is some bigger picture issue that I'm not seeing leading to these problems. I would appreciate some perspective from the team.

## How was this patch tested?

See the use case above.


machichima:
Hi @salotz ,
Would you mind adding DCO and fix CI error?
Thanks!

machichima:
> @machichima what is DCO?

DCO is like signing up your commit. You can have a look here to see how to add it: https://github.com/src-d/guide/blob/master/developer-community/fix-DCO.md

Moreover, to fix the lint CI, you can do `make fmt`

salotz:
@machichima I've got DCO sorted out.

I don't see any CI running from my end until approval, but I did the formatting.

flytekit/core/type_engine.py
```diff
@@ -1839,6 +1843,14 @@ class UnionTransformer(AsyncTypeTransformer[T]):
     def __init__(self):
         super().__init__("Typed Union", typing.Union)
 
+    @staticmethod
+    def is_union(t: Type[Any] | types.UnionType) -> bool:
+        return _is_union_type(t)
+
+    @staticmethod
+    def in_union(t: Type[Any], union: types.UnionType) -> bool:
+        return t in typing.get_args(union)
```
machichima, line 1852: Can we use `get_args` here?  Which is already imported

flytekit/core/type_engine.py
```diff
@@ -568,7 +566,13 @@ def assert_type(self, expected_type: Type[DataClassJsonMixin], v: T):
                         original_type = type(v)
                         if UnionTransformer.is_optional_type(expected_type):
                             expected_type = UnionTransformer.get_sub_type_in_optional(expected_type)
-                        if original_type != expected_type:
+
+                        if UnionTransformer.is_union(expected_type) and UnionTransformer.in_union(
+                            original_type, expected_type
+                        ):
+                            pass
```
machichima, line 573: How about putting this before `if UnionTransformer.is_optional_type(expected_type)` and use continue to make the code structure better? Like below:

```python
if UnionTransformer.is_union(expected_type) and UnionTransformer.in_union(
    original_type, expected_type
):
    continue
if UnionTransformer.is_optional_type(expected_type):
    expected_type = UnionTransformer.get_sub_type_in_optional(expected_type)
elif original_type != expected_type:
    raise TypeTransformerFailedError(
        f"Type of Val '{original_type}' is not an instance of {expected_type}"
    )

```

flytekit/core/type_engine.py
```diff
@@ -1839,6 +1843,14 @@ class UnionTransformer(AsyncTypeTransformer[T]):
     def __init__(self):
         super().__init__("Typed Union", typing.Union)
 
+    @staticmethod
+    def is_union(t: Type[Any] | types.UnionType) -> bool:
+        return _is_union_type(t)
+
+    @staticmethod
+    def in_union(t: Type[Any], union: types.UnionType) -> bool:
```
machichima, line 1851: Can we use `typing.Union` here? So that we do not need to `import types`

flytekit/core/type_engine.py
```diff
@@ -1839,6 +1843,14 @@ class UnionTransformer(AsyncTypeTransformer[T]):
     def __init__(self):
         super().__init__("Typed Union", typing.Union)
 
+    @staticmethod
+    def is_union(t: Type[Any] | types.UnionType) -> bool:
```
machichima, line 1847: Can we use `typing.Union` here? So that we do not need to `import types`

machichima:
Could you also merge master, which contains fix to the current failed CI test `build-with-pandas`

flytekit/core/type_engine.py
salotz: If you are maintaining strict type checking, unfortunately no. As you need to check for `UnionType` and `Union` separately and `UnionType` doesn't fall under `Type[Any]` (union types are weird in Python...).

I noticed a couple other problems that were due to type mismatches so I'll defer to your preferences here as I don't want to overhaul things.

PR #3273, opened 2025-06-11
support passing optional date through CLI

blaketastic2:

Currently, if we have a `workflow` with an optional date parameter, we cannot call this from the CLI.

Workflow definition:
```python
@workflow
def ingest_data(
    start_date: Optional[date] = None,
) -> None:
    pass
```

Executing the workflow:
```bash
pyflyte run --copy all -p flyte-snacks -d development --remote workflow.py ingest_data --start_date 2025-06-10
```

Produces this error:
```text
Invalid value for '--start_date': Failed to convert param: <Option start_date>, value: 2025-06-10 00:00:00 to type: typing.Optional[datetime.date]. Reason
Cannot convert from 2025-06-10 00:00:00 to typing.Optional[datetime.date]
```

We are now making sure to use the non-None variant when trying to do the conversion.

### DIFF
--- a/flytekit/interaction/click_types.py
+++ b/flytekit/interaction/click_types.py
@@ -558,8 +558,17 @@ class FlyteLiteralConverter(object):
         if isinstance(value, ArtifactQuery):
             return value
         try:
+            # Handle datetime.date conversion for union types (e.g., Optional[datetime.date])
+            target_type = self._python_type
+            if hasattr(self._python_type, "__origin__") and self._python_type.__origin__ is typing.Union:
+                # For union types like Optional[datetime.date], extract the non-None type
+                union_args = typing.get_args(self._python_type)
+                non_none_types = [arg for arg in union_args if arg is not type(None)]
+                if len(non_none_types) == 1:
+                    target_type = non_none_types[0]
+
             # If the expected Python type is datetime.date, adjust the value to date
-            if self._python_type is datetime.date:
+            if target_type is datetime.date:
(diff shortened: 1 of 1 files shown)


flytekit/interaction/click_types.py
```diff
@@ -559,7 +585,7 @@ def convert(
             return value
         try:
             # If the expected Python type is datetime.date, adjust the value to date
-            if self._python_type is datetime.date:
+            if self._python_type is datetime.date and isinstance(value, datetime.datetime):
                 # Click produces datetime, so converting to date to avoid type mismatch error
                 value = value.date()
```
machichima, line 590: I think here is the only place we need to modify. While it already deal with the conversion for `datetime.date` type, what we need to do is to also trigger this conversion for `Optional[datetime.date]` and `datetime.date | None` type.

The following should work:

```python
            python_type = self._python_type
            origin = typing.get_origin(python_type)
            is_union = origin in (typing.Union, types.UnionType)
            args = typing.get_args(python_type)
            is_date_type = (
                (is_union and datetime.date in args)
                or python_type is datetime.date
            )

            if is_date_type and isinstance(value, datetime.datetime):
                value = value.date()
```

machichima (changes requested):

machichima:
@blaketastic2 Could you fix CI failure? Thanks!

machichima (approved):
LGTM! Thank you

PR #3284, opened 2025-06-28
Support recursive `ImageSpec` build by `DefaultBuilder`

amitani:
## Tracking issue
None

`ImageBuildEngine` supports recursively build `ImageSpec` if `base_image` is also an `ImageSpec`.
https://github.com/flyteorg/flytekit/blob/147c405d11abb222f88933d5ef3ed258df92ba99/flytekit/image_spec/image_spec.py#L523
However, it doesn't work with current `DefaultBuilder`. This PR includes a few minor changes so that such recursive build works.

1. `micromamba create` falls back to `micromamba install` in case the micromamba environment already exists.
2. Set workdir to `/` to have consistent behavior in the second build. Otherwise, workdir is set to `/root` in the first build, and `uv run` itself uses `/root/.venv` instead of `UV_PYTHON` (if uv lock or poetry lock was used).  
3.  When there are `/root/.venv` from the first build (if uv lock or poetry lock was used), `chown -R flytekit /root` creates a duplicated layer, nearly doubling the size of the final image.

## How was this patch tested?

A small example was added, which can be run
```
_FLYTEKIT_TEST_IMAGE_BUILD_ENGINE=1 python -m pytest tests/flytekit/unit/core/image_spec/test_image_spec.py -k test_nested_build -s
```

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the `DefaultBuilder` to support recursive builds of `ImageSpec`, optimizes the ownership change process, and sets the working directory to root for consistency. A test case for nested builds has also been introduced to ensure functionality.

</div>

### DIFF
--- a/flytekit/image_spec/default_builder.py
+++ b/flytekit/image_spec/default_builder.py
@@ -86,9 +86,12 @@ RUN --mount=type=cache,sharing=locked,mode=0777,target=/opt/micromamba/pkgs,\
 id=micromamba \
     --mount=from=micromamba,source=/usr/bin/micromamba,target=/usr/bin/micromamba \
     micromamba config set use_lockfiles False && \
-    micromamba create -n runtime --root-prefix /opt/micromamba \
+    ( micromamba create -n runtime --root-prefix /opt/micromamba \
     -c conda-forge $CONDA_CHANNELS \
-    python=$PYTHON_VERSION $CONDA_PACKAGES
+    python=$PYTHON_VERSION $CONDA_PACKAGES \
+    || micromamba install -n runtime --root-prefix /opt/micromamba \
+    -c conda-forge $CONDA_CHANNELS \
+    python=$PYTHON_VERSION $CONDA_PACKAGES )
 """)
 
 DOCKER_FILE_TEMPLATE = Template("""\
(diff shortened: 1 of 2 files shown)


flytekit/image_spec/default_builder.py
```diff
@@ -98,14 +101,15 @@
...
+WORKDIR /
 USER root
 $APT_INSTALL_COMMAND
 RUN --mount=from=micromamba,source=/etc/ssl/certs/ca-certificates.crt,target=/tmp/ca-certificates.crt \
     [ -f /etc/ssl/certs/ca-certificates.crt ] || \
     mkdir -p /etc/ssl/certs/ && cp /tmp/ca-certificates.crt /etc/ssl/certs/ca-certificates.crt
 
 RUN id -u flytekit || useradd --create-home --shell /bin/bash flytekit
-RUN chown -R flytekit /root && chown -R flytekit /home
+RUN chown flytekit /root && chown -R flytekit /home
```
davidmirror-ops, line 112: Thanks for your contribution!
Could you elaborate a bit on why is this needed?

pingsutw (approved):

flytekit/image_spec/default_builder.py
thomasjpfan: When running [flyte interactive](https://www.union.ai/docs/flyte/integrations/external-service-backend-plugins/flyteinteractive-plugin/vscode/) where the runtime user is `flytekit` it's convenient to have `/root` be writable by the `flytekit` user.

PR #3294, opened 2025-07-13
[BUG] Local testing using optional dicts with optional values does not work

machichima:
## Tracking issue

Closes https://github.com/flyteorg/flyte/issues/6474

When running task locally, using optional list values types (i.e. `dict[str, list[str] | None] | None`) will cause error. However, when running such task on the cluster it works

Error occurs when we call `_are_types_castable(Union, List[Sample])`. In original code, the [check for collection type](https://github.com/machichima/flytekit/blob/9c720130e6558fa5172a8d4ca0f26c2dc2d36a2d/flytekit/core/type_engine.py#L1772-L1776) will directly return `False`, which indicates that `List[Sample]` is not castable to `Union[List[Sample], None]`, leading to the error.

In this PR, I deal with `union_type` first in  `_are_types_castable` function to overcome this problem.

## How was this patch tested?

Test with the following script. See result in Screenshots section.

```python
from flytekit import task

@task
def complex_task(data: dict[str, list[str] | None] | None) -> dict[str, list[str] | None] | None:
    return data

def test_complex_task() -> dict[str, list[str] | None] | None:
    return complex_task(data={"key": ["value1", "value2"]})


if __name__ == "__main__":
    print(test_complex_task())
```

### Screenshots

Successfully execute the above script:

image

### DIFF
--- a/flytekit/clis/sdk_in_container/register.py
+++ b/flytekit/clis/sdk_in_container/register.py
@@ -15,7 +15,8 @@ from flytekit.clis.sdk_in_container.utils import domain_option_dec, project_opti
 from flytekit.configuration import ImageConfig
 from flytekit.configuration.default_images import DefaultImages
 from flytekit.constants import CopyFileDetection
-from flytekit.interaction.click_types import key_value_callback
+from flytekit.core.resources import Resources, ResourceSpec
+from flytekit.interaction.click_types import key_value_callback, resource_callback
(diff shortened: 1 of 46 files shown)


kumare3 (approved):

kumare3 (approved):

pingsutw (approved):
Thank you @machichima

PR #1458, opened 2023-02-08
 Flytekit Auth system overhaul and pretty printing upgrade

kumare3:
# TL;DR
This PR introduces a new Oauth2 handling system in flytekit, that can be used in non flytekit python libraries that wish to use a standardized flow for flytekit. The standard flow is as defined in the RFC and a few additional supported methods.

The system uses client side - [grpc.Interceptors](https://grpc.github.io/grpc/python/grpc.html#client-side-interceptor). for fine control on the auth flow. It also introduces exception wrapping and retrying interceptor.

As a last time, it improves the terminal output when using pyflyte. All the output is not distilled and the error is traced correctly. Eventually we could link to docs for the errors.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [x] Code documentation added
 - [x] Any pending items have an associated Issue

PR #3303, opened 2025-07-24
Restore behavior passing FlyteClient kwargs to gRPC channel creation

clint-stripe:

The `FlyteRemote` class is documented as allowing gRPC options in the 
https://github.com/flyteorg/flytekit/blob/v1.16.3/flytekit/remote/remote.py#L284-L285.
These are still passed through to client construction, though we stopped passing these to the underlying gRPC channel creation in https://github.com/flyteorg/flytekit/pull/1458. (It's not clear to me if this was an oversight or intentional decision?)

This does mean that from v1.10.0 onward, there's no longer any way to provide additional options for the gRPC channel.
We have a custom proxy that requires additional gRPC options to be set at channel creation time (specifically, `grpc.default_authority`, though I can imagine others may be useful too). Unfortunately, it's not possible to configure this at any other point (e.g. with an interceptor).

This PR restores the previous (and still-documented) behavior.

## How was this patch tested?

I added a unit test to ensure we're keeping the default options as well as adding any user-provided options.

I've also tested with this change internally and it works with pre-1.10.0 code that has the desired behavior.

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request restores the ability to pass additional gRPC options when creating the gRPC channel in the FlyteClient, addressing a regression from a previous update. It ensures compatibility with both default and user-provided options, and includes a unit test to verify the functionality.

</div>

### DIFF
--- a/flytekit/clients/raw.py
+++ b/flytekit/clients/raw.py
@@ -49,12 +49,15 @@ class RawSynchronousFlyteClient(object):
         # StreamRemoved exception.
         # https://github.com/flyteorg/flyte/blob/e8588f3a04995a420559327e78c3f95fbf64dc01/flyteadmin/pkg/common/constants.go#L14
         # 32KB for error messages, 20MB for actual messages.
-        options = (("grpc.max_metadata_size", 32 * 1024), ("grpc.max_receive_message_length", 20 * 1024 * 1024))
+        options = [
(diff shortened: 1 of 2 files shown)


pingsutw (approved):

clint-stripe:
Just wanted to bump this PR @pingsutw, thanks!

PR #3311, opened 2025-08-11
[BUG] ImageSpec NoOpBuilder should always return True while calling should_build function

popojk:
## Tracking issue
closes [#6561](https://github.com/flyteorg/flyte/issues/6561)

The [NoOpBuilder](https://github.com/flyteorg/flytekit/blob/eb5a67f76aaef96a44bde04afb72b87592cc8b7a/flytekit/image_spec/noop_builder.py) should overrride [should_build](https://github.com/flyteorg/flytekit/blob/eb5a67f76aaef96a44bde04afb72b87592cc8b7a/flytekit/image_spec/image_spec.py#L454) function in order to correctly trigger [build_image](https://github.com/flyteorg/flytekit/blob/eb5a67f76aaef96a44bde04afb72b87592cc8b7a/flytekit/image_spec/noop_builder.py#L9) function in an environment that docker daemon is not running.

Make `NoOpBuilder` override `should_build` function.

## How was this patch tested?
Test workflow:
```
from flytekit.image_spec.noop_builder import NoOpBuilder

image_spec = ImageSpec(
    name="test_image",
    builder="noop",
    base_image="localhost:30000/test_image:latest",
    registry="ghcr.io/not_exist_registry"
    
)

def task_1(message: str) -> str:
    print(message)
    return message

@workflow
def wf() -> str:
    return task_1("hello")
```
We define a docker registry in `image_spec` that is not existed, and we expect that the SDK will still use the image user defined without actually communicating with the docker registry. After remote running the workflow, the SDK will use the user defined image as expected.
截圖 2025-08-11 上午11 44 59

### DIFF
--- a/flytekit/image_spec/noop_builder.py
+++ b/flytekit/image_spec/noop_builder.py
@@ -6,6 +6,20 @@ class NoOpBuilder(ImageSpecBuilder):
 
     builder_type = "noop"
 
+    def should_build(self, image_spec: ImageSpec) -> bool:
+        """
+        The build_image function of NoOpBuilder uses the image_spec name as defined by the user without
+        checking whether the image exists in the Docker registry. Therefore, the should_build function
+        should always return True to trigger the build_image function.
+
+        Args:
+            image_spec (ImageSpec): Image specification
+
+        Returns:
+            bool: Always returns True
+        """
+        return True
+
     def build_image(self, image_spec: ImageSpec) -> str:
         if not isinstance(image_spec.base_image, str):
             msg = "base_image must be a string to use the noop image builder"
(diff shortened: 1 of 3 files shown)


flytekit/image_spec/noop_builder.py
```diff
@@ -6,6 +6,20 @@ class NoOpBuilder(ImageSpecBuilder):
 
     builder_type = "noop"
 
+    def should_build(self, image_spec: ImageSpec) -> bool:
+        """
+        The build_image function of NoOpBuilder uses the image_spec name as defined by the user without
+        checking whether the image exists in the Docker registry. Therefore, the should_build function
+        should always return True to trigger the build_image function.
```
machichima, line 13: The docstring here is a bit unclear, I think what you mean is that the `build_image` of `NoOpBuilder` does not really build the image (but do...), so that we do not need to check if the docker daemon exists or docker image exists?

Could you please rewrite it to make it clearer?

pingsutw (approved):

machichima (approved):
LGTM

PR #3314, opened 2025-08-14
feat: Add timeout to container task

Narwhal-fish:
## Tracking issue

Related to flyteorg/flyte#6351

Currently, `ContainerTask` lacks built-in timeout functionality, which is available for regular Python tasks via the `@task` decorator. 

This PR adds a `timeout` parameter to `ContainerTask` that accepts `datetime.timedelta` objects:

## How was this patch tested?
1. **`test_container_task_timeout()`**: Tests local execution timeout with timedelta
2. **`test_container_task_timeout_k8s_serialization()`**: Verifies K8s `activeDeadlineSeconds` serialization
3. **`test_container_task_no_timeout()`**: Ensures normal execution works with sufficient timeout

## Docs link

 
 <div id='description'>
<h3>Summary by Bito</h3>
This pull request enhances the ContainerTask functionality by adding a timeout parameter for maximum execution duration, enabling it to accept a timeout parameter via a datetime.timedelta object. Modifications include updates to the ContainerTask class and Kubernetes pod specification, along with new unit tests to verify the feature's functionality during local execution and Kubernetes serialization, enhancing the task's capabilities to match those of standard Python tasks.

</div>

### DIFF
--- a/flytekit/core/container_task.py
+++ b/flytekit/core/container_task.py
@@ -76,6 +76,9 @@ class ContainerTask(PythonTask):
         metadata = metadata or TaskMetadata()
         metadata.pod_template_name = pod_template_name
 
+        if timeout is not None:
+            metadata.timeout = timeout
+
         super().__init__(
             task_type="raw-container",
             name=name,
@@ -283,7 +286,6 @@ class ContainerTask(PythonTask):
             self._image, command=commands, remove=True, volumes=volume_bindings, detach=True
         )
 
-        # Wait for the container to finish the task, with timeout if specified
         timeout_seconds = None
         if self._timeout is not None:
             timeout_seconds = self._timeout.total_seconds()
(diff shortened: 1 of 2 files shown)


Sovietaced:
Does this require backend changes as well?

Sovietaced:
I'm not sure if this is the correct approach. For regular Python tasks the timeout is passed through the task metadata and queueing time is counted towards the overall task timeout. It looks like this implementation only sets the active deadline seconds which will only affect the underlying container when it eventually runs. 

So I am a bit worried about the difference in semantics. Is it possible to instead or also plumb the timeout through the task metadata?

Narwhal-fish:
> I'm not sure if this is the correct approach. For regular Python tasks the timeout is passed through the task metadata and queueing time is counted towards the overall task timeout. It looks like this implementation only sets the active deadline seconds on the pod which will only affect the underlying container when it eventually runs.
> 
> So I am a bit worried about the difference in semantics. Is it possible to instead or also plumb the timeout through the task metadata?

Sorry for the late reply.
I hadn't considered the importance of using the standard task metadata timeout mechanism to ensure consistent behavior with regular Python tasks.

I'll modify the implementation to plumb the timeout through the task metadata so that it includes queueing time and follows the same execution semantics as regular Python tasks.

PR #3315, opened 2025-08-15
Dynamic platform detection for ImageSpec to support cross-platform development

BarryWu0812:
Closes flyteorg/flyte#6565

The ImageSpec should automatically detect the current machine's architecture and set an appropriate default platform:
"linux/arm64" for Apple Silicon Macs (ARM64)
"linux/amd64" for Intel Macs (x86_64)

Use platform.machine() to detect the current machine's architecture

## How was this patch tested?
Unit test

### DIFF
--- a/flytekit/image_spec/image_spec.py
+++ b/flytekit/image_spec/image_spec.py
@@ -4,6 +4,7 @@ import dataclasses
 import hashlib
 import os
 import pathlib
+import platform
 import re
 import sys
 import typing
@@ -135,6 +136,9 @@ class ImageSpec:
         if self.builder is None and builder_registry:
             # Use the builder with the highest priority by default
             self.builder = max(builder_registry, key=lambda name: builder_registry[name][1])
+
+        if platform.machine() == "arm64":
+            self.platform = "linux/arm64"
 
         parameters_str_list = [
             "packages",


flytekit/image_spec/image_spec.py
```diff
@@ -136,6 +137,9 @@ def __post_init__(self):
             # Use the builder with the highest priority by default
             self.builder = max(builder_registry, key=lambda name: builder_registry[name][1])
 
+        if platform.machine() == "arm64":
+            self.platform = "linux/arm64"
```
machichima, line 141: This line overwrite whatever user set in platform parameter if they are using arm64, which is not ideal

machichima:
I recommend using `field(default_factory=...)` for defining default value for `platform`

```python
platform: str = field(default_factory=_get_default_platform)
```

the get default platform function can be like:

```python
  def _get_default_platform():
      import platform
      machine = platform.machine().lower()
      if machine in ("arm64", "aarch64"):
          return "linux/arm64"
      return "linux/amd64"
```

flytekit/image_spec/image_spec.py
pingsutw: We should only change it if they push the images to the local registry

machichima (dismissed):
LGTM, thank you!

cc @pingsutw

pingsutw (commented):
LGTM, there is a lint error, could you take a look

pingsutw (approved):

PR #3331, opened 2025-09-29
add timestamp transformer

dansola:

This workflow
```
@fl.task(container_image=image)
def get_timestamp() -> pd.Timestamp:
    return pd.Timestamp("2024-09-29 15:30:45")


@fl.workflow
def timestamp_wf() -> pd.Timestamp:
    return get_timestamp()
```
gives this error:
```
TypeTransformerFailedError: Failed to convert 
outputs of task 'timestamp_wf.get_timestamp' at 
position 0.
Failed to convert type <class 
'pandas._libs.tslibs.timestamps.Timestamp'> to type 
<class 'pandas._libs.tslibs.timestamps.Timestamp'>.
Error Message: Expected value of type <class 
'datetime.datetime'> but got '2024-09-29 15:30:45' 
of type <class 
'pandas._libs.tslibs.timestamps.Timestamp'>.
```

We need a pandas timestamp type transformer. 

## How was this patch tested?

Tested locally and remotely. For remote run, just installed flytekit from the `da72500c93e1ff1107a24f1e67eb453a443e4d82` commit on the image. 

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1363,7 +1363,11 @@ class TypeEngine(typing.Generic[T]):
                 from flytekit.extras import pydantic_transformer  # noqa: F401
             if is_imported("pandas"):
                 try:
-                    from flytekit.types.schema.types_pandas import PandasSchemaReader, PandasSchemaWriter  # noqa: F401
+                    from flytekit.types.schema.types_pandas import (  # noqa: F401
+                        PandasSchemaReader,
+                        PandasSchemaWriter,
+                        PandasTimestampTransformer,
+                    )
                 except ValueError:
                     logger.debug("Transformer for pandas is already registered.")
             if is_imported("numpy"):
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
```diff
@@ -1363,7 +1363,11 @@ def lazy_import_transformers(cls):
                 from flytekit.extras import pydantic_transformer  # noqa: F401
             if is_imported("pandas"):
                 try:
-                    from flytekit.types.schema.types_pandas import PandasSchemaReader, PandasSchemaWriter  # noqa: F401
+                    from flytekit.types.schema.types_pandas import (  # noqa: F401
+                        PandasSchemaReader,
+                        PandasSchemaWriter,
+                        PandasTimestampTransformer,
```
wild-endeavor, line 1369: can you make this a separate import, and move the timestamp transformer into a separate file?

The "schema" folder name refers to the old Flyte Schema type, which was deprecated by the structureddataset type.  Could we put it in the `basic_dfs.py` file alongside the rest of the Dataframe stuff for pandas?

PR #3413, opened 2026-03-30
[Flytekit-7005] Adding ray autoscaler options

0yukali0:
## Tracking issue
Ralated [7005](https://github.com/flyteorg/flyte/issues/7005)

Ray plugin in Flytekit does not support ray autoscaler option when enable_autoscaling is settable.

1. Create autoscaler option dataclass
2. Add that to ray cluster message which sent to backend.

### Setup process

clone flyte and flytekit (branch flyte-7005 and flytekit-7005)
run the flyte backend in sandbox
2.1 flytectl demo start --dev
2.2 make compile && POD_NAMESPACE=flyte ./flyte start --config flyte-single-binary-local.yaml
Update ray example like [Ray example](https://github.com/0yukali0/flytesnacks/blob/flytesnack-7005/examples/ray_plugin/ray_plugin/ray_example.py#L68-L74)
3.1 uv venv -p 3.11 && source .venv/bin/activate && uv pip install -e ~/flytekit ~/flytekit/plugins/flytekit-ray && uv pip install -e ~/flyte/flyteidl
3.2 "pyflyte run --remote ray_example.py ray_workflow --n 10"

### Screenshots

image

image


flytekit/core/artifact.py
```diff
@@ -91,16 +91,16 @@ def bind_partitions(self, *args, **kwargs) -> ArtifactIDSpecification:
                 if not self.artifact.partition_keys or k not in self.artifact.partition_keys:
                     raise ValueError(f"Partition key {k} not found in {self.artifact.partition_keys}")
                 if isinstance(v, art_id.InputBindingData):
-                    p.partitions[k] = Partition(art_id.LabelValue(input_binding=v), name=k)
+                    p.partitions[k] = Partition(art_id.LabelValue(input_binding=v), name=k)  # type: ignore[index]
```
machichima, line 94: Could we fix the type issue rather than adding `type: ignore`?

flytekit/core/workflow.py
```diff
@@ -779,6 +779,10 @@ def __init__(
 
         self.compiled = False
 
+    @property
+    def name(self) -> str:  # type: ignore[override]
```
machichima, line 783: Why do we need `type: ignore` here?

plugins/flytekit-ray/flytekitplugins/ray/models.py
```diff
@@ -146,6 +148,73 @@ def from_flyte_idl(cls, proto):
...
+
+    def to_flyte_idl(self) -> _ray_pb2.AutoscalerOptions:
+        envs = []
+        if self.env:
+            for key, val in self.env.items():
+                envs.append(_literals_pb2.KeyValuePair(key=key, value=val))
+        return _ray_pb2.AutoscalerOptions(
+            upscaling_mode=_ray_pb2.AutoscalerOptions.UpscalingMode.Value(
+                "UPSCALING_MODE_" + self.upscaling_mode.upper()
+            )
```
machichima, line 194: Could we follow how we did for `Resources` enum without transforming to string. User can directly use enum value (e.g. `AutoscalerOptions.UpscalingMode.DEFAULT`)


https://github.com/flyteorg/flytekit/blob/7e289a20be6fa5caf6b230dd7c7c2c187b099ec9/flytekit/models/task.py#L24-L30

Same comment for `from_flyte_idl`

flytekit/core/workflow.py
0yukali0: hi @machichima 
I added this when linting indicates the problems.
Currently, should i update code to follow the lint rules in flytekit?
image

pingsutw (commented):
Thanks, LGTM, could you fix the lint errors

flytekit/core/artifact.py
```diff
@@ -348,7 +348,7 @@ def _repr_html_(self):
         return ", ".join([str(x) for x in self.__rich_repr__()])
 
     @property
-    def partitions(self) -> Optional[typing.Dict[str, Partition]]:
+    def partitions(self) -> typing.Dict[str, Partition]:
```
pingsutw: why do we need to change the type here?

flytekit/core/workflow.py
```diff
@@ -779,6 +779,10 @@ def __init__(
 
         self.compiled = False
 
+    @property
+    def name(self) -> str:
```
pingsutw: Do we use it anywhere? why expose it?

0yukali0:
I removed the modifications of lint errors which i used the wrong lint setting.
Currently linting that i ran  act -j lint passed.

pingsutw (approved):

PR #3388, opened 2026-02-12
Use setuptools<82.0.0

pingsutw:
## Tracking issue
NA

## Why are the changes needed?
unit tests are failing
```
/opt/hostedtoolcache/Python/3.12.12/arm64/lib/python3.12/importlib/__init__.py:90: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
tests/flytekit/unit/core/test_type_engine.py:24: in <module>
    from marshmallow_jsonschema import JSONSchema
/opt/hostedtoolcache/Python/3.12.12/arm64/lib/python3.12/site-packages/marshmallow_jsonschema/__init__.py:1: in <module>
    from pkg_resources import get_distribution
E   ModuleNotFoundError: No module named 'pkg_resources'
```

## What changes were proposed in this pull request?
The issue was that setuptools 82.0.0 (the latest version) removed pkg_resources entirely. I downgraded to setuptools 69.5.1, which still includes it

## Check all the applicable boxes 

- [ ] I updated the documentation accordingly.
- [ ] All new and existing tests passed.
- [ ] All commits are signed-off.

PR #3420, opened 2026-04-09
Move marshmallow-jsonschema to optional dependency to fix CVE-2024-6345

dwo:
## Tracking issue

https://www.cve.org/CVERecord?id=CVE-2024-6345

`setuptools` was pinned `<70` in #3388 as a runtime dependency solely because `marshmallow-jsonschema` imports `pkg_resources` at runtime.
This forces all flytekit users onto a setuptools version affected by [CVE-2024-6345](https://www.cve.org/CVERecord?id=CVE-2024-6345).

The `marshmallow-jsonschema` code path is only a fallback — `mashumaro` is the primary JSON Schema generator, and the marshmallow path only triggers when mashumaro fails for `DataClassJsonMixin` classes.
The imports are already guarded by try/except, so nothing breaks when the package is absent.

`marshmallow-jsonschema` itself appears unmaintained (last commit Oct 2023, multiple open PRs to remove `pkg_resources` unmerged).

An alternative would be to drop `marshmallow-jsonschema` support entirely, since `mashumaro` already covers the primary path.

- Remove `marshmallow-jsonschema`, `marshmallow-enum`, and `setuptools` from core dependencies
- Add a new `marshmallow-jsonschema` optional extra with those packages (`setuptools>=70,<82`)
- Bump build-system setuptools to `>=70`
- Update `dev-requirements.in` to install the extra for tests
- Update docstring to reflect mashumaro as the primary JSON Schema generator

## How was this patch tested?

Ran `test_type_engine.py` and `test_generice_idl_type_engine.py` — 416 passed, 2 skipped.

## Related PRs

#3388

### DIFF
--- a/dev-requirements.in
+++ b/dev-requirements.in
@@ -1,4 +1,4 @@
--e file:.
+-e file:.[marshmallow-jsonschema]
 
 coverage[toml]
 hypothesis
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -495,33 +495,35 @@ class DataclassTransformer(TypeTransformer[object]):
             (1) Convert MessagePack Bytes to a dataclass using mashumaro.
             (2) Handle dataclass attributes to ensure they are of the correct types.
 
-        For Json Schema, we use https://github.com/fuhrysteve/marshmallow-jsonschema library.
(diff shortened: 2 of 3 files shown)


Sovietaced:
I think this might be addressed in https://github.com/flyteorg/flytekit/pull/3430

PR #3270, opened 2025-06-10
Fix issue where pod template override pod spec was missing

Sovietaced:
## Tracking issue
Fixes https://github.com/flyteorg/flyte/issues/6463

## Why are the changes needed?
The pod_spec was not included in the pod template when not using fast registration. This would lead to bad task specification errors on the backend since there would be no pod spec but a primary container name was configured.

Although this comes from our custom Armada plugin, the plugin is just using common pod_helpers code. 

> RuntimeExecutionError: max number of system retry attempts [31/30] exhausted. Last known status message: failed at Node[n0-n0]. RuntimeExecutionError: failed during plugin execution, caused by: failed to execute handle for plugin [armada]: generating job request: [BadTaskSpecification] invalid TaskSpecification, primary container [primary] not defined

After debugging I found that this was logged from propeller pod_helpers.go code.
```
podTemplate=metadata:{} pod_spec:{} primary_container_name:"primary"
```

## What changes were proposed in this pull request?
Updates the translation code to render the pod template pod spec unconditionally.

## How was this patch tested?
Tested in a production environment.

## Check all the applicable boxes 

- [x] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [x] All commits are signed-off. 
 <div id='description'>
    <a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>
This pull request resolves an issue with the pod template override and the pod specification missing during non-fast registration, ensuring the pod spec is rendered unconditionally and enhancing task execution reliability. The changes were validated in a production environment.

</div>

PR #3428, opened 2026-04-27
Carry pod_template override through map_task serialization

### LINKED ISSUE #7076: [BUG] with_overrides(pod_template=...) silently ignored for map_task
### Flyte & Flytekit version

flytekit: 1.16.15
Python: 3.13.7

### Describe the bug

map_task(...).with_overrides(pod_template=PodTemplate(...)) silently drops the pod_template override during serialization. The override is accepted and stored on the Node object, but never serialized into the TaskNodeOverrides for array node map tasks.

This is a separate issue from #6463 / PR #3270, which fixed pod_template serialization for regular tasks (conditional should_fast_serialize() logic in get_serializable_task). The map_task serialization path has the same class of bug but in a different location.

### Expected behavior

Correctly override the map_task pod template

### Additional context to reproduce

_No response_

### Screenshots

_No response_

### Are you sure this issue hasn't been raised already?

- [x] Yes

### Have you read the Code of Conduct?

- [x] Yes

1fanwang:
## Tracking issue
Fixes https://github.com/flyteorg/flyte/issues/7076

`map_task(...).with_overrides(pod_template=PodTemplate(...))` silently drops the override at serialization. The override is accepted at workflow build time and stored on the `Node` (`node._pod_template`), but `get_serializable_array_node_map_task` in `flytekit/tools/translator.py` never reads that field — it builds `TaskNodeOverrides` with only `resources`, `extended_resources`, and `container_image`. The user thinks the override is applied, but the registered task spec has an empty `overrides {}` block.

This is the same class of bug as #6463, which #3270 fixed for the regular-task path of `get_serializable_node`. The array-node path has the same gap in a different function.

`flytekit/tools/translator.py` — `get_serializable_array_node_map_task`:
- Read `node._pod_template`, build the override pod spec via `_serialize_pod_spec`, and include `pod_template` in the emitted `TaskNodeOverrides`.
- Use `entity.get_container(settings)` (not `_get_container`) so `prepare_target()` substitutes the map-task command (`pyflyte-map-execute`) into the container args before the pod spec is built. `ArrayNodeMapTask` extends `PythonTask` directly (not `PythonAutoContainerTask`), so `_get_container` isn't on it anyway.
- For fast-serialization, prefix container args **post-build** (matching the existing pattern in `get_serializable_task` at `translator.py:170`) instead of swapping `command_fn` via `_fast_serialize_command_fn`. The latter wraps `task.get_default_command(settings)`, which on `ArrayNodeMapTask` returns the inherited `pyflyte-execute` — wrong shape for a map task.

`tests/flytekit/unit/test_translator.py`:
- New parametrized regression test `test_map_task_with_pod_template_override` covering both `fast_registration_enabled=True` and `False`. Mirrors the structure of the existing `test_task_with_pod_template_override` from #3270.

## How was this patch tested?

- New parametrized test passes (both cases).
- Full `tests/flytekit/unit/test_translator.py` (11 tests) passes.
- Adjacent suites that exercise the same surfaces all pass with no regressions:
  - `tests/flytekit/unit/core/test_array_node_map_task.py` (46 tests)
  - `tests/flytekit/unit/core/test_map_task.py` (23 tests)
  - `tests/flytekit/unit/core/test_node_creation.py` (26 tests)

- [ ] I updated the documentation accordingly. _(No user-facing behavior change beyond the bug fix; the documented behavior of `with_overrides(pod_template=...)` for `map_task` is what this PR makes work.)_

### DIFF
--- a/tests/flytekit/unit/test_translator.py
+++ b/tests/flytekit/unit/test_translator.py
@@ -1,8 +1,12 @@
 import typing
 from collections import OrderedDict
 
+import pytest
+from kubernetes import client
+from kubernetes.client import V1Container, V1PodSpec
+
 import flytekit.configuration
-from flytekit import ContainerTask, Resources, PodTemplate
+from flytekit import ContainerTask, PodTemplate, Resources, map_task
 from flytekit.configuration import FastSerializationSettings, Image, ImageConfig
 from flytekit.core.base_task import kwtypes
 from flytekit.core.launch_plan import LaunchPlan, ReferenceLaunchPlan
@@ -13,9 +17,6 @@ from flytekit.deck import Deck
 from flytekit.models.core import identifier as identifier_models
 from flytekit.models.task import Resources as resource_model
 from flytekit.tools.translator import get_serializable
-from kubernetes import client
-from kubernetes.client import V1PodSpec, V1Container
-import pytest
 
 default_img = Image(name="default", fqn="test", tag="tag")
 serialization_settings = flytekit.configuration.SerializationSettings(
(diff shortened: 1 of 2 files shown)


pingsutw (approved):
Thank you for fixing it. LGTM

tests/flytekit/unit/test_translator.py
```diff
@@ -234,3 +235,59 @@ def wf():
...
+                V1Container(
+                    name="primary",
+                    env=[client.V1EnvVar(name="MY_KEY", value="MY_VALUE")],
+                )
+            ]
+        ),
+    )
+
+    @task
+    def t(a: int) -> str:
```
pingsutw, line 265: Could we also add another test that `t` has a podTemplate config and we use with_overrides() to override it?

```python
 @task(pod_template=base_pod_template)
    def t(a: int) -> str:
        return str(a)
```

1fanwang: Done in 9dde14668 — added `test_map_task_with_pod_template_override_replaces_task_pod_template` covering the case where `@task(pod_template=base_pod_template)` is then overridden via `map_task(t)(...).with_overrides(pod_template=override_pod_template)`.

While writing the test I caught a real bug: when the underlying task already has `pod_template` set, `entity.get_container(settings)` returns `None` (it intentionally defers to `get_k8s_pod`), and `_serialize_pod_spec` would dereference `primary_container.image` on `None`. Fixed in the same commit by falling back to the inner task's `_get_container` under `prepare_target()` so the override pod spec still resolves image/command/args. Test asserts the override's `OVERRIDE_KEY` env wins over the base.

pingsutw (approved):
Looks good, thanks

Dribblelife:
When will this official version be released?

PR #3430, opened 2026-05-01
chore: Allow later version of setuptools (<82)

honnix:
If 82 is the first version removing `pkg_resources`, it doesn't feel necessary to pin to such a low version that could easily conflict with other packages that requires higher version of setuptool.

Change to `<82`.

## How was this patch tested?

CI

## Related PRs

#3388

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools<70", "setuptools_scm"]
+requires = ["setuptools<82", "setuptools_scm"]
 build-backend = "setuptools.build_meta"
 
 [project]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -38,7 +38,7 @@ dependencies = [
     "marshmallow>=3,<4",
     "marshmallow-enum",
     "marshmallow-jsonschema>=0.12.0",
-    "setuptools<70",
+    "setuptools<82",
```
pingsutw, line 41: Actually, we could remove the pin here. It was pinned to <70 because there was a bug between versions 70–80

pingsutw (approved):

pyproject.toml
honnix: Would 82 work because of the `pkg_resources` thing?

pingsutw: Ah, good point. I’m not sure, so let’s leave it as is for now. (<82)

honnix:
@pingsutw Is it the time to drop a new release? We would like to get this fix to get around a security issue https://nvd.nist.gov/vuln/detail/cve-2024-6345. Thanks.

pingsutw:
@honnix here you go. https://github.com/flyteorg/flytekit/releases/tag/v1.16.20

PR #3338, opened 2025-10-10
Add remote workflow version execution support to pyflyte run

CtfChan:
## Tracking issue
N/A

## Why are the changes needed?

Currently, the `pyflyte run` command supports executing remote tasks (`remote-task`) and remote launch plans (`remote-launchplan`), but lacks support for executing exact version of remote workflows directly. This creates an inconsistent user experience and limits flexibility when users want to execute workflows that are already registered in a Flyte backend.

## What changes were proposed in this pull request?

This PR adds support for executing remote workflows via the `pyflyte run remote-workflow` command, bringing feature parity with the existing remote task and launch plan execution capabilities.

**Changes include:**
1. Added `WORKFLOW_LAUNCHER` constant to `DynamicEntityLaunchCommand` class
2. Updated `_fetch_entity()` method to handle the workflow launcher type by calling `fetch_workflow()` on the remote instance
3. Added support for version-specific workflow execution using the `name:version` syntax
4. Updated `get_command()` in `RemoteEntityGroup` to create a `DynamicEntityLaunchCommand` with the workflow launcher when the command is `remote-workflow`
5. Added comprehensive unit tests (`test_remote_workflow` and `test_remote_workflow_with_version`) to verify the functionality

## How was this patch tested?

Added two new unit tests in `tests/flytekit/unit/cli/pyflyte/test_run.py`:

1. **`test_remote_workflow`**: Tests basic remote workflow execution with parameters
2. **`test_remote_workflow_with_version`**: Tests version-specific workflow execution using the `name:version` syntax

Both tests:
- Mock `FlyteRemote` and `run_remote` to verify correct behavior
- Use the `CliRunner` to simulate command-line execution
- Verify that `fetch_workflow()` is called with the correct arguments
- Ensure the workflow executes successfully with provided inputs

All tests pass successfully:
```
pytest tests/flytekit/unit/cli/pyflyte/test_run.py::test_remote_workflow tests/flytekit/unit/cli/pyflyte/test_run.py::test_remote_workflow_with_version -v
```

### Setup process
N/A

### Screenshots
N/A

## Check all the applicable boxes

- [ ] I updated the documentation accordingly.
- [x] All new and existing tests passed.
- [ ] All commits are signed-off.

## Related PRs
N/A

## Docs link
N/A 
 <div id='description'>
    <a href="https://bito.ai#summarystart"></a>
<h3>Summary by Bito</h3>
[...]

PR #3432, opened 2026-05-15
Make optional arguments not required in click option

Sovietaced:
This pull request addresses a regression introduced in https://github.com/flyteorg/flytekit/pull/3338

Running a remote-workflow used to involved retrieving an existing launch plan and crafting click options based on the launch plan defaults and the values provided on the CLI. When the code was changed to retrieve an existing workflow it does not include any possible default values from the launch plan so we update the code to introspect whether arguments are required or not based on the signature of the workflow. 

## How was this patch tested?

Tested in production as well as unit tests.

### DIFF
--- a/flytekit/clis/sdk_in_container/run.py
+++ b/flytekit/clis/sdk_in_container/run.py
@@ -866,7 +866,7 @@ class DynamicEntityLaunchCommand(click.RichCommand):
         for name, var in inputs.items():
             if fixed and name in fixed:
                 continue
-            required = True
+            required = not is_optional(native_inputs[name])
             default_val = None
             if defaults and name in defaults:
                 if not defaults[name].required:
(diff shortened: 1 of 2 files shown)


machichima:
I tested with following script, there's no error when using ` pyflyte run remote-workflow`. Thank you!

image


image


```py
import os
from typing import Optional

@task()
def example_task(flag: bool, docker_tag: Optional[str] = None) -> bool:
    print(f"docker_tag: {docker_tag}")
    return flag

@workflow
def wf(
    flag: bool = True,
    docker_tag: Optional[str] = None,
) -> None:
    example_task(flag=flag, docker_tag=docker_tag)


if __name__ == "__main__":

    runner = CliRunner()
    result = runner.invoke(pyflyte.main, ["run", "--remote", path, "wf"])
```

machichima (approved):

pingsutw (approved):

PR #3439, opened 2026-06-11
fix: EnumTransformer.to_literal accepts a string matching an enum value

1fanwang:

`EnumTransformer.assert_type` accepts a raw string that matches one of the enum's values (for instance an enum default supplied as a string), but `EnumTransformer.to_literal` rejects that same string with `Expected an enum`. So a matching-string value passes type-checking and then fails at serialization.

`to_literal` now accepts a string matching one of the enum's values and serializes it to the corresponding literal, consistent with `assert_type`. A non-matching string is still rejected.

## How was this patch tested?

Added `test_enum_to_literal_accepts_matching_string` — it fails on the current code with `TypeTransformerFailedError: Expected an enum` and passes with the fix.

### DIFF
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1081,8 +1081,7 @@ class EnumTransformer(TypeTransformer[enum.Enum]):
     def to_literal(
         self, ctx: FlyteContext, python_val: Union[enum.Enum, str], python_type: Type[T], expected: LiteralType
     ) -> Literal:
-        # Accept a raw string matching one of the enum's values: assert_type already allows it (e.g. an enum
-        # default supplied as a string), so to_literal must too, else such a value type-checks but fails to serialize.
+        # Accept either an enum member or a raw string matching one of its values.
         val = python_val.value if isinstance(python_val, enum.Enum) else python_val
         if not isinstance(val, str):
             raise TypeTransformerFailedError(f"Expected an enum or matching string, got {type(python_val)}")
--- a/flytekit/core/type_engine.py
+++ b/flytekit/core/type_engine.py
@@ -1081,6 +1081,14 @@ class EnumTransformer(TypeTransformer[enum.Enum]):
     def to_literal(
         self, ctx: FlyteContext, python_val: enum.Enum, python_type: Type[T], expected: LiteralType
     ) -> Literal:
+        # Accept a raw string that matches one of the enum's values. assert_type already
+        # allows this (e.g. an enum default supplied as a string), so to_literal must too,
+        # otherwise such a value passes type-checking but fails serialization.
+        if isinstance(python_val, str):
+            enum_type = cast(Type[enum.Enum], python_type)
+            if python_val not in [item.value for item in enum_type]:
+                raise TypeTransformerFailedError(f"Value {python_val} is not in Enum {python_type}")
+            return Literal(scalar=Scalar(primitive=Primitive(string_value=python_val)))
         if type(python_val).__class__ != enum.EnumMeta:
             raise TypeTransformerFailedError("Expected an enum")
         if type(python_val.value) != str:
@@ -2079,6 +2087,8 @@ class UnionTransformer(AsyncTypeTransformer[T]):
         is_ambiguous = False
         res = None
         res_type = None
+        # Result of the ``str`` variant, if it matched, used to disambiguate a bare-string value.
+        str_match: typing.Optional[typing.Tuple[Literal, LiteralType]] = None
         t = None
         for i in range(len(get_args(python_type))):
             try:
@@ -2095,6 +2105,8 @@ class UnionTransformer(AsyncTypeTransformer[T]):
                 res_type = _add_tag_to_type(trans.get_literal_type(t), trans.name)
                 found_res = True
                 potential_types.append(t)
+                if t is str:
+                    str_match = (res, res_type)
             except Exception as e:
                 logger.debug(
                     f"UnionTransformer failed attempt to convert from {python_val} to {t} error: {e}",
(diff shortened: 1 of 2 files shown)


flytekit/core/type_engine.py
machichima: Could we simplify this function to:

```suggestion
def to_literal(
    self, ctx: FlyteContext, python_val: Union[enum.Enum, str], python_type: Type[T], expected: LiteralType
) -> Literal:
    val = python_val.value if isinstance(python_val, enum.Enum) else python_val
    if not isinstance(val, str):
        raise TypeTransformerFailedError(f"Expected an enum or matching string, got {type(python_val)}")
    if val not in [item.value for item in python_type]:
        raise TypeTransformerFailedError(f"Value {python_val} is not in Enum {python_type}")
    return Literal(scalar=Scalar(primitive=Primitive(string_value=val)))
```

flytekit/core/type_engine.py
```diff
@@ -2079,6 +2087,8 @@ async def async_to_literal(
         is_ambiguous = False
         res = None
         res_type = None
+        # Result of the ``str`` variant, if it matched, used to disambiguate a bare-string value.
+        str_match: typing.Optional[typing.Tuple[Literal, LiteralType]] = None
```
machichima: Adding `str_match` can make the logic here more complex. Could we update this function to:

```py
    async def async_to_literal(
        self, ctx: FlyteContext, python_val: T, python_type: Type[T], expected: LiteralType
    ) -> typing.Union[Literal, asyncio.Future]:
        python_type = get_underlying_type(python_type)

        matches: list[tuple[type, Literal, LiteralType]] = []
        for i, t in enumerate(get_args(python_type)):
            try:
                trans: TypeTransformer[T] = TypeEngine.get_transformer(t)
                if isinstance(trans, AsyncTypeTransformer):
                    res = await trans.async_to_literal(ctx, python_val, t, expected.union_type.variants[i])
                else:
                    res = trans.to_literal(ctx, python_val, t, expected.union_type.variants[i])
                res_type = _add_tag_to_type(trans.get_literal_type(t), trans.name)
                matches.append((t, res, res_type))
            except Exception as e:
                logger.debug(f"UnionTransformer failed attempt to convert from {python_val} to {t} error: {e}")

        if not matches:
            raise TypeTransformerFailedError(f"Cannot convert from {python_val} to {python_type}")

        if len(matches) > 1:
            # NOTE: More than one variant accepted the value. Pick the one whose type is
            # exactly the same as python_val's type. For example, given Union[str, Color]
            # and python_val="red", both str and Color accept it (Color accepts a matching
            # string), but str is the exact match so it wins.
            exact = [m for m in matches if type(python_val) is m[0]]
            if len(exact) != 1:
                raise TypeError(
                    f"Ambiguous choice of variant for union type.\n"
                    f"Potential types: {[m[0] for m in matches]}\n"
                    "These types are structurally the same, because it's attributes have the same names and associated types."
                )
            matches = exact

        _, res, res_type = matches[0]
        return Literal(scalar=Scalar(union=Union(value=res, stored_type=res_type)))

```

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -1532,6 +1531,39 @@ def test_enum_type():
...
+    lt = TypeEngine.to_literal_type(Color)
+
+    with pytest.raises(TypeTransformerFailedError):
+        TypeEngine.to_literal(ctx, python_val, Color, lt)
+
+
+def test_enum_string_in_union_prefers_str():
+    # A bare string matching an enum value would otherwise match both the ``str`` and enum
+    # variants of a union. The string is most specifically a ``str``, so the ``str`` variant
+    # is chosen instead of raising an ambiguity error.
```
machichima: nit: simplify and put in docstring

```suggestion
    ﻿"""Test that a bare string in Union[Color, str] resolves to str, not the enum, without ambiguity error."""
```

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -1532,6 +1531,39 @@ def test_enum_type():
         TypeEngine.to_literal_type(UnsupportedEnumValues)
 
 
+@pytest.mark.parametrize("python_val", ["red", "green", "blue"])
```
machichima: nit: I think we do not need `parametrize` here, as they are all going through the same code path

tests/flytekit/unit/core/test_type_engine.py
```diff
@@ -1532,6 +1531,39 @@ def test_enum_type():
...
+def test_enum_to_literal_accepts_matching_string(python_val):
+    # A string matching an enum value is accepted by assert_type, so to_literal must
+    # accept it too (e.g. an enum default supplied as a string).
```
machichima: nit: put into the docstring

```py
"""Test that a string matching an enum value is accepted by to_literal."""
```

flytekit/core/type_engine.py
1fanwang: Done in 39c76dc5.

flytekit/core/type_engine.py
1fanwang: Done in 39c76dc5. Heads-up: exact-type-wins now resolves `Union[int, UnsignedInt]` + `3` to `int` instead of raising, so I updated `test_union_custom_transformer_sanity_check` to assert that (it matches `test_union_custom_transformer`, which already expects `int` to win for `3`).

tests/flytekit/unit/core/test_type_engine.py
1fanwang: Done in 39c76dc5.

tests/flytekit/unit/core/test_type_engine.py
1fanwang: Done in 39c76dc5.

tests/flytekit/unit/core/test_type_engine.py
1fanwang: Done in 39c76dc5.

flytekit/core/type_engine.py
1fanwang: Correction to the above: the general exact-type tie-break broke `test_ambiguous_union_transformer_to_literal` and the `sanity_check` (in 2 files) -- it silently resolves structurally-identical variants (`Union[A, B]`, `Union[int, UnsignedInt]`) that must stay ambiguous, since they are indistinguishable on read-back. Narrowed it to just the str-vs-enum case in 033d1cc9, which keeps that guard. Happy to drop the guard and update those tests instead if you prefer.

flytekit/core/type_engine.py
```diff
@@ -1079,14 +1079,16 @@ def get_literal_type(self, t: Type[T]) -> LiteralType:
...
+        self, ctx: FlyteContext, python_val: Union[enum.Enum, str], python_type: Type[T], expected: LiteralType
     ) -> Literal:
-        if type(python_val).__class__ != enum.EnumMeta:
-            raise TypeTransformerFailedError("Expected an enum")
-        if type(python_val.value) != str:
-            raise TypeTransformerFailedError("Only string-valued enums are supported")
-
-        return Literal(scalar=Scalar(primitive=Primitive(string_value=python_val.value)))  # type: ignore
+        # Accept a raw string matching one of the enum's values: assert_type already allows it (e.g. an enum
+        # default supplied as a string), so to_literal must too, else such a value type-checks but fails to serialize.
```
machichima: Those details should be on PR description rather than comment. Let's simply to:

```suggestion
        # Accept either an enum member or a raw string matching one of its values.
```

machichima (approved):
LGTM! Thank you. Just a small nit left

machichima:
Hi @1fanwang ,
I merged your CI fixed https://github.com/flyteorg/flytekit/pull/3448, could you please rebase/merge master?
Thanks!

flytekit/core/type_engine.py
1fanwang: Done in `ad015c604` — the rationale's already in the PR description's *Why* section.

PR #3442, opened 2026-06-20
fix: merge LaunchPlan security context per field on registration

1fanwang:

Registering a launch plan that declares `secrets` (or `tokens`) while passing registration `Options` that only set a service account silently drops the secrets. `get_serializable_launch_plan` does `options.security_context or entity.security_context` — a wholesale replace — so any non-empty options context wipes whatever the launch plan authored.

Merge the two security contexts per field instead of replacing wholesale: options win field-by-field (`run_as`, `secrets`, `tokens`), so a service-account-only override stops clobbering authored secrets.

## How was this patch tested?

`test_get_serializable_launch_plan` drives the same path. Also end-to-end against a real flyteadmin (`flyte demo`): register the secrets launch plan with service-account-only options and read it back — before, `secrets` returns `None` (dropped); after, it's preserved alongside the overridden service account.

### DIFF
--- a/flytekit/tools/translator.py
+++ b/flytekit/tools/translator.py
@@ -30,6 +30,7 @@ from flytekit.core.workflow import ReferenceWorkflow, WorkflowBase
 from flytekit.models import common as _common_models
 from flytekit.models import interface as interface_models
 from flytekit.models import launch_plan as _launch_plan_models
+from flytekit.models import security as _security_models
 from flytekit.models.admin import workflow as admin_workflow_models
 from flytekit.models.admin.workflow import WorkflowSpec
 from flytekit.models.concurrency import ConcurrencyPolicy
@@ -317,6 +318,26 @@ def get_serializable_workflow(
     )
 
 
+def _merge_security_context(
+    entity_sc: Optional[_security_models.SecurityContext],
+    options_sc: Optional[_security_models.SecurityContext],
+) -> Optional[_security_models.SecurityContext]:
+    """Merge the launch plan's authored security context with the one supplied via registration options.
+
+    Registration options override the authored launch plan per field (not wholesale), so e.g. registering with
+    only a service account does not drop secrets/tokens that were authored on the launch plan.
+    """
+    if options_sc is None:
+        return entity_sc
+    if entity_sc is None:
+        return options_sc
+    return _security_models.SecurityContext(
+        run_as=options_sc.run_as or entity_sc.run_as,
+        secrets=options_sc.secrets or entity_sc.secrets,
+        tokens=options_sc.tokens or entity_sc.tokens,
+    )
+
+
 def get_serializable_launch_plan(
     entity_mapping: OrderedDict,
     settings: SerializationSettings,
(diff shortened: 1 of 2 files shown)


tests/flytekit/unit/core/test_serialization.py
```diff
@@ -1272,3 +1270,31 @@ async def t1_eager(a: int) -> int:
         ],
         limits=[]
     )
+
+
+def test_launch_plan_security_context_merge():
+    # Registration options override the authored launch plan per field; registering with only a
+    # service account must not drop secrets/tokens that were authored on the launch plan.
```
machichima: let's make this the test function docstring

tests/flytekit/unit/core/test_serialization.py
```diff
@@ -1272,3 +1270,31 @@ async def t1_eager(a: int) -> int:
...
+    from flytekit import LaunchPlan
+    from flytekit.core.options import Options
+    from flytekit.models.security import Identity, Secret, SecurityContext
```
machichima: Could we move the import to the top of the file?

flytekit/tools/translator.py
```diff
@@ -317,6 +318,26 @@ def get_serializable_workflow(
...
+    only a service account does not drop secrets/tokens that were authored on the launch plan.
+    """
+    if options_sc is None:
+        return entity_sc
+    if entity_sc is None:
+        return options_sc
+    return _security_models.SecurityContext(
+        run_as=options_sc.run_as or entity_sc.run_as,
+        secrets=options_sc.secrets or entity_sc.secrets,
+        tokens=options_sc.tokens or entity_sc.tokens,
```
machichima, line 337: nit: not a blocker, just wondering if a user explicitly sets `secrets=[]` or `tokens=[]` in options to clear the entity value, we'll ignore them here and use the one from entity. Have you observed any usage of setting them as empty list?

tests/flytekit/unit/core/test_serialization.py
1fanwang: ack, done in https://github.com/flyteorg/flytekit/pull/3442/commits/f1ffdda0529ec6f1ddc2f4f8f3958747ba1e6dda

tests/flytekit/unit/core/test_serialization.py
1fanwang: done in https://github.com/flyteorg/flytekit/pull/3442/commits/f1ffdda0529ec6f1ddc2f4f8f3958747ba1e6dda

flytekit/tools/translator.py
1fanwang: Good point — right now the or treats an explicit empty list the same as "unset", so options.secrets=[]/tokens=[] can't clear an authored list; the launch-plan value wins. I haven't seen []-to-clear used in practice (registration options have been additive in what I've hit — service account, labels), so I kept the simpler or. If you'd rather honor an explicit empty list as "clear", I can switch to an is not None check per field — happy to do whichever you prefer.

machichima: Thanks, we can keep it simpler for now

machichima (approved):
LGTM! Thank you. Please rebase/merge master for the CI fix

PR #3443, opened 2026-06-20
fix(auth): don't delete cached credentials when token refresh fails

1fanwang:

The authenticators delete the cached credentials whenever a token refresh fails, then fall back to a full re-auth. But `KeyringStore.store()` overwrites in place, so the `delete()` is redundant — and harmful: a transient refresh failure (a network blip, an IdP hiccup) wipes the last-good tokens, including a still-valid `refresh_token`, forcing the user through an unnecessary browser re-login.

Drop the premature `KeyringStore.delete(self._endpoint)` from the refresh-failure paths in the PKCE, device-code, and GCP identity-aware-proxy authenticators. Each fall-through still calls `store()`, so a successful re-auth persists. At the device-flow poll site the surrounding `try/except` existed only to delete-and-re-raise, so it's removed and exceptions propagate unchanged.

## How was this patch tested?

Added `test_pkce_refresh_failure_keeps_cached_credentials` and tightened `test_device_flow_authenticator`, both asserting `delete` isn't called when a refresh fails. Reverting the fix makes both fail with `Expected 'delete' to not have been called. Called 1 times.`

### DIFF
--- a/tests/flytekit/unit/clients/auth/test_authenticator.py
+++ b/tests/flytekit/unit/clients/auth/test_authenticator.py
@@ -12,7 +12,7 @@ from flytekit.clients.auth.authenticator import (
     PKCEAuthenticator,
     StaticClientConfigStore,
 )
-from flytekit.clients.auth.exceptions import AuthenticationError, AccessTokenNotFoundError
+from flytekit.clients.auth.exceptions import AccessTokenNotFoundError, AuthenticationError
 from flytekit.clients.auth.keyring import Credentials
 from flytekit.clients.auth.token_client import DeviceCodeResponse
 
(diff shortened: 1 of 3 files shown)


tests/flytekit/unit/clients/auth/test_authenticator.py
```diff
@@ -130,6 +151,9 @@ def test_device_flow_authenticator(mock_refresh: MagicMock, poll_mock: MagicMock
     device_mock.assert_called()
     mock_refresh.assert_called()
 
+    # The failed refresh must not have deleted the cached credentials.
+    mock_keyring.delete.assert_not_called()
```
machichima: Seems like this test is testing the case which it will succeed in this line: https://github.com/flyteorg/flytekit/blob/ccf65aae23db62de22ed81ece82fdc44ac333f84/flytekit/clients/auth/authenticator.py#L342-L342

I think we do not need to add this assertion here?

tests/flytekit/unit/clients/auth/test_authenticator.py
```diff
@@ -51,6 +51,27 @@ def test_pkce_authenticator(mock_refresh: MagicMock, mock_get_creds: MagicMock,
...
+    mock_keyring.retrieve.return_value = Credentials("access", "refresh", ENDPOINT)
+    mock_refresh.side_effect = AccessTokenNotFoundError("expired")
+
+    authn = PKCEAuthenticator(ENDPOINT, static_cfg_store, verify=False)
+    authn.refresh_credentials()
+
+    # A failed refresh must not delete the cached credentials. The authenticator falls back to a
+    # full authorization flow and overwrites them via store(), so the delete was both redundant and
+    # harmful -- a transient failure would force an unnecessary re-login.
+    mock_keyring.delete.assert_not_called()
```
machichima:
```suggestion
    # A failed refresh must not delete the cached credentials
    mock_keyring.delete.assert_not_called()
```
nit: simpler comment

machichima (commented):
Thank you! Just two small nits

tests/flytekit/unit/clients/auth/test_authenticator.py
1fanwang: Done in `db2cc95ce` — removed; the device-flow path never calls `delete()`, so it was a no-op (the PKCE test covers the fix).

tests/flytekit/unit/clients/auth/test_authenticator.py
1fanwang: Done in `db2cc95ce`.

machichima (approved):
LGTM! Thank you

